PR #22, opened 2019-02-11
[WIP] Support Python 2.

ueshin:
Modifies to support Python 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,11 +6,11 @@ cache:
 
 matrix:
   include:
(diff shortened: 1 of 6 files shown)


rxin:
I think it is probably OK to have this be Python 3 only to start with.

@thunterdb WDYT?

thunterdb:
Let's keep it python 3 for now, until there is a clear need

ueshin:
Sure, I'd close this.

PR #44, opened 2019-03-29
setup: Make spark an extra-dependency

AbdealiJK:
Spark is installed in many different ways in an environment.
Sometimes using Cloudera parcels, homebrew, binary zips, etc.

Getting the pypi spark package can make it complicated
as it is pulled during a pip install and can cause multiple
sparks to be installed by mistake, and even confuse the configs
to use - potentially breaking a system.

Move spark dependency in an extra in case a user needs us to
use pip to install spark for them. But in general avoid that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -18,6 +18,7 @@
 
 from setuptools import setup
 
+
 install_requires = [
     'pandas>=0.23',
     'decorator',
--- a/setup.py
+++ b/setup.py
@@ -18,15 +18,20 @@
 
 from setuptools import setup
 
+install_requires = [
+    'pandas>=0.23',
+    'decorator',
+    'pyarrow>=0.10,<0.11',  # See https://github.com/databricks/spark-pandas/issues/26
+]
+
 setup(
     name='databricks-koala',
     version='0.0.5',
     packages=['databricks', 'databricks.koala', 'databricks.koala._dask_stubs'],
-    install_requires=[
-        'pyspark>=2.4.0',
-        'pandas>=0.23',
-        'decorator',
-        'pyarrow>=0.10,<0.11'],  # See https://github.com/databricks/spark-pandas/issues/26
+    extras_require={
+        'spark': ['pyspark>=2.4.0'],
+    },
+    install_requires=install_requires,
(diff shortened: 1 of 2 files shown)


thunterdb:
Hello @AbdealiJK , thank you for checking this issue. That sounds sensible, I did not know about the extras_require.

One piece that would be nice to have is a warning/error message if someone tries to use spark < 2.4.0, like in the `__init__.py` for example. Once you add that, I will merge this. Thanks!

AbdealiJK:
Could you help me a bit with this one ? 
The only way i know of checking pyspark version is:
```
import pyspark
spark = pyspark.sql.SparkSession.builder.getOrCreate()
spark.sparkContext.version
```

Do you want to create a spark context during the `pip install` to check this ?

AbdealiJK:
Oh, and also im assuming koala would have to be installed in all spark workers.
Correct me if I'm wrong - but I can configure a node to be a spark worker but not a spark gateway (i.e. it cannot create spark Contexts)

AbdealiJK:
Forgot that 2.1 onwards, `pyspark.__version__` existed. I've added a try catch to fetch it and handle 2.0 and lower versions.

setup.py
```diff
@@ -18,15 +18,38 @@
 
 from setuptools import setup
 
+
+def assert_pyspark_version():
```
thunterdb: @AbdealiJK this logic looks good to me, but I would put this function instead in this file: 

https://github.com/databricks/spark-pandas/blob/master/databricks/koala/__init__.py#L22

I do not think it would work as expected here since this file is the setup file.

AbdealiJK: I did not keep it on Line 22 - as if we're using an import that is not valid in an older pyspark, it would throw an ImportError and the warning would not be shown.
Added it to the top.

AbdealiJK:
Added the assert to the top of the file to avoid errors during imports.
As it is now happening inside the `__init__.py` - Now import pyspark should work, so if that raises an error instead of a warning - we just send a better error message.

thunterdb:
@AbdealiJK thank you for the patch, merging

PR #46, opened 2019-03-29
structures.PandasLikeDataFrame: Add rename()

AbdealiJK:
Allow renaming columns using the rename() function.

### DIFF
--- a/databricks/koala/structures.py
+++ b/databricks/koala/structures.py
@@ -511,6 +511,65 @@ class PandasLikeDataFrame(_Frame):
         df._metadata = self._metadata.copy()
         return df
 
+    def rename(self, mapper=None, index=None, columns=None, axis=None, inplace=False,
+               level=None):
+        """
+        Alter axes labels.
+
+        Function / dict values must be unique (1-to-1). Labels not contained in
+        a dict / Series will be left as-is. Extra labels listed don't throw an
+        error.
+
+        :param mapper: dict-like or function, optional
+                       dict-like or function transformations to apply to the axis given by `axis`.
+        :param index: dict-like or function, optional
+                      dict-like or function transformations to apply to the `index` axis.
+        :param columns: dict-like or function, optional
+                        dict-like or function transformations to apply to the `columns` axis.
+        :param axis: int or str, optional
+                     Axis to target with ``mapper``. Can be either the axis name ('index',
+                     'columns') or number (0, 1). The default is 'index'.
+        :param inplace: boolean, default False
+                        Whether to return a new DataFrame.
+        :return: :class: `DataFrame` if inplace=False else None
+
+        Examples
+        --------
+        >>> df = spark.createDataFrame([ [1, 2, 3], [4, 5, 6] ], ['A', 'B', 'C'])
+        >>> df.rename(columns={"A": "a", "B": "b"})
+        DataFrame[a: bigint, C: bigint, C: bigint]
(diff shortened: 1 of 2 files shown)


thunterdb:
Thanks a lot for this patch, this is very useful.

I have not reviewed the code, but one quick comment: would you mind adding some tests? The easiest is to copy/paste the tests from Dask:
https://github.com/dask/dask/blob/master/dask/dataframe/tests/test_dataframe.py#L191
https://github.com/dask/dask/blob/master/dask/dataframe/tests/test_dataframe.py#L219
https://github.com/dask/dask/blob/master/dask/dataframe/tests/test_dataframe.py#L237

(not all of them might be applicable). You can look into the other tests here for the small modifications you need to do:
https://github.com/databricks/spark-pandas/blob/master/databricks/koala/tests/test_dataframe.py

AbdealiJK:
These seem to be more like that this solves
https://github.com/dask/dask/blob/master/dask/dataframe/tests/test_dataframe.py#L1969

databricks/koala/tests/test_dataframe.py
```diff
@@ -130,6 +130,31 @@ def test_index_names(self):
...
+        df = pd.DataFrame({'a': [1, 2, 3, 4, 5, 6, 7],
+                           'b': [7, 6, 5, 4, 3, 2, 1]})
+        ddf = self.spark.from_pandas(df)
+
+        ddf2 = ddf.rename(columns={'a': 'x'})
+        df2 = df.rename(columns={'a': 'x'})
+        self.assert_eq(ddf2.columns, pd.Index(['x', 'b']))
+        self.assert_eq(ddf2, df2)
+
+        if pd.__version__ > '0.21':
```
thunterdb: we can assume this is the case based on the package bounds.

thunterdb: No need to check for it explicitly

databricks/koala/structures.py
```diff
@@ -511,6 +511,29 @@ def head(self, n=5):
         df._metadata = self._metadata.copy()
         return df
 
+    def rename(self, mapper=None, index=None, columns=None, axis=None):
+        if axis is not None and (index is not None or columns is not None):
```
thunterdb: Did you get inspired from a specific source of code to implement this logic? If this is the case, it would be nice to mention it here.

databricks/koala/tests/test_dataframe.py
AbdealiJK: I had to add this because in Travis there is a check for Spark 2.3 and Pandas 0.19

databricks/koala/structures.py
AbdealiJK: Not per se - I just tried the corner cases mentioned in pandas docs and implemented it here trying to match the same error message and error type

AbdealiJK:
@thunterdb I added the doctests. And apparently travis is trying to run them.
But the `spark` is not defined in them

thunterdb:
@AbdealiJK yes you need a specific setup to enable the doc tests, @ueshin should be able to help. See this file: 
https://github.com/apache/spark/blob/master/python/pyspark/ml/stat.py#L409

If it seems to require more work to set up properly pytest to check the docstring, then let's make a separate task to enable testing the docstring and merge this PR.

thunterdb:
@AbdealiJK the `rename` function has been subsumed in #90 . I am going to close this PR. Thanks again!

PR #48, opened 2019-03-29
PandasLikeDataFrame: Add sample()

AbdealiJK:
Mimic the pandas sample() functionality for spark.

### DIFF
--- a/databricks/koala/structures.py
+++ b/databricks/koala/structures.py
@@ -588,6 +588,24 @@ class PandasLikeDataFrame(_Frame):
     def shape(self):
         return len(self), len(self.columns)
 
+    def sample(self, n=None, frac=None, replace=False, random_state=None):
+        if n is not None and frac is not None:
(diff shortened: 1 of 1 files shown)


databricks/koala/structures.py
```diff
@@ -588,6 +588,24 @@ def pipe(self, func, *args, **kwargs):
     def shape(self):
         return len(self), len(self.columns)
 
+    def sample(self, n=None, frac=None, replace=False, random_state=None):
```
garawalid: I think it's better to add the same parameters as Pandas. And, we raise a `NotImplementedError`.
```python
def(self, n=None, frac=None, replace=False, weights=None, random_state=None, axis=None)
```

HyukjinKwon:
sorry for delay, @AbdealiJK, can you rebase this? let me take a look and be responsive.

rxin:
I'm closing this one. Will continue the work in https://github.com/databricks/koalas/pull/327

PR #53, opened 2019-03-30
PandasLikeDataFrame: Add ascending to sort_values

AbdealiJK:


databricks/koala/structures.py
```diff
@@ -560,8 +560,10 @@ def get(self, key, default=None):
         except (KeyError, ValueError, IndexError):
             return default
 
-    def sort_values(self, by):
-        df = self._spark_sort(by)
+    def sort_values(self, by, ascending=True):
```
ueshin: Could you add a doc?

databricks/koala/tests/test_dataframe.py
```diff
@@ -217,6 +217,23 @@ def test_to_datetime(self):
...
+        df2 = df.sort_values('b')
+        self.assert_eq(ddf2, df2)
+
+        ddf2 = ddf.sort_values(['b', 'a'])
+        df2 = df.sort_values(['b', 'a'])
+        self.assert_eq(ddf2, df2)
+
+        ddf2 = ddf.sort_values(['b', 'a'], ascending=[False, True])
+        df2 = df.sort_values(['b', 'a'], ascending=[False, True])
+        self.assert_eq(ddf2, df2)
```
ueshin: What if the size of `by` and `ascending` are different?
Could you add the tests and follow the pandas behavior?

databricks/koala/structures.py
ueshin: Also adding `inplace` here would be great, otherwise could you add a TODO comment to support `inplace` later?

ueshin: I guess we can support `na_position` as well?

AbdealiJK: Regarding `na_position` - is there any easy way for me to add support for that ?
https://spark.apache.org/docs/2.4.0/api/python/pyspark.sql.html#pyspark.sql.DataFrame.sort
Spark doesn't have any direct equivalent for it so, handling all cases seemed a bit complicated

ueshin: Actually I didn't try it, but I was thinking of using `Column.asc_nulls_first`, `.asc_nulls_last`, `.desc_nulls_first`, and `.desc_nulls_last`.
You can leave it. Thanks.

AbdealiJK: Had forgotten about those as I use Spark 2.3 as my default spark :D
Implemented it using those column-functions

rxin:
@AbdealiJK can you update the patch? This would be useful to include in the next release.

PR #54, opened 2019-03-30
PandasLikeDataFrame: Add .values property

AbdealiJK:
This uses the pandas implementation, so involves converting
to pandas and then into the underlying numpy NDArray
representation. While this may be a bit suboptimal
it is easier as it does not require us to manage spark
to numpy interoperability.


thunterdb:
@AbdealiJK I have started a discussion about this function in #146 , what are your thoughts about it? Your feedback there will be much appreciated.

thunterdb:
@AbdealiJK as discussed in #146 , we are adding `to_numpy()` instead with #130 . Closing this PR.

PR #56, opened 2019-03-30
_Frame: Add abs() for Series and DataFrame

AbdealiJK:
A DataFrame.abs() will go through all columns and apply abs()
on all the columns.

Minor differences:
 - pd.DataFrame.abs() on int,float converts both columns to float
   Here it retains the int type.
 - pd.DataFrame.abs() on str or array throws error.
   Here it will return a None for string. But throw an error for array.

### DIFF
--- a/databricks/koala/structures.py
+++ b/databricks/koala/structures.py
@@ -146,6 +146,15 @@ class _Frame(object):
     def max(self):
         return _reduce_spark(self, F.max)
 
+    @derived_from(pd.DataFrame)
+    def abs(self):
+        """
+        Return a Series/DataFrame with absolute numeric value of each element.
+
+        :return: :class:`Series` or :class:`DataFrame` with the absolute value of each element.
+        """
+        return _spark_col_apply(self, F.abs)
+
     def compute(self):
         """Alias of `toPandas()` to mimic dask for easily porting tests."""
         return self.toPandas()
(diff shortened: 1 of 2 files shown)


databricks/koala/structures.py
```diff
@@ -146,6 +146,9 @@ class _Frame(object):
     def max(self):
         return _reduce_spark(self, F.max)
 
+    def abs(self):
```
thunterdb: You can add the decorator `@derived_from(pd.DataFrame)`. This will add the documentation of the pandas.DataFrame object corresponding to that method automatically.

AbdealiJK: I'm going to copy the docstring instead of using the one in Pandas.
Cause the pandas one has a bunch of info about:
```
For ``complex`` inputs, ``1.2 + 1j``, the absolute value is` :math:`\sqrt{ a^2 + b^2 }`.
And also to some examples on `pd.Timedelta('1 days')` - which I dont think spark supports
```

thunterdb:
@AbdealiJK I am going to merge this pull request. You had extra comments about refining the documentation. This is something we can do as a follow-up.

AbdealiJK:
@thunterdb - I had modified the documentation as per the comment. Was something still pending?

thunterdb:
@AbdealiJK oh yes I had missed that, thank you. Merging this pull request.

PR #57, opened 2019-03-30
Handle common statistic functions

AbdealiJK:
Handles the stat functions that are common to both Spark and Pandas

### DIFF
--- a/databricks/koala/structures.py
+++ b/databricks/koala/structures.py
@@ -144,7 +144,7 @@ class _Frame(object):
     """
 
     def max(self):
-        return _reduce_spark(self, F.max)
+        return self._reduce_for_stat_function(F.max)
 
     def compute(self):
         """Alias of `toPandas()` to mimic dask for easily porting tests."""
@@ -265,6 +265,9 @@ class PandasLikeSeries(_Frame):
         # Pandas wants a series/array-like object
         return _col(self.to_dataframe().unique())
 
+    def _reduce_for_stat_function(self, sfun):
+        return _unpack_scalar(self._spark_ref_dataframe._spark_select(sfun(self)))
+
     @property
     def _pandas_anchor(self) -> DataFrame:
         """
@@ -335,6 +338,14 @@ class PandasLikeDataFrame(_Frame):
     def _index_columns(self):
         return [anchor_wrap(self, self._spark_getitem(field))
                 for field in self._metadata.index_fields]
+
+    def _reduce_for_stat_function(self, sfun):
+        df = self._spark_select([sfun(self[col]).alias(col) for col in self.columns])
+        pdf = df.toPandas()
+        assert len(pdf) == 1, (df, pdf)
+        row = pdf.iloc[0]
+        row.name = None
+        return row  # Return first row as a Series
 
     def set_index(self, keys, drop=True, append=False, inplace=False):
         """Set the DataFrame index (row labels) using one or more existing columns. By default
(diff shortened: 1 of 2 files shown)


AbdealiJK:
Note here that currently `sdf.a0.min` is the same as `sdf.a0.getItem('min')` in spark. Now, we would not support that because `sdf.a0.min` is a method now.

databricks/koala/tests/test_dataframe.py
```diff
@@ -203,6 +203,33 @@ def test_rename_series_method(self):
...
+            getattr(ddf.A, funcname)()
+            getattr(ddf, funcname)()
+
+    def test_count(self):
+        df = pd.DataFrame({'A': [1, 2, 3, 4],
+                           'B': [1.0, 2.1, 3, 4]})
+        ddf = self.spark.from_pandas(df)
+
+        # NOTE: This does not patch the pandas API, but maintains compat with spark
+        self.assertEqual(ddf.count(), len(df))
```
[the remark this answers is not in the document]
AbdealiJK: It finds the count per column - similar to how he mean() max() etc find the stat function per column

databricks/koala/structures.py
```diff
@@ -336,6 +360,14 @@ def _index_columns(self):
         return [anchor_wrap(self, self._spark_getitem(field))
                 for field in self._metadata.index_fields]
 
+    def _reduce_for_stat_function(self, sfun):
+        df = self._spark_select([sfun(self[col]).alias(col) for col in self.columns])
+        pdf = df.toPandas()
+        assert len(pdf) == 1, (df, pdf)
+        row = pdf.iloc[0]
+        row.name = None
+        return row  # Return first row as a Series
```
ueshin: @thunterdb Is it okay to return pandas Series here, or should we return `Column`?

thunterdb: Since this is a row, it makes sense to return here either a Spark row or a pandas Series. Unless people have do complex operations with the Row, I like the idea of just returning a pandas series: it is compact and it should work both as a spark row substitute and as the expected pandas output.

thunterdb:
I am fine with the fact that now the names of the methods shadow the fields, these are fairly short and obvious names. I will add it in the documentation ticket.

AbdealiJK:
I'll modify this PR in a bit.
I want to add some of the other args that are provided in pandas to all of them

databricks/koala/structures.py
```diff
@@ -336,6 +360,14 @@ def _index_columns(self):
         return [anchor_wrap(self, self._spark_getitem(field))
                 for field in self._metadata.index_fields]
 
+    def _reduce_for_stat_function(self, sfun):
+        df = self._spark_select([sfun(self[col]).alias(col) for col in self.columns])
```
rxin: maybe call it _apply_stat_function_to_all_cols?

ueshin:
Hi @AbdealiJK, any updates on this?
I think it's okay to add options in following separate PRs.
How about just fixing the conflicts here and merging this for now? cc @thunterdb 
Thanks!

thunterdb:
I agree we can have a preliminary version in, and add options alter.

thunterdb:
By the way, @AbdealiJK , the struture of the files has been changed. Would you mind resolving the conflicts? Thanks!

ueshin:
Hi @AbdealiJK, I'd take this over for now since there's been a lot of structure and code changes since the last commit here. Let's address adding options in separate PRs.

ueshin:
Submitted the PR #96.

AbdealiJK:
Sure, I did get busy with a few other things and wasn't able to follow up - sorry.

ueshin:
@AbdealiJK No problem. Thanks for tackling this!

PR #58, opened 2019-03-30
_Frame: Add corr() similar to pandas

AbdealiJK:


databricks/koala/structures.py
```diff
@@ -560,6 +564,15 @@ def get(self, key, default=None):
         except (KeyError, ValueError, IndexError):
             return default
 
+    def corr(self, col1=None, col2=None, method=None):
```
thunterdb: I think that not all the methods of Pandas are supported (only Spearman).

databricks/koala/structures.py
```diff
@@ -560,6 +564,15 @@ def get(self, key, default=None):
...
+        if col1 is None and col2 is None:
+            from pyspark.mllib.stat import Statistics
+            colnames = list(self.columns)
+            corr_values = Statistics.corr(self._spark_select(colnames).rdd.map(list), method=method)
```
thunterdb: The `rdd.map(list)` function is programmatically correct but it has poor performance characteristics. I had a first pass on that a while back using the spark.ml functions and the vector assembler:
https://github.com/thunterdb/spark-pandas-1/blob/pandas-profiler/databricks/koala/ml.py

I have not tried it yet, but this is the recommended way when using dataframes.

thunterdb:
@AbdealiJK thanks! We just added the sparkML-based correlation with #149 . Closing this PR.

PR #61, opened 2019-04-08
Add dropna().

ueshin:


ueshin:
cc @thunterdb

databricks/koala/structures.py
```diff
@@ -258,6 +260,17 @@ def to_dataframe(self):
     def toPandas(self):
         return _col(self.to_dataframe().toPandas())
 
+    @derived_from(pd.Series)
+    def dropna(self, axis=0, inplace=False, **kwargs):
```
thunterdb: I do not think we can accept `inplace=True`, because this is not going to work as expected. Also, I just tried it and the behavior is strange and inconsistent so I am in favor of refusing `inplace=True`.

Edit: I saw that you are testing for it and the test work as pandas would expect. I just find this behaviour confusing, but it is according to the spec. In general, I am in favour of  not implementing the in-place flavors of the functions.

PR #67, opened 2019-04-10
Add pyspark.get_dummies.

ueshin:


ueshin:
cc @thunterdb

databricks/koala/namespace.py
```diff
@@ -123,3 +126,59 @@ def _to_datetime2(arg_year, arg_month, arg_day,
...
+    else:
+        if isinstance(prefix, string_types):
+            raise ValueError("get_dummies currently does not support prefix as string types")
+        df = data.copy()
+        if columns is None:
+            columns = [column for column in df.columns
+                       if not isinstance(data.schema[column].dataType, NumericType)]
+        if prefix is None:
+            prefix = columns
+        remaining_columns = [df[column] for column in df.columns if column not in columns]
```
thunterdb: columns is a list so that makes an O(N^2) algorithm. Let's take all the column names into a set, just in case someone has 1000 columns.

databricks/koala/namespace.py
```diff
@@ -123,3 +126,59 @@ def _to_datetime2(arg_year, arg_month, arg_day,
...
+
+    if prefix is not None and len(columns) != len(prefix):
+        raise ValueError(
+            "Length of 'prefix' ({}) did not match the length of the columns being encoded ({})."
+            .format(len(prefix), len(columns)))
+
+    all_values = df._spark_agg(*[F._spark_collect_set(F._spark_col(column)).alias(column)
```
thunterdb: using `_unpack_scalar` in `stuctures.py` does not work?

thunterdb: If there are issues of circular imports, we can move `_unpack_scalar` to utils.

databricks/koala/namespace.py
```diff
@@ -123,3 +126,59 @@ def _to_datetime2(arg_year, arg_month, arg_day,
...
+        columns = [data.name]
+        df = data.to_dataframe()
+        remaining_columns = []
+    else:
+        if isinstance(prefix, string_types):
+            raise ValueError("get_dummies currently does not support prefix as string types")
+        df = data.copy()
+        if columns is None:
+            columns = [column for column in df.columns
+                       if not isinstance(data.schema[column].dataType, NumericType)]
```
thunterdb: Do you know if this is what pandas does? I thought it would accept also numeric values. In fact, I would filter for now for whitelisted types (string and maybe a few others), because I do not think this would work well with, say structures, in Spark (I have not tried).

databricks/koala/namespace.py
thunterdb: I see, you do multiple aggregations. Let's have a `_reduce_spark_multi` that takes a list of aggregators then.

databricks/koala/utils.py
```diff
@@ -94,7 +95,7 @@ def ref_df(x):
 
 def _wrap_operators():
     attrs = ["__neg__", "__add__", "__sub__", "__mul__", "__div__", "__truediv__", "__mod__",
-             "__eq__", "__ne__", "__lt__", "__le__", "__ge__", "__gt__"]
+             "__eq__", "__ne__", "__lt__", "__le__", "__ge__", "__gt__", "__and__", "__or__"]
```
thunterdb: oh yes, thank you.

databricks/koala/namespace.py
ueshin: Yes, I check the pandas behavior and it only accepts non-numeric values by default.
If we explicitly specify the `columns` argument, it accepts numeric values as well.

```py
>>> import pandas as pd
>>> pdf = pd.DataFrame({'a': [0,1,2,0], 'b': list('abcb')})
>>> pd.get_dummies(pdf)
   a  b_a  b_b  b_c
0  0    1    0    0
1  1    0    1    0
2  2    0    0    1
3  0    0    1    0
>>> pd.get_dummies(pdf, columns=['a'])
   b  a_0  a_1  a_2
0  a    1    0    0
1  b    0    1    0
2  c    0    0    1
3  b    1    0    0
>>> pd.get_dummies(pdf, columns=['a', 'b'])
   a_0  a_1  a_2  b_a  b_b  b_c
0    1    0    0    1    0    0
1    0    1    0    0    1    0
2    0    0    1    0    0    1
3    1    0    0    0    1    0
```

I'll use the whitelist for the acceptable types.

thunterdb:
Cool thanks

databricks/koala/namespace.py
```diff
@@ -39,6 +45,9 @@ def from_pandas(pdf):
     return default_session().from_pandas(pdf)
 
 
+SparkSessionPatches.from_pandas.__doc__ = from_pandas.__doc__
```
ueshin: Moved here to avoid circular imports.

thunterdb:
@ueshin it looks gr

PR #69, opened 2019-04-11
Fixed 3 issues of type: PYTHON_E402 throughout 1 file in repo.

deepcommit:
PYTHON_E402: 'module level import not at top of file'.  This is a pep8 error code.          See <a href='https://pep8.readthedocs.io/en/latest/intro.html#error-codes'>        here for a complete list of error codes</a>.

It was fixed with <a href='https://github.com/hhatto/autopep8'>autopep8</a>.  The fix is completely safe.

### DIFF
--- a/databricks/koala/__init__.py
+++ b/databricks/koala/__init__.py
@@ -15,6 +15,9 @@
 #
 
 
+from .utils import *
+from .namespace import *
+from .typing import Col, pandas_wrap
 def assert_pyspark_version():
     import logging
     pyspark_ver = None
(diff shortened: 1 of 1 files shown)


AbdealiJK:
This can cause issues
https://github.com/databricks/spark-pandas/pull/44#issuecomment-480466652

thunterdb:
@deepcommit thanks for the commit. I agree with @AbdealiJK , the test for the spark import should happen first, otherwise spark will be imported at a random place and not give a nice error message if it is not installed. That  means we probably need to disable the E402 check in this file.

ueshin:
Actually E402 is disabled from our pycodestyle now.
https://github.com/databricks/spark-pandas/blob/cde47082dc0e58db2027414d488f227310541d2f/dev/tox.ini#L17

PR #71, opened 2019-04-11
Refine around naming in metadata.

ueshin:


databricks/koala/structures.py
```diff
@@ -195,6 +195,10 @@ def getField(self, name):
                     "Field {} not found, possible values are {}".format(name, ", ".join(fnames)))
             return anchor_wrap(self, self._spark_getField(name))
 
+    # TODO: automate the process here
+    def alias(self, name):
```
thunterdb: With this fix, I am not even sure we need to overwrite this function.

ueshin: I just removed for now, but if we have this, `col.alias("x").name` will return `x`.
Currently the original `Column` methods are not anchor the dataframe reference except for methods in `_wrap_operators`.
We might need to wrap them automatically as the same.

ueshin:
@thunterdb Do you mean the changes with the automatic wrapping or just with `alias()` for now?

thunterdb:
@ueshin  I meant just making sure that chaining method calls with .alias works. This should fail without this PR and then work with this PR.

thunterdb:
@ueshin thanks, merging.

thunterdb:
@ueshin I reverted this PR, since it broke the build. Can you try again? There were some errors with the master merge.

ueshin:
I see, I'll check and submit a PR again.

PR #72, opened 2019-04-12
Fix value_count.

ueshin:
The name of the result of `value_count` should be the column name, but was always `"count"`.

### DIFF
--- a/databricks/koala/structures.py
+++ b/databricks/koala/structures.py
@@ -327,8 +327,9 @@ class PandasLikeSeries(_Frame):
             sum = df_dropna._spark_count()
             df = df._spark_withColumn('count', F._spark_col('count') / F._spark_lit(sum))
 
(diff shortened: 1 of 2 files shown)


databricks/koala/structures.py
```diff
@@ -327,7 +327,10 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
             sum = df_dropna._spark_count()
             df = df._spark_withColumn('count', F._spark_col('count') / F._spark_lit(sum))
 
-        return _col(df.set_index([self.name]))
+        hidden_name = 'index' if self.name != 'index' else 'level_0'
```
thunterdb: why hidden name? How about `index_name`?

thunterdb:
@ueshin it looks good, thanks!

PR #73, opened 2019-04-12
Add index property.

ueshin:
This pr adds `index` property for `DataFrame` and `Column`, which returns a `Column` representing the index column, whereas pandas returns `pd.Index` instead of `pd.Series`.

### DIFF
--- a/databricks/koala/structures.py
+++ b/databricks/koala/structures.py
@@ -239,8 +239,7 @@ class PandasLikeSeries(_Frame):
             self._pandas_metadata = ref._metadata.copy(column_fields=[self.name])
         return self._pandas_metadata
 
-    @_metadata.setter
-    def _metadata(self, metadata):
+    def _set_metadata(self, metadata):
         self._pandas_metadata = metadata
 
     @property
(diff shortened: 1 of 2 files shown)


ueshin:
We might want to wrap the returned column with something like `pd.Index` instead.

databricks/koala/structures.py
```diff
@@ -239,6 +239,18 @@ def _metadata(self):
             self._pandas_metadata = ref._metadata.copy(column_fields=[self.name])
         return self._pandas_metadata
 
+    @_metadata.setter
+    def _metadata(self, metadata):
```
thunterdb: instead of setters, get we just have `_set_metadata`? These functions are not meant to be user-facing, there is no advantage in having a setter.

databricks/koala/structures.py
```diff
@@ -239,6 +239,18 @@ def _metadata(self):
...
+        self._pandas_metadata = metadata
+
+    @property
+    def index(self):
+        """The index (axis labels) Column of the Series.
+
+        Currently supported only when the DataFrame has a single index.
+        """
+        return self._pandas_anchor.index
```
thunterdb: shouldn't there be a check for a single index? The message should also mention what happens when there is more than one index.

databricks/koala/structures.py
ueshin: Sure, I don't have a preference for that. Should we replace with it for `DataFrame` as well?

thunterdb:
@ueshin that looks good, thank you for tackling that task. I think that we will have to do more design around indexes: some code checks about the existence of specific indexes, like `RangeIndex` which is returned as the default index. This is fine for the time being, though.

PR #75, opened 2019-04-15
Reverting latest commit

thunterdb:
The latest commit #71 seems to have broken the build, reverting for the time bing.


thunterdb:
cc @ueshin

PR #78, opened 2019-04-16
Add fillna

garawalid:
+ Only axis = 0 is supported in this method 
+ Dataframe and Series are not supported as value

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -686,37 +686,38 @@ class DataFrame(_Frame):
 
         Examples
         --------
-        >>> df = ks.DataFrame([[np.nan, 2, np.nan, 0],
-        ...                    [3, 4, np.nan, 1],
-        ...                    [np.nan, np.nan, np.nan, 5],
-        ...                    [np.nan, 3, np.nan, 4]],
-        ...                    columns=list('ABCD'))
+        >>> df = ks.DataFrame({
+        ...     'A': [None, 3, None, None],
+        ...     'B': [2, 4, None, 3],
+        ...     'C': [None, None, None, 1],
+        ...     'D': [0, 1, 5, 4]
+        ...     })
         >>> df
-             A    B   C  D
-        0  NaN  2.0 NaN  0
-        1  3.0  4.0 NaN  1
-        2  NaN  NaN NaN  5
-        3  NaN  3.0 NaN  4
+             A    B    C  D
+        0  NaN  2.0  NaN  0
+        1  3.0  4.0  NaN  1
+        2  NaN  NaN  NaN  5
+        3  NaN  3.0  1.0  4
 
         Replace all NaN elements with 0s.
 
         >>> df.fillna(0)
-            A   B   C   D
-        0   0.0 2.0 0.0 0
-        1   3.0 4.0 0.0 1
-        2   0.0 0.0 0.0 5
-        3   0.0 3.0 0.0 4
+             A    B    C  D
+        0  0.0  2.0  0.0  0
+        1  3.0  4.0  0.0  1
+        2  0.0  0.0  0.0  5
+        3  0.0  3.0  1.0  4
 
         Replace all NaN elements in column 'A', 'B', 'C', and 'D', with 0, 1,
         2, and 3 respectively.
 
         >>> values = {'A': 0, 'B': 1, 'C': 2, 'D': 3}
         >>> df.fillna(value=values)
-            A   B   C   D
-        0   0.0 2.0 2.0 0
-        1   3.0 4.0 2.0 1
-        2   0.0 1.0 2.0 5
-        3   0.0 3.0 2.0 4
+             A    B    C  D
+        0  0.0  2.0  2.0  0
+        1  3.0  4.0  2.0  1
+        2  0.0  1.0  2.0  5
+        3  0.0  3.0  1.0  4
         """
 
         if axis is None:
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -685,6 +685,40 @@ class DataFrame(_Frame):
         :param inplace: boolean, default False
                     Fill in place (do not create a new object)
         :return: :class:`DataFrame`
+
+        Examples
+        --------
+        >>> df = ks.DataFrame([[np.nan, 2, np.nan, 0],
+        ...                    [3, 4, np.nan, 1],
+        ...                    [np.nan, np.nan, np.nan, 5],
+        ...                    [np.nan, 3, np.nan, 4]],
+        ...                    columns=list('ABCD'))
+        >>> df
+             A    B   C  D
+        0  NaN  2.0 NaN  0
+        1  3.0  4.0 NaN  1
+        2  NaN  NaN NaN  5
+        3  NaN  3.0 NaN  4
+
+        Replace all NaN elements with 0s.
+
+        >>> df.fillna(0)
+            A   B   C   D
+        0   0.0 2.0 0.0 0
+        1   3.0 4.0 0.0 1
+        2   0.0 0.0 0.0 5
+        3   0.0 3.0 0.0 4
+
+        Replace all NaN elements in column 'A', 'B', 'C', and 'D', with 0, 1,
+        2, and 3 respectively.
+
+        >>> values = {'A': 0, 'B': 1, 'C': 2, 'D': 3}
+        >>> df.fillna(value=values)
+            A   B   C   D
+        0   0.0 2.0 2.0 0
+        1   3.0 4.0 2.0 1
+        2   0.0 1.0 2.0 5
+        3   0.0 3.0 2.0 4
         """
 
         if axis is None:
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -671,6 +671,41 @@ class DataFrame(_Frame):
         else:
             raise NotImplementedError("dropna currently only works for axis=0 or axis='index'")
 
+    @derived_from(pd.DataFrame)
+    def fillna(self, value=None, axis=None, inplace=False):
+        """Fill NA/NaN values.
+
+        :param value: scalar, dict, Series
+                    Value to use to fill holes. alternately a dict/Series of values
+                    specifying which value to use for each column.
+                    DataFrame is not supported.
+        :param axis: {0 or `index`}
+                    1 and `columns` are not supported.
+        :param inplace: boolean, default False
+                    Fill in place (do not create a new object)
+        :return: :class:`DataFrame`
+        """
+
+        if axis is None:
+            axis = 0
+        if value is not None:
+            if axis == 0 or axis == "index":
+                if isinstance(value, (float, int, str, bool, dict)):
+                    sdf = self._sdf.fillna(value)
+                if isinstance(value, pd.Series):
+                    sdf = self._sdf.fillna(value.to_dict())
+                elif isinstance(value, pd.DataFrame):
+                    raise NotImplementedError("Dataframe value is not supported")
+            else:
+                raise NotImplementedError("fillna currently only works for axis=0 or axis='index'")
+        else:
+            raise ValueError('Must specify value')
+
+        if inplace:
+            self._sdf = sdf
+        else:
+            return DataFrame(sdf, self._metadata.copy())
+
     def head(self, n=5):
         """
         Return the first `n` rows.
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -37,9 +37,7 @@ from databricks.koalas.metadata import Metadata
 from databricks.koalas.missing.frame import _MissingPandasLikeDataFrame
 from databricks.koalas.ml import corr
 from databricks.koalas.selection import SparkDataFrameLocator
-from databricks.koalas.typedef import dict_sanitizer
-from databricks.koalas.typedef import infer_pd_series_spark_type
-from databricks.koalas.utils import default_session
+from databricks.koalas.typedef import infer_pd_series_spark_type, dict_sanitizer
 
 
 class DataFrame(_Frame):
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -21,8 +21,8 @@ from functools import partial, reduce
 
 import numpy as np
 import pandas as pd
-from pandas.api.types import is_datetime64_dtype, is_datetime64tz_dtype
 from decorator import dispatch_on
+from pandas.api.types import is_datetime64_dtype, is_datetime64tz_dtype
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Column
 from pyspark.sql.types import StructField, StructType, to_arrow_type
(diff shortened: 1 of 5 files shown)


garawalid:
According to Pandas documentation, `fillna()` uses index in the case of Series.
>value : scalar, dict, Series, or DataFrame
Value to use to fill holes (e.g. 0), alternately a dict/Series/DataFrame of values specifying which value to use for each index (for a Series) or column (for a DataFrame). (values not in the dict/Series/DataFrame will not be filled). This value cannot be a list.

But in the [test code](https://github.com/pandas-dev/pandas/blob/master/pandas/tests/frame/test_missing.py), Series and dict have the same behaviour. 

```python
def test_fillna_dict_series(self):
        df = DataFrame({'a': [np.nan, 1, 2, np.nan, np.nan],
                        'b': [1, 2, 3, np.nan, np.nan],
                        'c': [np.nan, 1, 2, 3, 4]})

        result = df.fillna({'a': 0, 'b': 5})

        expected = df.copy()
        expected['a'] = expected['a'].fillna(0)
        expected['b'] = expected['b'].fillna(5)
        assert_frame_equal(result, expected)

        # it works
        result = df.fillna({'a': 0, 'b': 5, 'd': 7})

        # Series treated same as dict
        result = df.fillna(df.max())
        expected = df.fillna(df.max().to_dict())
        assert_frame_equal(result, expected)

        # disable this for now
        with pytest.raises(NotImplementedError, match='column by column'):
            df.fillna(df.max(1), axis=1)
```

databricks/koala/structures.py
```diff
@@ -156,6 +155,29 @@ def compute(self):
...
+                elif isinstance(value, (pd.DataFrame, pd.Series)):
+                    raise NotImplementedError("Dataframe and Series value are not supported")
+            else:
+                raise NotImplementedError("fillna currently only works for axis=0 or axis='index'")
+        else:
+            raise ValueError('Must specify value')
+
+        if inplace:
+            df._metadata = self._metadata.copy()
+            _reassign_jdf(self, df)
```
thunterdb: when `inplace=True`, what is the return? I would expect it to be `self`

databricks/koala/tests/test_dataframe.py
```diff
@@ -249,11 +277,7 @@ def test_value_counts(self):
         df = pd.DataFrame({'x': [1, 2, 1, 3, 3, np.nan, 1, 4]})
         ddf = self.spark.from_pandas(df)
 
-        exp = df.x.value_counts()
```
thunterdb: I reverted a PR that was causing problems and I think you forked the code at that moment. Can you merge with the most recent master? You should not have to include these changes.

databricks/koala/structures.py
```diff
@@ -156,6 +155,29 @@ def compute(self):
         """Alias of `toPandas()` to mimic dask for easily porting tests."""
         return self.toPandas()
 
+    @derived_from(pd.DataFrame)
+    def fillna(self, value=None, axis=None, inplace=False):
```
thunterdb: Can you add also a docstring that mentions the limitations compared to pandas?
 - Series are not supported (yet) for values
 - not all the options are supported

garawalid:
@thunterdb  Thanks for your comments.
> Regarding the series: we can support pandas Series, like in pandas, it is just a matter to turning them into a dict and using that. If you prefer not to support it for the time being, this is fine with me.

I tried to use Series, it works fine with python3.6 but with python3.5 I ended up with this error. Do you have any idea ?
```
ERROR: test_fillna (databricks.koala.tests.test_dataframe.DataFrameTest)
----------------------------------------------------------------------
  File "/home/travis/build/garawalid/pandas-sparkTest/databricks/koala/tests/test_dataframe.py", line 220, in test_fillna
    self.assert_eq(ddf.fillna(df_nan), df.fillna(df_nan))
  File "/home/travis/build/garawalid/pandas-sparkTest/databricks/koala/structures.py", line 168, in fillna
    df = df._spark_fillna(value.to_dict())
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/pyspark/sql/dataframe.py", line 1523, in fillna
    return DataFrame(self._jdf.na().fill(value), self.sql_ctx)
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1248, in __call__
    args_command, temp_args = self._build_args(*args)
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1212, in _build_args
    (new_args, temp_args) = self._get_args(args)
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1199, in _get_args
    temp_arg = converter.convert(arg, self.gateway_client)
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_collections.py", line 513, in convert
    java_map[key] = object[key]
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_collections.py", line 72, in __setitem__
    self.put(key, value)
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1248, in __call__
    args_command, temp_args = self._build_args(*args)
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1218, in _build_args
    [get_command_part(arg, self.pool) for arg in new_args])
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/java_gateway.py", line 1218, in <listcomp>
    [get_command_part(arg, self.pool) for arg in new_args])
  File "/home/travis/.cache/spark-versions/spark-2.3.3-bin-hadoop2.7/python/lib/py4j-0.10.7-src.zip/py4j/protocol.py", line 298, in get_command_part
    command_part = REFERENCE_TYPE + parameter._get_object_id()
AttributeError: 'numpy.int64' object has no attribute '_get_object_id'
```

databricks/koala/structures.py
garawalid: When `inplace=True`  it will return the result in the same dataframe. 

```
print(df.show())
+-----------------+----+----+----+
|__index_level_0__|   x|   y|   z|
+-----------------+----+----+----+
|               10|null| 1.0| 1.0|
|               20| 2.0| 2.0| 2.0|
|               30| 3.0|null| 3.0|
|               40| 4.0| 4.0| 4.0|
|               50|null|null|null|
|               60| 6.0|null|null|
+-----------------+----+----+----+

ddf.fillna({'x': -1, 'y': -2, 'z': -5}, inplace=True)
print(ddf.show())
+-----------------+----+----+----+
|__index_level_0__|   x|   y|   z|
+-----------------+----+----+----+
|               10|-1.0| 1.0| 1.0|
|               20| 2.0| 2.0| 2.0|
|               30| 3.0|-2.0| 3.0|
|               40| 4.0| 4.0| 4.0|
|               50|-1.0|-2.0|-5.0|
|               60| 6.0|-2.0|-5.0|
+-----------------+----+----+----+

```

databricks/koala/structures.py
garawalid: Done !

databricks/koala/structures.py
garawalid: I changed it to 
```python
if inplace:
    self._jdf = df._jdf
```

databricks/koala/tests/test_dataframe.py
garawalid: Fixed.

databricks/koala/generic.py
```diff
@@ -46,6 +46,40 @@ def compute(self):
...
+                    Value to use to fill holes. alternately a dict of values
+                    specifying which value to use for each column.
+                    Dataframe and Series are not supported.
+        :param axis: {0 or `index`}
+                    1 and `columns` are not supported.
+        :param inplace: boolean, default False
+                    Fill in place (do not create a new object)
+        :return: :class:`Series` or :class:`DataFrame`
+        """
+        df = self.copy()
```
thunterdb: You should not do a copy when you are not doing it in place, all the columns that you may have declared elswhere will not be anchored to the new dataframe.

thunterdb:
Regarding the error you are having with python 3.5, I suspect this is because the contains numpy ints, and currently spark cannot deal with np.int64. The message you are getting is a typical spark serialization error message. We may need to sanitize the input to _spark_fillna to make sure it is given types it recognizes. Let's do that in a subsequent PR. I suspect that you can trigger the error in all python versions by trying to fill in with np.int64 values.

databricks/koala/generic.py
garawalid: I agree with you.

garawalid:
I think the issue is related to pandas. In fact, when I run this code with two different versions of python, I got different results. 
```python
df_nan = pd.Series([-1, -2, -5], index=['x', 'y', 'z'], dtype=int)
print(type(df_nan.to_dict()['x']))

# python 3.6 
<class 'int'>
# python 3.5
<class 'numpy.int64'>
``` 
I suggest that we don't support `pd.Series` for instance.

garawalid:
It turns out that pandas=0.19.2 have this issue. pandas=0.23.4 returns `int` for both versions of python.

garawalid:
@thunterdb 
The issue of pandas disappears starting from version `0.21.0`. Also, we require `pandas>=0.23`. 
I think we can support `Series` and update Travis CI configuration `PANDAS_VERSION=0.21.0` in python 3.5 to correctly build Koalas

garawalid:
@thunterdb  
I added support for Series as input and solved the conflict.

databricks/koalas/generic.py
```diff
@@ -77,6 +77,42 @@ def compute(self):
         """Alias of `toPandas()` to mimic dask for easily porting tests."""
         return self.toPandas()
 
+    @derived_from(pd.DataFrame)
```
thunterdb: This is only going to work with dataframes (and not series), even if it is currently in `Frame`. Let's move this function in `DataFrame` for now, because the implementation for rows is going to be quite different.

databricks/koalas/generic.py
```diff
@@ -77,6 +77,42 @@ def compute(self):
...
+                    Fill in place (do not create a new object)
+        :return: :class:`Series` or :class:`DataFrame`
+        """
+
+        if axis is None:
+            axis = 0
+        if value is not None:
+            if axis == 0 or axis == "index":
+                if isinstance(value, (float, int, str, bool, dict)):
+                    df = self._spark_fillna(value)
```
thunterdb: Now that #90 has been merged, it should now directly call the underlying Spark dataframe directly and not through the shim. For dataframes: `self._sdf.na.fill(value)` (if I am correct).

databricks/koalas/generic.py
garawalid: Ok!

garawalid:
@thunterdb 
Thanks for your comments. I updated the PR.

thunterdb:
@garawalid it looks like one of the tests failed?

garawalid:
@thunterdb 
I think that here also we need to perform data sanitization.

garawalid:
@thunterdb 
I add a method that checks if elements inside a given `dict` are supported, otherwise it raises a `TypeError`. 

Supported types are `int, str, float, np.float64, np.float and bool`. I found these types by running this experiment. (The code is quite a mess)

I used python 3.6 & pyspark 2.4.1

Please review it and let me know if anything is missing.

```python
import numpy as np
import pandas as pd
from pyspark.sql import SparkSession

spark = SparkSession.builder.getOrCreate()

pdf = pd.DataFrame({'x': [np.nan, 2, 3, 4, np.nan, 6],
                    'y': [1, 2, np.nan, 4, np.nan, np.nan],
                    'z': [1, 2, 3, 4, np.nan, np.nan]},
                   index=[10, 20, 30, 40, 50, 60])

df = spark.createDataFrame(pdf)

s_nan = {'x': np.int64(-1), 'y': np.int64(-2), 'z': np.int64(-5)}  # not working

s_nan = {'x': np.int32(-1), 'y': np.int32(-2), 'z': np.int32(-5)}  # not working

s_nan = {'x': np.float64(-1), 'y': np.float64(-2), 'z': np.float64(-5)}  # working

s_nan = {'x': np.int8(-1), 'y': np.int8(-2), 'z': np.int8(-5)}  # not working

s_nan = {'x': np.int16(-1), 'y': np.int16(-2), 'z': np.int16(-5)}  # not working

s_nan = {'x': np.float(-1), 'y': np.float(-2), 'z': np.float(-5)}  # working

df = df.fillna(s_nan)
print(df.show())

## Test bool

pdf = pd.DataFrame({'x': [None, False, False],
                    'y': [False, False, None],
                    'z': [False, False, False]},
                   index=[10, 20, 30])

df = spark.createDataFrame(pdf)

s_nan = {'x': np.bool(True), 'y': np.bool(True), 'z': np.bool(True)}  # working

df = df.fillna(s_nan)

print(df.show())

## Test Datetime

data = np.arange('2005-02', '2005-03', dtype='datetime64[D]')
data[2] = None

df = spark.createDataFrame(pd.DataFrame(data))

print(df.show())

s_nan = {'0': np.datetime64('2005-01-01')}  # not working

df = df.fillna(s_nan)

print(df.show())

```

garawalid:
I'll update #121 once we validate `dict_sanitizer()`

databricks/koalas/frame.py
```diff
@@ -399,6 +400,44 @@ def dropna(self, axis=0, how='any', thresh=None, subset=None, inplace=False):
...
+        :param value: scalar, dict, Series
+                    Value to use to fill holes. alternately a dict/Series of values
+                    specifying which value to use for each column.
+                    DataFrame is not supported.
+        :param axis: {0 or `index`}
+                    1 and `columns` are not supported.
+        :param inplace: boolean, default False
+                    Fill in place (do not create a new object)
+        :return: :class:`DataFrame`
+        """
```
HyukjinKwon: Can we copy and paste doctests?

HyukjinKwon: I mean here https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.fillna.html#pandas.DataFrame.fillna

garawalid: Do you mean examples?

HyukjinKwon: Yea, doc link above.https://github.com/pandas-dev/pandas/blob/master/pandas/core/generic.py#L5984

Ideally if we replace pd to ks, it should work.if it doesn't work we can fix the doctest/docstrings or do it later

garawalid: Good! I skipped `See also` part.

databricks/koalas/typedef.py
```diff
@@ -172,6 +170,18 @@ def _check_compatible(arg, sig_arg: X):
     assert False, (arg, sig_arg)
 
 
+def dict_sanitizer(input_dict):
+    """
+    This function checks if elements inside a given dict are supported by Spark.
+    :param input_dict: dict
+    :raises TypeError if any element is not supported by Spark
+    """
+    _possible_type_values = [int, str, float, np.float64, np.float, bool]
```
HyukjinKwon: Hm, actually I don't think we support np.float ones. Let me check and make a small followup

HyukjinKwon:
Thanks, @garawalid

Issue #77, opened 2019-04-16
Add most of the missing pandas methods as stubs


thunterdb:
Since the documentation step in an external spreadsheet is proving to be too tedious to be respected, let's simply add most of the missing methods as stubs. These methods should:
 - return a `NotImplementedError` (with a message for getting contributions and that includes the name of the missing function)
 - be documented as not implemented yet.

It should be pretty easy to implement a method that always throws for these functions.

PR #79, opened 2019-04-17
Split structures.py.

ueshin:
Prior to #77, we should split `structures.py` file into several files, otherwise the file will be very large to maintain.
Each separated file contains the following and its related util functions.

- session.py: `SparkSessionPatches`
- generic.py: `_Frame`
- frame.py: `PandasLikeDataFrame`
- series.py: `PandasLikeSeries`


thunterdb:
@ueshin thank you for doing that. I am going to merge this PR immediately since it is breaking every other PR.

PR #80, opened 2019-04-17
Add most of the missing pandas methods as stubs.

### LINKED ISSUE #77: Add most of the missing pandas methods as stubs

ueshin:
Added missing pandas methods as stubs and modified the current methods to follow pandas API.

Closes #77.


ueshin:
cc @thunterdb

databricks/koala/frame.py
```diff
@@ -65,8 +66,18 @@ def iteritems(self):
         return list((col_name, self[col_name]) for col_name in cols)
 
     @derived_from(pd.DataFrame)
-    def to_html(self, index=True, classes=None):
-        return self.toPandas().to_html(index=index, classes=classes)
+    def to_html(self, buf=None, columns=None, col_space=None, header=True, index=True,
```
thunterdb: nice, thank you.

dev/_make_missing_functions.py
```diff
@@ -0,0 +1,154 @@
...
+
+import inspect
+
+import pandas as pd
+
+from databricks.koala.frame import PandasLikeDataFrame
+from databricks.koala.series import PandasLikeSeries
+
+
+def inspect_missing_functions(original_type, target_type):
```
thunterdb: please put some documentation about what these functions are expected to do

thunterdb: and return

dev/_make_missing_functions.py
```diff
@@ -0,0 +1,154 @@
...
+                if str(original_signature) != str(target_signature):
+                    updatings.append((name, original_signature, target_signature))
+                continue
+
+        missings.append((name, original_signature))
+
+    return missings, updatings
+
+
+def format_arguments(name, signature):
```
thunterdb: can you explain what this function does?

ueshin:
@thunterdb I reverted the changes for the existing methods.
One thing, as for `Series.rename`, the signature is completely different from pandas'.
I guess I referred dask's signature when I introduced the method.
I think we should fix it now. WDYT?

pandas:
```py
def rename(self, index=None, **kwargs):
```

koala:
```py
def rename(self, name, inplace=False):
```

databricks/koala/dask/utils.py
```diff
@@ -62,8 +62,8 @@ def derived_from(original_klass, version=None, ua_args=[]):
     version : str
         Original package version which supports the wrapped method
     ua_args : list
-        List of keywords which Dask doesn't support. Keywords existing in
-        original but not in Dask will automatically be added.
+        List of keywords which Koala doesn't support. Keywords existing in
```
thunterdb: I forgot. Let's keep the dask code as close as possible to the original code. This will help when trying to backport upstream fixes.

databricks/koala/dask/utils.py
```diff
@@ -88,7 +88,7 @@ def wrapper(method):
 
             if len(not_supported) > 0:
                 note = ("\n        Notes\n        -----\n"
-                        "        Koala doesn't support the following argument(s).\n\n")
+                        "        Dask doesn't support the following argument(s).\n\n")
```
ueshin: @thunterdb At least this should be `Koala`? otherwise "Dask" appears in the document.

thunterdb:
Good point about this message, There is no need to translate all the other
comments.

PR #82, opened 2019-04-18
Switch relative imports to absolute imports in the codebase.

HyukjinKwon:
This PR proposes two things.

**Switch relative import to absolute imports**

1. PySpark codebase uses absolute imports, as far as I know, for the reason that each file has doctests and they are ran via PySpark's testing script infra. In this case, absolute imports should better be used because when the doctests are ran, we're in `__main__` module by default.

    See https://docs.python.org/3/tutorial/modules.html#intra-package-references

    > Note that relative imports are based on the name of the current module. Since the name of the main module is always "__main__", modules intended for use as the main module of a Python application must always use absolute imports.

    I assume we'll want the similar approach since this project is closely related with PySpark.

2. I would like to encourage to keep the styles similar with PySpark side as possible as we can because this project is closely related to PySpark. Pandas also looks preferring absolute import approach in general. See also `https://github.com/pandas-dev/pandas/pull/3827` for instance.

3. _In my humble opinion_, (explicit) relative imports are still practically less preferred in general (but it's fine).

Note: there are pros and cons about relative imports and absolute imports. See https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module for instance.

**Rename `_dask_stubs` to `dask`**

1. PySpark also has a similar case, `cloudpickle`. It also has the name as it. I would like to propose to keep it similar so that we can fix it together similarly if any issue happens.

2. I think it's a little slightly better to keep the name same to make it easier to read (to me, I stopped for a while when I read `_dask_stubs` ... ) 

(Maybe we can move those stubs into `stubs` package?)


HyukjinKwon:
cc @thunterdb and @ueshin, I guess this can be a bit arguable but I would like to go in this way if there's nothing I missed and if you guys are fine with it.

databricks/koala/dask/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""
+Some files have been adapted from the Dask project, under the terms of the Apache 2.0 license
+and the Dask license. See the file licenses/LICENSE-dask.txt for more details.
+"""
```
HyukjinKwon: Maybe, we can move those under `stubs` package?

thunterdb:
@HyukjinKwon thanks a lot for resolving the style differences with Spark. I am going to merge this PR immediately so that it does not conflict with other PRs.

HyukjinKwon:
Thanks, @thunterdb!

PR #94, opened 2019-04-19
Simplify stub code.

### LINKED ISSUE #91: Should we simplify stub code?
Right now they look like:

```
    def add(self, other, axis='columns', level=None, fill_value=None):
        """A stub for the equivalent method to `pd.DataFrame.add()`.
        The method `pd.DataFrame.add()` is not implemented yet.
        """
        raise PandasNotImplementedError(class_name='pd.DataFrame', method_name='add')
```

Should we simplify it to something like

```
def add(**kwargs):
  unsupported_function()
```

or

```
add = unsupported_function()
```

or even

```
unsupported_function('add')
```

Reason I think we should simplify:

1. I don't see a strong need to show all the arguments, in addition to the function name. (Correct me if I'm wrong).
2. Perception wise, it'd be better to have a shorter missing file than a very long file.

ueshin:
Closes #91.

### DIFF
--- a/dev/_make_missing_functions.py
+++ b/dev/_make_missing_functions.py
@@ -216,8 +216,8 @@ def _main():
 
         print('MISSING functions for {}'.format(original_type.__name__))
         for name, signature in missing:
-            print(make_misssing_function(original_type, name, signature))
-            # print("""    {0} = unsupported_function('{0}')""".format(name))
+            # print(make_misssing_function(original_type, name, signature))
+            print("""    {0} = unsupported_function('{0}')""".format(name))
 
         print()
         print('MODIFIED functions for {}'.format(original_type.__name__))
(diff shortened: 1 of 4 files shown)


dev/_make_missing_functions.py
```diff
@@ -217,6 +217,7 @@ def _main():
         print('MISSING functions for {}'.format(original_type.__name__))
         for name, signature in missing:
             print(make_misssing_function(original_type, name, signature))
+            # print("""    {0} = unsupported_function('{0}')""".format(name))
```
rxin: why is this commented out?

rxin:
LGTM other than that one comment. Please merge if tests pass.

dev/_make_missing_functions.py
ueshin: ah, not a special reason but only one of the two will be needed.

ueshin:
I'd merge this now. Thanks!

PR #96, opened 2019-04-19
Handle common statistic functions.

ueshin:
Handles the stat functions kurt, kurtosis, mean, min, skew, std, sum, var, that are common to both Spark and Pandas.


databricks/koalas/generic.py
```diff
@@ -30,8 +30,41 @@ class _Frame(object):
...
+    @derived_from(pd.DataFrame, ua_args=['axis', 'skipna', 'level', 'numeric_only'])
+    def skew(self):
+        return self._reduce_for_stat_function(F.skewness)
+
+    @derived_from(pd.DataFrame, ua_args=['axis', 'skipna', 'level', 'numeric_only'])
+    def kurtosis(self):
+        return self._reduce_for_stat_function(F.kurtosis)
+
+    @derived_from(pd.DataFrame, ua_args=['axis', 'skipna', 'level', 'numeric_only'])
+    def kurt(self, *args, **kwargs):
```
rxin: would it be better to do kurt = kurtosis ?

Similar to how Pandas does it:
```
        cls.kurt = _make_stat_function(
            cls, 'kurt', name, name2, axis_descr,
            "Return unbiased kurtosis over requested axis using Fisher's "
            "definition of\nkurtosis (kurtosis of normal == 0.0). Normalized "
            "by N-1.",
            nanops.nankurt)
        cls.kurtosis = cls.kurt
```

rxin:
LGTM other than that one minor comment.

PR #109, opened 2019-04-20
Add release instruction

### LINKED ISSUE #18: Script to package and release to pypi
We have enough content to make a release onto PyPI. The goal of this ticket is to:
- document the process for future releases, in `docs/RELEASE.md` for example
- make a first release with version `0.1.0`, because this is already pretty usable.

rxin:
Closes #18

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Pandas is the de facto standard (single-node) dataframe implementation in Python
   - [Running Tests](#running-tests)
   - [Contributions](#contributions)
   - [Coding Conventions](#coding-conventions)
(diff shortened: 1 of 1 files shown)


rxin:
@thunterdb the instruction here is quite different from yours in #18. Can you confirm `python setup.py sdist upload -r pypi` works?

Also how do we make sure the wheel files on GitHub are the same as the PyPi ones?

HyukjinKwon:
(FYI, here's cloudpickle release script that I wrote, https://github.com/cloudpipe/cloudpickle/blob/master/dev/make-distribution.sh to refer)

PR #112, opened 2019-04-20
Upgrade PySpark version for travis-ci.

ueshin:
The PySpark `2.4.0` package is no longer available on the download mirrors.


ueshin:
I'd merge this to unblock other PRs.

PR #113, opened 2019-04-20
conda activate in README

rxin:


rxin:
@ueshin can you take a look at this? I want to make sure we are not doing anything wrong switching from source activate to conda activate.

ueshin:
Sure. Actually my local conda was old. Let me check with the new conda.

rxin:
I was reading somewhere that in modern conda, conda activate is preferred.

ueshin:
yeah, it should work and is recommended with conda >= `4.4`.
https://github.com/conda/conda/blob/a4c4feae404b2b378e106bd25f62cc8be15c768f/CHANGELOG.md#440-2017-12-20

PR #117, opened 2019-04-20
Use encapsulation instead of monkey patching / inheritance.

### LINKED ISSUE #23: koalas.DataFrame()
based on a conversation with Brooke, we should have a function called `pyspark.DataFrame` that takes the same arguments as `pandas.DataFrame` https://github.com/pandas-dev/pandas/blob/master/pandas/core/frame.py#L378 but returns a Spark dataframe. It should be a simple wrapper that calls pyspark.createDataFrame

### LINKED ISSUE #90: Use encapsulation instead of monkey patching / inheritance
After spending some time thinking about the issue, I think it would be better for the project to create its own entry point and its own DataFrame / Series class, rather than doing it through monkey patching or inherence to change the existing Spark DataFrame API.

The main reason I am leaning this way now is because Spark and Pandas DataFrames already have quite a few functions with the same name but different semantics, e.g. `sample`, `head`. Keeping existing Spark behavior does not accomplish the goal of the project, while changing existing Spark behavior based on some import statement is an anti-pattern of Python code. Doing it through encapsulation also avoids the issue of Koalas polluting the internal state of Spark's DataFrame through monkey patching.

So here's an alternative design:

koalas: parallel to pandas namespace. Includes I/O functions (`read_csv`, `read_parquet`) as well as general functions (e.g. `concat`, `get_dummies`). Functions here would only work if there is an active SparkSession.

koalas.DataFrame: A completely new interface based on Pandas' DataFrame API. KoalaDataFrame wraps around a Spark DataFrame along with additional internal states.

koalas.Series: Based on Pandas' DataFrame API.

The only thing I'd monkey patch into Spark's DataFrame is a toKoalas method, which turns an existing Spark DataFrame into a Koalas DataFrame.

While this change will bring more code, we will get the following benefits:

- Very clear when users will get the Pandas behavior vs Spark behavior.
- Documentation also becomes clear, as we just provide documentation on Koalas.

The main tradeoff is:

If we go with encapsulation, there is a separate class hierarchy in addition to the existing Spark ones, and the new hierarchy follows strictly Pandas semantics. The downside is that users cannot directly use a Spark DataFrame with Pandas code, and they need to call "spark_ds.toKoalas".

If we go with the current monkey patching approach, it's the opposite of the above. Users can directly use a Spark DataFrame with Pandas code, but existing Spark code's behavior might change as soon as they import koalas package. For example, "head" now returns 5 rows, rather than 1 row. "columns' return a Pandas Index, rather than a Python list. "sample" in the future might also change.

ueshin:
Also addressed #23 while implementing `__init__` for wrappers.

Closes #23, #90.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -15,7 +15,7 @@
 #
 
 """
-A base class to be monkey-patched to DataFrame to behave similar to pandas DataFrame.
+A wrapper class for Spark DataFrame to behave similar to pandas DataFrame.
 """
 from decorator import dispatch_on
 from functools import reduce
@@ -29,7 +29,7 @@ from pyspark.sql.utils import AnalysisException
 
 from databricks.koalas.dask.compatibility import string_types
 from databricks.koalas.dask.utils import derived_from
-from databricks.koalas.generic import _Frame, anchor_wrap, max_display_count
+from databricks.koalas.generic import _Frame, max_display_count
 from databricks.koalas.metadata import Metadata
 from databricks.koalas.missing.frame import _MissingPandasLikeDataFrame
 from databricks.koalas.selection import SparkDataFrameLocator
@@ -45,17 +45,17 @@ class DataFrame(_Frame, _MissingPandasLikeDataFrame):
     @dispatch_on('data')
     def __init__(self, data=None, index=None, columns=None, dtype=None, copy=False):
         pdf = pd.DataFrame(data=data, index=index, columns=columns, dtype=dtype, copy=copy)
-        self.__init__from_pandas(pdf)
+        self._init_from_pandas(pdf)
 
     @__init__.register(pd.DataFrame)
-    def __init__from_pandas(self, pdf, *args):
+    def _init_from_pandas(self, pdf, *args):
         metadata = Metadata.from_pandas(pdf)
         reset_index = pdf.reset_index()
         reset_index.columns = metadata.all_fields
-        self.__init__internal__(default_session().createDataFrame(reset_index), metadata)
+        self._init_from_spark(default_session().createDataFrame(reset_index), metadata)
 
     @__init__.register(spark.DataFrame)
-    def __init__internal__(self, sdf, metadata=None, *args):
+    def _init_from_spark(self, sdf, metadata=None, *args):
         self._sdf = sdf
         self._pandas_metadata = metadata
 
@@ -67,14 +67,13 @@ class DataFrame(_Frame, _MissingPandasLikeDataFrame):
 
     @property
     def _index_columns(self):
-        from databricks.koalas.series import Series
-        return [Series(self._sdf.__getitem__(field), self)
+        return [self._sdf.__getitem__(field)
                 for field in self._metadata.index_fields]
 
     def _reduce_for_stat_function(self, sfun):
-        df = self._sdf.select([sfun(self[col]).alias(col) for col in self.columns])
-        pdf = df.toPandas()
-        assert len(pdf) == 1, (df, pdf)
+        sdf = self._sdf.select([sfun(self._sdf[col]).alias(col) for col in self.columns])
+        pdf = sdf.toPandas()
+        assert len(pdf) == 1, (sdf, pdf)
         row = pdf.iloc[0]
         row.name = None
         return row  # Return first row as a Series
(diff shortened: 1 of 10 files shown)


databricks/koalas/series.py
```diff
@@ -17,33 +17,80 @@
 """
 A base classes to be monkey-patched to Column to behave similar to pandas Series.
```
rxin: need to update this

rxin:
Once you do this, we will lose all the existing Spark methods like join, select, filter right? Can you creat ea list of the listing methods? We'd need to quickly implement them.

rxin:
Can you guys create a separate branch for this? Once this is merged, a lot of stuff will break (all existing Spark methods will be gone). I think we should create a new branch and then once we finish most of the basic stuff, merge that branch into master.

Creating a new branch and submitting PRs against that branch will allow us to still do proper code reviews.

databricks/koalas/utils.py
```diff
@@ -22,9 +22,6 @@
 import types
```
HyukjinKwon: @ueshin, I thought we can remove this entire file. Does that require some more works?

databricks/koalas/frame.py
```diff
@@ -15,55 +15,65 @@
...
+class DataFrame(_Frame, _MissingPandasLikeDataFrame):
+
+    @derived_from(pd.DataFrame)
+    @dispatch_on('data')
+    def __init__(self, data=None, index=None, columns=None, dtype=None, copy=False):
+        pdf = pd.DataFrame(data=data, index=index, columns=columns, dtype=dtype, copy=copy)
+        self._init_from_pandas(pdf)
+
+    @__init__.register(pd.DataFrame)
+    def _init_from_pandas(self, pdf, *args):
```
HyukjinKwon: Looks we don't need `*args`. `_init_from_spark` too.

databricks/koalas/frame.py
```diff
@@ -15,55 +15,65 @@
...
+        metadata = Metadata.from_pandas(pdf)
+        reset_index = pdf.reset_index()
+        reset_index.columns = metadata.all_fields
+        self._init_from_spark(default_session().createDataFrame(reset_index), metadata)
+
+    @__init__.register(spark.DataFrame)
+    def _init_from_spark(self, sdf, metadata=None, *args):
+        self._sdf = sdf
+        self._pandas_metadata = metadata
```
HyukjinKwon: How about .. 

```diff
diff --git a/databricks/koalas/frame.py b/databricks/koalas/frame.py
index b5fb148..2e1bd65 100644
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -57,12 +57,13 @@ class DataFrame(_Frame, _MissingPandasLikeDataFrame):
     @__init__.register(spark.DataFrame)
     def _init_from_spark(self, sdf, metadata=None, *args):
         self._sdf = sdf
-        self._pandas_metadata = metadata
+        if metadata is None:
+            self._pandas_metadata = Metadata(column_fields=self._sdf.schema.fieldNames())
+        else:
+            self._pandas_metadata = metadata

     @property
     def _metadata(self):
-        if self._pandas_metadata is None:
-            self._pandas_metadata = Metadata(column_fields=self._sdf.schema.fieldNames())
         return self._pandas_metadata

     @property
```

We could even remove `_metadata` by renaming `_pandas_metadata` to that but I guess we can do this cleanup separately too.

Seems like it's a bug if `_pandas_metadata` becomes `None` somehow.

databricks/koalas/testing/utils.py
```diff
@@ -22,10 +22,13 @@
 from contextlib import contextmanager
 
 import pandas as pd
+from pyspark import sql as spark
```
HyukjinKwon: Seems unused :).

HyukjinKwon:
Actually, let me make a separate branch and make a followup by myself.

HyukjinKwon:
I made a branch - https://github.com/databricks/koalas/commits/refactoring
I will address my own comments and fix a bit further.

databricks/koalas/frame.py
ueshin: Actually it's a convention of `@dispatch_on`. Seems like it invokes the registered method with the same number of arguments as the original method.

ueshin:
@HyukjinKwon Thanks for taking this over! I'll wait for the PR for `refactoring` branch.

PR #122, opened 2019-04-21
Rename GitHub repo from spark-pandas to koalas

rxin:


rxin:
Thanks - I fixed those two too.

Issue #124, opened 2019-04-22
Basic functionalities to support


rxin:
I quickly went through Pandas and Spark DataFrame docs and came up with a list of basic functions to add:

The following is a list of methods that we should make available in koalas.DataFrame and koalas.Series.

(In general, ignore "axis" argument for now, and assume axis = 0, i.e. row wise)

koalas.DataFrame/Series:

- to_pandas
- spark_schema
- select
- selectExpr

- describe
- dtypes
- explain
- head

- agg
- aggregate
- groupby

- isna
- dropna
- fillna
- notna

- abs
- all
- any
- max
- min
- mean
- kurt
- std
- skew
- sum
- var

- sub
- add
- mul
- div
- mod
- floordiv
- pow
- lt
- gt
- le
- ge
- ne
- eq

- append
- replace

- sample
- sort_values (done in #204)
- sort_index
- nlargest (#302)
- nsmallest (#302)

koalas.Series:

- unique  -> koalas.DataFrame
- nunique  -> same as Pandas
- is_unique
- value_counts

- str.upper
- str.lower
- str ...

koalas.DataFrame:

- append
- join
- merge

HyukjinKwon:
@rxin, we have implemented everything except two:

- select
- selectExpr

*`spark_schema` is partially implemented at Series instead as `spark_type`.

For `select`, it conflicts with [pandas' `select`](https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.DataFrame.select.html) so it might be better to don't implement for now.

`selectExpr` case is a bit arguable too because we have the external and internal column names are different. For example, the column name can be a tuple `("a", "b", "c")`; however, that's not allowed in Spark side.

Let me close this one for now but let me know if you have a different thought.

PR #125, opened 2019-04-22
Restores APIs that exists in both Pandas and Spark DataFrame

HyukjinKwon:
This PR partially addresses https://github.com/databricks/koalas/issues/124

This PR restores Spark DataFrame APIs missing by the refactoring:

- `schema`

- `to_pandas`

- `select`

- `selectExpr `

- `describe`

- `dtypes`

- `explain`

- `agg`, `aggregate`

- `fillna`

- `subtract`, `sub`

- `replace`

- `sample`

- `join`

Note that it looks other methods in https://github.com/databricks/koalas/issues/124 should newly be implemented.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -457,16 +457,6 @@ class DataFrame(_Frame, _MissingPandasLikeDataFrame):
         return DataFrame(spark.DataFrame.fillna(self._sdf, value, subset))
 
     @derived_from(spark.DataFrame)
-    def subtract(self, other):
-        return DataFrame(spark.DataFrame.subtract(self._sdf, other))
-
-    sub = subtract
-
-    @derived_from(spark.DataFrame)
-    def subtract(self, other):
-        return DataFrame(spark.DataFrame.subtract(self._sdf, other))
-
-    @derived_from(spark.DataFrame)
     def replace(self, to_replace, value=_NoValue, subset=None):
         return DataFrame(spark.DataFrame.replace(self._sdf, to_replace, value, subset))
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -22,7 +22,7 @@ from functools import reduce
 
 import numpy as np
 import pandas as pd
-from pyspark import sql as spark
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -420,6 +422,62 @@ def pipe(self, func, *args, **kwargs):
     def shape(self):
         return len(self), len(self.columns)
 
+    @property
+    def schema(self):
+        return spark.DataFrame.schema.fget(self._sdf)
+
+    @derived_from(spark.DataFrame)
+    def select(self, *cols):
+        return DataFrame(spark.DataFrame.select(self._sdf, *cols))
```
rxin: should we keep indexes?

databricks/koalas/frame.py
```diff
@@ -420,6 +422,62 @@ def pipe(self, func, *args, **kwargs):
...
+
+    @derived_from(spark.DataFrame)
+    def selectExpr(self, *cols):
+        return DataFrame(spark.DataFrame.selectExpr(self._sdf, *cols))
```
rxin: same as above -- should we keep indexes

databricks/koalas/frame.py
```diff
@@ -420,6 +422,62 @@ def pipe(self, func, *args, **kwargs):
...
+        return DataFrame(spark.DataFrame.agg(self._sdf, *exprs))
+
+    aggregate = agg
+
+    @derived_from(spark.DataFrame)
+    def fillna(self, value, subset=None):
+        return DataFrame(spark.DataFrame.fillna(self._sdf, value, subset))
+
+    @derived_from(spark.DataFrame)
+    def subtract(self, other):
```
rxin: sry when i mention sub i was referring to subtracting one column from another, i.e. arithmetic operation -.

HyukjinKwon:
Actually, even before refactoring, Spark DataFrame APIs were not supported:

```
databricks.koalas.exceptions.PandasNotImplementedError: The method `pd.DataFrame.select()` is not implemented yet.
```

Let me do this after finishing the refactoring.

Issue #118, opened 2019-04-20
Design of lazy evaluation for some properties


garawalid:
#### The idea behind this issue is to discuss the design of lazy evaluation for properties. 

I think we need this behavior especially with a property like [values](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.values.html).  We will load all the data in `values` when we create a new DataFrame and this becomes heavy when the data is huge. 

I found a good trick [here](https://stackoverflow.com/questions/20535342/lazy-evaluation-in-python) and I think it can be useful. We just need to override `__get__()`. 

Any idea related to the topic and we can use it in Koalas?

For instance, we can solve `values` by replacing it by `to_numpy()`. 
Pandas documentation suggests.
```
Warning: We recommend using DataFrame.to_numpy() instead.
 ```

**Comment by rxin** [comment]
Does Python automatically materialize all the properties when the class is created? If yes, I definitely agree we shouldn't create "values" ...

**Comment by garawalid** [comment]
Yes, it does. 
Here is an example. I implemented `values` as property in `_Frame` class. 
```python 
    @property
    def values(self):
        return self.toPandas().values
```
Running this code will show us that the property `values` contains already the data as numpy array.
 ```python 
df = pd.DataFrame({'a': [4, 2, 3, 4, 8, 6],
                    'b': [1, 2, 9, 4, 2, 4],
                    'c': ["one", "three", "six", "seven", "one", "5"]},
                    index=[10, 20, 30, 40, 50, 60])

ddf = self.spark.from_pandas(df)
```

A screen capture from the debugger stopped after creating `ddf`

![values](https://user-images.githubusercontent.com/20029252/56464049-17d2e780-63e1-11e9-9add-2e62c7662fd2.png)

**Comment by AbdealiJK** [comment]
A debugging tool would typically get all attributes in dir() and request for every attribute. Checking it in a debugger may not be an accurate test

**Comment by garawalid** [comment]
@AbdealiJK 
Thanks! You are right, it's not an accurate test. 
I tried it with this simple test, and the message doesn't appear when we create the dataframe. 

```python
@property
def values(self):
    print("executed !")
    return self.toPandas().values
```

garawalid:
So, we can say that properties are lazy by default!

rxin:
Thanks. Looks like we are safe here. I'm going to close this ticket. Feel free to reopen if there is an issue in the future.

PR #130, opened 2019-04-22
add to_numpy()

garawalid:
This PR contains `values` property and `get_values()` method. 
The lazy evaluation tested in this issue #118

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -30,12 +30,7 @@ class _Frame(object):
     The base class for both dataframes and series.
     """
 
-    @property
-    def values(self):
-        return self.toPandas().values
-
-    @derived_from(pd.DataFrame)
-    def get_values(self):
+    def to_numpy(self):
         """
         A NumPy ndarray representing the values in this DataFrame
         :return: numpy.ndarray
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -30,6 +30,19 @@ class _Frame(object):
     The base class for both dataframes and series.
     """
 
+    @property
+    def values(self):
+        return self.toPandas().values
+
+    @derived_from(pd.DataFrame)
+    def get_values(self):
+        """
+        A NumPy ndarray representing the values in this DataFrame
+        :return: numpy.ndarray
+                 Numpy representation of DataFrame
+        """
+        return self.toPandas().values
+
     @derived_from(pd.DataFrame, ua_args=['axis', 'skipna', 'level', 'numeric_only'])
     def mean(self):
         return self._reduce_for_stat_function(F.mean)
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -35,6 +35,9 @@ class _Frame(object):
         A NumPy ndarray representing the values in this DataFrame
         :return: numpy.ndarray
                  Numpy representation of DataFrame
+
+        .. note:: This method should only be used if the resulting NumPy ndarray is expected
+            to be small, as all the data is loaded into the driver's memory.
         """
         return self.toPandas().values
 
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -371,6 +371,35 @@ class DataFrameTest(ReusedSQLTestCase, TestUtils):
                                         "Series.*{}.*not implemented".format(name)):
                 getattr(d.a, name)()
 
+    def test_values(self):
+        df = pd.DataFrame({'a': [4, 2, 3, 4, 8, 6],
+                           'b': [1, 2, 9, 4, 2, 4],
+                           'c': ["one", "three", "six", "seven", "one", "5"]},
+                          index=[10, 20, 30, 40, 50, 60])
+
+        ddf = koalas.from_pandas(df)
+
+        np.testing.assert_equal(ddf.values, df.values)
+
+        s = pd.Series([1, 2, 3, 4, 5, 6, 7], name='x')
+
+        ddf = koalas.from_pandas(s)
+        np.testing.assert_equal(ddf.values, s.values)
+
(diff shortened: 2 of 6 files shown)


garawalid:
@rxin 

> I'm not sure if we should put values and get_values there, because it might blow up memory. One way we can do this is for each function that requires collecting all the data to a single node, we require users to run toPandas explicitly, so they would know what they are doing and not shoot themselves in the foot.

I absolutely agree with you, collecting the data in the driver in the case of big data is tricky.

But, this will happen in any method that uses `toPandas()`. I think `values` and `get_values()` are wrapper method for `toPandas()`, so we will end up in the same problem. 

I suggest that we add a warning sentence in the documentation to remind users that this method uses `toPandas()` and it can blow up memory.

thunterdb:
@garawalid I started a quick discussions about pros and cons of this API in #146

thunterdb:
@garawalid based on the discussion in #146, I think that we just need to rename `get_values` to `to_numpy`, and to not include `.values`?

Also, regarding documentation: can you copy/paste the pandas documentation and add a warning that this will put all the data in memory, see the warning in https://spark.apache.org/docs/latest/api/python/pyspark.sql.html#pyspark.sql.DataFrame.toPandas

databricks/koalas/tests/test_dataframe.py
```diff
@@ -371,10 +371,26 @@ def test_missing(self):
...
 
+    def test_to_numpy(self):
+        df = pd.DataFrame({'a': [4, 2, 3, 4, 8, 6],
+                           'b': [1, 2, 9, 4, 2, 4],
+                           'c': ["one", "three", "six", "seven", "one", "5"]},
+                          index=[10, 20, 30, 40, 50, 60])
+
+        ddf = koalas.from_pandas(df)
+
+        np.testing.assert_equal(ddf.to_numpy(), df.values)
```
garawalid: Since Pandas=23.4 doesn't have `to_numpy()` method. I used `values` property instead to pass the test.

databricks/koalas/generic.py
```diff
@@ -30,6 +30,17 @@ class _Frame(object):
...
+    def to_numpy(self):
+        """
+        A NumPy ndarray representing the values in this DataFrame
+        :return: numpy.ndarray
+                 Numpy representation of DataFrame
+
+        .. note:: This method should only be used if the resulting NumPy ndarray is expected
+            to be small, as all the data is loaded into the driver's memory.
+        """
+        return self.toPandas().values
```
garawalid: The same here,  Pandas=0.23  doesn't have `to_numpy()`

garawalid:
@thunterdb 
What do you think about using `values` inside `to_numpy()` ?

docs/source/reference/series.rst
```diff
@@ -70,3 +70,10 @@ Missing data handling
    Series.isna
    Series.notna
    Series.dropna
+
+Serialization / IO / Conversion
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. autosummary::
+   :toctree: api/
+
+   DataFrame.to_numpy
```
rxin: Series.to_numpy?

Also Pandas have this in the "Conversion" section. Maybe we should just follow that.

garawalid: Series has [to_numpy()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.to_numpy.html)
I agree with you, we can follow Pandas documentation!

rxin: But you are adding DataFrame.to_numpy doc in series.rst?

thunterdb: @garawalid I think you meant `Series.to_numpy` here

garawalid: You are right, it should be Series.to_numpy

docs/source/reference/series.rst
```diff
@@ -70,3 +70,10 @@ Missing data handling
    Series.isna
    Series.notna
    Series.dropna
+
+Serialization / IO / Conversion
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. autosummary::
+   :toctree: api/
+
+   Series.to_numpy
```
ueshin: Seems like pandas' doc has `to_num` in "Conversion" section above in this file.
Shall we move this up there?

docs/source/reference/frame.rst
```diff
@@ -113,3 +113,4 @@ Serialization / IO / Conversion
    :toctree: api/
 
    DataFrame.to_html
+   DataFrame.to_numpy
```
ueshin: Although pandas doesn't have `to_numpy` in `DataFrame`, I think we should have this in the same section as our `Series`, "Conversion" section.

docs/source/reference/series.rst
garawalid: I think it's better we keep it in `Serialization / IO / Conversion` because the `Conversion` section contains methods used to clean data like `isna()` or make a copy `copy()`. Somehow they are not relevant to `to_numpy()`.

In `Conversion` section, we have methods used to convert the DataFrame/Series like `DataFrame.to_parquet()`, `DataFrame.to_json()`

@ueshin  what do you think ?

ueshin: I'd prefer to follow pandas' way, but not a strong opinion. cc @thunterdb

thunterdb: I think that either points are valid and I do not have a strong opinion because the name of the method is so explicit and because it is very easy to change later. Since the pull request is structured as it is currently, I would be of the opinion that we merge it as it is. Also, nothing prevents us from adding the reference in both places. @ueshin I will let you decide.

ueshin: then I'd merge this now. thanks for the opinion!

ueshin:
Thanks! merging.

PR #133, opened 2019-04-22
Implement bigint dividing by timedelta64

rxin:
This patch adds support for dividing bigints by timedelta64. We automatically assume the integral value is the number of seconds for a time interval. This would make an upcoming blog post work.

However, after this I think we should look at holistically how we want to handle date/time and intervals.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -19,11 +19,12 @@ A wrapper class for Spark Column to behave similar to pandas Series.
 """
 from decorator import decorator, dispatch_on
 
+import numpy as np
 import pandas as pd
 
 from pyspark import sql as spark
 from pyspark.sql import functions as F
-from pyspark.sql.types import FloatType, DoubleType, StructType, TimestampType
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -101,11 +102,21 @@ def __sub__(self, other):
...
-        return _column_op(spark.Column.__sub__)(self, other)
+            return self.astype('bigint') - other.astype('bigint')
+        else:
+            # Fallback to Spark's division
+            return _column_op(spark.Column.__sub__)(self, other)
+
+    def __div__(self, other):
+        if isinstance(self.spark_type, LongType) and isinstance(other, np.timedelta64):
+            # Convert timedelta64[ns] to number of seconds.
+            num_secs = other / np.timedelta64(1, 's')
```
rxin: this needs to be `(other / np.timedelta64(1, 's')).astype(int)`

HyukjinKwon:
Yup, let me track this in https://github.com/databricks/koalas/pull/136

PR #136, opened 2019-04-23
 Support NumPy timedelta64 in Series div operations

HyukjinKwon:


databricks/koalas/series.py
```diff
@@ -53,6 +54,12 @@ def _column_op(f, self, *args):
     # To cover this case, explicitly check if the argument is Koalas Series and
     # extract Spark Column. For other arguments, they are used as are.
     args = [arg._scol if isinstance(arg, Series) else arg for arg in args]
+
+    # PySpark does not support NumPy type out of the box. For now, we convert NumPy types
+    # into some primitive types understandable in PySpark.
+    args = [float(arg / np.timedelta64(1, 's')) if isinstance(arg, np.timedelta64)
```
rxin: this might be problematic in the future for other operations that we don't intend to, e.g. timestamp + int might throw a weird error.

I'd for now patch the specific operations we have, rather than just making it work on everything.

HyukjinKwon: okie, that works too.

rxin:
LGTM

PR #138, opened 2019-04-23
Restructure documentation and exclude stub methods

### LINKED ISSUE #137: Hide stub methods from auto complete suggestions and docs
After looking at the generated docs and trying to use Koalas (with the refactoring branch), I'd definitely hide the (unimplemented) stub methods from docs and auto complete.

If it's difficult to do, perhaps we can still define them in the existing files, but don't mix them into the real exposed interfaces. Instead, we write a test case that'd fail if a method is defined in both the DataFrame class and the _MissingPandasLikeDataFrame class.

HyukjinKwon:
This PR proposes to restructure documentation and exclude stub methods.

1. Clarify what to expose as APIs in `__all__` in Koalas

2. Make the doctest roughly compatible to Koalas.
    For example, see Screen Shot 2019-04-23 at 10 58 34 AM

3. Remove all stub methods by only documenting method that has docstring. Namely docstrings in stub were removed.

4. Restructure:

    ```
    Koalas  # koalas.* APIs
    ├── DataFrame
    ├── Exceptions
    └── Series
    ```

Closes #137

### DIFF
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -36,7 +36,9 @@ def assert_pyspark_version():
 assert_pyspark_version()
 
 from databricks.koalas.namespace import *
+from databricks.koalas.frame import DataFrame
+from databricks.koalas.series import Series
 from databricks.koalas.typing import Col, pandas_wrap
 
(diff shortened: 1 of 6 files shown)


HyukjinKwon:
cc @rxin

HyukjinKwon:
Thanks, @ueshin

PR #142, opened 2019-04-23
Encapsulation refactoring

HyukjinKwon:


HyukjinKwon:
@rxin, @thunterdb, actually, I would like to avoid squash the commits since each commit logically address each issue. i will address @thunterdb's comment separately in a separate PR, and push those 7 commits into master.

PR #134, opened 2019-04-22
Encapsulation refactoring

rxin:

PR #143, opened 2019-04-23
Add some documentation, todo, make databricks.koalas.utils

HyukjinKwon:
This PR mainly address the comments in https://github.com/databricks/koalas/pull/134

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -65,6 +65,7 @@ def _numpy_column_op(f, self, *args):
     # into some primitive types understandable in PySpark.
     new_args = []
     for arg in args:
+        # TODO: This is a quick hack to support NumPy type. We should revisit this.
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
cc @thunterdb

databricks/koalas/series.py
```diff
@@ -354,9 +364,9 @@ def _unpack_scalar(sdf):
     Takes a dataframe that is supposed to contain a single row with a single scalar value,
     and returns this value.
     """
-    l = sdf.head(2)
-    assert len(l) == 1, (sdf, l)
-    row = l[0]
+    ret = sdf.head(1)
```
ueshin: Actually I thought this was because we wanted to check the sdf returns exactly one row.

Btw, we have one more similar place in `frame.py`.

HyukjinKwon: Hmm .. let me revert this change back for now.

PR #147, opened 2019-04-23
Add readthedocs configuration file

HyukjinKwon:
This PR proposes to add a readthedocs configuration file. (See https://docs.readthedocs.io/en/stable/config-file/v2.html)

I experimentally tested this after creating `stable` branch for testing purpose (and I removed it back).
The documentation will be generated as below:

See https://databricks-koalas.readthedocs.io/en/stable/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,39 @@
+#
+# Copyright (C) 2019 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -0,0 +1,39 @@
...
+
+# Optionally set the version of Python and requirements required to build your docs
+python:
+  version: 3.5
+  install:
+    - requirements: requirements-dev.txt
+    - method: pip
+      path: .
+      extra_requirements:
+        - spark
```
HyukjinKwon: This refers below ..

https://github.com/databricks/koalas/blob/master/setup.py#L53

rxin:
LGTM. Merging.

PR #149, opened 2019-04-23
Add corr().

ueshin:


databricks/koalas/tests/test_dataframe.py
```diff
@@ -345,6 +345,51 @@ def test_abs(self):
         self.assert_eq(ddf[['B', 'C']].abs(), df[['B', 'C']].abs())
         # self.assert_eq(ddf.select('A', 'B').abs(), df[['A', 'B']].abs())
 
+    def test_corr(self):
```
rxin: i'd create a new file for these functions. in the future we can parallelize these tests and if they are different files it's easier to do that.

databricks/koalas/missing/series.py
```diff
@@ -55,7 +55,6 @@ class _MissingPandasLikeSeries(object):
     compress = unsupported_function('compress')
     convert_objects = unsupported_function('convert_objects')
     copy = unsupported_function('copy')
-    corr = unsupported_function('corr')
```
rxin: how is this change adding corr to Series? Do all the methods that are added to Frame automatically get added to Series?

databricks/koalas/ml.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def corr(kdf, method='pearson'):
+    """
+    The correlation matrix of all the numerical columns of this dataframe.
+
+    Only accepts scalar numerical values for now.
+    :param kdf: the koalas dataframe.
+    :param method:
+    :return:
```
rxin: document return type

databricks/koalas/missing/series.py
thunterdb: Yes, Frame is the base class for both `Series` and `DataFrame`. In practice, though, the implementations are usually different enough that functions are added in each of `frame.py` and `series.py`.

databricks/koalas/ml.py
```diff
@@ -0,0 +1,64 @@
...
+    :return: a pair of dataframe, list of strings (the name of the columns
+             that were converted to numerical types)
+    """
+    # TODO, it should be more robust.
+    accepted_types = {np.dtype(dt) for dt in [np.int8, np.int16, np.int32, np.int64,
+                                              np.float32, np.float64, np.bool_]}
+    numeric_fields = [fname for fname in kdf._metadata.column_fields
+                      if kdf[fname].dtype in accepted_types]
+    numeric_df = kdf._sdf.select(*numeric_fields)
+    va = VectorAssembler(inputCols=numeric_fields, outputCol="_1")
```
ueshin: @thunterdb Is it okay to remove `handleInvalid` parameter here? Seems like it does not exist in pyspark 2.3.

thunterdb:
Yes that sounds good to me. We should just make clear in the doc that NaNs
will give invalid results.

ueshin:
@thunterdb I added a notes at the bottom of the doc. Is it enough?

thunterdb:
@ueshin this is great, I am merging.

ueshin:
@thunterdb Thanks for a lot of help!

PR #163, opened 2019-04-24
Use `import databricks.koalas as ks` style.

ueshin:
This is a follow-up of #160.


HyukjinKwon:
Looks good - I have no preference either way.

thunterdb:
LGTM. Thanks.

Issue #162, opened 2019-04-24
Enable static type checking in the tests


thunterdb:
The best type checker these days seems to be `mypy`: http://mypy-lang.org/

It is fully compliant with PEP-484. We should enable it for all the commits before we start adding type hints.

The goal of this task is the following:
 - enable mypy on Travis runs
 - making sure that the spark typing hints are installed. I am not quite sure what they are, but PyCharm asks to also install them as an extra package, so it may be something to check
 - resolve the existing errors, if any.

thunterdb:
My mistake, closing the issue. Further work is in #166

PR #165, opened 2019-04-24
Add Mypy type hint linter in Koalas lint check

HyukjinKwon:
This PR adds Mypy type hint linter in our tests and Python linter process.

Closes https://github.com/databricks/koalas/issues/162

### DIFF
--- a/databricks/__init__.py
+++ b/databricks/__init__.py
@@ -15,4 +15,4 @@
 #
 
 # https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages
-__path__ = __import__('pkgutil').extend_path(__path__, __name__)
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)  # type: ignore
--- a/databricks/koalas/dask/utils.py
+++ b/databricks/koalas/dask/utils.py
@@ -1,56 +1,6 @@
 import functools
 from databricks.koalas.dask.compatibility import get_named_args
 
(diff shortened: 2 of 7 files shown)


databricks/koalas/dask/utils.py
```diff
@@ -1,56 +1,6 @@
 import functools
 from databricks.koalas.dask.compatibility import get_named_args
 
-_method_cache = {}
-
-
-class methodcaller(object):
```
HyukjinKwon: I think these are not used. mypy complained so I just removed.

databricks/__init__.py
```diff
@@ -15,4 +15,4 @@
 #
 
 # https://packaging.python.org/guides/packaging-namespace-packages/#pkgutil-style-namespace-packages
-__path__ = __import__('pkgutil').extend_path(__path__, __name__)
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)  # type: ignore
```
thunterdb: Yes, not much we can do here.

databricks/koalas/dask/utils.py
thunterdb: That sounds great, thanks for the cleanup.

thunterdb:
As discussed outside this thread:
 - strict mode is too strict for now
 - we should eventually add `spark-stubs` for more precise typing. This is to be done in subsequent work.

Merging the current work. Thanks!

PR #170, opened 2019-04-24
Reference PyPi instead of pip

gstaubli:
The installation on Databricks is handled through installing the library from PyPi, not pip


rxin:
Thanks! Merging in.

PR #172, opened 2019-04-24
[168] Enables doctest tests

### LINKED ISSUE #168: Enable doctest
We currently disable all the doctests. It'd be great to be able to run them. The main blocker right now is we call "ks" directly, so we need to figure out a way to automatically import koalas as ks in the namespace.

thunterdb:
Fixes #168 .

This PR does the minimal amount of changes to enable doctests in `frame.py`, and could be expanded to support other modules.

cc @ueshin


databricks/koalas/frame.py
```diff
@@ -542,6 +542,9 @@ def _pd_getitem(self, key):
             return DataFrame(self._sdf.filter(bcol), self._metadata.copy())
         raise NotImplementedError(key)
 
+    def __repr__(self):
```
thunterdb: this is how doctest gets the results back, it should be added for Series too.

databricks/koalas/testing/doctest_main.py
```diff
@@ -0,0 +1,6 @@
+
+if __name__ == "__main__":
+    import doctest
+    import databricks.koalas as ks
+    from databricks.koalas import frame, series
+    doctest.testmod(frame, extraglobs={"ks": ks})
```
thunterdb: This line should be repeated for every module that needs testing. One cannot rely on nosetest to automatically discover all the modules. On the bright side, there is no need to change any of the other files.

databricks/koalas/testing/doctest_main.py
```diff
@@ -0,0 +1,9 @@
+
```
HyukjinKwon: license :).

databricks/koalas/testing/doctest_main.py
```diff
@@ -0,0 +1,9 @@
...
+
+if __name__ == "__main__":
+    import doctest
+    import databricks.koalas as ks
+    from databricks.koalas import frame, series, namespace
+    modules = [frame, series, namespace]
+    for m in modules:
+        doctest.testmod(m, extraglobs={"ks": ks})
```
HyukjinKwon: This is fine too, for now.

One advantage of following PySpark one (which script has its own main that runs doctests) is, we can just run that the script we're fixing, for instance, in PyCharm. In this way, it's tricky to conditionally run each.

The other way might be to switch to pytest (which pandas uses), with [`conftest.py`](https://docs.pytest.org/en/latest/doctest.html#the-doctest-namespace-fixture).

HyukjinKwon:
Let me merge this one first. We can make a followup.

PR #177, opened 2019-04-25
Use pyarrow to infer data type.

### LINKED ISSUE #167: Nullable string/object columns not supported
When trying to load a pandas DataFrame with a nullable string column, I always get 

```
TypeError: field header: Can not merge type <class 'pyspark.sql.types.StringType'> and <class 'pyspark.sql.types.DoubleType'>
```

It can be reproduced in two different ways. We can create an in-memory Series and load it:

```
import numpy as np
import pandas as pd

ks.from_pandas(pd.Series(['foo', np.nan, 'bar']))
```

Or we can populate a CSV file with at least one NULL:

```
header
value
N/A
nothing above
```

And then try to read it:

```
import pandas as pd

ks.from_pandas(pd.read_csv('foo.csv'))
```

Either way, koalas things that the `np.nan` signifies a float, but it's actually a NULL for an object column.

ueshin:
PySpark without arrow execution sometimes fails to infer data type.
This PR uses pyarrow infer prior to create DataFrame.

Closes #167.

### DIFF
--- a/databricks/koalas/typedef.py
+++ b/databricks/koalas/typedef.py
@@ -22,6 +22,7 @@ import typing
 from decorator import decorate
 from decorator import getfullargspec
 import numpy as np
+import pandas as pd
 from pandas.api.types import is_datetime64_dtype, is_datetime64tz_dtype
 import pyarrow as pa
 from pyspark.sql import Column
@@ -135,7 +136,7 @@ def as_python_type(spark_tpe):
     return _py_conversions.get(spark_tpe, None)
 
 
-def infer_pd_series_spark_type(s):
+def infer_pd_series_spark_type(s: pd.Series) -> types.DataType:
     """Infer Spark DataType from pandas Series dtype.
 
     :param s: :class:`pandas.Series` to be inferred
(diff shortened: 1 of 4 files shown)


databricks/koalas/typedef.py
```diff
@@ -133,6 +136,23 @@ def as_python_type(spark_tpe):
...
+    :param s: :class:`pandas.Series` to be inferred
+    :return: the inferred Spark data type
+    """
+    dt = s.dtype
+    if dt == np.dtype('object'):
+        if len(s) == 0 or s.isnull().all():
+            raise ValueError("can not infer schema from empty or null dataset")
+        return types.from_arrow_type(pa.Array.from_pandas(s).type)
+    elif is_datetime64_dtype(dt) or is_datetime64tz_dtype(dt):
+        return types.TimestampType()
```
HyukjinKwon: @ueshin, sorry if I missed something but why do we need this manual map?

ueshin: Seems like `pa.from_numpy_dtype(dt)` does not work for timestamp types.

HyukjinKwon: Ah, given from my rough test, it worked with arrow 0.12. Maybe we can remove it after the upgrade.

PR #183, opened 2019-04-26
Use pytest instead of nose to run tests

HyukjinKwon:
This PR propose to use `pytest` instead of `nose` for the reasons below:

- We can remove the manual doctest / unitests running and integrate it with one testing framework `pytest`

- It's more popular in terms of download stats (PyPi) and Github stars
  - https://pypistats.org/packages/nose vs https://pypistats.org/packages/pytest
  - https://github.com/nose-devs/nose vs https://github.com/pytest-dev/pytest

- It give a better output with a little bit of more info and colored text:

    `nose`:

    ```
    ...
    test_read_csv (databricks.koalas.tests.test_csv.CsvTest) ... ok
    test_read_csv_with_comment (databricks.koalas.tests.test_csv.CsvTest) ... ok
    test_read_csv_with_mangle_dupe_cols (databricks.koalas.tests.test_csv.CsvTest) ... ok
    ...
    ```

    `pytest`:

    ```
    ...
    databricks/koalas/frame.py::databricks.koalas.frame.DataFrame.count PASSED    [  3%] 
    databricks/koalas/frame.py::databricks.koalas.frame.DataFrame.drop SKIPPED    [  4%]
    databricks/koalas/frame.py::databricks.koalas.frame.DataFrame.dropna SKIPPED  [  5%]
    ...
    ```


databricks/koalas/testing/utils.py
```diff
@@ -53,6 +55,8 @@ def conf(cls):
 
     @classmethod
     def setUpClass(cls):
+        if SparkContext._active_spark_context is not None:
+            SparkContext._active_spark_context.stop()
```
HyukjinKwon: Seems multiple contexts can be initiated when running doctests first.
I think It does not happen in PySpark testing because it runs them in each process.

rxin:
Can we now run individual files for doctests? If yes, I'd update README.md's testing section to give an example command (on how to run that).

HyukjinKwon:
Let me check and update.

rxin:
I definitely like this test output better.

dev/run-tests.sh
```diff
@@ -29,9 +31,11 @@ FWDIR="$( cd "$DIR"/.. && pwd )"
...
+    ARGS="--verbose --showlocals --doctest-modules databricks"
 else
     ARGS="$@"
 fi
-# Runs the main program that finds and runs the docstrings.
-python databricks/koalas/testing/doctest_main.py && nosetests $ARGS --where "$FWDIR"
+
+pushd "$FWDIR" > /dev/null
+pytest $ARGS
+popd > /dev/null
```
thunterdb: why do you need that?

HyukjinKwon: Maybe we can add some more commands in this script after this, for instance, coverage report. So, I came back to the original location although it might not affect anything.

(honestly, I just copied and pasted this pattern from Spark's scripts .. and it made me feel complete when something goes in and come back .. :D..)

README.md
```diff
@@ -100,21 +100,34 @@ conda activate koalas-dev-env
...
-./dev/run-tests.sh
+# Run all unittest and doctest
+./dev/pytest
 ```
 
 To run a specific test file:
 ```bash
-python databricks/koalas/tests/test_dataframe.py
+# Run unittest
+./dev/pytest -k test_dataframe.py
```
HyukjinKwon: Strictly, there look other ways to specify as below:

```
pytest databricks/koalas/tests/test_dataframe.py
```

but I think it's easier for us just to stick to `-k` expression for both doctest and unittest. It's even shorter.

I also renamed `run-tests.sh` to `pytest` as it is exactly same as `pytest` but only have some configuration and/or default settings.

HyukjinKwon: I think in this way, we can just forget about documentation about how to test but we can just redirect the details into `pytest`.

HyukjinKwon:
I believe this is ready for a look.

dev/pytest
```diff
@@ -15,9 +15,15 @@
...
+set -o pipefail
+set -e
 
-# Runs both doctests and unit tests by default, otherwise hands arguments over to nose.
+if ! hash pytest 2> /dev/null; then
+    echo "The pytest command was not found. Please install 'pytest' Python package."
+    exit 1
+fi
 
+# Runs both doctests and unit tests by default, otherwise hands arguments over to pytest.
```
ueshin: nit: this comment should be at the previous line of the current line 37 (`if [ "$#" = 0 ]; then`)?

README.md
```diff
@@ -100,21 +100,34 @@ conda activate koalas-dev-env
...
 
-To run a specific test method:
+To run a specific doctest/unittest:
 ```bash
-python databricks/koalas/tests/test_dataframe.py DataFrameTest.test_Dataframe
+# Run unittest
+./dev/pytest -k "DataFrameTest and test_Dataframe"
+
+# Run doctest
+./dev/pytest -k DataFrame.corr --doctest-modules databricks
```
ueshin: super nit: an extra space at the end of the line.

HyukjinKwon:
Will address them thanks!

README.md
HyukjinKwon: How did you find this :D?

HyukjinKwon:
Thanks all for review. I am pretty confident of this change :D. merged.

README.md
ueshin: haha, just by accident. :)

PR #188, opened 2019-04-26
DataFrame.count()

### LINKED ISSUE #184: DataFrame.count()
DataFrame.count() currently returns a single number, i.e. the number of rows, which is Spark's behavior and doesn't even match the doctest's behavior. (The doctest passes because we are still using pd.DataFrame rather than testing ks.DataFrame.)

We should fix it to return a new DataFrame consistent with the doctest.

```

**********************************************************************
File "/Users/rxin/workspace/koalas/databricks/koalas/frame.py", line 530, in count
Failed example:
    df.count()
Expected:
    Person    5
    Age       4
    Single    5
    dtype: int64
Got:
    5

```

rxin:
This patch implements DataFrame.count based on Pandas' syntax. In this PR, DataFrame.count returns a pd.Series, rather than a ks.Series.

In the future, we should look into what we do with Inf and -Inf values, and Series.count.

Resolves #184

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -86,6 +86,13 @@ class DataFrame(_Frame):
                 for field in self._metadata.index_fields]
 
     def _reduce_for_stat_function(self, sfun):
+        """
+        Applies sfun to each column and returns a pd.Series where the number of rows equal the
+        number of columns.
+
+        :param sfun: either an 1-arg function that takes a Column and returns a Column, or
+        a 2-arg function that takes a Column and its DataType and returns a Column.
+        """
         from inspect import signature
         exprs = []
         num_args = len(signature(sfun).parameters)
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -86,9 +86,18 @@ class DataFrame(_Frame):
                 for field in self._metadata.index_fields]
 
     def _reduce_for_stat_function(self, sfun):
+        from inspect import signature
         exprs = []
+        num_args = len(signature(sfun).parameters)
         for col in self.columns:
-            exprs.append(sfun(self._sdf[col], self._sdf.schema[col].dataType).alias(col))
+            if num_args == 1:
+                # Only pass in the column if sfun accepts only one arg
+                exprs.append(sfun(self._sdf[col]).alias(col))
+            else:  # must be 2
+                assert num_args == 2
+                # Pass in both the column and its data type if sfun accepts two args
+                exprs.append(sfun(self._sdf[col], self._sdf.schema[col].dataType).alias(col))
+
         sdf = self._sdf.select(*exprs)
         pdf = sdf.toPandas()
         assert len(pdf) == 1, (sdf, pdf)
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -24,7 +24,7 @@ import numpy as np
 import pandas as pd
 from pyspark import sql as spark
 from pyspark.sql import functions as F
-from pyspark.sql.types import DoubleType, FloatType, StructType, to_arrow_type
+from pyspark.sql.types import DataType, DoubleType, FloatType, StructType, to_arrow_type
 from pyspark.sql.utils import AnalysisException
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
@@ -74,7 +74,10 @@ class DataFrame(_Frame):
                 for field in self._metadata.index_fields]
 
     def _reduce_for_stat_function(self, sfun):
-        sdf = self._sdf.select([sfun(self._sdf[col]).alias(col) for col in self.columns])
+        exprs = []
+        for col in self.columns:
+            exprs.append(sfun(self._sdf[col], self._sdf.schema[col].dataType).alias(col))
+        sdf = self._sdf.select(*exprs)
         pdf = sdf.toPandas()
         assert len(pdf) == 1, (sdf, pdf)
         row = pdf.iloc[0]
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -497,9 +497,7 @@ def count(self):
 
         Returns
         -------
-        Series or DataFrame
-            For each column/row the number of non-NA/null entries.
-            If `level` is specified returns a `DataFrame`.
+        pandas.Series
```
rxin: should we return a ks.Series here?

databricks/koalas/frame.py
```diff
@@ -532,10 +530,20 @@ def count(self):
...
+            spark_type = self._sdf.schema[col].dataType
+            # Special handle floating point types because Spark's count treats nan as a valid value,
+            # whereas Pandas count doesn't include nan.
+            if isinstance(spark_type, DoubleType) or isinstance(spark_type, FloatType):
+                count_exprs.append(F.count(F.nanvl(col, F.lit(None))).alias(col))
+            else:
+                count_exprs.append(F.count(col).alias(col))
+
+        row = self._sdf.select(*count_exprs).collect()[0]
+        return pd.Series(row.asDict())
```
ueshin: We might be able to reuse `DataFrame._reduce_for_stat_function`.
The `NaN` handling is slightly different but I guess we should do the same thing in the `_reduce_for_stat_function`, and then we can reuse it.

databricks/koalas/frame.py
ueshin: I had the same question before. We picked pandas Series at that time, but we can rethink now. https://github.com/databricks/koalas/pull/57#discussion_r271578421
cc @thunterdb

thunterdb: We should return a pandas series in my opinion. Here is the reasoning:
 - the series is guaranteed to be 'small'
 - it is the output of an aggregation. if it was spark, we would return a Row object
 - koalas series are supposed to be for 'big' data and have different performance expectations when doing a point lookup. It should be instant and not trigger a spark job.

databricks/koalas/frame.py
thunterdb: I agree, `_reduce_for_stat_function` should be the standard path.

databricks/koalas/frame.py
```diff
@@ -655,6 +663,7 @@ def __setitem__(self, key, value):
         else:
             kdf = self.assign(**{key: value})
 
+        # Spark DataFrame
         self._sdf = kdf._sdf
```
thunterdb: Can you do a type annotation instead? Either as a comment:
```
self._sdf = kdf._sdf  # type: pyspark.sql.DataFrame
```
or as an annotation:
```
self._sdf: pyspark.sql.DataFrame = kdf._sdf
```
Both should be picked up by mypy

rxin:
Ok that makes sense. I will add this to design principles too.

Where should I take the test cases from?

thunterdb:
Here are a couple:
https://github.com/pandas-dev/pandas/blob/master/pandas/tests/frame/test_analytics.py#L1322
https://github.com/pandas-dev/pandas/blob/master/pandas/tests/series/test_analytics.py#L434

I could not find tests in dask, but I did not look long.

databricks/koalas/frame.py
garawalid: If we return a `pd.Series` we need to cast it again to `ks.Series` in case we want to use Koalas methods.
Since the result is guaranteed to be small, we can just return `ks.Series` and I think there is no big impact.

thunterdb: @garawalid  there is still an impact in the sense that accessing any of the fields is going to trigger a spark job (> 0.1 sec), let alone the cost of displaying the info, or doing things like iterating. From the code bases I have looked at (seaborn, pandas-profiling) these sections are used with regular python loops and would be very inefficient.

What do you mean 'in case we want to use the koalas methods'? All the functions in the `koalas.*` namespaces should eventually accept both koalas or pandas series / dataframes, because they follow the same specs. Similarly, all the methods of the classes should have the same behaviour.

garawalid: @thunterdb 
I agree about Spark overhead. 

What I meant by koalas methods, when we return `pd.Series` and we apply `filter()`, we will use `pd.Series.filter()`, in this case, there is no problem. But let's assume that we have an other method in Koalas `m()` that doesn't exist in `pd.Series`, users have to cast the result to `ks.Series`  in order to use `m()`

```python
res = df.count()
res = pd2ks(res) # casting 
res.m()
```

thunterdb: @garawalid I agree that if we add other methods that are not in the pandas api, then they will not be available. In general so far, we have gone to great lengths to reproduce the pandas API, and there is debate about even adding other functions. This is why for the time being, and because the scenario you are mentioning is hypothetical, I am comfortable returning a pandas object. We can always change that later with no change to the user if we see the benefits outweigh costs in performance. Also, theres is always the possibility of explicitly converting the pandas series to a koalas series.

databricks/koalas/frame.py
rxin: OK in the future please add docstring comment for every function. It took me a while to understand what _reduce_for_stat_function does.

databricks/koalas/frame.py
garawalid: @thunterdb Now, I am convinced that `pd.Series` is a better choice in this case.

databricks/koalas/frame.py
rxin: _reduce_for_stat_function passes a spark.Column in, and there is no way to get the type. Should I refactor that to pass also the column data type in?

rxin:
PTAL

databricks/koalas/frame.py
```diff
@@ -85,7 +86,19 @@ def _index_columns(self):
                 for field in self._metadata.index_fields]
 
     def _reduce_for_stat_function(self, sfun):
-        sdf = self._sdf.select([sfun(self._sdf[col]).alias(col) for col in self.columns])
+        from inspect import signature
```
thunterdb: @rxin  this is very magical, and it is hard to debug what is happening here. There are 2 ways you could go here, I believe:
 - have a separate `_reduce_for_stat_function_with_type` , for which `sfun` is expected to take 2 arguments like you are doing with `count` below
 - use `k.sSeries` (which can resolve their types).

I think that the first option is the easiest, given that this is a low-level wrapper for spark functions.

rxin: I'm tempted to just keep it as in, and add some documentation.

The reason is I wouldn't create a whole new function because only one function depends on it...

rxin:
Going to merge this so I can submit another PR. @thunterdb if you still think we should remove that reflection, we can work on it later.

PR #189, opened 2019-04-26
Add docs to metadata.

ueshin:
Resolves #186.

PR #190, opened 2019-04-26
Add type annotations to metadata.

ueshin:
This is a follow up of #189 to add type annotations.

### DIFF
--- a/databricks/koalas/metadata.py
+++ b/databricks/koalas/metadata.py
@@ -25,7 +25,7 @@ import pandas as pd
 from databricks.koalas.dask.compatibility import string_types
 
 
-IndexInfo = Tuple[str, str]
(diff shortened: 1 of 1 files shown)


ueshin:
oh, mypy is awesome!

databricks/koalas/metadata.py
```diff
@@ -31,7 +37,8 @@ class Metadata(object):
                        and the index name to be seen in Koalas DataFrame.
     """
 
-    def __init__(self, column_fields, index_info=None):
+    def __init__(self, column_fields: List[str],
+                 index_info: Optional[List[IndexInfo]] = None) -> None:
```
thunterdb: can you also add type info to the internal fields below (`_column_fields`, ...)? Thanks!

rxin:
Thanks @ueshin

PR #197, opened 2019-04-28
Inline docs for value_counts()

abishekganesh72:
Hi @thunterdb Take a look at this PR, If it looks good then we work other function

### DIFF
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -23,6 +23,7 @@ from databricks import koalas
 def add_ks(doctest_namespace):
     doctest_namespace['ks'] = koalas
 
+
 @pytest.fixture(autouse=True)
 def add_np(doctest_namespace):
     doctest_namespace['np'] = np
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -14,9 +14,10 @@
 # limitations under the License.
 #
 import pytest
+import numpy as np
 
 from databricks import koalas
-import numpy as np
+
 
 @pytest.fixture(autouse=True)
 def add_ks(doctest_namespace):
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -16,8 +16,12 @@
 import pytest
 
 from databricks import koalas
-
+import numpy as np
 
 @pytest.fixture(autouse=True)
 def add_ks(doctest_namespace):
     doctest_namespace['ks'] = koalas
+
+@pytest.fixture(autouse=True)
+def add_np(doctest_namespace):
+    doctest_namespace['np'] = np
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -348,7 +348,7 @@ class Series(_Frame):
 
         Examples
         --------
-        >>> df = ks.DataFrame({'x':[ ]})
+        >>> df = ks.DataFrame({'x':[0, 0, 1, 1, 1, np.nan ]})
         >>> df.x.value_counts()
         1.0    3
         0.0    2
@@ -368,7 +368,7 @@ class Series(_Frame):
         >>> df.x.value_counts(dropna=False)
         1.0    3
         0.0    2
-        NaN     1
+        NaN    1
         Name: x, dtype: int64
         """
         if bins is not None:
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -348,33 +348,27 @@ class Series(_Frame):
 
         Examples
         --------
-        >>> df = ks.DataFrame({'x':[1, 1,1 ,np.nan, 2, 3,2, 2, 2, 3, 0]})
+        >>> df = ks.DataFrame({'x':[ ]})
         >>> df.x.value_counts()
-        2.0    4
         1.0    3
-        3.0    2
-        0.0    1
+        0.0    2
         Name: x, dtype: int64
 
         With `normalize` set to `True`, returns the relative frequency by
         dividing all values by the sum of values.
 
         >>> df.x.value_counts(normalize=True)
-        2.0    0.4
-        1.0    0.3
-        3.0    0.2
-        0.0    0.1
+        1.0    0.6
+        0.0    0.4
         Name: x, dtype: float64
 
         **dropna**
         With `dropna` set to `False` we can also see NaN index values.
 
         >>> df.x.value_counts(dropna=False)
-        2.0    4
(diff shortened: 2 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -319,6 +319,64 @@ def unique(self):
 
     @derived_from(pd.Series)
```
ueshin: We don't need this anymore.

databricks/koalas/series.py
```diff
@@ -319,6 +319,64 @@ def unique(self):
...
+        -------
+        counts : Series
+        See Also
+        --------
+        Series.count: Number of non-NA elements in a Series.
+        Examples
+        --------
+        import pandas as pd
+        import databricks.koalas as ks
+        import numpy as np
```
ueshin: We don't need to import these here since they are already imported at the top of this file.

databricks/koalas/series.py
```diff
@@ -319,6 +319,64 @@ def unique(self):
...
+        >>> pdf = pd.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})
+        >>> df = ks.from_pandas(pdf)
```
ueshin: `df = ks.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})` should work.

abishekganesh72:
Let me work on this.

thunterdb:
@abishekganesh72 if you need `numpy` and `pandas` in your docstring, they are common enough that we can add in to the list of standard imports in the docstrings, see how `koalas` is imported here:
https://github.com/databricks/koalas/blob/master/databricks/conftest.py#L23

databricks/koalas/series.py
abishekganesh72: Updated code

abishekganesh72:
Did the changes let me know if I have change something else.

databricks/conftest.py
```diff
@@ -16,8 +16,12 @@
 import pytest
 
 from databricks import koalas
-
+import numpy as np
```
HyukjinKwon: Let's fix the import order too per PEP8

```python
import pytest
import numpy as np

from databricks import koalas
```

HyukjinKwon:
I think this is ready for review.

databricks/koalas/series.py
```diff
@@ -317,8 +317,61 @@ def unique(self):
...
+        ----------
+        normalize : boolean, default False
+            If True then the object returned will contain the relative
+            frequencies of the unique values.
+        sort : boolean, default True
+            Sort by values.
+        ascending : boolean, default False
+            Sort in ascending order.
+        bins : Not Yet Supported
+        # TODO: Update Documentation when its supported
```
HyukjinKwon: Can we leave this todo out of the docstring?

HyukjinKwon: Actually, let's do this here. We can exclude what we don't support in this docstring, and leave a short todo comment what's excluded right above of `value_counts` definition.

databricks/koalas/series.py
```diff
@@ -317,8 +317,61 @@ def unique(self):
...
+            Don't include counts of NaN.
+        Returns
+        -------
+        counts : Series
+        See Also
+        --------
+        Series.count: Number of non-NA elements in a Series.
+        Examples
+        --------
+        >>> pdf = ks.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})
```
shril: I think the name of the variable will be df instead of pdf for the rest of the examples to work.

`>>> pdf = ks.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})`
`>>> df = ks.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})`

abishekganesh72:
@ueshin made all the requested changes. let me know if this looks good

databricks/koalas/series.py
abishekganesh72: Good catch @shril. Thanks for that

abishekganesh72:
I am not clear why the build is failing. can you help me out here @HyukjinKwon  ?

databricks/koalas/series.py
```diff
@@ -317,8 +317,61 @@ def unique(self):
...
+        --------
+        Series.count: Number of non-NA elements in a Series.
+        Examples
+        --------
+        >>> df = ks.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})
+        >>> df.x.value_counts()
+         3.0    2
+         0.0    2
+         1.0    1
+         2.0    1
```
HyukjinKwon:
```
398         >>> df = ks.DataFrame({'x':[0, 1, np.nan, 0, 2, np.nan, 3, 3]})
399         >>> df.x.value_counts()
    @@ -1,5 +1,5 @@
    - 3.0    2
    - 0.0    2
    - 1.0    1
    - 2.0    1
    +0.0    2
    +3.0    2
    +1.0    1
    +2.0    1
     Name: x, dtype: int64
```

Seems doctests are being failed due to spacing diff

HyukjinKwon: and the order of values

HyukjinKwon: You can remove the spaces and the order of expected result in the doctest.

abishekganesh72: Did the changes. should be good to go.

abishekganesh72:
please take a look again.

databricks/koalas/series.py
```diff
@@ -317,8 +317,60 @@ def unique(self):
...
+        >>> df.x.value_counts(normalize=True)
+        1.0    0.6
+        0.0    0.4
+        Name: x, dtype: float64
+
+        **dropna**
+        With `dropna` set to `False` we can also see NaN index values.
+
+        >>> df.x.value_counts(dropna=False)
+        1.0    3
```
HyukjinKwon: In the case the test results are different due to white spaces, you can add `  # doctest: +NORMALIZE_WHITESPACE`:

```python
>>> df.x.value_counts(dropna=False)  # doctest: +NORMALIZE_WHITESPACE
...
```

abishekganesh72:
All test cleared

HyukjinKwon:
Thanks, @abishekganesh72

PR #199, opened 2019-04-29
spark.DataFrame.to_koalas()

### LINKED ISSUE #175: Monkey patch Spark's DataFrame to add to_koalas method
Create a simple and convenient way to turn a Spark DataFrame into a Koalas DataFrame.

rxin:
Introduces a new method to_koalas that can be used to convert a Spark DataFrame into a Koalas DataFrame. This method is monkey patched into Spark dynamically when the koalas package is imported.

Resolves #175

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -303,6 +303,47 @@ class DataFrame(_Frame):
 
     notna = notnull
 
+    def to_koalas(self):
+        """
+        Converts the existing DataFrame into a Koalas DataFrame.
+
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -303,6 +303,55 @@ def notnull(self):
 
     notna = notnull
 
+    def to_koalas(self):
+        """
+        Converts the existing DataFrame into a Koalas DataFrame.
```
HyukjinKwon: tiny nit .. nit: Return -> Returns or Converts -> Convert

PR #200, opened 2019-04-30
DataFrame/Series.to_pandas

rxin:
Rename existing toPandas methods to_pandas, add docs, and alias toPandas to to_pandas for backward compatibility with Spark.


rxin:
I also need to update Series'.

PR #203, opened 2019-04-30
Add docstring in Series.rename and Series.alias

### LINKED ISSUE #131: Can't use Spark's .alias() to rename column
I would like to rename a column using .alias() instead of the pandas syntax (it's cleaner in this case), but I get the error message seen below. .toDF() works for renaming all the columns. Can we implement .alias() please?
Screen Shot 2019-04-22 at 7 47 50 AM

HyukjinKwon:
Resolves #131

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -206,8 +206,8 @@ class Series(_Frame):
                     "Field {} not found, possible values are {}".format(name, ", ".join(fnames)))
             return Series(self._scol.getField(name), self._kdf, self._index_info)
 
-    # TODO: automate the process here
     def alias(self, name):
+        """An alias for :meth:`Series.rename`."""
         return self.rename(name)
 
     @property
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
cc @rxin

databricks/koalas/series.py
```diff
@@ -206,8 +206,8 @@ def getField(self, name):
                     "Field {} not found, possible values are {}".format(name, ", ".join(fnames)))
             return Series(self._scol.getField(name), self._kdf, self._index_info)
 
-    # TODO: automate the process here
     def alias(self, name):
+        """An alias for :meth:`Series.rename`."""
```
rxin: we should also add alias to the Series doc page.

Two ways:

One is to create a Spark API section, and just put alias there. The other is to put it in the same section as rename. WDYT? cc @thunterdb

rxin:
Merging now. Please submit a follow-up PR for adding alias to the doc page once we decided on which section to put it.

HyukjinKwon:
Will do. Thanks.

databricks/koalas/series.py
thunterdb: Nothing prevents you from putting it into both sections. We will eventually have a spark section anyway.

rxin: Yes that's a brilliant idea Tim.

PR #206, opened 2019-04-30
Support limit push down in to_html

rxin:
This patch inlines to_html docs from pandas, and supports limit push down when max_rows is specified, so we don't need to pull in all the data.


HyukjinKwon:
LGTM. Let me cherry-pick workaround separately right way at https://github.com/databricks/koalas/pull/209

PR #210, opened 2019-05-01
validate_arguments_and_invoke_function

rxin:
This patch adds validate_arguments_and_invoke_function for compatibility with older pandas versions. It is currently used in to_html, but I think we can apply that to a lot of other functions too.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -250,6 +250,7 @@ class DataFrame(_Frame):
         --------
         to_string : Convert DataFrame to a string.
         """
+        # Make sure locals() call is at the top of the function so we don't capture local variables.
         args = locals()
         if max_rows is not None:
             kdf = self.head(max_rows)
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -250,24 +250,14 @@ def to_html(self, buf=None, columns=None, col_space=None, header=True, index=Tru
         --------
         to_string : Convert DataFrame to a string.
         """
+        args = locals()
```
HyukjinKwon: @rxin, can we add a comment that this should be located at the top always?

rxin: done

databricks/koalas/utils.py
```diff
@@ -17,8 +17,56 @@
...
+
+    koalas_params = inspect.signature(koalas_func).parameters
+    pandas_params = inspect.signature(pandas_func).parameters
+
+    for param in koalas_params.values():
+        if param.name not in pandas_params:
+            if args[param.name] == param.default:
+                del args[param.name]
+            else:
+                raise TypeError(
```
HyukjinKwon: nit .. 

```python
                raise TypeError(
                    ("The pandas available [%s] does not support parameter '%s' " +
                        "for function '%s'.") % (pd.__version__, param.name, pandas_func.__name__))
```

rxin: good message

Issue #169, opened 2019-04-24
Implement the first batch of Serialization / IO / Conversion functions


rxin:
See https://pandas.pydata.org/pandas-docs/stable/reference/frame.html#serialization-io-conversion

Looks like we can easily implement almost all of them by calling `toPandas().func_name()`.

One thing is that some of the functions support `max_rows`. When that argument is specified, we should add a `limit` call in Spark to avoid moving all the data to the driver.

The list to add in the first batch are:

- [x] to_dict (see #169)
- [x] to_excel (#288)
- [x] to_html (we already have this, but let's add a limit when max_rows is set), done in #206
- [x] to_latex (#297)
- [x] to_records (#298)
- [x] to_string (done in #211 and #213)
- [x] to_clipboard (#257)

Skipping the following because I don't know how popular they are:

- to_pickle
- to_hdf
- to_stata
- to_msgpack
- to_records
- to_sparse
- to_dense

The following might require parallelization with Pandas UDFs, rather than collecting all the data to the driver, so leaving them for the future:

- to_sql
- to_gbq

I'm also not adding json and csv here. We need to design those properly because both Spark and Pandas have those.

rxin:
For contributors -- in general this is a pretty easy one to get started. Just look at how we did https://github.com/databricks/koalas/pull/211 and https://github.com/databricks/koalas/pull/213

arjun180:
Can I work on this?

HyukjinKwon:
please go ahead.

shril:
@rxin , @HyukjinKwon Can I take this up as well?

rxin:
Go for it.

HyukjinKwon:
I think we should do `to_json` too.

garawalid:
@rxin 
> I'm also not adding json and csv here. We need to design those properly because both Spark and Pandas have those.

I agree with @HyukjinKwon, we can add `to_json` & `to_csv`. 
In fact, both methods need to collect data in the driver through `toPandas()`. And if we pass the path as a parameter, the output will be saved in the driver. This is what Pandas users will expect, at least, those who are not familiar with Spark. 

In the case of Spark, I think we should keep `to_csv` and `to_json` in a separate section than `Serialization / IO / Conversion` and we explain the behaviour in the documentation. These methods will be useful if we want to write the output in HDFS without impacting the performance.

Btw, `to_clipboard` uses `to_csv`.

HyukjinKwon:
[...]

PR #211, opened 2019-05-01
DataFrame.to_string

rxin:
Part of #169

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -258,6 +258,53 @@ class DataFrame(_Frame):
         return validate_arguments_and_invoke_function(
             kdf.to_pandas(), self.to_html, pd.DataFrame.to_html, args)
 
+    def to_string(self, buf=None, na_rep='NaN', float_format=None, header=True,
+                  index=True, length=False, dtype=False, name=False,
+                  max_rows=None):
+        """
+        Render a string representation of the Series.
+
+        .. note:: This method should only be used if the resulting Pandas DataFrame is expected
+            to be small, as all the data is loaded into the driver's memory. If the input DataFrame
+            is large, set max_rows parameter.
+
+        Parameters
+        ----------
+        buf : StringIO-like, optional
+            buffer to write to
+        na_rep : string, optional
+            string representation of NAN to use, default 'NaN'
+        float_format : one-parameter function, optional
+            formatter function to apply to columns' elements if they are floats
+            default None
+        header : boolean, default True
+            Add the Series header (index name)
+        index : bool, optional
+            Add index (row) labels, default True
+        length : boolean, default False
+            Add the Series length
+        dtype : boolean, default False
+            Add the Series dtype
+        name : boolean, default False
+            Add the Series name if not None
+        max_rows : int, optional
+            Maximum number of rows to show before truncating. If None, show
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -258,6 +258,53 @@ def to_html(self, buf=None, columns=None, col_space=None, header=True, index=Tru
         return validate_arguments_and_invoke_function(
             kdf.to_pandas(), self.to_html, pd.DataFrame.to_html, args)
 
+    def to_string(self, buf=None, na_rep='NaN', float_format=None, header=True,
+                  index=True, length=False, dtype=False, name=False,
+                  max_rows=None):
+        """
+        Render a string representation of the Series.
+
+        .. note:: This method should only be used if the resulting Pandas DataFrame is expected
```
HyukjinKwon: Actually, we're mixing NumPy docstring `Paramters` with underscores, and typical sphinx docstring (like `.. note::`) ..

If I am not mistaken, we can do as:

```

        Notes
        -----
        This method should only be used if the resulting Pandas DataFrame is expected
```

databricks/koalas/frame.py
```diff
@@ -258,6 +258,53 @@ def to_html(self, buf=None, columns=None, col_space=None, header=True, index=Tru
...
+        name : boolean, default False
+            Add the Series name if not None
+        max_rows : int, optional
+            Maximum number of rows to show before truncating. If None, show
+            all.
+
+        Returns
+        -------
+        formatted : string (if not buffer passed)
+        """
```
HyukjinKwon: @rxin, can we add a doctest (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_string.html#pandas.DataFrame.to_string)?

databricks/koalas/frame.py
```diff
@@ -258,6 +258,53 @@ def to_html(self, buf=None, columns=None, col_space=None, header=True, index=Tru
...
+            Add the Series dtype
+        name : boolean, default False
+            Add the Series name if not None
+        max_rows : int, optional
+            Maximum number of rows to show before truncating. If None, show
+            all.
+
+        Returns
+        -------
+        formatted : string (if not buffer passed)
```
HyukjinKwon: Seems this isn't matched with Pandas's doc (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_string.html#pandas.DataFrame.to_string)

HyukjinKwon:
Looks good otherwise.

databricks/koalas/frame.py
rxin: Ah ok. They both work. We have other places too. I'm thinking we should just submit a separate PR to address all of them at the same time. Sounds good?

rxin: actually i copied the series one haha. going to fix it.

rxin:
@HyukjinKwon I pushed a new version. I'm actually thinking the other notes method is better, because it highlights the problem more.

See below:

Screen Shot 2019-04-30 at 8 12 04 PM

vs

Screen Shot 2019-04-30 at 8 12 09 PM

HyukjinKwon:
Yup, either way is fine to me.

PR #194, opened 2019-04-27
[WIP] Add encapsulated `GroupBy` and basic aggregate functions.

ueshin:
Switch over to the encapsulation approach for `groupby`, and add basic aggregate functions.
In this PR, only `by` argument will be supported.

The aggregate functions which will be addressed in this PR are:

- `agg`/`aggregate`
- `count`
- `first`
- `last`
- `max`
- `mean`
- `min`
- `std`
- `sum`
- `var`

TODOs:
- [ ] add `agg`/`aggregate` function
- [ ] docs

Resolves #181, #185.

Issue #185, opened 2019-04-26
Create docs for GroupBy


rxin:
Similar to Pandas: https://pandas.pydata.org/pandas-docs/stable/reference/groupby.html

Right now groupby itself returns nothing, and there is no docs for the methods available for groupby.

PR #215, opened 2019-05-01
Add encapsulated `GroupBy` and basic aggregate functions

### LINKED ISSUE #181: `groupby` method problem
Hi, I'm very appreciate of your nice work !

Here I found a problem of `groupby` function between `pandas.DataFrame` and `koalas.DataFrame`.

```python
import pandas as pd

# Create a pandas DataFrame
pdf = pd.DataFrame({'x':range(3), 'y':['a','b','b'], 'z':['a','b','b']})
# Create a Koalas DataFrame from pandas DataFrame
kdf = ks.from_pandas(pdf)

# pandas groupby, works well
print(pdf.groupby('y')['x'].mean())
## y
## a    0.0
## b    1.5
## Name: x, dtype: float64
```
However, when using `koalas`,
```python
# koalas groupby, throw error
print(kdf.groupby('y')['x'].mean())
```
```shell
ValueError                                Traceback (most recent call last)
<ipython-input-71-1d6e96373165> in <module>()
      3
      4 # koalas groupby, throw error
----> 5 print(kdf.groupby('y')['x'].mean())

/usr/install/anaconda3/lib/python3.6/site-packages/databricks/koalas/groups.py in __getitem__(self, key)
     49             if k not in fnames:
     50                 raise ValueError("{} not a field. Possible values are {}".format(k, fnames))
---> 51         return PandasLikeGroupBy(self._df, self._groupdata, key)
     52
     53     def __getattr__(self, key):

/usr/install/anaconda3/lib/python3.6/site-packages/databricks/koalas/groups.py in __getattr__(self, key)
     53     def __getattr__(self, key):
     54         try:
---> 55             return self[key]
     56         except KeyError as e:
     57             raise AttributeError(e)

/usr/install/anaconda3/lib/python3.6/site-packages/databricks/koalas/groups.py in __getitem__(self, key)
     48         for k in l:
     49             if k not in fnames:
---> 50                 raise ValueError("{} not a field. Possible values are {}".format(k, fnames))
     51         return PandasLikeGroupBy(self._df, self._groupdata, key)
     52

ValueError: _df not a field. Possible values are {'y', '__index_level_0__', 'x', 'z'}
```
<br/>

My environment:
- python 3.6.5
- spark 2.4.0

HyukjinKwon:
Switch over to the encapsulation approach for `groupby`, and add basic aggregate functions.
In this PR, only `by` argument will be supported.

Also note that this PR does not add `agg` and `aggregate` into `Series` - this has never been supported and looks a bit tricky to add. This could be done separately.

The aggregate functions which will be addressed in this PR are:

- `agg`/`aggregate`
- `count`
- `first`
- `last`
- `max`
- `mean`
- `min`
- `std`
- `sum`
- `var`
- `agg`/`aggregate`

Resolves #181, #185.
Closes #194


HyukjinKwon:
cc @rxin and @thunterdb

databricks/koalas/groupby.py
```diff
@@ -0,0 +1,312 @@
...
+
+        return kdf
+
+    agg = aggregate
+
+    def count(self):
+        """
+        Compute count of group, excluding missing values.
+        """
+        return self._reduce_for_stat_function(F.count, only_numeric=False)
```
rxin: do we need to handle nan values differently?

databricks/koalas/testing/utils.py
```diff
@@ -194,30 +194,53 @@ def assertPandasAlmostEqual(self, left, right):
             self.assertEqual(left.shape, right.shape, msg=msg)
             for lcol, rcol in zip(left.columns, right.columns):
                 self.assertEqual(str(lcol), str(rcol), msg=msg)
-                for lval, rval in zip(left[lcol], right[rcol]):
+                for lnull, rnull in zip(left[lcol].isnull(), right[rcol].isnull()):
```
rxin: add some comment explaining what's happening here ...

databricks/koalas/groupby.py
HyukjinKwon: I think that's handled in `_reduce_for_stat_function`. Yup, it's different from what we do in `frame.py`. Let me at least add a doctest here for now.

rxin:
LGTM - for expediency I think we can merge this once tests pass, and work on improving it later.

We should also make a release today!

databricks/koalas/groupby.py
```diff
@@ -0,0 +1,321 @@
...
+                if isinstance(spark_type, DoubleType) or isinstance(spark_type, FloatType):
+                    stat_exprs.append(sfun(F.nanvl(ks._scol, F.lit(None))).alias(ks.name))
+                    column_fields.append(ks.name)
+                elif isinstance(spark_type, NumericType) or not only_numeric:
+                    stat_exprs.append(sfun(ks._scol).alias(ks.name))
+                    column_fields.append(ks.name)
+            sdf = sdf.groupby(*groupkey_cols).agg(*stat_exprs)
+        else:
+            sdf = sdf.select(*groupkey_cols).distinct()
+        sdf = sdf.sort(*groupkey_cols)
```
ueshin: @HyukjinKwon Do you remember why we added `sort` here?

HyukjinKwon: It was from your PR (https://github.com/databricks/koalas/pull/194/files#diff-611bb17e37caad740e4b9943a69c3b4bR97) :-) ..

ueshin: Yeah, right. I don't remember the reason. haha

HyukjinKwon: Looks we can remove

PR #218, opened 2019-05-01
Enable statistics for boolean values

floscha:
In pandas it is possible to sum over boolean values which are implicitly converted to integers. For example, the following code can be used to compute to number of NA values per column.
```python
>>> df = pd.DataFrame({'A': [0, np.nan, np.nan], 'B': [0, 0, np.nan]})
>>> df.isna().sum()
A    2
B    1
dtype: int64
```

When trying to run the same code with a Koalas DataFrame, the following exception is thrown: `org.apache.spark.sql.AnalysisException: cannot resolve 'sum(`A`)' due to data type mismatch: function sum requires numeric types, not boolean`

To fix this, the PR casts boolean values to integers as required when using statistical functions. Furthermore, the unit tests have been extended to cover the following statistics with boolean values:
- min
- max
- sum
- mean
- var
- std

### DIFF
--- a/databricks/koalas/tests/test_stats.py
+++ b/databricks/koalas/tests/test_stats.py
@@ -99,6 +99,19 @@ class StatsTest(ReusedSQLTestCase, SQLTestUtils):
             # assert ddf.a.cov(ddf.b)._meta.dtype == 'f8'
             # assert ddf.a.corr(ddf.b)._meta.dtype == 'f8'
 
+    def test_stats_on_boolean(self):
+        df = pd.DataFrame({'A': [True, False]})
+        ddf = koalas.from_pandas(df)
+
+        self.assert_eq(ddf.min(), df.min())
+        self.assert_eq(ddf.max(), df.max())
+
+        self.assert_eq(ddf.sum(), df.sum())
+        self.assert_eq(ddf.mean(), df.mean())
+
+        self.assert_eq(ddf.var(), df.var())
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_stats.py
```diff
@@ -99,6 +99,19 @@ def test_cov_corr_meta(self):
             # assert ddf.a.cov(ddf.b)._meta.dtype == 'f8'
             # assert ddf.a.corr(ddf.b)._meta.dtype == 'f8'
 
+    def test_stats_on_boolean(self):
+        df = pd.DataFrame({'A': [True, False]})
```
garawalid: If you test the case of this `df`, asserts of `var()` and `std()` will fail. 
```python
        df = pd.DataFrame({'A': [True, False,True],
                           'B': [False,False,True]})
```

floscha: Thanks for your feedback. The error seems to be in Koalas' `assert_eq()` method. When using pandas' `pd.testing.assert_series_equal()` instead, the tests for `var()` and `std()` work as expected. However, this revealed another issue where `min()` and `max()` return boolean values for pandas and integers for Koalas which I have now fixed in 5992c6f.

rxin:
This is great. I tested it locally and it seemed to work. Merging now.

rxin:
@floscha can you also submit a PR to fix the same issue for Series?

Issue #152, opened 2019-04-23
Inline method docs, rather than using "derived_from"


rxin:
The problem is that the equivalent methods in pandas usually have a lot of arguments, and some of them don't make to support in Koalas or are temporarily not supported. Showing those in the examples are simply confusing to end users when they read the docs. We should just inline the docs. We can however create them using pandas docs as the baseline.

Here is an initial list:

```
frame.py-    def iteritems(self):
frame.py-    def to_html(self, buf=None, columns=None, col_space=None, header=True, index=True,
frame.py-    def isnull(self):
frame.py-    def notnull(self):
frame.py-    def toPandas(self):
frame.py-    def assign(self, **kwargs):
frame.py-    def dropna(self, axis=0, how='any', thresh=None, subset=None, inplace=False):
frame.py-    def drop(self, labels, axis=0, errors='raise'):
frame.py-    def get(self, key, default=None):
frame.py-    def pipe(self, func, *args, **kwargs):

series.py-    def rename(self, index=None, **kwargs):
series.py-    def reset_index(self, level=None, drop=False, name=None, inplace=False):
series.py-    def isnull(self):
series.py-    def notnull(self):
series.py-    def dropna(self, axis=0, inplace=False, **kwargs):
series.py-    def head(self, n=5):
series.py-    def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, dropna=True):
```

rxin:
cc @thunterdb what do you think about this one?

thunterdb:
I totally agree, we should not need to use derived_from. Personnally, I would rather have (slightly misleading) doc from pandas for now rather than no doc at all.

This is a great starter task if someone wants to do it on their favorite functions.

rxin:
A lot of them are probably blocked by https://github.com/databricks/koalas/issues/168

abishekganesh72:
Hi let me know if I can be of any help to for this

thunterdb:
@abishekganesh72 you are welcome to help. Here is an example, taking `Series.value_counts`:
It would be great to take the documentation from pandas:
https://github.com/pandas-dev/pandas/blob/v0.24.2/pandas/core/base.py#L1222
and copy the docstring into `series.py`:
https://github.com/databricks/koalas/blob/master/databricks/koalas/series.py#L321

Some changes will have to be made:
 - removing the argument that are not available yet (like `bins`) and making note in the doc that it is not supported yet
 - removing the example that uses `bins`
[...]

PR #222, opened 2019-05-01
Inline documentation for iteritems, dropna, drop, get

shril:
Inline documentation for drop function in frame.py file.


shril:
Hi @rxin , have a look at this.
I am working on the inline docs of other functions in frame.py file.

Happy to address any issues regarding my changes.

databricks/koalas/frame.py
```diff
@@ -881,8 +881,47 @@ def unique(self):
         sdf = self._sdf
         return DataFrame(spark.DataFrame(sdf._jdf.distinct(), sdf.sql_ctx), self._metadata.copy())
 
-    @derived_from(pd.DataFrame)
     def drop(self, labels, axis=0, errors='raise'):
```
rxin: cc @thunterdb why do we have "errors" parameter that seem unused?

databricks/koalas/frame.py
```diff
@@ -881,8 +881,47 @@ def unique(self):
...
+        Drop specified labels from columns.
+
+        Remove columns by specifying label names and axis=1.
+        Removing rows is yet to be implemented.
+
+        Parameters
+        ----------
+        labels : single label or list-like
+            Column labels to drop.
+        axis : {0 or 'index', 1 or 'columns'}, default 0
```
rxin: say something about only axis = 1 is supported at the moment.

rxin:
This looks great. For the next pull request, feel free to do more than one function.

shril:
@rxin I'll push the documentation of other functions as well in this PR itself.
I little caught up.

databricks/koalas/frame.py
shril: Added the information.

shril:
Hi @rxin , I have added documentation for several functions in this PR.

I was keeping a track of documentation of functions left in frame.py. Issue #152 
Unchecked ones are left.

- [x] to_html
- [x] isnull
- [x] notnull
- [x] toPandas
- [ ] assign
- [x] dropna
- [x] drop
- [x] get
- [ ] pipe
- [x] iteritems

databricks/koalas/frame.py
```diff
@@ -163,8 +163,44 @@ def corr(self, method='pearson'):
...
 
-    @derived_from(pd.DataFrame)
     def iteritems(self):
+        """
+        Iterator over (column name, Series) pairs.
+
+        Iterates over the DataFrame columns, returning a tuple with
+        the column name and the content as a Series.
+
+        Yields
```
HyukjinKwon: Let's make this API yield or fix the doc.

databricks/koalas/frame.py
```diff
@@ -691,8 +727,88 @@ def loc(self):
...
         return DataFrame(self._sdf, self._metadata.copy())
 
-    @derived_from(pd.DataFrame)
     def dropna(self, axis=0, how='any', thresh=None, subset=None, inplace=False):
+        """
+        Remove missing values.
+
+        Parameters
+        ----------
+        axis : {0 or 'index', 1 or 'columns'}, default 0
```
HyukjinKwon: I think we don;t support `columns` yet. Let's take out.

databricks/koalas/frame.py
HyukjinKwon: Seems like just wanted to match the signature (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.drop.html#pandas.DataFrame.drop) Let's remove.

databricks/koalas/frame.py
HyukjinKwon: I think we should fix this doc.

databricks/koalas/frame.py
shril: I am fixing the doc for now.

databricks/koalas/frame.py
shril: Doing it.

databricks/koalas/frame.py
shril: Removing it.

shril:
Hi @HyukjinKwon,

- Changed the docs for `dropna`, which only supports axis=0 at the moment.
- Changed the docs for `drop`, which only supports axis=1 at the moment.
- Changed from yields to returns in the documentation for `iteritems`.
- Removed error parameter from the `drop` function.

shril:
Hi @HyukjinKwon, I think my PR contains a large number of commits.
Should I fork another branch from master and transfer the changes to a single commit?

rxin:
It's ok when we merge into master we squash all commits into a single one.

databricks/koalas/frame.py
```diff
@@ -691,8 +727,86 @@ def loc(self):
...
+        Returns
+        -------
+        DataFrame
+            DataFrame with NA entries dropped from it.
+
+        See Also
+        --------
+        DataFrame.drop : Drop specified labels from columns.
+        DataFrame.isnull: Indicate missing values.
+        DataFrame.notnull : Indicate existing (non-missing) values.
```
garawalid: We don't have `notnull()`.
You can replace it by `notna()`.

rxin: We do have notnull, don't we? https://github.com/databricks/koalas/blob/master/databricks/koalas/frame.py#L535

rxin:
Thanks. I've merged this in master.

databricks/koalas/frame.py
shril: Line 571 is `notnull()` in frame.py.
Please correct me if I am making any mistake here.

HyukjinKwon:
Nice doc change. Thanks, @shril

databricks/koalas/frame.py
garawalid: I saw the documentation. You re right

PR #226, opened 2019-05-02
Cleanup and simplify the logics of fillna

HyukjinKwon:
This PR propose to simplify fillna and the tests of fillna.

### DIFF
--- a/databricks/koalas/typedef.py
+++ b/databricks/koalas/typedef.py
@@ -170,18 +170,6 @@ def _check_compatible(arg, sig_arg: X):
     assert False, (arg, sig_arg)
 
 
-def dict_sanitizer(input_dict):
-    """
-    This function checks if elements inside a given dict are supported by Spark.
-    :param input_dict: dict
-    :raises TypeError if any element is not supported by Spark
-    """
-    _possible_type_values = [int, str, float, np.float64, np.float, bool]
-    for e in input_dict.values():
-        if not type(e) in _possible_type_values:
-            raise TypeError("Dict contains unsupported type {}".format(type(e)))
-
-
 def make_fun(f, *args, **kwargs):
     """
     This function calls the function f while taking into account some of the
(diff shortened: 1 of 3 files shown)


databricks/koalas/typedef.py
```diff
@@ -170,18 +170,6 @@ def _check_compatible(arg, sig_arg: X):
     assert False, (arg, sig_arg)
 
 
-def dict_sanitizer(input_dict):
-    """
-    This function checks if elements inside a given dict are supported by Spark.
-    :param input_dict: dict
-    :raises TypeError if any element is not supported by Spark
-    """
-    _possible_type_values = [int, str, float, np.float64, np.float, bool]
```
HyukjinKwon: `np.float64` and `np.float` seems inheriting `float`:

```python
>>> import numpy as np
>>> np.float64(1.0)
1.0
>>> type(np.float64(1.0))
<class 'numpy.float64'>
>>> isinstance(np.float64(1.0), float)
True
>>> type(np.float(1.0))
<class 'float'>
>>> isinstance(np.float(1.0), float)
True
```

That seems why it works with Py4J.

thunterdb: Ah, this must be a recent change, I recall having some issues with this before.

thunterdb: what does it give with `np.float32` and `np.float16`?

HyukjinKwon: Interesting :D

```python
>>> import numpy as np
>>> type(np.float32(1.0))
<class 'numpy.float32'>
>>> isinstance(np.float32(1.0), float)
False
>>> type(np.float16(1.0))
<class 'numpy.float16'>
>>> isinstance(np.float16(1.0), float)
False
```

HyukjinKwon:
This is just a clean up. I merged.

garawalid:
Nice clean up :+1:

PR #227, opened 2019-05-02
Cleanups to groupy

HyukjinKwon:
This simplifies somewhat the calls to `GroupBy`.

Closes #225

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -29,7 +29,6 @@ from pyspark.sql.types import DataType, DoubleType, FloatType
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
 from databricks.koalas.dask.utils import derived_from
-from databricks.koalas.utils import _resolve_column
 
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
It says no for coverage decrease .. which I don't think it's quite true :D.

thunterdb:
@HyukjinKwon there is an escape hatch:
https://coverage.readthedocs.io/en/coverage-4.2/excluding.html

THis way you can exclude some of the `else` conditions that are errors anyway

HyukjinKwon:
oh, I see. I will do that. BTW, I think another reason why it says coverage is decreased is the amount of codes at `generic.py` was reduced:

> databricks/koalas/generic.py | 95% <87.5%> (-3.31%)

I think it complains spuriously anyway.

HyukjinKwon:
@thunterdb, let me get this in but please let me know if there are more to be done. I will fix it right away.

thunterdb:
This is great, thank you!

PR #230, opened 2019-05-03
Inline docs for get_dummies

rxin:
This also gets us some free test cases.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -24,8 +24,8 @@ from pyspark.sql import functions as F
 from pyspark.sql.types import ByteType, ShortType, IntegerType, LongType, FloatType, \
     DoubleType, BooleanType, TimestampType, DecimalType, StringType, DateType, StructType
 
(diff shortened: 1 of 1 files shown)


databricks/koalas/namespace.py
```diff
@@ -190,9 +190,89 @@ def to_datetime(arg, errors='raise', format=None, infer_datetime_format=False):
...
+    dtype : dtype, default np.uint8
+        Data type for new columns. Only a single dtype is allowed.
+
+    Returns
+    -------
+    dummies : DataFrame
+
+    See Also
+    --------
+    Series.str.get_dummies
```
HyukjinKwon: Looks we don't have it yet

rxin: Maybe it's not the end of the world to have it here so we won't forget in the future ...

HyukjinKwon: Yea, I don't mind merging it. Just wanted to note so that we can kind of remember together :D.

PR #234, opened 2019-05-04
refactor test_dataframe & test_series

garawalid:
as requested by @rxin, this PR refactor `test_dataframe.py` & `test_series.py`

### DIFF
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -39,9 +39,8 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
     def test_series(self):
         ks = self.ks
 
-        self.assertTrue(isinstance(ks.x, Series))
-        self.assertTrue(isinstance(ks.x + 1, Series))
-        self.assertTrue(isinstance(1 + ks.x, Series))
+        self.assertTrue(isinstance(ks['x'], Series))
+
         # TODO: self.assert_eq(d + 1, pdf + 1)
 
     def test_empty_series(self):
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -15,13 +15,90 @@
 #
 
 
+import numpy as np
 import pandas as pd
 
 from databricks import koalas
+from databricks.koalas.series import Series
 from databricks.koalas.testing.utils import ReusedSQLTestCase, SQLTestUtils
 
 
 class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
+    @property
+    def pdf(self):
+        return pd.DataFrame({
+            'a': [1, 2, 3, 4, 5, 6, 7, 8, 9],
+            'b': [4, 5, 6, 3, 2, 1, 0, 0, 0],
+        }, index=[0, 1, 3, 5, 6, 8, 9, 9, 9])
+
+    @property
+    def kdf(self):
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_series.py
```diff
@@ -15,13 +15,90 @@
...
+        return pd.DataFrame({
+            'a': [1, 2, 3, 4, 5, 6, 7, 8, 9],
+            'b': [4, 5, 6, 3, 2, 1, 0, 0, 0],
+        }, index=[0, 1, 3, 5, 6, 8, 9, 9, 9])
+
+    @property
+    def kdf(self):
+        return koalas.from_pandas(self.pdf)
+
+    def test_Series(self):
```
HyukjinKwon: I think we should stick to the one naming `test_series` while we're here.

garawalid: Agree !

garawalid:
@HyukjinKwon 
I 've been thinking about moving `test_series_datetime` tests to `test_series`.

What do you think?

garawalid:
> test_value_counts : kdf.x.value_counts()  

This method implemented only for series. I'll move it to `test_series.py` 

> test_fillna : kdf.fillna(s_nan)  

`fillna()` accepts series as input. We have to keep it!

rxin:
Let’s keep the test_series_datetime on its own, because we will be adding a lot more datetime tests and it’s best to keep each test suite still small.

rxin:
@HyukjinKwon I will leave this to you to merge.

HyukjinKwon:
Will check and merge tonight (KST)

garawalid:
> Looks these lines below have to be moved too.
> 
> ```
> # TODO: assert d.b.mean().compute() == pdf.b.mean()
> # TODO: assert np.allclose(d.b.var().compute(), pdf.b.var())
> # TODO: assert np.allclose(d.b.std().compute(), pdf.b.std())
> ```
>
I think we can keep them. They test operations on column of dataframe (which they are `Series`) 
> ```
>         self.assert_eq(kdf['a'].head(2), pdf['a'].head(2))
>         self.assert_eq(kdf['a'].head(3), pdf['a'].head(3))
> ```
> 
> ```
>         # TODO: self.assert_eq(d['a'].tail(2), pdf['a'].tail(2))
>         # TODO: self.assert_eq(d['a'].tail(3), pdf['a'].tail(3))
> ```

I'll remove them. I replaced them in `test_head_tail()`
```python
    def test_head_tail(self):
        ks = self.ks
        ps = self.ps

        self.assert_eq(ks.head(3), ps.head(3))

        # TODO: self.assert_eq(ks.tail(3), ps.tail(3))
```

> `test_fillna` seems having a series test,
> 
> ```
>         self.assert_eq(kdf.fillna(s_nan),
>                        pdf.fillna(s_nan))
> ```
> 
>
`fillna()` takes a series as input. I think we can keep it in `test_dataframe` because the method exists only for dataframe.

garawalid:
These tests are no longer working for `series` after #235 
```python 
        self.assertTrue(isinstance(ks['x'] + 1, Series))
        self.assertTrue(isinstance(1 + ks['x'], Series))

        self.assertTrue(isinstance(ks.x + 1, Series))
        self.assertTrue(isinstance(1 + ks.x, Series))
```

HyukjinKwon:
Merged. Thanks.

PR #235, opened 2019-05-04
Fix string addition for Koalas Series

### LINKED ISSUE #219: String concatenation does not work correctly
When concatenating columns and assigning to another column works for numeric columns, it does not work for strings:

Works for numeric columns:
```python
df["a_b"] = df["a_b"] + df["a_b"]
df.toPandas()

# Out[29]: 
#    id  a.b  a_b
# 0   0    1    4
```

Does not work for strings:
```python
from pyspark.sql.functions import lit
spark.range(1).withColumn("a", lit("abc")).withColumn("b", lit("def")).withColumn("c", lit("+")).coalesce(1).write.mode("overwrite").parquet("/mnt/ivan/sample2.parquet")

df = ks.read_parquet("/mnt/ivan/sample2.parquet")
df["c"] = df["a"] + df["b"]
df.toPandas()

# Out[36]: 
#    id    a    b     c
# 0   0  abc  def  None
```

However, this works in Pandas:
```python
df = ks.read_parquet("/mnt/ivan/sample2.parquet").toPandas()
df["c"] = df["a"] + df["b"]
df

# Out[38]: 
#    id    a    b       c
# 0   0  abc  def  abcdef
```

sadikovi:
This PR updates `__add__` and `__radd__` functions to handle string concatenation. Before this patch, when adding two string columns or string column and literal, you get `None`, because the library would use Spark columns addition.

I updated the function to use `concat` Spark SQL function instead, this applies for cases of two columns or column and literal. Added unit tests to verify the intended behaviour.

Pandas does not support 'str' + 'int' or 'int' + 'str', or other cross types additions; so I only added handling for strings.

> `__add__` function became a bit messy, because of that check. Also I found that error messages are not consistent between Koalas and Pandas when it comes to adding columns of different types.

Fixes #219

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -25,8 +25,8 @@ import pandas as pd
 
 from pyspark import sql as spark
 from pyspark.sql import functions as F
-from pyspark.sql.types import BooleanType, FloatType, DoubleType, LongType, StructType, \
-    TimestampType, to_arrow_type
+from pyspark.sql.types import BooleanType, FloatType, DoubleType, LongType, StringType, \
+    StructType, TimestampType, to_arrow_type
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
 from databricks.koalas.dask.utils import derived_from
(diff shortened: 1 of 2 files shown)


rxin:
Thanks Ivan. Merged.

PR #236, opened 2019-05-05
Documentation for Series.py

abishekganesh72:
Added Remaining documentation for series.py 
Ref #152  

- [x] series.py-    def rename(self, index=None, **kwargs):
- [x] series.py-    def reset_index(self, level=None, drop=False, name=None, inplace=False):
- [x] series.py-    def isnull(self):
- [x] series.py-    def notnull(self):
- [x] series.py-    def dropna(self, axis=0, inplace=False, **kwargs):
- [x] series.py-    def head(self, n=5):
- [x] series.py-    def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, dropna=True)

### DIFF
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -23,6 +23,7 @@ from databricks import koalas
 def add_ks(doctest_namespace):
     doctest_namespace['ks'] = koalas
 
+
 @pytest.fixture(autouse=True)
 def add_np(doctest_namespace):
     doctest_namespace['np'] = np
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -14,9 +14,10 @@
 # limitations under the License.
 #
 import pytest
+import numpy as np
 
 from databricks import koalas
-import numpy as np
+
 
 @pytest.fixture(autouse=True)
 def add_ks(doctest_namespace):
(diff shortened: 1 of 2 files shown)


abishekganesh72:
@thunterdb  @rxin  This is ready for CR. Let me know if anything need to be changed

databricks/koalas/series.py
```diff
@@ -227,8 +227,62 @@ def name(self):
...
+    #  1. function, changes labels
+    #  2.  mapping, changes labels
     def rename(self, index=None, **kwargs):
+        """
+        Alter axes input function or functions. Function / dict values must be unique (1-to-1).
+        Labels not contained in a dict / Series will be left as-is.
+        Extra labels listed don’t throw an error.
+        Alternatively, change Series.name with a scalar value (Series only).
+
+        Parameters
```
rxin: Delete reference to parameters that we don’t support.

databricks/conftest.py
```diff
@@ -14,10 +14,16 @@
 # limitations under the License.
 #
 import pytest
+import numpy as np
```
HyukjinKwon: not a big deal but let's match with koalas import style

```
import numpy

...

    doctest_namespace['np'] = numpy
```

abishekganesh72:
Updated the documentation.

PR #239, opened 2019-05-05
DataFrame.to_csv

garawalid:
Implementation of `to_csv()`. Part of #169

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -18,7 +18,6 @@
 A wrapper class for Spark DataFrame to behave similar to pandas DataFrame.
 """
 import warnings
-from distutils.version import LooseVersion
 from functools import partial, reduce
 from typing import Any, List, Tuple, Union
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -580,8 +580,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         Write object to a comma-separated values (csv) file.
 
-            .. note:: This method should only be used if the resulting CSV is expected
-                to be small, as all the data is loaded into the driver's memory.
+        .. note:: This method should only be used if the resulting CSV is expected
+            to be small, as all the data is loaded into the driver's memory.
 
         Parameters
         ----------
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -571,6 +571,104 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         return validate_arguments_and_invoke_function(
             kdf.to_pandas(), self.to_dict, pd.DataFrame.to_dict, args)
 
+    def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
+               columns=None, header=True, index=True, index_label=None,
+               mode='w', encoding=None, compression='infer', quoting=None,
+               quotechar='"', line_terminator=None, chunksize=None,
+               tupleize_cols=None, date_format=None, doublequote=True,
+               escapechar=None, decimal='.'):
+        r"""
+        Write object to a comma-separated values (csv) file.
+            The order of arguments for Series was changed.
+        Parameters
+        ----------
+        path_or_buf : str or file handle, default None
+            File path or object, if None is provided the result is returned as
+            a string.  If a file object is passed it should be opened with
+            `newline=''`, disabling universal newlines.
+               Was previously named "path" for Series.
+        sep : str, default ','
+            String of length 1. Field delimiter for the output file.
+        na_rep : str, default ''
+            Missing data representation.
+        float_format : str, default None
+            Format string for floating point numbers.
+        columns : sequence, optional
+            Columns to write.
+        header : bool or list of str, default True
+            Write out the column names. If a list of strings is given it is
+            assumed to be aliases for the column names.
+               Previously defaulted to False for Series.
+        index : bool, default True
+            Write row names (index).
+        index_label : str or sequence, or False, default None
+            Column label for index column(s) if desired. If None is given, and
+            `header` and `index` are True, then the index names are used. A
+            sequence should be given if the object uses MultiIndex. If
+            False do not print fields for index names. Use index_label=False
+            for easier importing in R.
+        mode : str
+            Python write mode, default 'w'.
+        encoding : str, optional
+            A string representing the encoding to use in the output file,
+            defaults to 'ascii' on Python 2 and 'utf-8' on Python 3.
+        compression : str, default 'infer'
+            Compression mode among the following possible values: {'infer',
+            'gzip', 'bz2', 'zip', 'xz', None}. If 'infer' and `path_or_buf`
+            is path-like, then detect compression from the following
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -492,6 +492,104 @@ def to_dict(self, orient='dict', into=dict):
         return validate_arguments_and_invoke_function(
             kdf.to_pandas(), self.to_dict, pd.DataFrame.to_dict, args)
 
+    def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
+               columns=None, header=True, index=True, index_label=None,
+               mode='w', encoding=None, compression='infer', quoting=None,
+               quotechar='"', line_terminator=None, chunksize=None,
+               tupleize_cols=None, date_format=None, doublequote=True,
+               escapechar=None, decimal='.'):
+        r"""
```
rxin: what's up with the "r" here?

databricks/koalas/frame.py
```diff
@@ -492,6 +492,104 @@ def to_dict(self, orient='dict', into=dict):
...
+        Write object to a comma-separated values (csv) file.
```
rxin: for some reason all the blank lines are gone. can you copy paste again from the raw file, rather than directly on github? i noticed that github's copy paste would sometimes result in loss of blank lines.

databricks/koalas/frame.py
```diff
@@ -492,6 +492,104 @@ def to_dict(self, orient='dict', into=dict):
...
+            The order of arguments for Series was changed.
```
rxin: also need a note warning saying this could blow up. you can copy the warning from to_pandas

databricks/koalas/frame.py
```diff
@@ -492,6 +492,104 @@ def to_dict(self, orient='dict', into=dict):
...
+        ...                    'mask': ['red', 'purple'],
+        ...                    'weapon': ['sai', 'bo staff']})
+        >>> df.to_csv(index=False)
+        'name,mask,weapon\nRaphael,red,sai\nDonatello,purple,bo staff\n'
+        """
+        # Make sure locals() call is at the top of the function so we don't capture local variables.
+        args = locals()
+        kdf = self
+
+        if line_terminator is None and pd.__version__ < "0.24":
```
rxin: maybe just change our signature to have the line_terminator be \n?

ueshin: Could you use `LooseVersion` instead of comparing directly?

garawalid: I'm not sure if all methods of pandas (version less than 0.24 ) are affected or only this one, that's why I kept it outside `validate_arguments_and_invoke_function`

rxin: If you choose `\n` explicitly as the default value for us, then we wouldn't have any problem right regardless of the pandas version?

garawalid: I tested with Pandas 0.24 and it works fine! Thanks.

databricks/koalas/frame.py
garawalid: Without the "r" docstring throws this error `inconsistent leading whitespace`. 
I solved in a different way!

databricks/koalas/frame.py
garawalid: Removing the "r" will solve the problem.

databricks/koalas/frame.py
garawalid: Agree !

garawalid:
@rxin @HyukjinKwon 
Thanks for your comments. Would you mind reviewing the method?

databricks/koalas/frame.py
```diff
@@ -492,6 +492,104 @@ def to_dict(self, orient='dict', into=dict):
...
+            Character recognized as decimal separator. E.g. use ',' for
+            European data.
+
+        Returns
+        -------
+        None or str
+            If path_or_buf is None, returns the resulting csv format as a
+            string. Otherwise returns None.
+
+            .. note:: This method should only be used if the resulting CSV is expected
```
rxin: i'd put this note at the top similar to other functions, so it's more obvious.

databricks/koalas/frame.py
garawalid: Putting "\n" as default value breaks the test for Pandas 0.23. The result is different!
I think it's better we keep it like before. What do you think? 
```
588         Examples
589         --------
590         >>> df = ks.DataFrame({'name': ['Raphael', 'Donatello'],
591         ...                    'mask': ['red', 'purple'],
592         ...                    'weapon': ['sai', 'bo staff']})
593         >>> df.to_csv(index=False)
Expected:
    'name,mask,weapon\nRaphael,red,sai\nDonatello,purple,bo staff\n'
Got:
    'mask,name,weapon\nred,Raphael,sai\npurple,Donatello,bo staff\n'
```

garawalid:
I skipped the doctest, the result is different because of python3.5. 
I test it locally with Pandas 0.23.4 and python 3.6 and it works well! 
```
Expected:
    'name,mask,weapon\nRaphael,red,sai\nDonatello,purple,bo staff\n'
Got:
    'mask,name,weapon\nred,Raphael,sai\npurple,Donatello,bo staff\n'
```

databricks/koalas/frame.py
```diff
@@ -500,6 +501,107 @@ def to_dict(self, orient='dict', into=dict):
...
+        None or str
+            If path_or_buf is None, returns the resulting csv format as a
+            string. Otherwise returns None.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'name': ['Raphael', 'Donatello'],
+        ...                    'mask': ['red', 'purple'],
+        ...                    'weapon': ['sai', 'bo staff']})
+        >>> df.to_csv(index=False) # doctest: +SKIP
```
ueshin: Could you try with the explicit `columns=[...]`?

```py
    >>> df = ks.DataFrame({'name': ['Raphael', 'Donatello'],
    ...                    'mask': ['red', 'purple'],
    ...                    'weapon': ['sai', 'bo staff']},
    ...                   columns=['name', 'mask', 'weapon'])
```

garawalid: @ueshin thanks for the tip! It works.

databricks/koalas/frame.py
garawalid: @rxin now `\n` is the default value of `line_terminator`. 
The previous error is due to python 3.5

databricks/koalas/frame.py
```diff
@@ -571,6 +571,105 @@ def to_dict(self, orient='dict', into=dict):
...
+    def to_csv(self, path_or_buf=None, sep=",", na_rep='', float_format=None,
+               columns=None, header=True, index=True, index_label=None,
+               mode='w', encoding=None, compression='infer', quoting=None,
+               quotechar='"', line_terminator="\n", chunksize=None,
+               tupleize_cols=None, date_format=None, doublequote=True,
+               escapechar=None, decimal='.'):
+        """
+        Write object to a comma-separated values (csv) file.
+
+            .. note:: This method should only be used if the resulting CSV is expected
```
HyukjinKwon: I think we can remove 4 leading spaces for this note

PR #241, opened 2019-05-06
Fix python 3.5 build.

ueshin:
Currently Python 3.5 build in travis-ci is not working properly since the Python is upgraded while setting up the environment because the dev dependency is strictly versioned but some of them depends only Python 3.6 or upper.

E.g., this (https://travis-ci.com/databricks/koalas/jobs/197751790) is the 3.5 build, but we can see:

Screen Shot 2019-05-06 at 20 05 04

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -183,6 +183,7 @@ class DataFrame(_Frame):
         >>> df = ks.DataFrame({'species': ['bear', 'bear', 'marsupial'],
         ...                    'population': [1864, 22000, 80000]},
         ...                     index=['panda', 'polar', 'koala'])
+        >>> df = df[['species', 'population']]
         >>> df
                  species  population
         panda       bear        1864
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -1346,7 +1346,7 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
               for colname, asc in zip(by, ascending)]
         kdf = DataFrame(self._sdf.sort(*by), self._metadata.copy())
         if inplace:
-            self._sdf: spark.DataFrame = kdf._sdf
+            self._sdf = kdf._sdf
```
ueshin: Variable Annotations is supported from Python 3.6. https://www.python.org/dev/peps/pep-0526/

arisehkawamura:
When will this fix be released to PyPI?

HyukjinKwon:
0.3.0 is released now.

arisehkawamura:
Thank you!

thunterdb:
@ueshin we lost a few annotations in the process. Can you add them as comments (`... # type: ks.DataFrame`) so that mypi can still pick up the types? Thanks!

PR #242, opened 2019-05-06
Cleanup isin and remove list type check util for now

HyukjinKwon:
This PR cleans up isin and removes list type checker introduced together.
Let's do the type check one separately for other APIs too.


HyukjinKwon:
I am merging this. This is just a cleanup.

Issue #151, opened 2019-04-23
Enforce type annotation in the code base


rxin:
Python can get very difficult to understand for large projects, and one of the main reasons for that is the lack of explicit & static typing. We can mitigate that with type annotations. I think we should update the existing code and enforce that in Koalas.

This is important to do before more contributors come to the project.

HyukjinKwon:
Actually, we enforced that the given type hints are correct at least. Do we target to write all type hint against this issue? if so I think we should delay after 0.7.0 ....

itholic:
I think we can close this since we added type annotation to almost all important API files such as `frame.py`, `series.py`, `indexes.py`, etc.

WDYT, @HyukjinKwon , @ueshin and @xinrong-databricks ??

HyukjinKwon:
We already finished typing and plan to finish most of them in Q3. Let's open a JIRA in Apache Spark to document contributors to do the typing.

HyukjinKwon:
Contribution guides are all done in PySpark documentation side too. Finally I am tentatively closing this.

PR #245, opened 2019-05-06
Refactor ml module

floscha:
This PR does some basic refactoring of the `koalas.ml` module. For example, it adds type annotations (as suggested in #151), moves constants to respective variables, fixes docstrings, and removes unnecessary code.


databricks/koalas/ml.py
```diff
@@ -14,13 +14,24 @@
...
 
-def corr(kdf, method='pearson'):
+if TYPE_CHECKING:
+    import databricks.koalas as ks
+
+
+CORRELATION_OUTPUT_COLUMN = '_correlation_output'
+
+
+def corr(kdf: 'ks.DataFrame', method: str = 'pearson') -> pd.DataFrame:
```
rxin: one thing that i think would really help with making this more understandable is to add an example (doctest?).

databricks/koalas/ml.py
```diff
@@ -43,13 +54,13 @@ def corr(kdf, method='pearson'):
     return arr
 
 
-def to_numeric_df(kdf):
+def to_numeric_df(kdf: 'ks.DataFrame') -> Tuple[pyspark.sql.DataFrame, List[str]]:
     """
     Takes a dataframe and turns it into a dataframe containing a single numerical
     vector of doubles. This dataframe has a single field called '_1'.
```
rxin: same here ... an example would be great

rxin:
Looks great. Thanks @floscha. I left some comments, mostly about using doctest to help with readability.

floscha:
Hi @rxin, I've added doctests for both methods and made sure they pass. Tbh, I believe the `ml` module could need some further refactoring but I'd leave those to a future PR :wink:

rxin:
Thanks. Merging.

databricks/koalas/ml.py
```diff
@@ -31,10 +42,15 @@ def corr(kdf, method='pearson'):
...
                    * spearman : Spearman rank correlation
     :return: :class:`pandas.DataFrame`
+
+    >>> ks.DataFrame({'A': [0, 1], 'B': [1, 0], 'C': ['x', 'y']}).corr()
+         A    B
+    A  1.0 -1.0
+    B -1.0  1.0
     """
-    assert method in ('pearson', 'spearman'), method
+    assert method in ('pearson', 'spearman')
```
thunterdb: why?

databricks/koalas/ml.py
```diff
@@ -14,13 +14,24 @@
...
 from pyspark.ml.feature import VectorAssembler
 from pyspark.ml.stat import Correlation
 
 
-def corr(kdf, method='pearson'):
+if TYPE_CHECKING:
+    import databricks.koalas as ks
+
+
+CORRELATION_OUTPUT_COLUMN = '_correlation_output'
```
thunterdb: it is good style to prefix it with `_koalas` to make sure it does not clash

databricks/koalas/ml.py
thunterdb: That function is more meant to be internal

databricks/koalas/ml.py
floscha: What was the point of the `, method` at the end of the line?

databricks/koalas/ml.py
floscha: Fair enough. I might change it to `_koalas_corr` in a future pull request.

PR #251, opened 2019-05-07
Remove decorator dependency

HyukjinKwon:
Looks we can remove `decorator` dependency. This library looks not that much popular (https://github.com/micheles/decorator) and also every usecase seems able to be rewritten Python-natively.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -52,14 +52,14 @@ class DataFrame(_Frame):
     @derived_from(pd.DataFrame)
     def __init__(self, data=None, index=None, columns=None, dtype=None, copy=False):
         if isinstance(data, pd.DataFrame):
-            self._init_from_pandas(data, index, columns, dtype, copy)
+            self._init_from_pandas(data)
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -51,12 +50,15 @@ class DataFrame(_Frame):
...
+        if isinstance(data, pd.DataFrame):
+            self._init_from_pandas(data, index, columns, dtype, copy)
+        elif isinstance(data, spark.DataFrame):
+            self._init_from_spark(data, index, columns, dtype, copy)
+        else:
+            pdf = pd.DataFrame(data=data, index=index, columns=columns, dtype=dtype, copy=copy)
+            self._init_from_pandas(pdf)
 
-    @__init__.register(pd.DataFrame)
     def _init_from_pandas(self, pdf, *args):
```
ueshin: I guess we can remove `*args` now. It was a convention of `@dispatch_on`.

ueshin:
Otherwise LGTM.

HyukjinKwon:
Thanks, @ueshin . Merged.

PR #253, opened 2019-05-07
Add 'columns' parameter to DataFrame's drop method

floscha:
This PR adds a `columns` parameter to the DataFrame's `drop()` method that behaves exactly like the corresponding parameter in pandas. While it is certainly a matter of taste, I prefer writing `df.drop(columns=['y', 'z'])` over `df.drop(['y', 'z'], axis=1)` and appreciate having compatible APIs between pandas and Koalas. Apart from implementing the described functionality, the PR adds docstests and makes sure no existing tests fail.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1168,7 +1168,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         Drop specified labels from columns.
 
-        Remove columns by specifying label names and axis=1.
+        Remove columns by specifying label names and axis=1 or columns.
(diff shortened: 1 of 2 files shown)


rxin:
Can you add a test case that both labels and columns are specified? Not in a doctest but in the tests directory. And maybe add a line to the docstring indicating what the behavior is when both are specified.

floscha:
I've now added a bunch of unit tests as well as the requested comment in the docstring.

rxin:
Looks like it failed the pep8 check.

floscha:
I've fixed the pep8 issues now.

rxin:
Thanks. Merged!

PR #255, opened 2019-05-07
Design principles

rxin:
Created a design principles section in README.md, and moved the old development guide into CONTRIBUTING.md.

Also added a new FAQ question: Is it Koalas or koalas?


README.md
```diff
@@ -83,6 +92,77 @@ Project docs are published here: https://koalas.readthedocs.io
...
+
+- Koalas docs' style and infrastructure simply follow rest of the PyData projects'.
+
+### Unify small data (pandas) API and big data (Spark) API, but pandas first
+
+The Koalas DataFrame is meant to provide the best of pandas and Spark under a single API, with easy and clear conversions between each API when necessary. When Spark and pandas have similar APIs with subtle differences, the principle is to honor the contract of the pandas API first.
+
+There are 4 different classes of functions:
+
+ 1. Functions that are only found in Spark (`select`, `selectExpr`). These functions should also beavailable in Koalas.
```
garawalid: A small type: also be available

README.md
```diff
@@ -74,97 +79,92 @@ df.columns = ['x', 'y', 'z1']
...
-# Run all unittest and doctest
-./dev/pytest
-```
+Koalas targets Python data scientists. We want to stick to the convention that users are already familiar with as much as possible. Here are some examples:
 
-To run a specific test file:
-```bash
-# Run unittest
-./dev/pytest -k test_dataframe.py
+- Function names and parameters use snake_case, rather than CamelCase. This is different from PySpark's design. For example, Koalas has `to_pandas()`, whereas PySpark has `toPandas()` for converting a DataFrame into a pandas DataFrame. In limited cases, to maintain compatibility with Spark, we also provide Spark's variant as an alias.
```
HyukjinKwon: Just to add some official reasoning, this actually duplicates PEP 8 (https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions).

With this exception:

> mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility.

`threading.py` introduced [Java style `mixedCase` to be similar with Java's `Thread`](https://github.com/python/cpython/blob/master/Lib/threading.py#L17-L19), which is similar case with PySpark.

Therefore, allowing `mixedCase` to match with PySpark makes sense in a way with PEP 8 too.

README.md
```diff
@@ -74,97 +79,92 @@ df.columns = ['x', 'y', 'z1']
...
 
-# Run doctest
-./dev/pytest -k series.py --doctest-modules databricks
-```
+- Koalas respects to the largest extent the conventions of the Python numerical ecosystem, and allows the use of numpy types, etc. that are supported by Spark.
```
HyukjinKwon: not a big deal but I believe numpy is NumPy (https://www.numpy.org/)

HyukjinKwon: "that are supported by Spark.". I don't think officially PySpark supports NumPy types tho .. Maybe I am too worried but how about say, instead, "that can be supported by Spark"?

README.md
```diff
@@ -74,97 +79,92 @@ df.columns = ['x', 'y', 'z1']
...
+### Unify small data (pandas) API and big data (Spark) API, but pandas first
 
-Note that `-k` is used for simplicity although it takes an expression. You can use `--verbose` to check what to filter. See `pytest --help` for more details.
+The Koalas DataFrame is meant to provide the best of pandas and Spark under a single API, with easy and clear conversions between each API when necessary. When Spark and pandas have similar APIs with subtle differences, the principle is to honor the contract of the pandas API first.
 
-### Building Documentation
+There are 4 different classes of functions:
 
-To build documentation via Sphinx:
+ 1. Functions that are only found in Spark (`select`, `selectExpr`). These functions should also be available in Koalas.
```
HyukjinKwon: I still am not sure if we should allow those. `select` conflicts with Pandas's (https://github.com/pandas-dev/pandas/blob/88062f75dbca929ec082295c936edd07cc912dbf/pandas/core/generic.py#L3653-L3669):

```python
>>> pd.DataFrame([{'a': 1}]).select
<bound method NDFrame.select of    a
0  1>
```

`selectExpr` means we might face later if we should allow, for instance, `filter` should accept string expressions as well, which conflicts with Pandas's.

I was wondering if we, instead, say:

Spark APIs can be added only with strong reasons when they don't conflict with Pandas.

README.md
```diff
@@ -74,97 +79,92 @@ df.columns = ['x', 'y', 'z1']
...
 
-```bash
-cd docs && make clean html
-```
+ 2. Functions that are found in Spark but that have a clear equivalent in pandas, e.g. `alias` and `rename`. These functions will be implemented as the alias of the pandas function, but should be marked that they are aliases of the same functions. They are provided so that existing users of PySpark can get the benefits of Koalas without having to adapt their code.
```
HyukjinKwon: I think `alias` and `rename` are not necessary. Existing users of PySpark will be confused what APIs are exposed and not. I was thinking it's unclear what Koalas API expects. This can be worked around via `to_spark`. We can expose another API that truncates Spark's index.

PySpark doesn't have index so when PySpark users use `to_spark(index='truncate').alias(...)` won't get confused.

README.md
```diff
@@ -74,97 +79,92 @@ df.columns = ['x', 'y', 'z1']
...
-### Coding Conventions
-We follow [PEP 8](https://www.python.org/dev/peps/pep-0008/) with one exception: lines can be up to 100 characters in length, not 79.
+ 4. Functions that are only found in pandas. When these functions are appropriate for distributed datasets, they should become available in Koalas.
 
-### Release Instructions
-Only project maintainers can do the following.
+### Return Koalas data structure for big data, and pandas data structure for small data
 
-Step 1. Make sure version is set correctly in `databricks/koalas/version.py`.
+Often developers face the question whether a particular function should return a Koalas DataFrame/Series, or a pandas DataFrame/Series. The principle is: if the returned object can be large, use a Koalas DataFrame/Series. If the data is bound to be small, use a pandas DataFrame/Series. For example, `DataFrame.dtypes` return a pandas Series, because the number of columns in a DataFrame is bounded and small, whereas `DataFrame.head()` or `Series.unique()` returns a Koalas DataFrame, because the resulting object can be large.
```
HyukjinKwon: > whereas `DataFrame.head()` or `Series.unique()` returns a Koalas DataFrame, because the resulting object can be large.

typos ..

whereas `DataFrame.head()` or `Series.unique()` returns a Koalas DataFrame/Series, because the resulting object can be large.

README.md
```diff
@@ -74,97 +79,92 @@ df.columns = ['x', 'y', 'z1']
...
+- [`DataFrame.to_pandas()`](https://koalas.readthedocs.io/en/stable/reference/api/databricks.koalas.DataFrame.to_pandas.html) : returns a pandas DataFrame, koalas only
+- [`DataFrame.to_numpy()`](https://koalas.readthedocs.io/en/stable/reference/api/databricks.koalas.DataFrame.to_numpy.html): returns a numpy array, works with both pandas and Koalas
+
+Note that it is clear from the names that these functions return some local data structure that would require materializing data in a single node's memory. For these functions, we also explicitly document them with a warning note that the resulting data structure must be small.
+
+### Be a lean API layer and move fast
+
+Koalas is designed as an API overlay layer on top of Spark. The project should be lightweight, and most functions should be implemented as wrappers around Spark or pandas. Koalas does not accept heavyweight implementations, e.g. execution engine changes.
+
+This approach enables us to move fast. For the considerable future, we aim to be making weekly releases.
```
HyukjinKwon: Wow .. weekly release ..

README.md
rxin: sounds good

README.md
thunterdb: There is indeed a select function, but this is fine because the expected input is different and it is deprecated, see the doc here:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.select.html#pandas.DataFrame.select

It does not conflict with pandas in the sense that we can add the full pandas behaviour later. For now, like we do with functions that are partially implemented, we throw a notimplementederror for the pandas arguments.

Given the ubiquity of select in spark, I believe we should add it.

README.md
thunterdb: These functions are used everywhere in pyspark user code, so I would hate to have to rename everything to use koalas. Also, kdf.to_spark(index='truncate').alias(...).to_koalas() is also way too long to be useful.

README.md
thunterdb: Making a release is a very quick process. We could even aim at every 3-4 days.

README.md
```diff
@@ -180,10 +180,14 @@ You should expect the following differences:
...
    if you come across any discrepancies
-   
+
  - because Spark is lazy in nature, some operations like creating new columns only get 
    performed when Spark needs to print or write the dataframe.
-   
+
+### Is it Koalas or koalas?
+
+It's Koalas. Unlike pandas, we use upper case here.
```
thunterdb: Thanks :)

README.md
HyukjinKwon: I think at least we shouldn't pick `select` as an example in this doc. and .. if you won't mind, I would stick to "Spark APIs can be added only with strong reasons when they don't conflict with Pandas." ..

README.md
HyukjinKwon: If this is added since it's commonly used, I am okie.

README.md
rxin: sounds great.

README.md
rxin: I changed the wording to say 

```There are different classes of functions:

 1. Functions that are found in both Spark and pandas under the same name (`count`, `dtypes`, `head`). The return value is the same as the return type in pandas (and not Spark's).
    
 2. Functions that are found in Spark but that have a clear equivalent in pandas, e.g. `alias` and `rename`. These functions will be implemented as the alias of the pandas function, but should be marked that they are aliases of the same functions. They are provided so that existing users of PySpark can get the benefits of Koalas without having to adapt their code.
 
 3. Functions that are only found in pandas. When these functions are appropriate for distributed datasets, they should become available in Koalas.
 
 4. Functions that are only found in Spark that are essential to controlling the distributed nature of the computations, e.g. `cache`. These functions should be available in Koalas.

We are still debating whether data transformation functions only available in Spark should be added to Koalas, e.g. `select`. We would love to hear your feedback on that.```

HyukjinKwon:
LGTM. Thanks, @rxin and all.

rxin:
I made a mistake that the merge didn't include my final commit. I had to push that commit directly to master.

PR #258, opened 2019-05-08
repr and repr_html improvements

rxin:
1. Fixed the bug that `DataFrame. __repr__` was turning the entire DataFrame into pandas DataFrame, rather than simply taking the top 1000 rows.

2. Added a cache so repeated calls of repr and repr_html don't trigger multiple Spark actions. But then I undid that because cache invalidation is difficult. I did leave some of the infrastructure still in place and added test cases to make sure repr and repr_html are correct when it comes to cache invalidation, if we ever implement any form of caching.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -29,8 +29,7 @@ from pyspark.sql.types import BooleanType, StructField, StructType, to_arrow_typ
 from pyspark.sql.utils import AnalysisException
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
-from databricks.koalas.utils import default_session, lazy_property, \
-    validate_arguments_and_invoke_function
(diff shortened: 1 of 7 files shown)


databricks/koalas/frame.py
```diff
@@ -86,6 +87,11 @@ def _index_columns(self):
         return [self._sdf.__getitem__(field)
                 for field in self._metadata.index_fields]
 
+    @lazy_property
```
HyukjinKwon: @rxin, I think we should just leave it as `property` for now. I think some APIs have `inplace` options and I suspect that won't work with cache.

rxin: Yea you are right. I got rid of it completely. There's no point doing property since it just shortens a little bit of code.

rxin:
Eh the test cases fail due to an existing flakiness issue: sort_values are not deterministic when there are two elements that are the same.

Issue #252, opened 2019-05-07
got ImportError when 


paulxll:
I cannot run import statement `import databricks.koalas as ks` on Databricks(Databricks Runtime 5.3).

I tried to run that codes but it doesn't work. 
```
dbutils.library.installPyPI("Koalas","0.3.0")
dbutils.library.installPyPI("numpy","1.12.0")
dbutils.library.restartPython()
```
```
import databricks.koalas as ks
```

I got error message below.
```

ImportError: numpy.core.multiarray failed to import
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
RuntimeError: module compiled against API version 0xc but this version of numpy is 0xa---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<command-962463> in <module>()
----> 1 import databricks.koalas as ks

/local_disk0/pythonVirtualEnvDirs/virtualEnv-d969741a-2a67-4210-b86b-1ec4d29760f1/lib/python3.5/site-packages/databricks/koalas/__init__.py in <module>()
     36 assert_pyspark_version()
     37 
---> 38 from databricks.koalas.namespace import *
     39 from databricks.koalas.frame import DataFrame
     40 from databricks.koalas.series import Series

/local_disk0/pythonVirtualEnvDirs/virtualEnv-d969741a-2a67-4210-b86b-1ec4d29760f1/lib/python3.5/site-packages/databricks/koalas/namespace.py in <module>()
     28 from databricks.koalas.dask.compatibility import string_types
     29 from databricks.koalas.utils import default_session
---> 30 from databricks.koalas.frame import DataFrame, _reduce_spark_multi
     31 from databricks.koalas.typedef import Col, pandas_wrap
     32 from databricks.koalas.series import Series

/local_disk0/pythonVirtualEnvDirs/virtualEnv-d969741a-2a67-4210-b86b-1ec4d29760f1/lib/python3.5/site-packages/databricks/koalas/frame.py in <module>()
     39 from databricks.koalas.ml import corr
     40 from databricks.koalas.selection import SparkDataFrameLocator
---> 41 from databricks.koalas.typedef import infer_pd_series_spark_type
     42 
     43 

/local_disk0/pythonVirtualEnvDirs/virtualEnv-d969741a-2a67-4210-b86b-1ec4d29760f1/lib/python3.5/site-packages/databricks/koalas/typedef.py in <module>()
     25 import pandas as pd
[...]

PR #260, opened 2019-05-08
Explicitly specify numpy minimum version.

ueshin:
Although `pyarrow==0.10` says numpy minimum version is `1.10.0`, it does not work properly with `numpy<0.14`, so we should explicitly specify the numpy minimum version. This is related to #252.


setup.py
```diff
@@ -55,8 +55,8 @@
     python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*',
     install_requires=[
         'pandas>=0.23',
-        'decorator',
```
HyukjinKwon: :)

PR #262, opened 2019-05-08
Inline doc for DataFrame.pipe

nitlev:
This PR has inline documentation for the `koalas.DataFrame.pipe` method.
Feedbacks are welcome :-)


nitlev:
Hmmm looks like the current doctests are not deterministic. I'll fix that.

rxin:
@nitlev do you think this is good to go now?

nitlev:
@rxin yes it should be fine now

rxin:
Thanks. Merging.

PR #264, opened 2019-05-08
Add basic merge functionality to DataFrame

### LINKED ISSUE #232: koalas.merge
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.merge.html#pandas.merge

This is basically join in Spark.

floscha:
This PR adds the basic [merge](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.merge.html) functionality from pandas to Koalas DataFrames as requested in #232. While the PR does not cover 100% of pandas' functionality, it covers the basic use cases to be useful enough to start with.

Resolves #232

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1474,16 +1474,28 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         Parameters
         ----------
-        right: the DataFrame for the right side of the join operation
-        how: a join method out of ['inner', 'left', 'right', 'full', 'outer']
-        on: the column name to be joined on. Defaults to the index if not specified
-        suffixes: suffix to apply to overlapping column names in the left and right side,
-            respectively
+        right: Object to merge with.
+        how: Type of merge to be performed.
+            {‘left’, ‘right’, ‘outer’, ‘inner’}, default ‘inner’
+
+            left: use only keys from left frame, similar to a SQL left outer join; preserve key
+                order.
+            right: use only keys from right frame, similar to a SQL right outer join; preserve key
+                order.
+            outer: use union of keys from both frames, similar to a SQL full outer join; sort keys
+                lexicographically.
+            inner: use intersection of keys from both frames, similar to a SQL inner join;
+                preserve the order of the left keys.
+        on: Column or index level names to join on. These must be found in both DataFrames. If on
+            is None and not merging on indexes then this defaults to the intersection of the
+            columns in both DataFrames.
+        suffixes: Suffix to apply to overlapping column names in the left and right side,
+            respectively.
 
         Returns
         -------
         DataFrame
-            The joined DataFrame
+            A DataFrame of the two merged objects.
 
         Examples
         --------
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1470,7 +1470,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
     def merge(self, right: 'DataFrame', how: str = 'inner', on: str = None,
               suffixes: Tuple[str, str] = ('_x', '_y')) -> 'DataFrame':
         """
-        Perform a database (SQL) merge operation between two DataFrame objects.
+        Merge DataFrame objects with a database-style join.
 
         Parameters
         ----------
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1549,7 +1549,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             how = 'full'
         if how not in ('inner', 'left', 'right', 'full'):
             raise ValueError("The 'how' parameter has to be amongst the following values: ",
-                             "['inner', 'left', 'right', 'full', 'outer']")
+                             "['inner', 'left', 'right', 'outer']")
 
         if on is None:
             # FIXME Move index string to constant?
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -18,7 +18,7 @@
 A wrapper class for Spark DataFrame to behave similar to pandas DataFrame.
 """
 from functools import partial, reduce
-from typing import Any, List, Union
+from typing import Any, List, Tuple, Union
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -1467,6 +1467,101 @@ def shape(self):
...
+
+        Notes
+        -----
+        As described in #263, joining string columns currently returns None for missing values
+            instead of NaN.
+        """
+        if how == 'full':
+            print("Warning: While Koalas will accept 'full', you should use 'outer' instead to",
+                  "be compatible with the pandas merge API")
+            pass
```
HyukjinKwon: Seems `pass` unnecessary.

databricks/koalas/frame.py
```diff
@@ -1467,6 +1467,101 @@ def shape(self):
         """
         return len(self), len(self.columns)
 
+    def merge(self, right: 'DataFrame', how: str = 'inner', on: str = None,
+              suffixes: Tuple[str, str] = ('_x', '_y')) -> 'DataFrame':
+        """
+        Perform a database (SQL) merge operation between two DataFrame objects.
```
HyukjinKwon: Can we inline the doc to Pandas'?

databricks/koalas/frame.py
floscha: Fair enough. I've removed it now

databricks/koalas/frame.py
floscha: Does it work for you now after 5472af1 and dde5b9c?

databricks/koalas/tests/test_dataframe.py
```diff
@@ -380,3 +380,41 @@ def test_isin(self):
...
+        right_kdf = koalas.DataFrame({'B': ['x', 'y']}, index=[1, 2])
+
+        msg = ("The 'how' parameter has to be amongst the following values: ['inner', 'left', " +
+               "'right', 'full', 'outer']")
+        with self.assertRaises(ValueError, msg=msg):
+            left_kdf.merge(right_kdf, how='foo')
+
+        # Assert inner join
+        res = left_kdf.merge(right_kdf)
+        self.assert_eq(res, pd.DataFrame({'A': [2], 'B': ['x']}))
```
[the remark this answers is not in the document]
floscha: I've now added the `SparkPandasMergeError` with abfe0c9.

databricks/koalas/frame.py
```diff
@@ -1467,6 +1468,124 @@ def shape(self):
...
+        -------
+        DataFrame
+            A DataFrame of the two merged objects.
+
+        Examples
+        --------
+        >>> left_kdf = ks.DataFrame({'A': [1, 2]})
+        >>> right_kdf = ks.DataFrame({'B': ['x', 'y']}, index=[1, 2])
+
+        >>> left_kdf.merge(right_kdf)
```
rxin: in this case what is the join key?

floscha: This is outdated. Now it should be
```python
left_kdf.merge(right_kdf, left_index=True, right_index=True)
```
to join on the indices.

databricks/koalas/frame.py
```diff
@@ -1467,6 +1468,124 @@ def shape(self):
...
+            instead of NaN.
+        """
+        if on is None and not left_index and not right_index:
+            raise SparkPandasMergeError("At least 'on' or 'left_index' and 'right_index' have ",
+                                        "to be set")
+        if on is not None and (left_index or right_index):
+            raise SparkPandasMergeError("Only 'on' or 'left_index' and 'right_index' can be set")
+
+        if how == 'full':
+            print("Warning: While Koalas will accept 'full', you should use 'outer' instead to",
```
HyukjinKwon: Can we use warnings package like `warnings.warn("...", UserWarning)`

HyukjinKwon:
Looks good otherwise. Let me leave the final check to @ueshin .

databricks/koalas/frame.py
ueshin: Need one more space at the end of the string: `instead to "`

databricks/koalas/frame.py
```diff
@@ -1467,6 +1468,124 @@ def shape(self):
...
+                  "be compatible with the pandas merge API")
+        if how == 'outer':
+            # 'outer' in pandas equals 'full' in Spark
+            how = 'full'
+        if how not in ('inner', 'left', 'right', 'full'):
+            raise ValueError("The 'how' parameter has to be amongst the following values: ",
+                             "['inner', 'left', 'right', 'full', 'outer']")
```
ueshin: I don't think we should include `'full'` in the message since it is optional.

floscha: Fair enough. I've removed the `full` option with 0ad39b8.

databricks/koalas/frame.py
floscha: Good point. I've changed it to `warnings.warn(...)` with 4e7c429.

floscha: When comma-separating strings in a `print` statement, Python will automatically add spaces in between. However, since 4e7c429 moves from `print()` to `warnings.warn()`, this has been taken care of.

ueshin:
Thanks! merging to master.

databricks/koalas/frame.py
```diff
@@ -1467,6 +1469,124 @@ def shape(self):
...
+        Notes
+        -----
+        As described in #263, joining string columns currently returns None for missing values
+            instead of NaN.
+        """
+        if on is None and not left_index and not right_index:
+            raise SparkPandasMergeError("At least 'on' or 'left_index' and 'right_index' have ",
+                                        "to be set")
+        if on is not None and (left_index or right_index):
+            raise SparkPandasMergeError("Only 'on' or 'left_index' and 'right_index' can be set")
```
rxin: @floscha should we just raise ValueError rather than introducing a new exception type just for this?

floscha: I think the main reason for introducing a new `SparkPandasMergeError` was the fact that pandas also has a dedicated `MergeError` instead of using the generic `ValueError`. That being said, I personally wouldn't mind using a `ValueError` in this case either 😉

rxin: Got it. Can you submit a PR to just remove this? I find it odd to have a special Error type just for this check .... since it's a different error from pandas' anyway, it's not going to make a difference from the api compatibility point of view.

floscha: Sure, I've just opened #323 to remove it.

PR #268, opened 2019-05-08
Revamp the UDF wrapper to make it more usable

thunterdb:
This revamps and significantly improves / simplifies the pandas wrapper (which is a much more flexible pandas udf wrapper). The exact naming of some of the content like `pandas_wrap` or `Col` can be debated. See the docstring examples for usage and motivation.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -94,7 +94,8 @@ class Series(_Frame):
     """
 
     @derived_from(pd.Series)
-    def __init__(self, data=None, index=None, dtype=None, name=None, copy=False, fastpath=False, anchor=None):
+    def __init__(self, data=None, index=None, dtype=None, name=None, copy=False, fastpath=False,
+                 anchor=None):
         if isinstance(data, pd.Series):
             self._init_from_pandas(data)
         elif isinstance(data, spark.Column):
@@ -162,7 +163,8 @@ class Series(_Frame):
     def __radd__(self, other):
         # Handle 'literal' + df['col']
         if isinstance(self.spark_type, StringType) and isinstance(other, str):
-            return Series(F.concat(F.lit(other), self._scol), anchor=self._kdf, index=self._index_info)
+            return Series(F.concat(F.lit(other), self._scol), anchor=self._kdf,
+                          index=self._index_info)
         else:
             return _column_op(spark.Column.__radd__)(self, other)
 
@@ -466,7 +468,8 @@ class Series(_Frame):
     @derived_from(pd.Series)
     def isnull(self):
         if isinstance(self.schema[self.name].dataType, (FloatType, DoubleType)):
-            return Series(self._scol.isNull() | F.isnan(self._scol), anchor=self._kdf, index=self._index_info)
+            return Series(self._scol.isNull() | F.isnan(self._scol), anchor=self._kdf,
+                          index=self._index_info)
         else:
             return Series(self._scol.isNull(), anchor=self._kdf, index=self._index_info)
 
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -125,6 +125,8 @@ class Series(_Frame):
         :param index_info: index information of this Series.
         """
         assert index_info is not None
+        assert kdf is not None
+        assert isinstance(kdf, ks.DataFrame), type(kdf)
         self._scol = scol
         self._kdf = kdf
         self._index_info = index_info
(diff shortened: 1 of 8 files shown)


databricks/koalas/series.py
```diff
@@ -94,11 +94,11 @@ class Series(_Frame):
     """
 
     @derived_from(pd.Series)
-    def __init__(self, data=None, index=None, dtype=None, name=None, copy=False, fastpath=False):
+    def __init__(self, data=None, index=None, dtype=None, name=None, copy=False, fastpath=False, anchor=None):
         if isinstance(data, pd.Series):
             self._init_from_pandas(data)
         elif isinstance(data, spark.Column):
-            self._init_from_spark(data, index, dtype)
+            self._init_from_spark(data, anchor, index)
```
thunterdb: I noticed that we currently reassign different values and different types to the arguments of the dataframe. This is not the spirit and or recommended style in python. Instead of having a complex mechanism for the constructor, it is much simpler to have a function called `Series` in `namespace.py`.

databricks/koalas/typedef.py
```diff
@@ -151,76 +164,61 @@ def infer_pd_series_spark_type(s: pd.Series) -> types.DataType:
...
-                "Expected a column argument, but got argument of type {} instead".format(type(arg)))
-        s = arg.schema
-        if s != sig_arg.inner:
-            raise ValueError("Passing an argument {} of type {}, but the function only accepts "
-                             "columns of type {} for this argument".format(arg, s, sig_arg))
-    assert False, (arg, sig_arg)
-
-
-def make_fun(f, *args, **kwargs):
+def _make_fun(f, return_type, *args, **kwargs):
```
rxin: can you add type hints?

databricks/koalas/typedef.py
```diff
@@ -255,35 +270,135 @@ def f(*args, **kwargs):
...
+
+    Consider a simple function that operates on pandas series of integers
+    >>> def fun(col1):
+    ...     return col1.apply(lambda x: x * 2) # Arbitrary pandas code.
+    >>> fun(pdf.col1)
+    0    2
+    1    4
+    Name: col1, dtype: int64
+
+    Koalas needs to know the return type in order to make this function accessible to Spark. The following function
```
rxin: we should wrap this line. it's 112 chars

databricks/koalas/typedef.py
```diff
@@ -235,55 +236,162 @@ def clean_fun(*args2):
...
+    1    4
+    Name: col1, dtype: int64
+
+    Koalas needs to know the return type in order to make this function accessible to Spark.
+    The following function uses python built-in typing hint system to hint that this function
+    returns a Series of integers:
+
+    >>> @pandas_wrap
+    ... def fun(col1) -> Col[np.int64]:
+    ...     return col1.apply(lambda x: x * 2) # Arbitrary pandas code.
```
HyukjinKwon: not a big deal but at least two spaces before inlined comment (per pep8)

databricks/koalas/typedef.py
```diff
@@ -29,6 +29,13 @@
 from pyspark.sql.functions import pandas_udf
 import pyspark.sql.types as types
 
+from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
+
+
+__all__ = ['Col', '_make_fun', 'pandas_wrap', 'as_spark_type',
```
HyukjinKwon: I think `__all__` usually doesn't include `_*` functions.

databricks/koalas/typedef.py
```diff
@@ -235,55 +236,162 @@ def clean_fun(*args2):
...
+
+    >>> from databricks.koalas import pandas_wrap, Col
+    >>> pdf = pd.DataFrame({"col1": [1, 2], "col2": [10, 20]}, dtype=np.int64)
+    >>> df = ks.DataFrame(pdf)
+
+    Consider a simple function that operates on pandas series of integers
+
+    >>> def fun(col1):
+    ...     return col1.apply(lambda x: x * 2) # Arbitrary pandas code.
+    >>> fun(pdf.col1)
```
HyukjinKwon: @thunterdb, actually, this looks more like Pandas's `apply`. How about we add `apply` with the restriction that it should be wrapped with `pandas_wrap`? I will take a look for that separately and see if that's possible if you think it's a good idea.

HyukjinKwon:
Actually, let me make a followup. Merged.

PR #273, opened 2019-05-09
Inline docs for all functions in generic.py

rxin:
This is part of #152.

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -19,75 +19,287 @@ A base class to be monkey-patched to DataFrame/Column to behave similar to panda
 """
 from collections.abc import Iterable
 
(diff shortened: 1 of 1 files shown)


databricks/koalas/generic.py
```diff
@@ -19,75 +19,287 @@
...
+        ...     'c': [100, 50, -30, -50]
+        ... })
+        >>> df.abs()
+           a   b    c
+        0  4  10  100
+        1  5  20   50
+        2  6  30   30
+        3  7  40   50
         """
+        # TODO: The first example above should not have "Name: abs(0)".
```
rxin: cc @HyukjinKwon @ueshin there's a bug here maybe you guys can fix later ...

PR #275, opened 2019-05-09
Add Series.apply

HyukjinKwon:
This PR adds `apply` to Series. It works identically with Pandas except that Koalas requires to add a type hint to the input `func`'s return type.

```python
>>> import pandas as pd
>>> import numpy as np
>>>
>>> pser = pd.Series([20, 21, 12], index=['London', 'New York', 'Helsinki'])
>>> kser = ks.Series(pser)
>>> def square(x) -> np.int64:
...     return x ** 2
...
```

```
>>> kser.apply(square)
London      400
New York    441
Helsinki    144
Name: square(0), dtype: int64
```

```
>>> pser.apply(square)
London      400
New York    441
Helsinki    144
dtype: int64
```

While I am here,
 - I renamed `pandas_wrap` -> `pandas_wraps` just to match with `functools.wraps`.
 - Addressed few nits.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -731,9 +731,9 @@ class Series(_Frame):
         """
         Invoke function on values of Series.
 
-        Can be a Python function that only works on the series.
+        Can be a Python function that only works on the Series.
 
-        .. note:: unlike Pandas, it is required for `func` to specify return type hint.
+        .. note:: unlike pandas, it is required for `func` to specify return type hint.
 
         Parameters
         ----------
@@ -750,7 +750,7 @@ class Series(_Frame):
 
         Examples
         --------
-        Create a series with typical summer temperatures for each city.
+        Create a Series with typical summer temperatures for each city.
 
         >>> s = ks.Series([20, 21, 12],
         ...               index=['London', 'New York', 'Helsinki'])
@@ -775,7 +775,7 @@ class Series(_Frame):
 
         Define a custom function that needs additional positional
         arguments and pass these additional arguments using the
-        ``args`` keyword by using `pandas_wraps`.
+        ``args`` keyword
 
         >>> def subtract_custom_value(x, custom_value) -> np.int64:
         ...     return x - custom_value
@@ -788,7 +788,7 @@ class Series(_Frame):
 
 
         Define a custom function that takes keyword arguments
-        and pass these arguments to ``apply``  by using `pandas_wraps`.
+        and pass these arguments to ``apply``
 
         >>> def add_custom_values(x, **kwargs) -> np.int64:
         ...     for month in kwargs:
@@ -802,7 +802,7 @@ class Series(_Frame):
         Name: add_custom_values(0), dtype: int64
 
 
-        Use a function from the Numpy library by using `pandas_wraps`.
+        Use a function from the Numpy library
 
         >>> def numpy_log(col) -> np.float64:
         ...     return np.log(col)
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -17,6 +17,7 @@
 """
 A wrapper class for Spark Column to behave similar to pandas Series.
 """
+import inspect
 from functools import partial, wraps
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


rxin:
hm - I wonder if in the future we can just take a few rows and apply that function to figure out the return type. It's not going to be the most efficient on large datasets but will probably be very handy.

databricks/koalas/series.py
```diff
@@ -725,6 +726,100 @@ def count(self):
...
+        >>> s.apply(square)
+        London      400
+        New York    441
+        Helsinki    144
+        Name: square(0), dtype: int64
+
+
+        Define a custom function that needs additional positional
+        arguments and pass these additional arguments using the
+        ``args`` keyword by using `koalas_wraps`.
```
rxin: what's up with koalas_wraps here? It is not part of the example.

HyukjinKwon:
>  I wonder if in the future we can just take a few rows and apply that function to figure out the return type

Yes .. I was wondering about that too. I guess it's practically useful.

databricks/koalas/series.py
HyukjinKwon: Ahh.. it's mixed with my another PR in my local. Let me revert..

HyukjinKwon:
Tests passed. It looks spurious no.

rxin:
btw the pandas version also supports returning a DataFrame right? As long as the udf returns a series.

HyukjinKwon:
Yup, Pandas version does. Current Koalas' version doesn't.

thunterdb:
@HyukjinKwon this is very cool, thanks! 
@rxin yes `.apply` should accept eventually numpy ufuncs and functions on pandas series. The logic for that is already there in `typedef.py` for pandas series at least.

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
...
+        ...     return np.log(col)
+        >>> s.apply(numpy_log)
+        London      2.995732
+        New York    3.044522
+        Helsinki    2.484907
+        Name: numpy_log(0), dtype: float64
+        """
+        spec = inspect.getfullargspec(func)
+        return_sig = spec.annotations.get("return", None)
+        if return_sig is None:
```
thunterdb: actually, if you want to handle series, you should be able to check if `return_sig` is an instance of `typedef.Col` and in that case, directly call ks.pandas_wraps instead of first calling `wraps`. No need to pass return_col, it will be recalculated.

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
         """
         return self._reduce_for_stat_function(_Frame._count_expr)
 
+    def apply(self, func, args=(), **kwds):
+        """
+        Invoke function on values of Series.
+
+        Can be a Python function that only works on the series.
```
ueshin: nit: `series` -> `Series`?

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
...
+
+        .. note:: unlike Pandas, it is required for `func` to specify return type hint.
```
ueshin: nit: `Pandas` -> `pandas`?

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
...
+        **kwds
+            Additional keyword arguments passed to func.
+
+        Returns
+        -------
+        Series
+
+        Examples
+        --------
+        Create a series with typical summer temperatures for each city.
```
ueshin: nit: `series` -> `Series`?

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
...
+        >>> s.apply(square)
+        London      400
+        New York    441
+        Helsinki    144
+        Name: square(0), dtype: int64
+
+
+        Define a custom function that needs additional positional
+        arguments and pass these additional arguments using the
+        ``args`` keyword by using `pandas_wraps`.
```
ueshin: Do we need to mention about `pandas_wraps` here?

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
...
+
+        >>> s.apply(subtract_custom_value, args=(5,))
+        London      15
+        New York    16
+        Helsinki     7
+        Name: subtract_custom_value(0), dtype: int64
+
+
+        Define a custom function that takes keyword arguments
+        and pass these arguments to ``apply``  by using `pandas_wraps`.
```
ueshin: ditto.

databricks/koalas/series.py
```diff
@@ -726,6 +727,100 @@ def count(self):
...
+        ...     return x
+
+        >>> s.apply(add_custom_values, june=30, july=20, august=25)
+        London      95
+        New York    96
+        Helsinki    87
+        Name: add_custom_values(0), dtype: int64
+
+
+        Use a function from the Numpy library by using `pandas_wraps`.
```
ueshin: ditto.

HyukjinKwon:
Thanks guys. let me address them soon.

databricks/koalas/series.py
HyukjinKwon: Actually .. I tried that and then the problem was return type. The return type should be a dataframe if `func` returns a series - so return type should struct-type equivalent (like `struct<_1: int, _2: int, _3: int, ...>`).

```python
>>> pdf = pd.Series([1, 2, 3]).apply(lambda x: pd.Series([1, 2, 3, 4]))
>>> pdf.dtypes
0    int64
1    int64
2    int64
3    int64
dtype: object
>>> pdf
   0  1  2  3
0  1  2  3  4
1  1  2  3  4
2  1  2  3  4
```

I will take a look later about this and try it when I handle series - seems non-trivial.

PR #283, opened 2019-05-09
Inline doc for DataFrame.assign 

HyukjinKwon:
If the test pass, I will just merge and make a followup separately.


shril:
> If the test pass, I will just merge and make a followup separately.

@HyukjinKwon thanks for the help. :)

PR #284, opened 2019-05-09
Complete inlining and remove DASK stubs out

### LINKED ISSUE #152: Inline method docs, rather than using "derived_from"

HyukjinKwon:
This PR proposes two things:

- Get rid of DASK things out in Koalas
- Complete inlining docs

Resolves #152


databricks/koalas/frame.py
```diff
@@ -47,9 +45,73 @@ class DataFrame(_Frame):
...
+        Data type to force. Only a single dtype is allowed. If None, infer
+    copy : boolean, default False
+        Copy data from inputs. Only affects DataFrame / 2d ndarray input
+
+    Examples
+    --------
+    Constructing DataFrame from a dictionary.
+
+    >>> d = {'col1': [1, 2], 'col2': [3, 4]}
+    >>> df = ks.DataFrame(data=d)
```
ueshin: This could be flaky with Python 3.5.

databricks/koalas/frame.py
```diff
@@ -47,9 +45,73 @@ class DataFrame(_Frame):
...
+    >>> df
+       col1  col2
+    0     1     3
+    1     2     4
+
+    Constructing DataFrame from Pandas DataFrame
+
+    >>> df = ks.DataFrame(pd.DataFrame(data=d))
```

HyukjinKwon:
Let me get this in first but please let me know if there are more things to be done.

databricks/koalas/frame.py
```diff
@@ -73,15 +73,15 @@ class DataFrame(_Frame):
...
+    >>> df = ks.DataFrame(data=d, columns=['col1', 'col2'])
     >>> df
        col1  col2
     0     1     3
     1     2     4
 
     Constructing DataFrame from Pandas DataFrame
 
-    >>> df = ks.DataFrame(pd.DataFrame(data=d))
+    >>> df = ks.DataFrame(pd.DataFrame(data=d), columns=['col1', 'col2'])
```
ueshin: `ks.DataFrame(pd.DataFrame(data=d, columns=['col1', 'col2']))`?

HyukjinKwon: Hm, let me address them with adding some assertions.

PR #288, opened 2019-05-09
DataFrame.to_excel

shril:


shril:
Hi @HyukjinKwon,

I am getting this error in Travis CI.
How can I install `openpyxl` module in the build process?

```
self = <pandas.io.excel._OpenpyxlWriter object at 0x7fd307735978>
path = 'output.xlsx', engine = None, engine_kwargs = {}
    def __init__(self, path, engine=None, **engine_kwargs):
        # Use the openpyxl module as the Excel writer.
>       from openpyxl.workbook import Workbook
E       ImportError: No module named 'openpyxl'
```

garawalid:
@shril you need to include `openpyxl` in `requirements-dev.txt`

```
# Dependencies in Koalas
pandas>=0.23
pyarrow>=0.10,<0.11
numpy>=0.14
openpyxl
```

shril:
> @shril you need to include `openpyxl` in `requirements-dev.txt`
> 
> ```
> # Dependencies in Koalas
> pandas>=0.23
> pyarrow>=0.10,<0.11
> numpy>=0.14
> openpyxl
> ```

Just figured that out. Btw thanks @garawalid. :)

databricks/koalas/tests/test_dataframe_conversion.py
```diff
@@ -70,3 +70,34 @@ def test_to_html(self):
             """)
         got = self.strip_all_whitespace(self.kdf.to_html(max_rows=2))
         self.assert_eq(got, expected)
+
+    def test_to_excel(self):
+        pdf = self.pdf
+        kdf = self.kdf
+        excel_writer = "output.xlsx"
+
+        self.assert_eq(kdf.to_excel(excel_writer), pdf.to_excel(excel_writer))
```
rxin: should we assert the output.xlsx is the same, rather than asserting on the return value?

garawalid:
@shril  you are welcome!

databricks/koalas/tests/test_dataframe_conversion.py
shril: Doing that.

shril: @rxin, should I load the values of the generated excel files in 2 dataframes and check the values?

rxin: Yes that'd work.

rxin: Make sure you write it to a temporary location that's unique too.

shril: @rxin, I have made the changes you asked for.
Please have a look.

databricks/koalas/frame.py
```diff
@@ -950,6 +950,108 @@ def assign(self, **kwargs):
...
+        To write a single object to an Excel .xlsx file it is only necessary to
+        specify a target file name. To write to multiple sheets it is necessary to
+        create an `ExcelWriter` object with a target file name, and specify a sheet
+        in the file to write to.
+
+        Multiple sheets may be written to by specifying unique `sheet_name`.
+        With all data written to the file it is necessary to save the changes.
+        Note that creating an `ExcelWriter` object with a file name that already
+        exists will result in the contents of the existing file being erased.
+
```
HyukjinKwon: This one would also need a note that says it will collect data into driver side which can end up with OOO

databricks/koalas/tests/test_dataframe_conversion.py
```diff
@@ -70,3 +71,76 @@ def test_to_html(self):
...
+
+    @staticmethod
+    def teardown_location(location1, location2, directory):
+        """Helper function to remove the temporary directory and it's contents."""
+        if os.path.isfile(location1):
+            os.remove(location1)
+        if os.path.isfile(location2):
+            os.remove(location2)
+        if os.path.exists(directory):
+            os.rmdir(directory)
```
ueshin: We can use `TestUtils.temp_dir` or `TestUtils.temp_file`.

```py
class DataFrameConversionTest(ReusedSQLTestCase, SQLTestUtils, TestUtils):

    ...

    def test_xxx(self):
        with self.temp_dir() as tmp:
            ...
```

requirements-dev.txt
```diff
@@ -2,6 +2,8 @@
 pandas>=0.23
 pyarrow>=0.10,<0.11
 numpy>=1.14
+openpyxl>=2.6.2
+xlrd>=0.9.0
```
ueshin: Maybe these are under `Test` category?
And do we need to explicitly specify the version? I'm afraid it causes to break the Python 3.5 build like before (#241).

databricks/koalas/frame.py
shril: I am not getting you here @HyukjinKwon, what do you mean by OOO?

requirements-dev.txt
shril: Moved the libraries and removed the versions.

databricks/koalas/tests/test_dataframe_conversion.py
shril: Working on it.

shril: @ueshin, I have changed it according to your suggestions.
Please have a look.

databricks/koalas/frame.py
shril: @rxin, @ueshin can you help me out here?

garawalid: @shril you should add this note.
```
.. note:: This method should only be used if the resulting DataFrame is expected
```

shril: Once again thanks @garawalid :)

garawalid: Peace!

databricks/koalas/testing/utils.py
```diff
@@ -272,6 +273,25 @@ def temp_file(self):
         with self.temp_dir() as tmp:
             yield tempfile.mktemp(dir=tmp)
 
+    @contextmanager
+    def cd(self, newdir, cleanup=lambda: True):
```
ueshin: I couldn't get why we need these two.
Didn't `temp_dir` work for any reason?

shril: Removing my functions and adding `temp_dir`.
Thanks for pointing out.

shril:
Hi @ueshin, I have changed according to your specifications.
Please have a look.

databricks/koalas/testing/utils.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 
+import os
```
ueshin: nit: We can revert this now.

databricks/koalas/tests/test_dataframe_conversion.py
```diff
@@ -14,15 +14,16 @@
 # limitations under the License.
 #
 
+import os
```
ueshin: ditto.

databricks/koalas/testing/utils.py
shril: My bad. Doing it right away.

databricks/koalas/tests/test_dataframe_conversion.py
shril: Removed.

ueshin:
Thanks! merging to master.

PR #290, opened 2019-05-10
Cleanup metadata and add documentations for it

HyukjinKwon:
This PR:

- adds documentation at `Metadata` with some doctests.
- renames:
  - `all_fields` -> `columns`
  - `index_fields` -> `index_columns`
  - `column_fields` -> `data_columns`
  - `index_names` -> `index_names` (no change)
  - `index_info` -> `index_map`

  I tried to keep the original intention and avoid name conflict between field and column.
Also, `index_info` -> `index_map`, for now (see https://github.com/databricks/koalas/pull/290#issuecomment-491450583)

- fixes a bug (one line fix) at `set_index`:

    ```python
    kdf = ks.DataFrame({
        'A': [1, 2, 3, 4],
        'B': [5, 6, 7, 8],
        'C': [9, 10, 11, 12],
        'D': [13, 14, 15, 16],
        'E': [17, 18, 19, 20]})
    kdf1 = kdf.set_index("A")
    kdf1.to_spark().show()
    ```

   ```
    Failed example:
        kdf1.to_spark().show()  # doctest: +NORMALIZE_WHITESPACE
    Expected:
        +---+---+---+---+---+
        |  A|  B|  C|  D|  E|
        +---+---+---+---+---+
        |  1|  5|  9| 13| 17|
        |  2|  6| 10| 14| 18|
        |  3|  7| 11| 15| 19|
        |  4|  8| 12| 16| 20|
        +---+---+---+---+---+
    Got:
        +-----------------+---+---+---+---+---+
        |__index_level_0__|  A|  B|  C|  D|  E|
        +-----------------+---+---+---+---+---+
        |                0|  1|  5|  9| 13| 17|
        |                1|  2|  6| 10| 14| 18|
        |                2|  3|  7| 11| 15| 19|
        |                3|  4|  8| 12| 16| 20|
        +-----------------+---+---+---+---+---+
   ```

    Note that `__index_level_0__` doesn't exist in metadata anymore. To make it synced, it selects all fields in metadata at `set_index`.

### DIFF
--- a/databricks/koalas/metadata.py
+++ b/databricks/koalas/metadata.py
@@ -33,7 +33,7 @@ class Metadata(object):
     Manages column names and index information.
 
     :ivar _data_columns: list of the Spark field names to be seen as columns in Koalas DataFrame.
-    :ivar _index_pairs: list of pair holding the Spark field names for indexes,
+    :ivar _index_map: list of pair holding the Spark field names for indexes,
                        and the index name to be seen in Koalas DataFrame.
 
     .. note:: this is an internal class. It is not supposed to be exposed to users and users
@@ -47,7 +47,7 @@ class Metadata(object):
     ...     'B': [5, 6, 7, 8],
     ...     'C': [9, 10, 11, 12],
     ...     'D': [13, 14, 15, 16],
-    ...     'E':[17, 18, 19, 20]})
+    ...     'E': [17, 18, 19, 20]}, columns = ['A', 'B', 'C', 'D', 'E'])
     >>> kdf  # doctest: +NORMALIZE_WHITESPACE
        A  B   C   D   E
     0  1  5   9  13  17
@@ -79,7 +79,7 @@ class Metadata(object):
 
     * `index_names` represents the external index name
 
-    * `index_pairs` is zipped pairs of `index_columns` and `index_names`
+    * `index_map` is zipped pairs of `index_columns` and `index_names`
 
     >>> metadata = kdf._metadata
     >>> metadata.data_columns
@@ -90,7 +90,7 @@ class Metadata(object):
     ['__index_level_0__', 'A', 'B', 'C', 'D', 'E']
     >>> metadata.index_names
     [None]
-    >>> metadata.index_pairs
+    >>> metadata.index_map
     [('__index_level_0__', None)]
 
     In case that index is set to one of the existing column as below:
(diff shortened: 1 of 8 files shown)


databricks/koalas/metadata.py
```diff
@@ -30,69 +32,189 @@ class Metadata(object):
     """
     Manages column names and index information.
 
-    :ivar _column_fields: list of the Spark field names to be seen as columns in Koalas DataFrame.
-    :ivar _index_info: list of pair holding the Spark field names for indexes,
+    :ivar _data_columns: list of the Spark field names to be seen as columns in Koalas DataFrame.
```
thunterdb: This is super useful doc, thank you for your time writing it. Can you add a comment to say this is an internal class and that it is not meant to be directly accessed by the users (only developers)?

HyukjinKwon: sure

databricks/koalas/metadata.py
```diff
@@ -30,69 +32,192 @@ class Metadata(object):
...
+
+    Metadata represents the index information for a DataFrame it belongs to. For instance,
+    if we have a Koalas DataFrame as below, Pandas DataFrame does not store the index as columns.
+
+    >>> kdf = ks.DataFrame({
+    ...     'A': [1, 2, 3, 4],
+    ...     'B': [5, 6, 7, 8],
+    ...     'C': [9, 10, 11, 12],
+    ...     'D': [13, 14, 15, 16],
+    ...     'E':[17, 18, 19, 20]})
```
ueshin: This could be flaky with Python 3.5.

rxin: also need a space after :

databricks/koalas/metadata.py
```diff
@@ -30,69 +32,192 @@ class Metadata(object):
...
+    >>> kdf1 = kdf.set_index("A")
+    >>> kdf1  # doctest: +NORMALIZE_WHITESPACE
+       B   C   D   E
+    A
+    1  5   9  13  17
+    2  6  10  14  18
+    3  7  11  15  19
+    4  8  12  16  20
+
+    >>> kdf1.to_spark().show()  # doctest: +NORMALIZE_WHITESPACE
```
rxin: do we just lose the index column when to_spark is called? i think we should keep those. cc @thunterdb

rxin: oh nvm it's A

rxin:
I like the rest, but I don't think index_pairs is the right name. "pairs" does not describe what that variable is, but rather the current data structure (a pair of objects). It's less intuitive than a name that describes the purpose of the variable. Also if we ever add more information to make it not a "pair", which is pretty likely, we will need to rename a lot of code.

HyukjinKwon:
How about `index_mapping` (since it maps Spark column to Koalas index name?).  I dont mind reverting the naming change too.

rxin:
Yea maybe just index_map.

Issue #285, opened 2019-05-09
Allow querying DataFrame directly in sql method


floscha:
While I really like the idea of @rxin's recent #256 PR, he uses an (in my opinion) over-simplistic example of `ks.sql("select * from range(10) where id > 7")`. I believe that the ability to query actual Koalas DataFrames through SQL can prove really valuable to many users. However, when trying to use `ks.sql` with a Koalas DataFrame, the following exception occurs:

```python
kdf = ks.DataFrame({'A': [1,2,3]})
>>> ks.sql("select * from kdf")
...
org.apache.spark.sql.AnalysisException: Table or view not found: kdf; line 1 pos 14
...
```

This is not surprising to someone with PySpark knowledge who knows that `kdf` has to be registered as a temporary table before being able to use it with SparkSQL. Unfortunately, (as I understand it) the target group of the Koalas library should not be expected to be Spark experts. To get the above example working, the following workaround is needed, which requires the usage of the lower-level (Py)Spark API, thus somewhat defeating the purpose of Koalas.

```python
>>> from pyspark import SparkContext
>>> from pyspark.sql import SQLContext
>>> sc = SparkContext.getOrCreate()
>>> sql = SQLContext(sc)
>>> sql.registerDataFrameAsTable(kdf._sdf, "kdf")
>>> ks.sql("select * from kdf")
   __index_level_0__  A                                                         
0                  0  1
1                  1  2
2                  2  3
# Optionally clean-up by dropping the temporary table
>>> sql.dropTempTable("kdf")
```

Wouldn't it be much more convenient if this "temporary table magic" would instead be handled by Koalas behind the scenes or are there any design objections against such an approach?

HyukjinKwon:
I think we shouldn't mix PySpark APIs and Koalas APIs. Maybe we should add a method called `to_spark(..)` and delegate everything to there.

As you said, I think `ks.sql`'s purpose is mainly to generate some data or read data from Spark source only. We maybe should document this limitation clearly.

So, when they want to switch to PySpark context, they should explicitly call `to_spark(...)`. Once we start to mix both, it might be difficult for Koalas to target its original goal, Pandas corresponding APIs.

floscha:
I see the point of not mixing the two libraries' APIs. Then again, how does `ks.sql` relate to the pandas API where no `pandas.sql()` exists?

Also, the mentioned `to_spark()` method already exists but I don't quite understand how it relates to the `ks.sql` method.

[...]

PR #291, opened 2019-05-10
Add SQL queries for Koalas DataFrames

floscha:
As discussed in #285, this PR extends the `ks.sql` method to be used with Koalas DataFrames. While it does not yet cover more sophisticated queries involving for example joins, this PR should rather serve as a first proof of concept to collect feedback.


HyukjinKwon:
I don't think we should open this SQL variant in this way. Did the discussion reach to the conclusion for this support?

floscha:
Hey @HyukjinKwon, as explained above, this PR is mainly supposed to be a proof of concept how the `ks.sql()` method could be implemented. If #285 comes to the conclusion that this method should not be part of Koalas, the PR can simply be closed.

rxin:
I left a comment in https://github.com/databricks/koalas/issues/285#issuecomment-491349872

thunterdb:
I left my thoughts on #285  as well.

HyukjinKwon:
@floscha, can you update this PR based upon #285? Since the issue targets 0.5.0 release, I will get this in first after reviewing as planned.

We can revert or take it out after some more discussion later if there's something wrong with it.

rxin:
I'm going to close this one. Let's work on https://github.com/databricks/koalas/pull/360.

Issue #281, opened 2019-05-09
String handling functions (Series.str)


rxin:
The set of Series.str functions are super useful for manipulating text data: http://pandas.pydata.org/pandas-docs/stable/reference/series.html#string-handling

We can probably start with building a simple framework that can wrap around pandas' Series.str functions using pandas UDF. Then for the functions that have native Spark ones, change the implementation to use the Spark native ones instead for higher performance.

icexelloss:
I am taking a look a this

HyukjinKwon:
Nice! @icexelloss!

icexelloss:
I am working with Hyoojee @hjoo. She will work on this instead :). I am going to take a stab at dt functions.

hjoo:
I'm working on this.

hjoo:
Opting not to implement `Series.str.encode()` and `Series.str.decode()` for now because Spark 2.3 doesn't support BinaryType. Issue should be addressed separately from #429.

hjoo:
#429 implements the following `Series.str` functions:

- `capitalize()`, `upper()`, `lower()`, `swapcase()`
- `get()`
- `startswith()`, `endswith()`
- `strip()`, `lstrip()`, `rstrip()`
- `isalnum()`, `isalpha()`, `isdigit()`, `isspace()`, `islower()`, `isupper()`, `istitle()`, `isnumeric()`, `isdecimal()`

Functions that still need to be implemented are the remaining functions listed here: https://pandas.pydata.org/pandas-docs/stable/reference/series.html#api-series-str.

hjoo:
Implementing the remaining functions.

HyukjinKwon:
Fixed in #496

PR #292, opened 2019-05-10
[WIP] Koalas Series.str functions

hjoo:
As discussed in #281, this PR creates Series.str functions for manipulating strings in Koalas Series.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -258,6 +258,36 @@ class Series(_Frame):
         return self.schema.fields[-1].dataType
 
     def astype(self, dtype):
+        """
+        Cast a Koalas object to a specified dtype ``dtype``.
+
+        Parameters
+        ----------
+        dtype : data type
+            Use a numpy.dtype or Python type to cast entire pandas object to
+            the same type.
+
+        Returns
+        -------
+        casted : same type as caller
+
+        See Also
+        --------
+        to_datetime : Convert argument to datetime.
+
+        Examples
+        --------
+        >>> ser = ks.Series([1, 2], dtype='int32')
+        >>> ser
+        0    1
+        1    2
+        Name: 0, dtype: int32
+
+        >>> ser.astype('int64')
+        0    1
+        1    2
+        Name: 0, dtype: int64
+        """
         from databricks.koalas.typedef import as_spark_type
         spark_type = as_spark_type(dtype)
         if not spark_type:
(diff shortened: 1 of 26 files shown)


hjoo:
Any mypy experts? I think str might be a special attribute name, and I'm getting this failure with the linter.

```
mypy checks failed:
databricks/koalas/series.py: note: In member "__getattr__" of class "Series":
databricks/koalas/series.py:1089: error: Invalid type "databricks.koalas.series.Series.str"
databricks/koalas/series.py:1090: error: str? has no attribute "startswith"
```

rxin:
I don't know what the right way is ... but this worked:

```
str_type = str # in the beginning of the file

...
    def __getattr__(self, item: str_type) -> Any:
```

databricks/koalas/series.py
```diff
@@ -80,6 +81,30 @@ def wrapper(self, *args):
     return wrapper
 
 
+def _pandas_column_op(f, spark_return_type, f_name=''):
```
rxin: @icexelloss and @hjoo you two should synchronize to use the same infrastructure ... we probably shouldn't create two completely different infra for dt vs str.

rxin: and btw I'm super excited about these two PRs! Will be a huge productivity boost.

docs/source/reference/series.rst
```diff
@@ -25,6 +25,11 @@ Attributes
 
    Series.dtype
 
+.. autosummary::
+   :toctree:  api/
+
+   Series.str
```
rxin: instead of listing just the str function, we should do it similar to how pandas organizes it. pandas just shows the list all str.xxx functions in the same page as other Series functions.

icexelloss: The thing that pandas does is:
https://github.com/pandas-dev/pandas/blob/master/doc/source/conf.py#L481

It uses lots of sphinx magic to accomplish this. Maybe can we keep this out the scope of this PR? I think it's actually pretty tricky to figure this out.

I think we can just keep make the doc show sth like:

```
StringMethods.capticalize()
StringMethods.substr(...)
```
For this PR?

databricks/koalas/series.py
icexelloss: Yes we will sync up and use the same infrastructure for this.

HyukjinKwon:
Argh, seems it's wrongly merged. Can you rebase or make another pr after closing this?

hjoo:
Ah yes, I'll try making another pr after closing this one.

Just a heads up, I'm going to be gone on vacation for a week and a half, but will revisit this when I get back!

PR #295, opened 2019-05-10
Series.dt functions

icexelloss:
Initial attempt to implement pd.Series.dt functions on koalas.Series

### DIFF
--- a/databricks/koalas/datetime.py
+++ b/databricks/koalas/datetime.py
@@ -23,43 +23,12 @@ import functools
 import databricks.koalas as ks
 import pyspark.sql.functions as F
 from pyspark.sql.types import DateType, TimestampType, LongType, StringType
-from databricks.koalas.series import Series, _column_op
-from databricks.koalas.typedef import pandas_wraps
+from databricks.koalas.series import (
+    Series,
+    _wrap_accessor_pandas,
+    _wrap_accessor_spark
+)
 from databricks.koalas.utils import lazy_property
-
-
-def defer_to_pandas(output_type):
-    """Wraps a function that operates on pd.Series to work with DatetimeMethods.
-
-    The origin function operates on pd.Series and the wrapped function operates
-    on DatetimeMethods.
-    """
-    def decorator(func):
-        @functools.wraps(func)
-        def wrapper(self, *args):
-            scol = F.pandas_udf(
-                lambda series: func(series, *args),
-                output_type
-            )(self._data._scol)
-            return Series(
-                scol, anchor=self._data._kdf, index=self._data._index_info)
-        return wrapper
-    return decorator
-
-
-def defer_to_spark(output_type):
-    """Wraps a function that operates on spark.sql.Column to work with DatetimeMethods.
-
-    The origin function operates on spark.sql.Column and the wrapped function
-    operates on DatetimeMethods.
-    """
-    def decorator(func):
-        @functools.wraps(func)
-        def wrapper(self, *args):
-            scol = func(self._data._scol, *args).cast(output_type)
-            return Series(scol, anchor=self._data._kdf, index=self._data._index_info)
-        return wrapper
-    return decorator
 
 
 class DatetimeMethods(object):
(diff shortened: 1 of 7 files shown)


icexelloss:
Not every dt functions is implemented yet. But I'd like to get some feedback on whether this approach looks good before finishing it off.

cc @HyukjinKwon @ueshin

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+    def decorator(func):
+        @functools.wraps(func)
+        def wrapper(self, *args):
+            scol = func(self._data._scol, *args).cast(output_type)
+            return Series(scol, anchor=self._data._kdf, index=self._data._index_info)
+        return wrapper
+    return decorator
+
+
+class DatetimeMethods:
```
icexelloss: This might not be the best name. In pandas they are called sth like `DatetimelikeProperties` but it's really a combination of properties and methods I think.

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+        return s.dt.dayofweek
+
+    @lazy_property
+    @defer_to_spark(output_type=LongType())
+    def dayofyear(col):
+        return F.dayofyear(col)
+
+    # Methods
+    @defer_to_pandas(output_type=StringType())
+    def strftime(s, date_format):
```
icexelloss: The decorator supports multiple arguments, but the first argument must be the "parent" series of the DatetimeMethods object.

databricks/koalas/series.py
```diff
@@ -163,6 +163,11 @@ def _init_from_spark(self, scol, kdf, index_info):
     # arithmetic operators
     __neg__ = _column_op(spark.Column.__neg__)
 
+    @property
+    def dt(self):
+        from databricks.koalas.datetime import DatetimeMethods
```
icexelloss: This is to get around circular imports.

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+import functools
+from pyspark.sql.types import DateType, TimestampType, LongType, StringType
+
+from databricks.koalas.series import Series
+from databricks.koalas.utils import lazy_property
+
+import pyspark.sql.functions as F
+
+
+def defer_to_pandas(output_type):
```
rxin: add type annotation for output_type?

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+                lambda series: func(series, *args),
+                output_type
+            )(self._data._scol)
+            return Series(
+                scol, anchor=self._data._kdf, index=self._data._index_info)
+        return wrapper
+    return decorator
+
+
+def defer_to_spark(output_type):
```
rxin: same here

databricks/koalas/datetime.py
rxin: i think the name is fine ...

databricks/koalas/datetime.py
rxin: also these two functions are not specific to datetime right? perhaps we should put that in utils.py.

and I'd name it either `apply_pandas_function` and `apply_spark_function`. 

One thing is we should make sure the generated doc is still readable, i.e. it shows the parameters for the functions.

rxin: you should try add this to the api reference in docs directory and see how they render

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+Date/Time related functions on Koalas Series
+"""
+
+import functools
+from pyspark.sql.types import DateType, TimestampType, LongType, StringType
+
+from databricks.koalas.series import Series
+from databricks.koalas.utils import lazy_property
+
+import pyspark.sql.functions as F
```
ueshin: This should be in the upper import block.

databricks/koalas/datetime.py
ueshin: I guess these are reusable for other series functions? If so, I'd move these to `series.py` or somewhere more global.

databricks/koalas/datetime.py
HyukjinKwon: not a big deal but let's stick to new class definition style `class DatetimeMethods(object):` :D.

databricks/koalas/datetime.py
icexelloss: Sounds good. I haven't been able to generate docs (missing usr/bin/xcrun). I will spend some time figuring it out.

icexelloss: Makes sense. Will move to some shared place.

databricks/koalas/datetime.py
icexelloss: Ah, thanks! Didn't know that.

databricks/koalas/datetime.py
rxin: You might need to install the xcode command line tools on mac.

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+    return decorator
+
+
+def defer_to_spark(output_type):
+    """Wraps a function that operates on spark.sql.Column to work with DatetimeMethods.
+
+    The origin function operates on spark.sql.Column and the wrapped function
+    operates on DatetimeMethods.
+    """
+    def decorator(func):
```
thunterdb: Sorry for coming late to the review, but I would like to make a few improvement before we merge this. This is more complicated than it ought to and it is going to be harder to refactor in the future. We should be able to make the whole body of this function simpler along this sketch:

```py
class DateTimeMethods(object):
  ...
   @lazy_property
    def year(self) -> ks.Series: # Let's get the type hints too.
      return defer_to_spark(self, LongType(), F.year)

def defer_to_spark(kseries, return_type, spark_function):
  scol = func(kseries._data._scol, *args).cast(return_type)
  return Series(scol, anchor=kseries._kdf, index=kseries._index_info)
```

I have not tried this code, but it should be pretty much all you need.

databricks/koalas/datetime.py
```diff
@@ -0,0 +1,154 @@
...
+    """Date/Time methods for Koalas Series"""
+    def __init__(self, series):
+        if not isinstance(series.spark_type, (DateType, TimestampType)):
+            raise ValueError(
+                "Cannot call DatetimeMethods on type {}"
+                .format(series.spark_type))
+        self._data = series
+
+    # Properties
+    @lazy_property
```
thunterdb: why do you need lazy property? The performance difference is negligible.

thunterdb: I meant compared to regular `@property`

databricks/koalas/datetime.py
thunterdb: You are copying here most of the code in `typedef.py` and you do not need any of that. I have not tried the following but it should work, let me know if it does not:

```py
class DateTimeMethods(object):
  ...
    def strftime(self, date_format) -> ks.Series:
       return pandas_wrap(lambda col: col.strftime(col, date_format), return_col=StringType())(self._data)
```

This is all that you should need. There is no need to think about the parameters because they get embeded in the closure.

databricks/koalas/datetime.py
icexelloss: I am not too sure about using property here. When user creates a koalas Series, I think it's a little too much to also create lots of accessor series along with it. It increases the number of Series we create by a lot I think.

databricks/koalas/datetime.py
icexelloss: I have reimplemented some of the function using pandas_wrap and Series._column_op. 

It looks a little more verbose than before but it doesn't need extra decorator now. If this is what you want I can go ahead and convert the rest of them and remove the two decorators `defer_to_spark` and `defer_to_pandas`.

databricks/koalas/datetime.py
icexelloss: Added

docs/source/reference/series.rst
```diff
@@ -93,3 +93,26 @@ Serialization / IO / Conversion
    Series.to_dict
    Series.to_clipboard
    Series.to_latex
+
+Datetime Methods
```
rxin: take a look at how pandas does it for docs: https://raw.githubusercontent.com/pandas-dev/pandas/master/doc/source/reference/series.rst

They show up pretty nicely.

icexelloss: Yeah let me play with it to see how far I can get. Will update here.

databricks/koalas/datetime.py
thunterdb: @icexelloss it looks a bit verbose indeed. Since most of the return types are of the type, I think we can refactor it this way, taking your example below:

```py
class DateTimeMethods(object):

    def strftime(self, date_format) -> ks.Series:
        """
        Convert to String Series using specified date_format.
        """
        return self._wrap_string(lamba col: col.dt.strftime(date_format))

    @lazy_property
    def weekofyear(self) -> ks.Series:
        """
        The week ordinal of the year.
        """
        return self._wrap_spark(F.weekofyear)

    def _wrap_string(self, pd_col_fun, return_dtype=StringType()):
       return  pandas_wraps(
            function=lambda x: pd_col_fun(x),
            return_col=return_dtype
        )(self._data)

  def _wrap_spark(self, sp_col_fun, return_dtype=LongType()):
      return _column_op(
            lambda col: sp_col_fun(col).cast(return_dtype)
        )(self._data)
```

What do you think?

icexelloss: Yeah I think that looks good to me but I'd suggest removing the default return type for those two wrappers. And I assume `_wrap_string` should be `_wrap_pandas`?

icexelloss: Although, then what these two wrappers are really doing is just saving us from writing 
`(self._data)`...

icexelloss: Well, that and the `cast` call in `_wrap_spark` case. It looks fine to me either way. Having the wrappers does make the implementation of these functions cleaner so even if it's just saving the `cast` and `(self._data)` so it looks to me to having them. Maybe sth like this?

```
class DateTimeMethods(object):

    def strftime(self, date_format) -> ks.Series:
        """
        Convert to String Series using specified date_format.
        """
        return self._wrap_pandas(lamba col: col.dt.strftime(date_format), StringType())

    @lazy_property
    def weekofyear(self) -> ks.Series:
        """
        The week ordinal of the year.
        """
        return self._wrap_spark(F.weekofyear, LongType())

    def _wrap_pandas(self, pd_col_fun, return_dtype):
       return  pandas_wraps(
            function=lambda x: pd_col_fun(x),
            return_col=return_dtype
        )(self._data)

  def _wrap_spark(self, sp_col_fun, return_dtype):
      return _column_op(
            lambda col: sp_col_fun(col).cast(return_dtype)
        )(self._data)
```

docs/source/reference/series.rst
icexelloss: I spent a couple of hours on this (basically trying to copy what pandas is doing) but looks like it involves quite bit of sphinx magic and it's going to take non trivial amount of time to figure out..I am not familiar with sphinx custom Documenter and templating which is what pandas is using for this.

thunterdb:
This is exactly what I had in mind, yes.

icexelloss:
@thunterdb I have addressed your comments about the wrappers. Anything else you'd like to change?

HyukjinKwon:
@icexelloss, can you fix the tests? I will just push this and address other cleanups (if there's any)

HyukjinKwon:
I fixed the tests, cherry-picked the commit and merged into master. I will make a separate followup to clean up.

PR #296, opened 2019-05-11
Add a new env for travis-ci.

ueshin:
After new version of Spark 2.4 was released, we might be able to remove maximum pyarrow version boundary.
This PR is to add a new env for travis-ci to see whether it works or not.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
         - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
         - PYARROW_VERSION=0.10.0
-    - python: "3.7"
+    - python: "3.6"
       env:
         - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,9 +18,14 @@ matrix:
         - PYARROW_VERSION=0.10.0
     - python: "3.6"
       env:
-        - SPARK_VERSION=2.4.1
+        - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
         - PYARROW_VERSION=0.10.0
+    - python: "3.7"
+      env:
+        - SPARK_VERSION=2.4.3
+        - PANDAS_VERSION=0.24.2
+        - PYARROW_VERSION=0.13.0
 
 before_install:
  - ./dev/download_travis_dependencies.sh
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,9 +18,14 @@ matrix:
...
       env:
-        - SPARK_VERSION=2.4.1
+        - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
         - PYARROW_VERSION=0.10.0
+    - python: "3.6"
+      env:
+        - SPARK_VERSION=2.4.3
+        - PANDAS_VERSION=0.24.2
+        - PYARROW_VERSION=0.13.0
```
HyukjinKwon: @ueshin maybe we should fix this line too https://github.com/databricks/koalas/blob/master/setup.py#L57

ueshin:
I'd merge this for now.

HyukjinKwon:
Yea nice!

rxin:
Does this (latest pyarrow) work on Databricks? If yes, maybe we should just require pyarrow to be the latest version.

ueshin:
@rxin I ran doctests on Databricks and seems like they work with pyarrow=0.13. I didn't try all the unittests, though.
One thing is I got a warning `UserWarning: pyarrow.open_stream is deprecated, please use pyarrow.ipc.open_stream` caused by Spark 2.4, which will be fixed when we move to Spark 3.0.

ueshin:
Seems like pyarrow 0.13 installed with `conda` on mac has a problem for now, we can't upgrade the minimum version up to 0.13. Maybe 0.12 should be fine if we want to bump up it.

```
>>> import pyarrow as pa
  File "/.../lib/python3.7/site-packages/pyarrow/__init__.py", line 47, in <module>
    from pyarrow.lib import cpu_count, set_cpu_count
ImportError: dlopen(/.../lib/python3.7/site-packages/pyarrow/lib.cpython-37m-darwin.so, 2): Library not loaded: @rpath/libglog.0.dylib
  Referenced from: /.../lib/libarrow.13.dylib
  Reason: Incompatible library version: libarrow.13.dylib requires version 1.0.0 or later, but libglog.0.dylib provides version 0.0.0
```

HyukjinKwon:
Yes, I checked that the error happens too in my Mac. Spark master sets 0.12.1 as minimum version so should be fine I guess.

ueshin:
I filed an issue conda-forge/pyarrow-feedstock#74.

PR #299, opened 2019-05-13
Test against Python 3.7 too in CI

HyukjinKwon:
See https://docs.travis-ci.com/user/languages/python/ as well for how to setup Python 3.7 in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ matrix:
         - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
         - PYARROW_VERSION=0.10.0
-    - python: "3.6"
+    - python: "3.7"
       env:
         - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,8 +16,15 @@ matrix:
...
-    - python: "3.6"
+    # When we upgrade Spark 3.0 for this, remove 'oraclejdk8'.
+    - python: 3.7
+      dist: xenial
+      sudo: true
+      jdk:
+        - oraclejdk8
       env:
+        - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
+        - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
```
HyukjinKwon: This is a such ugly way to downgrade JDK 11 to JDK 8 at `xenial` image. We can maybe just remove all those stuff when we switch this to Spark 3 later I guess.

HyukjinKwon: I should see if it works too

HyukjinKwon:
Let me get this in.

PR #300, opened 2019-05-13
Improve docs for DataFrame.reset_index / set_index

rxin:
These two methods were previously missing examples and were also using the old doc style.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -903,7 +903,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         --------
         >>> df = ks.DataFrame({'month': [1, 4, 7, 10],
         ...                    'year': [2012, 2014, 2013, 2014],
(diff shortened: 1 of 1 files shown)


databricks/koalas/frame.py
```diff
@@ -870,17 +870,67 @@ def index(self):
...
-
-        :param keys: column label or list of column labels / arrays
-        :param drop: boolean, default True
-                     Delete columns to be used as the new index
-        :param append: boolean, default False
-                       Whether to append columns to existing index
-        :param inplace: boolean, default False
-                        Modify the DataFrame in place (do not create a new object)
-        :return: :class:`DataFrame`
+        """Set the DataFrame index (row labels) using one or more existing columns.
```
HyukjinKwon: not a big deal but this first line looks different from Pandas's - https://github.com/pandas-dev/pandas/blob/master/pandas/core/frame.py#L4054

PR #310, opened 2019-05-13
Define to_clipboard doc only once

rxin:
Avoid defining them twice by just setting `Series.to_clipboard.__doc__ = DataFrame.to_clipboard.__doc__`.

Also added skip doctest since those tests are not verified anyway.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -313,8 +313,8 @@ class DataFrame(_Frame):
         --------
         Copy the contents of a DataFrame to the clipboard.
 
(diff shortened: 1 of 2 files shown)


garawalid:
@rxin do you suggest that we use this method `Series.to_clipboard.__doc__ = DataFrame.to_clipboard.__doc__` or just `to_clipboard` is an exception?

HyukjinKwon:
Seems panas shares one doc as well.

rxin:
@garawalid only if they look nearly identical.

garawalid:
@rxin I see. Thanks!

PR #311, opened 2019-05-13
Remove dead code in test_stats.py

rxin:


rxin:
I guess this was just mistakenly left here? cc @HyukjinKwon

PR #313, opened 2019-05-13
Add clip method to Koalas Series and DataFrames

floscha:
This PR adds pandas' `clip()` method to Koalas Series and DataFrames.

For reference:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.clip.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.clip.html

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -312,7 +312,8 @@ class _Frame(object):
         # TODO: The first example above should not have "Name: abs(0)".
         return _spark_col_apply(self, F.abs)
 
-    def clip(self, lower: Union[float, int] = None, upper: Union[float, int] = None) \
+    def clip(self: Union['ks.Series', 'ks.DataFrame'],  # Use type hint on to avoid mypy issues
+             lower: Union[float, int] = None, upper: Union[float, int] = None) \
             -> Union['ks.Series', 'ks.DataFrame']:
         """
         Trim values at input threshold(s).
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -312,8 +312,7 @@ class _Frame(object):
         # TODO: The first example above should not have "Name: abs(0)".
         return _spark_col_apply(self, F.abs)
 
-    def clip(self: Union['ks.Series', 'ks.DataFrame'],  # Use type hint to avoid mypy issues
-             lower: Union[float, int] = None, upper: Union[float, int] = None) \
+    def clip(self, lower: Union[float, int] = None, upper: Union[float, int] = None) \
             -> Union['ks.Series', 'ks.DataFrame']:
         """
         Trim values at input threshold(s).
@@ -353,7 +352,7 @@ class _Frame(object):
 
         is_df = isinstance(self, ks.DataFrame)
 
-        df = self if is_df else self.to_dataframe()
+        df = self if is_df else self.to_dataframe()  # type: ignore
         sdf = df._sdf
 
         if lower is not None:
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -312,7 +312,7 @@ class _Frame(object):
         # TODO: The first example above should not have "Name: abs(0)".
         return _spark_col_apply(self, F.abs)
 
-    def clip(self: Union['ks.Series', 'ks.DataFrame'],  # Use type hint on to avoid mypy issues
+    def clip(self: Union['ks.Series', 'ks.DataFrame'],  # Use type hint to avoid mypy issues
              lower: Union[float, int] = None, upper: Union[float, int] = None) \
             -> Union['ks.Series', 'ks.DataFrame']:
         """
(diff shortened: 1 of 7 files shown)


floscha:
One difference between this implementation and pandas which is worth mentioning is that running `pd.DataFrame({'A': ['a', 'b']}).clip(0, 1)` will crash with `TypeError: '<=' not supported between instances of 'str' and 'int'` while `ks.DataFrame({'A': ['a', 'b']}).clip(0, 1)` will output the original DataFrame, simply ignoring the incompatible types.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -458,3 +458,16 @@ def test_merge(self):
         res = left_kdf.merge(koalas.DataFrame({'A': [3, 4]}), left_index=True, right_index=True,
                              suffixes=('_left', '_right'))
         self.assert_eq(res, pd.DataFrame({'A_left': [1, 2], 'A_right': [3, 4]}))
+
+    def test_clip(self):
+        pdf = pd.DataFrame({'A': [0, 2, 4]})
```
rxin: i'd add a test case for non-numeric type, e.g. string.

databricks/koalas/generic.py
```diff
@@ -311,6 +312,59 @@ def abs(self):
...
+        axis.
+
+        Parameters
+        ----------
+        lower : float or int, default None
+            Minimum threshold value. All values below this threshold will be set to it.
+        upper : float or int, default None
+            Maximum threshold value. All values above this threshold will be set to it.
+
+        Returns
```
rxin: add a small note section explaining the difference w.r.t. pandas.

databricks/koalas/generic.py
```diff
@@ -311,6 +312,59 @@ def abs(self):
         # TODO: The first example above should not have "Name: abs(0)".
         return _spark_col_apply(self, F.abs)
 
+    def clip(self, lower: Union[float, int] = None, upper: Union[float, int] = None) \
```
rxin: does the type hint prevent one from passing in an array_like value at runtime? If not, perhaps we should add a check in the code to throw unsupported exception for array_like values.

databricks/koalas/tests/test_dataframe.py
floscha: Done with 267e509.

databricks/koalas/generic.py
floscha: Added with 8f4876a.

databricks/koalas/generic.py
floscha: No, the type hint only warns in IDEs like PyCharm or when using mypy. So I've added a check with d5ea5b9 that throws an exception when list-like values (which include array-like values) are passed as 'lower' or 'upper'.

databricks/koalas/generic.py
```diff
@@ -311,6 +313,70 @@ def abs(self):
...
+        if lower is None and upper is None:
+            return self  # type: ignore
+
+        is_df = isinstance(self, ks.DataFrame)
+
+        df = self if is_df else self.to_dataframe()  # type: ignore
+        sdf = df._sdf
+
+        if lower is not None:
+            sdf = sdf.select(*(F.when(F.col(c) < lower, lower).otherwise(F.col(c)).alias(c)
```
rxin: Can we optimize this a little bit to only apply this on numeric columns? for non-numeric columns, just keep the original column. This would make the compute much faster on real datasets with mixed type columns.

databricks/koalas/generic.py
rxin: Got it. Thanks.

is there a way in the type hint framework to indicate if the class is a DataFrame return a DataFrame; otherwise return a Series? If there isn't, I'd probably prefer just defining clip twice, once in DataFrame and once in Series.

The Series one can just delegate to the DataFrame one and take the first column. They can also share the same doc by just doing `clip.__doc__ = DataFrame.clip.__doc__`.

rxin:
Looks pretty good. Just some minor comments left ...

databricks/koalas/generic.py
floscha: With ac66cdc clipping is only applied to numerical columns now.

databricks/koalas/generic.py
rxin: @floscha not sure if you see this comment. Let me know what you think.

floscha: I've seen your comment but I was thinking about a good way to use type hints with the *generic* module. However, I ended up splitting the method over the *frame* and *series* modules with d07c0a5 like you've suggested. This resulted in one line for the *series*  module where the `clip` method from *frame* is called. After all, I believe this implementation is much cleaner than the one before (no type checks everywhere), especially when dealing with mypy.

Regarding the docs, I decided to have them separate which I think works nicer when also having separated methods.

floscha:
@rxin do you have an idea what the problem with the code coverage is?

rxin:
I feel it's a codecov bug. The test coverage looks fine from my human eye inspection.

I'm going to merge this now. Thanks.

PR #316, opened 2019-05-13
Skip to_excel test when openpyxl is not installed

HyukjinKwon:


rxin:
Are there environments in which this would be difficult to install? This type of ignore is risky because we might accidentally not have this package installed on CI and then the test will be permanently disabled.

HyukjinKwon:
Hmm. Okie.

Issue #308, opened 2019-05-13
Series.fillna


rxin:
We implemented DataFrame.fillna but not Series.fillna.

pandas doc: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.fillna.html#pandas.Series.fillna

It seems like the doc can just reuse DataFrame's, as long as we change the Returns section and add an example doctest to verify Series.fillna behavior.

rxin:
cc @garawalid want to add this? You added DataFrame.fillna.

garawalid:
@rxin I'll take care of it.

PR #317, opened 2019-05-13
Series.fillna

garawalid:
+ Implementation of `Series.fillna()` #308 
+ Docstring with examples

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -50,6 +50,7 @@ def _column_op(f):
     :param self: Koalas Series
     :param args: arguments that the function `f` takes.
     """
+
     @wraps(f)
     def wrapper(self, *args):
         assert all((not isinstance(arg, Series)) or (arg._kdf is self._kdf) for arg in args), \
@@ -61,6 +62,7 @@ def _column_op(f):
         args = [arg._scol if isinstance(arg, Series) else arg for arg in args]
         scol = f(self._scol, *args)
         return Series(scol, anchor=self._kdf, index=self._index_map)
+
     return wrapper
 
 
@@ -77,6 +79,7 @@ def _numpy_column_op(f):
             else:
                 new_args.append(arg)
         return _column_op(f)(self, *new_args)
+
     return wrapper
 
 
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -50,6 +50,7 @@ def _column_op(f):
     :param self: Koalas Series
     :param args: arguments that the function `f` takes.
     """
+
```
rxin: It's fine for this PR, but in the future try to minimize unrelated changes.

databricks/koalas/tests/test_series.py
```diff
@@ -155,6 +155,18 @@ def test_isin(self):
         with self.assertRaisesRegex(TypeError, msg):
             ds.isin(1)
 
+    def test_fillna(self):
+        ps = pd.Series([np.nan, 2, 3, 4, np.nan, 6], name='x')
+        ks = koalas.from_pandas(ps)
+
+        self.assert_eq(ks.fillna(0), ps.fillna(0))
+        self.assert_eq(ks.fillna(0, inplace=True), ps.fillna(0, inplace=True))
```
rxin: i don't think this test case is correct. in both cases inpalce=True returns nothing. We need to compare the ks.

Also we probably need to make a copy of it. Otherwise you pollute the following "ks" because ks has been changed.

databricks/koalas/series.py
garawalid: @rxin you mean we use `Series.to_clipboard.__doc__ = DataFrame.to_clipboard.__doc__` ?

databricks/koalas/tests/test_series.py
garawalid: Agree! I'll fix it.

databricks/koalas/series.py
rxin: no i just meant don't add a blank line here since it's unrelated to your change. It's best if a commit is atomic and self-contained about a single feature or bug fix.

garawalid: @rxin sorry, didn't pay attention to blank lines! I'll clean it.

databricks/koalas/tests/test_series.py
```diff
@@ -155,6 +155,21 @@ def test_isin(self):
         with self.assertRaisesRegex(TypeError, msg):
             ds.isin(1)
 
+    def test_fillna(self):
+        ps = pd.Series([np.nan, 2, 3, 4, np.nan, 6], name='x')
+        ks = koalas.from_pandas(ps)
+
+        self.assert_eq(ks.fillna(0), ps.fillna(0))
+        self.assert_eq(ks.fillna({'x': 0}), ps.fillna(0))
```
garawalid: It turns out that `Pandas.Series.fillna` supports neither `Series` nor `dict` as value. 
`Koalas.Series.fillna` does!

databricks/koalas/series.py
```diff
@@ -648,6 +648,69 @@ def notnull(self):
...
+        1    2.0
+        2    3.0
+        3    4.0
+        4    0.0
+        5    6.0
+        Name: x, dtype: float64
+
+        Replace all NaN elements in column 'x' with 0.
+
+        >>> values = {'x': 0}
```
rxin: this is actually somewhat confusing because we only have one column here since it's a series ... that's probably why pandas doesn't support dict.

databricks/koalas/series.py
```diff
@@ -648,6 +648,69 @@ def notnull(self):
 
     notna = notnull
 
+    def fillna(self, value=None, axis=None, inplace=False):
+        """Fill NA/NaN values.
+
+        Parameters
+        ----------
+        value : scalar, dict, Series
```
rxin: i think we should just follow pandas here and drop dict/Series. This is a Series, so there's only one column really.

garawalid: @rxin 
[Pandas doc](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.fillna.html) says that it supports `dict` & `Series`. 
But I think the doc is a copy paste of `DataFrame.fillna`. 

Anyway, we can remove them from the docstring.

garawalid:
@rxin I left only `scalar` as values.

rxin:
Thanks. I've merged this. This will be available as part of release 0.4.

PR #323, opened 2019-05-14
Remove SparkPandasMergeError

floscha:
As requested in #264, this PR removes the specific `SparkPandasMergeError` and replaces its usage with the more generic `ValueError`.


rxin:
Thanks for the quick response. Merging.

PR #325, opened 2019-05-14
Add Series.is_unique property

### LINKED ISSUE #305: Series.is_unique
Although an expensive function, this should be pretty easy to add. We can do a count distinct and compare it with the total cardinality.

pandas doc: http://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.is_unique.html

HyukjinKwon:
Resolves #305

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -360,8 +360,6 @@ class Series(_Frame):
         """
         Return boolean if values in the object are unique
 
-        .. note:: if the values contain `None`, the result is `False`.
-
         Returns
         -------
         is_unique : boolean
@@ -371,14 +369,21 @@ class Series(_Frame):
         >>> ks.Series([1, 2, 2]).is_unique
         False
         >>> ks.Series([1, 2, 3, None]).is_unique
-        False
+        True
         """
         sdf = self._kdf._sdf.select(self._scol)
-        # Note that we compare the distinct count of values ignoring None
-        # to the count of values containing None.
-        # Therefore, it results to False if there is None.
+        col = self._scol
+
+        # Here we check:
+        #   1. the distinct count without nulls and count without nulls for non-null values
+        #   2. count null values and see if null is a distinct value.
+        #
+        # This workaround is in order to calculate the distinct count including nulls in
+        # single pass. Note that COUNT(DISTINCT expr) in Spark is designed to ignore nulls.
         return sdf.select(
-            F.countDistinct(self._scol) == F.count(F.lit(1))).collect()[0][0]
+            (F.count(col) == F.countDistinct(col)) &
+            (F.count(F.when(col.isNull(), 1).otherwise(None)) <= 1)
+        ).collect()[0][0]
 
     def reset_index(self, level=None, drop=False, name=None, inplace=False):
         """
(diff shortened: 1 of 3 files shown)


databricks/koalas/series.py
```diff
@@ -348,6 +348,25 @@ def index(self):
...
+        Returns
+        -------
+        is_unique : boolean
+
+        >>> ks.Series([1, None, None]).is_unique
+        False
+        >>> ks.Series([1, 2, 3, None]).is_unique
+        True
+        """
+        sdf = self._kdf._sdf.select(self._scol)
```
rxin: the way i'd do this is to do a single pass, e.g.

select(count(col), countDistinct(col))

and get the two values back and compare them at once.

HyukjinKwon: I actually tried one pass first. The problem is countDistinct ignore null and `expr` with COUNT(DISTINCT *) ignore null by design as well.

rxin: It seems like we are fine?

```
>>> ks.DataFrame([1, 2, 3, 4, None, None]).to_spark().selectExpr("count(distinct `0`)").show()
+-----------------+
|count(DISTINCT 0)|
+-----------------+
|                4|
+-----------------+
```

This is the same as pandas:
```
>>> pd.Series([1, 2, 3, 4, None, None]).nunique()
4
```

HyukjinKwon: Yes, I will stick to the original one.

databricks/koalas/series.py
```diff
@@ -348,6 +348,25 @@ def index(self):
             raise KeyError('Currently supported only when the Column has a single index.')
         return self._kdf.index
 
+    @property
+    def is_unique(self):
+        """
+        Return boolean if values in the object are unique
+
+        .. note:: this is an expensive operation that has to count all values in the Series.
```
rxin: i don't know if you need this note here ... of course it has to count all the values :)

databricks/koalas/namespace.py
```diff
@@ -33,27 +33,29 @@
 from databricks.koalas.series import Series
 
 
-def from_pandas(pdf):
+def from_pandas(pobj: Union[pd.DataFrame, pd.Series]):
```
HyukjinKwon: Ah, I missed you submitted a PR already about this  @rxin. Sorry, let me revert

databricks/koalas/series.py
```diff
@@ -355,6 +355,31 @@ def index(self):
...
+
+        Returns
+        -------
+        is_unique : boolean
+
+        >>> ks.Series([1, 2, 3]).is_unique
+        True
+        >>> ks.Series([1, 2, 2]).is_unique
+        False
+        >>> ks.Series([1, 2, 3, None]).is_unique
```
rxin: this should be true right?

HyukjinKwon: Actually, it gives `False` in my local:

```
>>> pd.Series([1, 2, 3, None]).is_unique
False
```

Maybe version difference. let me check

databricks/koalas/series.py
HyukjinKwon: Actually, Pandas seems using:

```
>>> pd.Series([1, 2, 3, 4, None, None]).nunique(dropna=False)
5
```

PR #326, opened 2019-05-14
DataFrame nlargest and nsmallest

### LINKED ISSUE #302: nsmallest / nlargest
Two basic functions to support in both DataFrame and Series.

Note that pandas doc says: "This method is equivalent to df.sort_values(columns, ascending=True).head(n), but more performant." In the case of Koalas there shouldn't be a difference, because Spark's query optimizer will collapse a sort_values followed by head into a topk operator internally.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.nlargest.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.nsmallest.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.nlargest.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.nsmallest.html

abishekganesh72:
Added nlargest and nsmallest methods. 

Resolves #302


databricks/koalas/frame.py
```diff
@@ -2145,6 +2145,159 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
         else:
             return kdf
 
+    # TODO: Not Supported:  keep = First
+    def nlargest(self, n, columns, keep='first'):
+        """
+        Return the first `n` rows ordered by `columns` in descending order.
+        Return the first `n` rows with the largest values in `columns`, in
```
rxin: insert a blank line here

databricks/koalas/frame.py
```diff
@@ -2145,6 +2145,159 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
...
+        descending order. The columns that are not specified are returned as
+        well, but not used for ordering.
+        This method is equivalent to
```
rxin: this is not true, in Koalas the performance will be the same. So we should just add a note section saying the

```
"In pandas, ``nlargest(n)`` is more performant than ``sort_values(columns, ascending=False).head(n)``. In Koalas, thanks to Spark's lazy execution and query optimizer, the two would have same performance.
```

databricks/koalas/frame.py
```diff
@@ -2145,6 +2145,159 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
...
+        Italy     59000000  1937894      IT
+        Malta       434001    12011      MT
+
+        >>> df.nlargest(3, ['population', 'GDP'])
+                population      GDP alpha-2
+        France    65000000  2583560      FR
+        Italy     59000000  1937894      IT
+        Malta       434001    12011      MT
+        """
+        return self.sort_values(by=columns, ascending=False).head(n=n)
```
rxin: we should throw an error if keep is set and is not the default value

databricks/koalas/frame.py
```diff
@@ -2145,6 +2145,159 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
...
+        ``df.sort_values(columns, ascending=False).head(n)``, but more
+        performant.
+
+        Parameters
+        ----------
+        n : int
+            Number of rows to return.
+        columns : label or list of labels
+            Column label(s) to order by.
+        keep : Not yet supported
```
rxin: this should be

```
keep :  {‘first’, ‘last’, ‘all’}, default ‘first’
    Only the default value first is supported
```

rxin:
Can you also add the same methods for Series in this PR?

databricks/koalas/frame.py
```diff
@@ -2145,6 +2145,159 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
         else:
             return kdf
 
+    # TODO: Not Supported:  keep = First
+    def nlargest(self, n, columns, keep='first'):
```
HyukjinKwon: Can we add type hints? see https://github.com/databricks/koalas/pull/327

databricks/koalas/frame.py
abishekganesh72: Shouldnt it be 'all' that is supported and not first?

abishekganesh72:
> Can you also add the same methods for Series in this PR?

Working on adding Series methods

abishekganesh72:
@HyukjinKwon  Can you please check if the functionality for the `keep` parameter. 
I am not very clear of how that needs to be done for Koalas?

HyukjinKwon:
We can leave it as unsupported. Let's keep that out here.

HyukjinKwon:
Can you fix the tests and rebase?

abishekganesh72:
Fixed all the Issues.

databricks/koalas/series.py
```diff
@@ -1085,6 +1085,143 @@ def corr(self, other, method='pearson'):
...
+            Return this many ascending sorted values.
+        keep : {'first', 'last', 'all'}, default 'first'
+            When there are duplicate values that cannot all fit in a
+            Series of `n` elements:
+            - ``first`` : take the first occurrences based on the index order
+            - ``last`` : take the last occurrences based on the index order
+            - ``all`` : keep all occurrences. This can result in a Series of
+                size larger than `n`.
+
+        Please note the Koalas: default behaviour is like "all"
```
HyukjinKwon: Is` keep`'s default first or all?

abishekganesh72: Keep in Pandas defaults to first. But we don't have a way to replicate that here so Koalas behaves like all

databricks/koalas/frame.py
```diff
@@ -2204,6 +2204,154 @@ def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
...
+        3  5.0   9
+
+        >>> df.nlargest(n=3, columns=['Y', 'X'])
+             X   Y
+        6  NaN  12
+        5  7.0  11
+        4  6.0  10
+
+        """
+        if keep != 'first':
```
HyukjinKwon: @abishekganesh72, does that mean this isn't actually working like `first` but works like `all`? Let's just remove `keep` parameter for now in the arguments, and the doc for now.

abishekganesh72: Yes, it works like all. Ok will do the changes and push the code.

HyukjinKwon: thanks.

HyukjinKwon:
Thanks @abishekganesh72

PR #336, opened 2019-05-15
Inline to_datetime documentation

HyukjinKwon:


databricks/koalas/namespace.py
```diff
@@ -278,6 +278,92 @@ def read_parquet(path, columns=None):
...
+
+    Passing errors='coerce' will force an out-of-bounds date to NaT,
+    in addition to forcing non-dates (or non-parseable dates) to NaT.
+
+    >>> ks.to_datetime('13000101', format='%Y%m%d', errors='ignore')
+    datetime.datetime(1300, 1, 1, 0, 0)
+    >>> ks.to_datetime('13000101', format='%Y%m%d', errors='coerce')
+    NaT
+
+    Passing infer_datetime_format=True does not affect performance much
```
rxin: is this true based on our implementation? we ran something on almost no data so i'm not sure if that's the best test.

rxin:
Looks great. Only astype left.

databricks/koalas/namespace.py
HyukjinKwon: sorry, my bad. let me fix it.

PR #337, opened 2019-05-15
Inline Series.astype documentation

HyukjinKwon:


databricks/koalas/series.py
```diff
@@ -258,6 +258,36 @@ def spark_type(self):
...
+        numpy.ndarray.astype : Cast a numpy array to a specified type.
+
+        Examples
+        --------
+        >>> ser = ks.Series([1, 2], dtype='int32')
+        >>> ser
+        0    1
+        1    2
+        Name: 0, dtype: int32
+        >>> ser.astype('int64')
```
rxin: add a new line here ...

in general i think we should separate doctests with blank lines to make them easier to read

databricks/koalas/series.py
```diff
@@ -258,6 +258,36 @@ def spark_type(self):
...
+            the same type.
+
+        Returns
+        -------
+        casted : same type as caller
+
+        See Also
+        --------
+        to_datetime : Convert argument to datetime.
+        numpy.ndarray.astype : Cast a numpy array to a specified type.
```
rxin: remove this

databricks/koalas/series.py
HyukjinKwon: okie

rxin:
LGTM other than the two minor comments .... and this reminds me, we should implement DataFrame.astype as well.

PR #341, opened 2019-05-15
Introduce `Index` and `MultiIndex`.

### LINKED ISSUE #306: Support more than one index for the index property
Both DataFrame and Series have this limitation:

```python
    @property
    def index(self):
        """The index (row labels) Column of the DataFrame.

        Currently supported only when the DataFrame has a single index.
        """
        if len(self._metadata.index_map) != 1:
            raise KeyError('Currently supported only when the DataFrame has a single index.')
        return Series(self._index_columns[0], anchor=self, index=[])
```

We should just support more than 1 here ... (and also 0).

ueshin:
This PR introduces `Index` and `MultiIndex` to behave similar to pandas' indexes.
Resolves #306.


ueshin:
I'll add docs later.

ueshin:
@rxin I agree that we can reuse many of code in `Series`, but there are still many differences between `Index` and `Series` according to my rough investigation.

no. of common functions | no. of only in Series | no. of only in Index
--------------- | -------------------- | ------------------
49 | 131 | 39

Note that even the common functions here might behave differently between `Series` and `Index`.

Also, I'm feeling weird if we return `Series` for a single index whereas `MultiIndex` for more than one index since they are hierarchically different.

How about introducing a mix-in for common functions like `IndexOpsMixin` (from pandas' name) or something to reuse code?

rxin:
Sgtm

ueshin:
@rxin Thanks! I will address them soon.

databricks/koalas/frame.py
```diff
@@ -972,6 +972,10 @@ def index(self):
         """The index (row labels) Column of the DataFrame.
 
         Currently not supported when the DataFrame has no index.
```
rxin: maybe we should return something ... like None?

rxin:
@HyukjinKwon @ueshin can you review each other's code? I will be gone for almost 2 wks :)

HyukjinKwon:
Oops, will do. sry.

HyukjinKwon:
LGTM

HyukjinKwon:
For https://github.com/databricks/koalas/pull/341#issuecomment-492917885, seems we (I) have added `any` and `all`, for instnace. let me try to put it in the IndexOpsMixin too.

PR #342, opened 2019-05-15
Deprecate clip_lower/clip_upper

### LINKED ISSUE #318: clip_lower/clip_upper deprecated
Both the *missing/frame.py* and *missing/series.py* files contain entries for the `clip_lower()` and `clip_upper()` methods. According to the [pandas documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.clip_upper.html), both methods are deprecated now. While it would be trivial to add those to Koalas once #313 is merged, I'm wondering if makes sense to still add them or otherwise add a note that they're deprecated and shouldn't be used.

floscha:
As discussed in #318, this PR extends Koalas to treat deprecated pandas functions/properties as special cases of *not implemented*. More specifically, this PR starts by handling the `clip_lower()` and `clip_upper()` methods. In the future, other deprecated methods should be treated the same way.

Resolves #318


floscha:
The PR decreases the test coverage by introducing compatibility for deprecated properties which is not yet tested as the PR focuses on the `clip_lower()` and `clip_upper()` functions only. Would it be okay to accept this for now, knowing that the coverage will increase again once the first deprecated property is added?

rxin:
@floscha I thought you were testing those? Did I read this wrong?

```

        deprecated_functions = [name for (name, type_) in missing_functions
                                if type_.__name__ == 'deprecated_function']
        for name in deprecated_functions:
            with self.assertRaisesRegex(PandasNotImplementedError,
                                        "method.*DataFrame.*{}.*is deprecated".format(name)):
                getattr(kdf, name)()
```

rxin:
Ah I see what you mean. It's property not method. Yes I think this is fine. But can you do one more thing ...

Update the `dev/_make_missing_functions.py` and `dev/_make_missing_properties.py` files to reflect this change. I think you can just search for `.. deprecated` in the docstring to identify whether a method is deprecated or not.

floscha:
I've updated the `dev/_make_missing_functions.py` and `dev/_make_missing_properties.py` files with  ee2c35d and 687c74a.

floscha:
Yes, there are many. I was originally creating this PR to fix #318 but I might as well add the other deprecated functions/properties as well.

dev/_make_missing_functions.py
```diff
@@ -221,17 +226,23 @@ def _main():
...
-        missing, modified = inspect_missing_functions(original_type, target_type)
+        missing, deprecated, modified = inspect_missing_functions(original_type, target_type)
 
         print('MISSING functions for {}'.format(original_type.__name__))
         for name, signature in missing:
             # print(make_misssing_function(original_type, name, signature))
             print("""    {0} = unsupported_function('{0}')""".format(name))
 
+        print()
+        print('DEPRECATED functions for {}'.format(original_type.__name__))
```
HyukjinKwon: Can you use `warnings.warn("...", DeprecationWarning)`?

HyukjinKwon:
+1 !

floscha:
I've added all deprecated functions/properties to dedicated sections in *frames.py*, *grouby.py* and *series.py*.

rxin:
Merging this in. @ueshin if you find anything else that can be improved please comment here still.

Thanks @floscha!

ueshin:
late LGTM.
Thanks @floscha!

PR #346, opened 2019-05-16
Add nunique() method from pandas

### LINKED ISSUE #303: nunique
Both DataFrame / Series should have nunique method. This should be pretty easy to implement by using the count_distinct expression in Spark.

dropna would require little bit of extra handling (perhaps by adding a filter?).

We might want to introduce a flag that does approximate unique, using HyperLogLog, which would be much faster.

pandas docs:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.nunique.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.nunique.html

floscha:
As suggested in #303, this PR adds pandas' `nunique()` method for both DataFrames and Series.

Resolves #303


databricks/koalas/frame.py
```diff
@@ -1283,6 +1283,41 @@ def notnull(self):
...
+        A    3
+        B    2
+        Name: 0, dtype: int64
+        """
+        if axis != 0:
+            raise ValueError("The 'nunique' method only works with axis=0 at the moment")
+        res = self._sdf.select(*(F.countDistinct(Column(c)).alias(c)for c in self.columns))
+        res = res.toPandas().T.iloc[:, 0]
+        if not dropna:
+            res += self.isnull().sum().clip(upper=1)
```
rxin: It'd be better if we can do this in a single pass, rather than triggering two jobs. Take a look at the current is_unique property and you can transfer that "algorithm" here.

databricks/koalas/frame.py
```diff
@@ -1283,6 +1283,41 @@ def notnull(self):
 
     notna = notnull
 
+    def nunique(self, axis=0, dropna=True):
```
rxin: type hint?

floscha: Sorry 🙈Added with 9514e16.

databricks/koalas/frame.py
floscha: I've fixed this now, so with 0f5b16b it's only a single job.

databricks/koalas/frame.py
```diff
@@ -1284,6 +1284,48 @@ def notnull(self):
...
+                                   .alias(c)
+                                    for c in self.columns])
+        else:
+            res = self._sdf.select([(F.countDistinct(Column(c))
+                                     + F.when(F.count(F.when(F.col(c).isNull(), 1).otherwise(None))
+                                              >= 1, 1).otherwise(0))
+                                   .alias(c)
+                                    for c in self.columns])
+        res = res.toPandas().T.iloc[:, 0]
+        return res
```
HyukjinKwon: Hm .. isn't it returning Pandas' ones, not Koalas series?

databricks/koalas/series.py
```diff
@@ -915,6 +915,34 @@ def unique(self):
...
+        --------
+        >>> ks.Series([1, 2, 3, np.nan]).nunique()
+        3
+
+        >>> ks.Series([1, 2, 3, np.nan]).nunique(dropna=False)
+        4
+        """
+        sdf = self.to_dataframe()._sdf
+        distinct_count = sdf.select(F.countDistinct(self._scol)).toPandas().iloc[0, 0]
+        if not dropna and self.isnull().sum() > 0:
```
HyukjinKwon: I think this won't be in single pass too .. due to `sum()`.

databricks/koalas/frame.py
floscha: It should probably return a Koalas Series, right?

databricks/koalas/series.py
floscha: You're right. I was only focused on the DataFrame implementation of `nunique()`. It's also fixed for Series with b78d187.

databricks/koalas/frame.py
```diff
@@ -1284,6 +1284,49 @@ def notnull(self):
...
+                                   .alias(c)
+                                    for c in self.columns])
+        else:
+            res = self._sdf.select([(F.countDistinct(Column(c))
+                                     + F.when(F.count(F.when(F.col(c).isNull(), 1).otherwise(None))
+                                              >= 1, 1).otherwise(0))
+                                   .alias(c)
+                                    for c in self.columns])
+        # TODO: Find a way to avoid converting to pandas and back
+        res = res.toPandas().T.iloc[:, 0]
```
HyukjinKwon: hm, out of curiosity, why converting to Pandas is required? I thought we can just return the Spark column without conversion.

floscha: The problem is that the select statement returns
```
   A  B
0  3  1
```

However, the expected output of the method is
```
A    3
B    1
Name: 0, dtype: int64
```

Do you know a practical way to get from the first to the latter using only PySpark?

thunterdb: Yes, we need to perform a transpose here. It should not matter performance-wise though, because there is only one row of data to convert. This will break the lineage in spark but we can always add a pure spark transpose later.

databricks/koalas/frame.py
```diff
@@ -1063,6 +1063,49 @@ def notnull(self):
...
+        return ks.Series(res)
```
thunterdb: Also, I missed this: it should return a pandas series, because it is guaranteed to be small and is typically used for point indexing, see note in the contributing guide https://github.com/databricks/koalas#return-koalas-data-structure-for-big-data-and-pandas-data-structure-for-small-data

thunterdb:
@floscha this is great, thank you for the contribution. Would you mind also adding an extra option to use spark's approx_distinct_count (https://spark.apache.org/docs/2.1.0/api/python/pyspark.sql.html#pyspark.sql.functions.approx_count_distinct)? In practice, this is what users will typically want because it makes such a big difference in runtime.

Here is how i would suggest to present it, based on  other APIs in numpy and pandas:

```py
def nunique(self, axis: int = 0, dropna: bool = True, method = 'auto') -> 'ks.Series':
  """
  ...
  method: if set to 'exact', will use the exact algorithm and return the exact number of unique. If set to 'approx', it uses Spark's approximate algorithm, which is faster in most circumstances. Default is 'exact'. Note: this parameter is specific to Spark and is not found in pandas.
  """ 
```

databricks/koalas/frame.py
floscha: Thanks for pointing out this section from the design principles. I believe this makes a strong case for using pandas Series instead of Koalas Series which I've done with 9c425a6. Also, this makes it somewhat more consistent with `Series.nunique()` which returns a simply Python int instead of some Spark-ish data type.

databricks/koalas/frame.py
floscha: Based on the comments below, `DataFrame.nunique()` now returns a pandas Series which solves the conversion issue.

floscha:
@thunterdb I agree that approximated counts would be a great thing to have and I will happily implement it. But from an API point of view, wouldn't it be better to use `approximate: bool = False` instead as it's less prone to typos?

thunterdb:
@floscha approximate: bool sounds good too. Having a string allows us to have multiple engines in the backend, but I do not foresee that happening anytime soon in Spark.

rxin:
I vote for approx to be a bit shorter. We also need an extra parameter to allow setting the error.

We should also be consistent across all the functions to use the same names.

floscha:
I'm fine with `approx` too so I changed it with 65f3007.

What error specifically do you mean @rxin?

databricks/koalas/frame.py
HyukjinKwon: Okie, but why don't we convert Pandas Series into Koalas Series as it was done before? It'd be nicer if we match the return type into Koalas series.

rxin:
Sorry for the late reply & greetings from Iceland.

I was referring to the "rsd" parameter in approx_count_distinct. We need to allow users to optionally set it.

floscha:
Greetings back to Iceland @rxin! :milky_way:

I've added the `rsd` parameter with 0ba5a09.

databricks/koalas/frame.py
floscha: Do you have good reasons for returning a Koalas Series @HyukjinKwon? Personally, I find @thunterdb's reference to the design principles (see comment below) quite compelling.

databricks/koalas/series.py
```diff
@@ -747,6 +747,44 @@ def unique(self):
...
+
+        Examples
+        --------
+        >>> ks.Series([1, 2, 3, np.nan]).nunique()
+        3
+
+        >>> ks.Series([1, 2, 3, np.nan]).nunique(dropna=False)
+        4
+        """
+        sdf = self.to_dataframe()._sdf
```
rxin: is there a way to reduce code duplication by simply calling the dataframe version of this series and just take the first number?

rxin: basically i worry as the logic becomes more complicated (which it already is), it will become difficult to maintain.

databricks/koalas/frame.py
```diff
@@ -1063,6 +1063,58 @@ def notnull(self):
...
+        if axis != 0:
+            raise ValueError("The 'nunique' method only works with axis=0 at the moment")
+        count_fn = partial(F.approx_count_distinct, rsd=rsd) if approx else F.countDistinct
+        if dropna:
+            res = self._sdf.select([count_fn(Column(c))
+                                   .alias(c)
+                                    for c in self.columns])
+        else:
+            res = self._sdf.select([(count_fn(Column(c))
+                                     + F.when(F.count(F.when(F.col(c).isNull(), 1).otherwise(None))
```
rxin: i find the when/when/otherwise pretty difficult to understand. can you add some comment explaining them above? maybe add some pseudocode ...

rxin:
I'm almost back at work (two days away). Finally had some time to read through this. Left some comments. Thanks for being patient with me @floscha!

databricks/koalas/frame.py
rxin: Since the number of columns is small, it'd make sense to follow our own principle to just return a pandas Series.

Note that an alternative design (which would require us changing the principle and every other place ...) is to always return a koalas Series everywhere.

databricks/koalas/series.py
floscha: Indeed it does :) I've implemented your proposed change with d80f6a9.

databricks/koalas/frame.py
floscha: Fair enough. 295c6c4 adds some comments that explain the rationale behind writing the select statement this way.

rxin:
Thanks @floscha. Merged this.

PR #353, opened 2019-05-19
Integration with MLflow

thunterdb:
This PR adds basic integration with MLflow, so that models that have the `pyfunc` flavor (which is, most of them), can be loaded as predictors. These predictors then works on both pandas and koalas dataframes with no code change. See the documentation example for details.

The goal of this PR is to introduce a simple interface for ML models, but keeping this dependency optional so that the base installation has minimal dependencies.

Note: travis uses pip because no version of mlflow is published in conda-forge for python 3.5


rxin:
This is super cool.

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,158 @@
...
+    Notes
+    -----
+    Currently, the model prediction can only be merged back with the existing dataframe.
+    Other columns have to be manually joined.
+    For example, this code will not work:
+
+    >>> df = ks.DataFrame({"x1": [2.0], "x2": [3.0], "z": [-1]})
+    >>> features = ["x1", "x2"]
+    >>> y = model.predict(df[features])
+    >>> #df["y"] = y # Will fail with a message about dataframes not aligned.
```
HyukjinKwon: @thunterdb, not a big deal but I think we can just do `# doctest: +SKIP`

```python
>>> # Will fail with a message about dataframes not aligned.
>>> df["y"] = y  # doctest: +SKIP
```

setup.py
```diff
@@ -40,6 +40,7 @@
     packages=['databricks', 'databricks.koalas', 'databricks.koalas.missing'],
     extras_require={
         'spark': ['pyspark>=2.4.0'],
+        'mlflow': ['mlflow>=0.2.0'],
```
HyukjinKwon: @thunterdb, can we increase the minimal version to almost the latest ones? I was thinking people will mostly use the latest version of mlflow

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,158 @@
...
+    Here is a full example that creates a model with scikit-learn and saves the model with
+     MLflow. The model is then loaded as a predictor that can be applied on a Koalas
+     dataframe.
+
+    We first initialize our MLflow environment:
+
+    >>> from mlflow.tracking import MlflowClient
+    >>> import mlflow.sklearn
+    >>> client = MlflowClient()
+    >>> print('doctest_ignore'); mlflow.set_experiment("my_experiment") # doctest: +ELLIPSIS
```
HyukjinKwon: we can do `_ = mlflow.set_experiment("my_experiment")` to ignore its output :).

.travis.yml
```diff
@@ -59,6 +59,7 @@ install:
   - conda config --env --add pinned_packages pyarrow==$PYARROW_VERSION
   - conda install -c conda-forge --yes codecov
   - conda install -c conda-forge --yes --file requirements-dev.txt
+  - pip install mlflow
```
HyukjinKwon: Out of curiosity, Is mlflow not in conda? If it;s uploaded, we can just write it into `requirements-dev.txt`.

HyukjinKwon:
Looks fine but I suspect I don't know enough of mlflow. I don't mind merging it. I will leave it to you guys.

.travis.yml
ueshin: Actually mlflow is in conda, but seems like it breaks dependency with Python 3.5 for some reason.
https://travis-ci.com/databricks/koalas/jobs/201362227

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,158 @@
...
+    Examples
+    --------
+    Here is a full example that creates a model with scikit-learn and saves the model with
+     MLflow. The model is then loaded as a predictor that can be applied on a Koalas
+     dataframe.
+
+    We first initialize our MLflow environment:
+
+    >>> from mlflow.tracking import MlflowClient
+    >>> import mlflow.sklearn
```
ueshin: `import mlflow`?

databricks/koalas/mlflow.py
ueshin: Seems like `mlflow.set_experiment("my_experiment")` does not return any values? So we don't need the trick?

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,158 @@
...
+    >>> print('doctest_ignore'); mlflow.set_experiment("my_experiment") # doctest: +ELLIPSIS
+    doctest_ignore...
+    >>> exp = client.get_experiment_by_name("my_experiment")
+
+    We aim at learning this numerical function using a simple linear regressor.
+
+    >>> from sklearn.linear_model import LinearRegression
+    >>> train = pd.DataFrame({"x1": np.arange(8), "x2": np.arange(8)**2,
+    ...                       "y": np.log(2 + np.arange(8))})
+    >>> train_x = train.drop(["y"], axis=1)
```
ueshin: `train[["x1", "x2"]]` is easier to read?

databricks/koalas/mlflow.py
HyukjinKwon: oh yea, we won't need that

.travis.yml
HyukjinKwon: hm, then can we use conda and skip it for Python 3.5 one? We can do it in a followup too.

databricks/koalas/mlflow.py
thunterdb: Actually this line is very annoying: using mlflow.set_experiment() prints a log message sometimes if the experiment does not exist already, the problem is not the result. There is no .set_or_create_experiment API, unfortunately.

.travis.yml
thunterdb: I tried hard, but currently mlflow is not packaged for python 3.5 and 3.7. I suggest that until the 1.0 release, we keep it this way?

databricks/koalas/mlflow.py
thunterdb: This model is loaded lazily, so it would not work, you need to import it explicitly

thunterdb: module

databricks/koalas/mlflow.py
thunterdb: I found a workaround in the end.

thunterdb:
@ueshin good point about the temp directory, now it puts the runs in the system temp directory, as it should.

.travis.yml
ueshin: FYI: mlflow for Python 3.7 is available on conda-forge.
If we keep this way, we need to add some note in `CONTRIBUTING.md` that we need to install mlflow separately for testing, especially for Python 3.5 users.

databricks/koalas/mlflow.py
ueshin: I mean, `import mlflow.sklearn` does not actually import the `sklearn` in the context but just `mlflow`, right?
So I think `import mlflow` is enough here.
It's not a big deal anyway, though.

ueshin: ah, I see what you mean. We need it explicitly.

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+            return self._model.predict(data)
+        if isinstance(data, DataFrame):
+            cols = [data._sdf[n] for n in data.columns]
+            return_col = self._model_udf(*cols)
+            return Series(data=return_col, anchor=data, index=data._metadata.index_map)
+
+
+def load_model(path, run_id=None, predict_type='infer') -> PythonModelWrapper:
+    """
+    Loads an MLflow model into an wrapper that can be used both for pandas and Koalas dataframes.
```
rxin: DataFrames

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Parameters
+    ----------
+    path : the path of the model, as logged when calling mlflow.log_model or mlflow.save_model
```
rxin: i think the docs convention is to just explain the data type and default/optional value in the first line, and leave the actual explanation of the parameter on the 2nd line. they are rendered differently.

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+    ...                       "y": np.log(2 + np.arange(8))})
+    >>> train_x = train[["x1", "x2"]]
+    >>> train_y = train[["y"]]
+    >>> with mlflow.start_run():
+    ...     lr = LinearRegression()
+    ...     lr.fit(train_x, train_y)
+    ...     mlflow.sklearn.log_model(lr, "model")
+    LinearRegression(copy_X=True, fit_intercept=True, n_jobs=None, normalize=False)
+
+    Now that our model is logged using MLflow, we load it back and apply it on a Koalas dataframe:
```
rxin: DataFrame

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+    Other columns have to be manually joined.
+    For example, this code will not work:
+
+    >>> df = ks.DataFrame({"x1": [2.0], "x2": [3.0], "z": [-1]})
+    >>> features = ["x1", "x2"]
+    >>> y = model.predict(df[features])
+    >>> # Will fail with a message about dataframes not aligned.
+    >>> df["y"] = y   # doctest: +SKIP
+
+    This is being tracked in the issue ticket https://github.com/databricks/koalas/issues/354.
```
rxin: are we actually going to address this issue? if not, perhaps we should create a function that simply adds a column to an existing dataframe. Then it is more obvious what can be done.

.travis.yml
tomasatdatabricks: Interesting, I was not aware we do put mlflow on conda. Maybe conda pulls it from Pip? I would definitely recommend loading it from pip for now.

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+        # TODO: do something smarter, for example when there is a sklearn.Classifier (it should
+        # return an integer or a categorical)
+        # We can do the same for pytorch/tensorflow/keras models by looking at the output types.
+        # However, this is probably better done in mlflow than here.
+        if hint == 'infer' or not hint:
+            hint = np.float64
+        return as_spark_type(hint)
+
+    @lazy_property
+    def _model(self) -> PythonModel:
```
tomasatdatabricks: The return value of load_pyfunc is PythonModel only for custom pyfunc models. Other flavors may return any object they like, the only requirement is it has a predict method.
I am not sure how the python type annotations work. Is this checked at runtime? 

Also, load_pyfunc is being renamed to load_model.

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+    def __str__(self):
+        return "PythonModelWrapper({})".format(str(self._model))
+
+    def __repr__(self):
+        return "PythonModelWrapper({})".format(repr(self._model))
+
+    def predict(self, data):
+        """
+        Returns a prediction on the data.
+        If the data is a pandas Dataframe, the return is a pandas series (TODO: np.array?)
```
tomasatdatabricks: The return value is pandas.Series.

databricks/koalas/mlflow.py
```diff
@@ -0,0 +1,166 @@
...
+        If the data is a koalas DataFrame, the return is a Koalas Series.
+        """
+        if isinstance(data, pd.DataFrame):
+            return self._model.predict(data)
+        if isinstance(data, DataFrame):
+            cols = [data._sdf[n] for n in data.columns]
+            return_col = self._model_udf(*cols)
```
tomasatdatabricks: Btw, when you apply the udf to *cols, the data is passed to the udf as an array fo columns and the column names are lost which breaks many models. Starting dbr 5.4 conda, you can pass the columns as as a struct in which case the udf gets a dataframe with column names.

tomasatdatabricks: E.g you can call it like this:
```
spark_df.withColumn("prediction", f(struct("x", "label"))).select("prediction")
```

Or you will be able to after we update our spark_udf call.

databricks/koalas/mlflow.py
thunterdb: This is being resolved as of https://github.com/databricks/koalas/pull/381 . Then users will have to  `.merge()` for the time being, but this should be acceptable enough for most use cases. I will update the documentation once that PR is merged.

databricks/koalas/mlflow.py
thunterdb: thanks for the info about the rename and the return type. Python's typing module allows to define structural subtyping but the syntax is not nice and it is not used too much by the typechecking tools anyway.

databricks/koalas/mlflow.py
thunterdb: Unfortunately, in the case of pandas dataframe input, the output can be whatever (and the doctest shows that sklearn returns a np.array instead).

databricks/koalas/mlflow.py
thunterdb: This is great to know. I just tried and it is not supported yet in OSS Spark <= 2.4.3 (currently the latest). I have put a comment to fix this, along with the code snippet, once spark 2.4.4+ is released.

databricks/koalas/mlflow.py
thunterdb: Documentation updated.

thunterdb:
@ueshin @tomasatdatabricks comments addressed. Let me know if you would like to merge.

.travis.yml
ueshin: FYI: Here's the mlflow recipe. https://github.com/conda-forge/mlflow-feedstock

databricks/koalas/mlflow.py
ueshin: Unfortunately, it won't be included in 2.4 series.

tomasatdatabricks: Makes sense.  

Unfortunately until that feature becomes available you may be returning different results for pandas dataframe vs spark dataframe. The spark_udf currently returns the left-most numeric column, so if your model predict ran on pandas df returns anything else than 1d numeric vector it's gonna be inconsistent. The return type of mlflow spark_udf  can be controlled by passing in explicit return_type argument. Unfortunately there is no way getting it back as a struct. I would consider copying the same logic here for consistency, but that's not great either.

tomasatdatabricks: Oh I've you are already passing the return type argument to spark_udf.

ueshin:
LGTM except for the Spark version in the comment related to https://github.com/databricks/koalas/pull/353/files#r287638021.

.travis.yml
HyukjinKwon: (I don't think conda pulls it from PIP out of the box)

HyukjinKwon:
Looks good to me too. I can take a separate look for conda.

thunterdb:
@ueshin comment fixed, I am merging. Thanks!

ueshin:
oops, the next Spark release will be 3.0. I'll fix it soon.

PR #359, opened 2019-05-20
Add Series.all and Series.any

HyukjinKwon:


databricks/koalas/series.py
```diff
@@ -416,6 +416,136 @@ def is_unique(self):
...
+        # if ret is None:
+        #     return False
+        # else:
+        #     return ret
+
+        # Note that we're ignoring `None`s here for now.
+        # Here we check if the count of `True`s is more than one in order to mimic `any`.
+        return sdf.select(
+            (F.count(F.when(col.cast('boolean'), 1).otherwise(None)) >= 1)
+        ).collect()[0][0]
```
HyukjinKwon: @hvanhovell do you remember this discussion about rewriting for `any` and `every`? (IIRC you suggested max/min approach but the hole was `None` handling). Can you quickly check if this rewriting makes sense to you?

Here I need to mimic both here to mimic Pandas' with rewriting since `any` and `every` only exists in the master.

HyukjinKwon:
All tests passed

databricks/koalas/series.py
HyukjinKwon: BTW, `count(col)` ignores null

HyukjinKwon:
Let me get this in. The output is as expected anyway.

databricks/koalas/series.py
hvanhovell: I can't entirely remember that comment. I did found an e-mail thread about this, and there I state the opposite: `ANY(col)` can be safely rewritten into `MAX(col)` you may need to add a coalesce if we need to return `false` for an empty dataset.

databricks/koalas/series.py
```diff
@@ -416,6 +416,136 @@ def is_unique(self):
...
+        # ret = sdf.select(F.expr("any(CAST(`%s` AS BOOLEAN))" % sdf.columns[0])).collect()[0][0]
+        # if ret is None:
+        #     return False
+        # else:
+        #     return ret
+
+        # Note that we're ignoring `None`s here for now.
+        # Here we check if the count of `True`s is more than one in order to mimic `any`.
+        return sdf.select(
+            (F.count(F.when(col.cast('boolean'), 1).otherwise(None)) >= 1)
```
hvanhovell: Two nits: `otherwise(None)` should be implied, and if you do a is not empty check why not check `> 0`?

HyukjinKwon: Yea, currently `None` is being simply ignored (that's what Pandas does by default - `skipna` argument). Yes .. `> 0` looks better. Will fix it soon. thanks :D.

HyukjinKwon: also let me give a shot with min/max ones too

PR #360, opened 2019-05-20
[285] SQL statements

### LINKED ISSUE #285: Allow querying DataFrame directly in sql method

thunterdb:
This PR builds on #291  and adds SQL statement with curly brace escaping in the query, as discussed in #285. In particular, it allows arbitrary mixing of Koalas, pandas dataframes and scalar types, see the docstring.

The doctests should give reasonable coverage.

Note: dates are not supported for now but can be added later.

Resolves #285


databricks/koalas/sql.py
```diff
@@ -0,0 +1,224 @@
...
+            if df_id not in self._temp_views:
+                sdf = var.to_spark()
+                # We should not capture extra index.
+                # TODO: document this
+                sdf = sdf[list(var.columns)]
+                sdf.createOrReplaceTempView(df_id)
+                self._temp_views[df_id] = sdf
+            return df_id
+        if isinstance(var, str):
+            return pymysql.escape_string(str)
```
HyukjinKwon: Do we need this `pymysql` dependency only to escape string?

databricks/koalas/sql.py
```diff
@@ -0,0 +1,224 @@
...
+            # Simply propagate PySpark exceptions
+            s = ("Could not execute statement: normalized_statement='{}' "
+                 "temp_views={}".format(self._normalized_statement, self._temp_views))
+            raise ValidationError(s, e)
+        finally:
+            for v in self._temp_views:
+                self._session.catalog.dropTempView(v)
+        return DataFrame(sdf)
+
+    def _convert(self, key) -> Any:
```
rxin: need doc

databricks/koalas/sql.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+class SQLProcessor(object):
+
+    def __init__(self, scope: Dict[str, Any], statement: str, session: SparkSession):
+        self._scope = scope
+        self._statement = statement
+        # All the temporary views created when executing this statement
+        self._temp_views = {}  # type: Dict[str, SDataFrame]
+        self._cached_vars = {}  # type: Dict[str, Any]
```
rxin: doc

databricks/koalas/sql.py
```diff
@@ -0,0 +1,224 @@
...
+        self._normalized_statement = None  # type: Optional[str]
```
rxin: doc on what normalized mean

requirements-dev.txt
```diff
@@ -2,6 +2,7 @@
 pandas>=0.23
 pyarrow>=0.10
 numpy>=1.14
+pymysql
```
rxin: is it just one function that we need? If that's the case, can we get rid of this dependency?

databricks/koalas/sql.py
```diff
@@ -0,0 +1,243 @@
...
+            return str(var)
+        if isinstance(var, Series):
+            return self._convert_var(var.to_dataframe())
+        if isinstance(var, pd.DataFrame):
+            return self._convert_var(ks.DataFrame(var))
+        if isinstance(var, DataFrame):
+            df_id = "koalas_" + str(id(var))
+            if df_id not in self._temp_views:
+                sdf = var.to_spark()
+                # We should not capture extra index.
```
rxin: i feel we need top keep the index, otherwise there is no way to query the index. Index is just a normal column in SQL.

rxin:
I'm going to close this one. I've taken over the work since Tim is busy with other stuff. 

New PR in: https://github.com/databricks/koalas/pull/448

Issue #146, opened 2019-04-23
Adding DataFrame.values?


thunterdb:
There has been discussion to add `DataFrame.values`, which in pandas returns a `numpy` array. I am of the opinion that we should not add this API, but other, more recent substitutes, for the following reasons:

This property is a low-level API that is meant to directly provide a view of the content of the dataframe, in the form of a numpy array. The expectation from the user and the developer (ex: pandas-profiling, seaborn) is that the result will be a numpy or numpy-like object. Spark does not provide low-level, high-performance APIs for representing arrays, so it would force the materialization of the data.

This behaviour is both unexpected from users, and not recommended by pandas itself, which has a warning in the documentation. Instead, I recommend we add the following APIs, which in practice also force a collect:
 - `to_numpy()`
 - `to_pandas()` (already implemented)
These functions have clear meaning that they convert data, and that they will return arrays of data or pandas dataframes. This is more clear to me than adding the naked `.values` property.

thunterdb:
@garawalid what are your thoughts? I think that `to_numpy()` will be what you want.

garawalid:
@thunterdb  I totally agree with you! `to_numpy` is better and gives an idea that result will be numpy array.

rxin:
I agree -- to_numpy is better!

PR #365, opened 2019-05-20
Place hint for using 'to_numpy' instead of 'values'

floscha:
As discussed in #54 and #146, Koalas will not support pandas' `values` property for the reasons discussed there. However, when currently calling `values`, the user will get an exception that the property is *not yet* implemented which gives the false impression that it will be implemented at some point. To avoid confusion, this PR "implements" the `values` property of Koalas DataFrames and Series with an exception that tells the user to call the `to_numpy()` method in order to get the data expected from pandas' `values`.


HyukjinKwon:
Looks fine. can you rebase @floscha ?

PR #366, opened 2019-05-20
Add sort_values method to Series

floscha:
This PR adds pandas' `sort_values()` method to Koalas Series after it has already been implemented for DataFrames. While at it, I also added type hints to the DataFrames method.


HyukjinKwon:
Nice. Will take a look soon

databricks/koalas/series.py
```diff
@@ -1003,6 +1003,38 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
...
+             the by.
+        inplace : bool, default False
+             if True, perform operation in-place
+        na_position : {'first', 'last'}, default 'last'
+             `first` puts NaNs at the beginning, `last` puts NaNs at the end
+
+        Returns
+        -------
+        sorted_obj : Series ordered by values.
+        """
```
HyukjinKwon: Eh, why don't we add the doctests as examples as well? (https://github.com/pandas-dev/pandas/blob/v0.24.2/pandas/core/series.py#L2668-L2831)

floscha: Good point. I've added them with fa32e67.

databricks/koalas/frame.py
```diff
@@ -1861,7 +1861,8 @@ def get(self, key, default=None):
         except (KeyError, ValueError, IndexError):
             return default
 
-    def sort_values(self, by, ascending=True, inplace=False, na_position='last'):
+    def sort_values(self, by: Union[str, List[str]], ascending: Union[bool, List[bool]] = True,
+                    inplace: bool = False, na_position: str = 'last') -> Union['DataFrame', None]:
```
HyukjinKwon: I think the return type can be`Optional['DataFrame']`

databricks/koalas/series.py
```diff
@@ -824,6 +825,158 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
...
+                                                   na_position=na_position))
+        if inplace:
+            self._kdf = ks_.to_dataframe()
+            self._scol = ks_._scol
+            self._index_map = ks_._index_map
+            return None
+        else:
+            return ks_
+
+    def isin(self, values):
```
HyukjinKwon: Nice, why don't we have two commits for isin and sort_values :)?

floscha: Oops, this somehow sneaked in while merging 'master' with 7637c6e :/

databricks/koalas/frame.py
floscha: Fair enough. Fixed it with 862b42f.

databricks/koalas/series.py
HyukjinKwon: Ah, this was moved to `IndexOpsMixin`. Let;s remove it here.

floscha: Removed with 18249a0

HyukjinKwon:
Merged. Thanks @floscha

PR #368, opened 2019-05-21
Add Index.any and Index.all

HyukjinKwon:


databricks/koalas/base.py
```diff
@@ -87,6 +89,9 @@ class IndexOpsMixin(object):
     def _with_new_scol(self, scol: spark.Column) -> IndexOpsMixin
         Creates new object with the new column
     """
+    def __init__(self, kdf, scol):
+        self._kdf = kdf
+        self._scol = scol
```
HyukjinKwon: this fixes typing test. it complains:

```
databricks/koalas/base.py: note: In member "all" of class "IndexOpsMixin":
databricks/koalas/base.py:371: error: "IndexOpsMixin" has no attribute "_kdf"
databricks/koalas/base.py:371: error: "IndexOpsMixin" has no attribute "_scol"
databricks/koalas/base.py:372: error: "IndexOpsMixin" has no attribute "_scol"
databricks/koalas/base.py: note: In member "any" of class "IndexOpsMixin":
databricks/koalas/base.py:441: error: "IndexOpsMixin" has no attribute "_kdf"
databricks/koalas/base.py:441: error: "IndexOpsMixin" has no attribute "_scol"
databricks/koalas/base.py:442: error: "IndexOpsMixin" has no attribute "_scol"
```

databricks/koalas/base.py
```diff
@@ -323,3 +328,126 @@ def notnull(self):
...
+            raise ValueError('axis should be either 0 or "index" currently.')
+
+        sdf = self._kdf._sdf.select(self._scol)
+        col = self._scol
+
+        # Note that we're ignoring `None`s here for now.
+        # any and every was added as of Spark 3.0
+        # ret = sdf.select(F.expr("any(CAST(`%s` AS BOOLEAN))" % sdf.columns[0])).collect()[0][0]
+        # Here we use max as its alternative:
+        ret = sdf.select(F.max(F.coalesce(col.cast('boolean'), F.lit(False)))).collect()[0][0]
```
HyukjinKwon: I switched previous ones to max. Seems it's simpler.

databricks/koalas/base.py
```diff
@@ -323,3 +328,126 @@ def notnull(self):
...
+            raise ValueError('axis should be either 0 or "index" currently.')
+
+        sdf = self._kdf._sdf.select(self._scol)
+        col = self._scol
+
+        # Note that we're ignoring `None`s here for now.
+        # any and every was added as of Spark 3.0
+        # ret = sdf.select(F.expr("every(CAST(`%s` AS BOOLEAN))" % sdf.columns[0])).collect()[0][0]
+        # Here we use min as its alternative:
+        ret = sdf.select(F.min(F.coalesce(col.cast('boolean'), F.lit(True)))).collect()[0][0]
```
HyukjinKwon: and, this uses min for all.

databricks/koalas/base.py
HyukjinKwon: @hvanhovell, does this make sense?

hvanhovell: How should we handle empty datasets? Should that return `False` instead of `None`?

databricks/koalas/base.py
hvanhovell: Shouldn't we use `False` as the default value here?

HyukjinKwon: it uses Pandas's behaviour .. (and appearntly that's what current `every` in Spark does too ..)

```
[null] -> true
[null, true] -> true
[true] -> true
[] -> true
```
```
[False] -> False
[null, False] -> False
```

databricks/koalas/base.py
HyukjinKwon: Ah, that's handled below if/else :).

HyukjinKwon:
I will get this in. 0.5.0 will be released today. Let me know if there are some more changes to fix. Will make a followup.

databricks/koalas/base.py
```diff
@@ -323,3 +329,125 @@ def notnull(self):
...
+        Returns True unless there at least one element within a series that is
+        False or equivalent (e.g. zero or empty)
+
+        Parameters
+        ----------
+        axis : {0 or 'index'}, default 0
+            Indicate which axis or axes should be reduced.
+
+            * 0 / 'index' : reduce the index, return a Series whose index is the
+              original column labels.
```
ueshin: nit: if we accept `'index'`, the type hint for `axis` should be `Union[int, str]`?

databricks/koalas/base.py
```diff
@@ -323,3 +329,125 @@ def notnull(self):
...
+        Returns False unless there at least one element within a series that is
+        True or equivalent (e.g. non-zero or non-empty).
+
+        Parameters
+        ----------
+        axis : {0 or 'index'}, default 0
+            Indicate which axis or axes should be reduced.
+
+            * 0 / 'index' : reduce the index, return a Series whose index is the
+              original column labels.
```
ueshin: ditto.

PR #374, opened 2019-05-22
Cleanup and inline documentation of dt functions to Pandas'

HyukjinKwon:
1. rename `datetime.py` to `datetimes.py` to match with Pandas'
2. Add some missing docstrings that has some examples
3. `lazy_property` -> `property`. I don't feel strongly but let's use `lazy_property` when it's really needed .. :-)..
4. Add the original name of the returned series.
5. Add a couple of minor tests and type hints.


HyukjinKwon:
Let me get this in. I am going to test this via notebook.

PR #380, opened 2019-05-23
Add sort_index() method

### LINKED ISSUE #304: sort_index
Should be available on both DataFrame and Series.

For backward compatibility, we might want to keep the "kind" option but state that its value is ignored.

pandas doc:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sort_index.html
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.sort_index.html

floscha:
This PR adds pandas' `sort_index()` method to both Koalas DataFrames and Series.

Resolves #304.


databricks/koalas/frame.py
```diff
@@ -1963,6 +1963,68 @@ def sort_values(self, by: Union[str, List[str]], ascending: Union[bool, List[boo
...
 
+    def sort_index(self, axis: int = 0, level: int = None, ascending: bool = True,
+                   inplace: bool = False, kind: str = None, na_position: str = 'last') \
+            -> Optional['DataFrame']:
+        """
+        Sort object by labels (along an axis)
+
+        Parameters
+        ----------
+        axis : index, columns to direct sorting
```
garawalid: I'd put that axis 1 is not supported.

databricks/koalas/frame.py
```diff
@@ -1963,6 +1963,68 @@ def sort_values(self, by: Union[str, List[str]], ascending: Union[bool, List[boo
...
+        level : int or level name or list of ints or list of level names
```
garawalid: The same here.

databricks/koalas/frame.py
```diff
@@ -1963,6 +1963,68 @@ def sort_values(self, by: Union[str, List[str]], ascending: Union[bool, List[boo
...
+            if not None, sort on values in specified index level(s)
+        ascending : boolean, default True
+            Sort ascending vs. descending
+        inplace : bool, default False
+            if True, perform operation in-place
+        kind : Koalas does not allow specifying the sorting algorithm at the moment, default None
```
garawalid: `Koalas does not allow specifying the sorting algorithm at the moment` should be in the next line.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -307,6 +307,26 @@ def test_sort_values(self):
         self.assert_eq(kdf.sort_values('b', inplace=True), pdf.sort_values('b', inplace=True))
         self.assert_eq(repr(kdf), repr(pdf))
 
+    def test_sort_index(self):
+        pdf = pd.DataFrame({'A': [2, 1, np.nan]}, index=['b', 'a', np.nan])
```
garawalid: I suggest that the DataFrame contains two columns.

databricks/koalas/series.py
```diff
@@ -929,6 +929,75 @@ def sort_values(self, ascending: bool = True, inplace: bool = False,
         else:
             return ks_
 
+    def sort_index(self, axis: int = 0, level: int = None, ascending: bool = True,
+                   inplace: bool = False, kind: str = None, na_position: str = 'last') \
+            -> Optional['Series']:
+        """
```
garawalid: The docstring is almost the same as `DataFrame.sort_index()`. You can use 
```python
sort_index.__doc__ = DataFrame.sort_index.__doc__
```

databricks/koalas/frame.py
```diff
@@ -1963,6 +1963,68 @@ def sort_values(self, by: Union[str, List[str]], ascending: Union[bool, List[boo
...
+        b    2.0
+        NaN  NaN
+        """
+        if axis != 0:
+            raise ValueError("No other axes than 0 are supported at the moment")
+        if level is not None:
+            raise ValueError("The 'axis' argument is not supported at the moment")
+        if kind is not None:
+            raise ValueError("Specifying the sorting algorithm is supported at the moment.")
+        return self.sort_values(by='__index_level_0__', ascending=ascending, inplace=inplace,
```
ueshin: The actual index column names are in the metadata. We should use it: `by=self._metadata.index_columns`.

databricks/koalas/series.py
```diff
@@ -929,6 +929,75 @@ def sort_values(self, ascending: bool = True, inplace: bool = False,
...
+        NaN    NaN
+        Name: 0, dtype: float64
+        """
+        if axis != 0:
+            raise ValueError("No other axes than 0 are supported at the moment")
+        if level is not None:
+            raise ValueError("The 'axis' argument is not supported at the moment")
+        if kind is not None:
+            raise ValueError("Specifying the sorting algorithm is supported at the moment.")
+        ks_ = _col(self.to_dataframe().sort_values(by='__index_level_0__', ascending=ascending,
```
ueshin: ditto.

databricks/koalas/frame.py
thunterdb: on that note, can you add a function in `utils.py` called `check_axis` for example that throws an exception if the axis is not 0? It will help us track everywhere where we add it eventually.

floscha: I've extended the docstrings with 0bf4542. The `check_axis` method sounds like a good idea. However, I will add it with a new PR to keep this one clean, if that's okay for you @thunterdb.

databricks/koalas/frame.py
floscha: Also resolved with 0bf4542.

databricks/koalas/frame.py
floscha: Fair point. I've fixed it with c899fde.

databricks/koalas/tests/test_dataframe.py
floscha: Changed with 0db87fd.

databricks/koalas/series.py
floscha: Fair point but how does this deal with the doctests that need to use Series instead of DataFrames?

databricks/koalas/frame.py
floscha: I didn't know that. Thanks a lot for pointing it out! I'm sure I've also used `'__index_level_0__'` in some other PRs so I'll make sure to fix it there as well but for now, 46fbd2e fixes it for this PR.

databricks/koalas/series.py
floscha: Also fixed with 46fbd2e.

databricks/koalas/frame.py
ueshin: If the PR in your mind is the `merge` one, I'm working on it with some improvements, otherwise please go ahead. Thanks!

floscha: Thanks for letting me know. I've just done a text search across the project and `merge` seems to be the only other place where '__index_level_0__' appears right now, except for the `meta` module.

databricks/koalas/series.py
garawalid: You can include one example of Series in the same doctest of DataFrame.

databricks/koalas/frame.py
```diff
@@ -1963,6 +1963,69 @@ def sort_values(self, by: Union[str, List[str]], ascending: Union[bool, List[boo
...
+        b    2.0
+        NaN  NaN
+        """
+        if axis != 0:
+            raise ValueError("No other axes than 0 are supported at the moment")
+        if level is not None:
+            raise ValueError("The 'axis' argument is not supported at the moment")
+        if kind is not None:
+            raise ValueError("Specifying the sorting algorithm is supported at the moment.")
+        return self.sort_values(by=self._metadata.index_columns[0], ascending=ascending,
```
ueshin: I don't think we need `[0]`, for multi-index support. Could you add tests for multi-index?

databricks/koalas/series.py
```diff
@@ -929,6 +929,76 @@ def sort_values(self, ascending: bool = True, inplace: bool = False,
...
+        NaN    NaN
+        Name: 0, dtype: float64
+        """
+        if axis != 0:
+            raise ValueError("No other axes than 0 are supported at the moment")
+        if level is not None:
+            raise ValueError("The 'axis' argument is not supported at the moment")
+        if kind is not None:
+            raise ValueError("Specifying the sorting algorithm is supported at the moment.")
+        ks_ = _col(self.to_dataframe().sort_values(by=self._metadata.index_columns[0],
```
ueshin: ditto.

databricks/koalas/tests/test_series.py
```diff
@@ -242,6 +242,27 @@ def test_sort_values(self):
...
+        # Assert default behavior without parameters
+        self.assert_eq(repr(ks.sort_index()), repr(ps.sort_index()))
+        # Assert sorting descending
+        self.assert_eq(repr(ks.sort_index(ascending=False)), repr(ps.sort_index(ascending=False)))
+        # # Assert sorting NA indices first
+        self.assert_eq(repr(ks.sort_index(na_position='first')),
+                       repr(ps.sort_index(na_position='first')))
+        # # Assert sorting inplace
+        self.assert_eq(ks.sort_index(inplace=True), ps.sort_index(inplace=True))
+        self.assert_eq(repr(ks), repr(ps))
```
ueshin: Do we need `repr` here?

databricks/koalas/frame.py
floscha: Good idea! I've added multi-index support with 95dcbdf.

databricks/koalas/series.py
floscha: See comment above.

databricks/koalas/tests/test_series.py
floscha: Unfortunately yes 😕Otherwise the tests will fail.

databricks/koalas/tests/test_series.py
```diff
@@ -242,6 +242,32 @@ def test_sort_values(self):
...
+
+        # Assert default behavior without parameters
+        self.assert_eq(repr(ks.sort_index()), repr(ps.sort_index()))
+        # Assert sorting descending
+        self.assert_eq(repr(ks.sort_index(ascending=False)), repr(ps.sort_index(ascending=False)))
+        # Assert sorting NA indices first
+        self.assert_eq(repr(ks.sort_index(na_position='first')),
+                       repr(ps.sort_index(na_position='first')))
+        # Assert sorting inplace
+        self.assert_eq(ks.sort_index(inplace=True), ps.sort_index(inplace=True))
```
ueshin: nit: this is only comparing `None` vs. `None`.

databricks/koalas/tests/test_series.py
ueshin: ah, I see. maybe `almost=True` should work.

floscha: It works indeed. So with a64999e I've replaced all instances of `repr()` with `almost=True`.

databricks/koalas/tests/test_series.py
floscha: I've replaced `assert_eq` with `assertEquals` in a64999e. However, comparing `None` with `None` is still necessary to make sure the Koalas implementation has the same return value as the pandas equivalent.

ueshin:
Thanks! merging to master!

PR #381, opened 2019-05-24
Improve merge function.

### LINKED ISSUE #354: DataFrame.merge issues with specifying both joins
The `.merge` function does not seem to work when the index is set through `set_index`. Here are two examples that works with pandas but fail with Koalas:

```py
df1 = pd.DataFrame({"a":[0], "x": [-1]})
df2 = pd.DataFrame({"b":[0], "y": [-1]})
df1.set_index("a").merge(df2.set_index("b"), left_index=True, right_index=True)
df1.merge(df2, left_on=["a"], right_on=["b"])
```

ueshin:
This PR improves `merge` function:

- add `left_on` and `right_on`
- fix index management

but still have some TODOs:

- in some case, indexes can be retained
- when `left_index=True` and `right_index=True` for multi-index, the behavior is not the same as pandas'

I'd leave them to the future work.

Resolves #354.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -24,7 +24,7 @@ from typing import Any, Optional, List, Tuple, Union
 
 import numpy as np
 import pandas as pd
-from pandas.api.types import is_datetime64_dtype, is_datetime64tz_dtype, is_list_like, \
+from pandas.api.types import is_bool, is_datetime64_dtype, is_datetime64tz_dtype, is_list_like, \
     is_dict_like
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Column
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -2356,41 +2426,55 @@ def merge(self, right: 'DataFrame', how: str = 'inner', on: str = None,
...
-            # Merge duplicate key columns
-            joined_table = joined_table.withColumnRenamed(on + left_suffix, on)
-            joined_table = joined_table.drop(on + right_suffix)
+        left_index_columns = set(left._metadata.index_columns)
+        right_index_columns = set(right._metadata.index_columns)
 
-        # Remove auxiliary index
-        # FIXME Move index string to constant?
-        joined_table = joined_table.drop('__index_level_0__')
+        # TODO: in some case, we can remain indexes.
```
thunterdb: remain -> keep

thunterdb:
@ueshin this is great!

Issue #382, opened 2019-05-24
Implement missing functions for 10-minute tutorial with Koalas


garawalid:
This is the list of missing functions for **10-minute tutorial with Koalas** (#196) 

+ [ ] DataFrame.pivot_table
+ [ ] DataFrame.reindex
+ [ ] DataFrame.apply
+ [x] DataFrame.append (#388)
+ [x] DataFrame.at (#384)
+ [ ] DataFrame.T
+ [x] DataFrame.sub (#427)

Ignored functions
+ DataFrame.to_hdf
+ DataFrame.iat
+ DataFrame.shift

Some functions could be implemented for both DataFrame and Series. 

For contributors, please mention that you're working on some functions here and go ahead.

garawalid:
@HyukjinKwon any thoughts?

thunterdb:
Let's focus on: pivot_table, reindex, apply, append, at, T, to_hdf (in this order)

The other functions have big performance impact and will require more careful integration with spark, by using window functions under the hood for example.

@garawalid, if you are looking at implementing some of them, I suggest that we go in this order. What are your thoughts?

floscha:
I can definitely start with `reindex`,  `append` and `at` as those three all seem pretty straight forward.

The others seem tricky to me for different reasons:
- `pivot_table` and `T` might require special care with their implementation as they potentially move around huge amounts of data between workers.
- `apply` should probably take UDFs since pandas/NumPy methods need to move the data from the Spark process to the Python process and back. Expecting the user to use PySpark UDFs however, would (again) break the separation between the (Py)Spark and pandas APIs.
- `to_hdf` should be easy to implement but requires a decision if the write to disk should happen from Spark or from Python.

rxin:
Does anybody still use hdf these days? Maybe I'm not familiar enough with the pandas use cases. I feel maybe it's ok to skip hdf given people rarely use that in "big data".

athena15:
@rxin, FWIW I think you are right that to_hdf is a very rarely used function. I'm sure there are edge cases, but I've never seen it used.

floscha:
I've seen people (including myself) using HDF5 files to store Keras models but I've also never seen anyone using it with pandas or Spark. So probably it makes most sense to leave it out for now and wait until someone requests this feature (if ever).

garawalid:
@thunterdb `pivot_table` sounds a little bit tricky, I'll start with a simple version and we will improve it with different PR. 

[...]

PR #384, opened 2019-05-25
Implement .at property for Koalas DataFrames and Series

floscha:
As requested in #382, this PR implements pandas' `.at` property for Koalas DataFrames and Series.

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -132,6 +132,10 @@ class AtIndexer(object):
         if self._ks is not None and len(key) != 1:
             raise TypeError("Use Series.at like .at[row_index]")
 
+        # TODO Maybe extend to multilevel indices in the future
+        if len(self._kdf._metadata.index_columns) != 1:
+            raise ValueError("'.at' only supports indices with level 1 right now")
+
         column = key[1] if len(key) > 1 else self._ks.name
         if column is not None and column not in self._kdf._metadata.data_columns:
             raise KeyError("'%s" % column)
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -88,7 +88,10 @@ class AtIndexer(object):
     Similar to ``loc``, in that both provide label-based lookups. Use ``at`` if you only need to
     get a single value in a DataFrame or Series.
 
-    Unlike pandas, Koalas only allows using ``at`` to get values but not to set them.
+    .. note:: Unlike pandas, Koalas only allows using ``at`` to get values but not to set them.
+
+    .. note:: Warning: If ``row_index`` matches a lot of rows, large amounts of data will be
+        fetched, potentially causing your machine to run out of memory.
 
     Raises
     ------
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -116,7 +116,7 @@ class AtIndexer(object):
 
     Get array if an index occurs multiple times
 
-    >>> kdf.to_pandas().at[5, 'B']
+    >>> kdf.at[5, 'B']
     array([ 4, 20])
     """
     def __init__(self, df_or_s):
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -130,19 +130,19 @@ class AtIndexer(object):
     def __getitem__(self, key):
         if self._ks is None and (not isinstance(key, tuple) or len(key) != 2):
             raise TypeError("Use DataFrame.at like .at[row_index, column_name]")
-        if self._ks is not None and len(key) != 1:
+        if self._ks is not None and not isinstance(key, str) and len(key) != 1:
             raise TypeError("Use Series.at like .at[row_index]")
 
         # TODO Maybe extend to multilevel indices in the future
         if len(self._kdf._metadata.index_columns) != 1:
             raise ValueError("'.at' only supports indices with level 1 right now")
 
-        column = key[1] if len(key) > 1 else self._ks.name
+        column = key[1] if self._ks is None else self._ks.name
         if column is not None and column not in self._kdf._metadata.data_columns:
             raise KeyError("%s" % column)
         series = self._ks if self._ks is not None else self._kdf[column]
 
-        row = key[0]
+        row = key[0] if self._ks is None else key
         pdf = (series._kdf._sdf
                .where(F.col(self._kdf._metadata.index_columns[0]) == row)
                .select(column)
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -127,7 +127,7 @@ class AtIndexer(object):
     def __getitem__(self, key):
         from databricks.koalas.frame import DataFrame
 
-        if self._ks is None and len(key) != 2:
+        if self._ks is None and (not isinstance(key, tuple) or len(key) != 2):
             raise TypeError("Use DataFrame.at like .at[row_index, column_name]")
         if self._ks is not None and len(key) != 1:
             raise TypeError("Use Series.at like .at[row_index]")
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -128,8 +128,6 @@ class AtIndexer(object):
             self._ks = df_or_s
 
     def __getitem__(self, key):
-        from databricks.koalas.frame import DataFrame
-
         if self._ks is None and (not isinstance(key, tuple) or len(key) != 2):
             raise TypeError("Use DataFrame.at like .at[row_index, column_name]")
         if self._ks is not None and len(key) != 1:
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -138,7 +138,7 @@ class AtIndexer(object):
 
         column = key[1] if len(key) > 1 else self._ks.name
         if column is not None and column not in self._kdf._metadata.data_columns:
-            raise KeyError("'%s" % column)
+            raise KeyError("%s" % column)
         series = self._ks if self._ks is not None else self._kdf[column]
 
         row = key[0]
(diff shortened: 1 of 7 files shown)


databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+        from databricks.koalas.frame import DataFrame
+
+        if self._ks is None and len(key) != 2:
+            raise TypeError("Use DataFrame.at like .at[row_index, column_name]")
+        if self._ks is not None and len(key) != 1:
+            raise TypeError("Use Series.at like .at[column_name]")
+
+        column = key[1] if len(key) > 1 else self._ks.name
+        if column is not None and column not in self._kdf._metadata.data_columns:
+            raise KeyError("'%s" % column)
```
ueshin: just out of curiosity, what's `'` for?

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+        series = self._ks if self._ks is not None else self._kdf[column]
+
+        row = key[0]
+        sdf = (series._kdf._sdf
+               .where(F.col(self._kdf._metadata.index_columns[0]) == row)
+               .select(column))
+        if sdf.count() < 1:
+            raise KeyError("'%s" % row)
```
ueshin: ditto.

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+
+        values = DataFrame(sdf).to_pandas().iloc[:, 0].values
```
ueshin: Actually we don't need to create Koalas `DataFrame` here. Spark DataFrame has `toPandas()`.

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+            self._kdf = df_or_s._kdf
+            self._ks = df_or_s
+
+    def __getitem__(self, key):
+        from databricks.koalas.frame import DataFrame
+
+        if self._ks is None and len(key) != 2:
+            raise TypeError("Use DataFrame.at like .at[row_index, column_name]")
+        if self._ks is not None and len(key) != 1:
+            raise TypeError("Use Series.at like .at[column_name]")
```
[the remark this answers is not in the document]
floscha: True. Fixed with 8137630.

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+        else:
+            # If df_or_col is Column, store both the DataFrame anchored to the Column and
+            # the Column itself.
+            self._kdf = df_or_s._kdf
+            self._ks = df_or_s
+
+    def __getitem__(self, key):
+        from databricks.koalas.frame import DataFrame
+
+        if self._ks is None and len(key) != 2:
```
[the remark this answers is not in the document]
floscha: Sure, added with 09df062.

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+            raise TypeError("Use Series.at like .at[column_name]")
+
+        column = key[1] if len(key) > 1 else self._ks.name
+        if column is not None and column not in self._kdf._metadata.data_columns:
+            raise KeyError("'%s" % column)
+        series = self._ks if self._ks is not None else self._kdf[column]
+
+        row = key[0]
+        sdf = (series._kdf._sdf
+               .where(F.col(self._kdf._metadata.index_columns[0]) == row)
```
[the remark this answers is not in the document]
floscha: I've added a check now that raises an exception if the index level doesn't equal 1 with a61f163. I'll have to look closer into how `.at` behaves for multilevel indices which generally seems to be rather uncommon. Also, is a Koalas DataFrame without an index even valid? Like, wouldn't it for example crash when calling `to_pandas()` on it?

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,72 @@ def _unfold(key, kseries):
...
+               .select(column))
+        if sdf.count() < 1:
```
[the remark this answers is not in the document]
floscha: Good point! I've implemented your suggestion in 2dc8fa5.

databricks/koalas/indexing.py
floscha: You're right. Fixed this with 2dc8fa5.

databricks/koalas/indexing.py
floscha: This is just to be consistent with the way pandas uses Python's `KeyError`, putting the missing key between `'`s. However, I've just realized I've missed the closing `'` and the `'`s are not even necessary to start with 🙈 So I've fixed this with 2dc8fa5.

databricks/koalas/indexing.py
floscha: See comment above.

databricks/koalas/indexing.py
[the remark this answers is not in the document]
ueshin: As for multi- or no index, currently `.loc` raises an exception for multi- or no index. Maybe we can follow the behavior for now, but please make sure to check it and raise the exception.
And yes, no index in Koalas DataFrame is valid, but I'm not sure whether we should crash in that case. E.g., `sdf.to_koalas()` doesn't have index, but should we explicitly `set_index` prior to `to_pandas()` in that case? This is the future work anyway. Let's file an issue to discuss it.
Thanks!

floscha:
Thanks for your thorough review @ueshin!

I've addressed all of your comments and also added a warning note for matching a lot of rows with f14d136.

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,80 @@ def _unfold(key, kseries):
...
+            # the Column itself.
+            self._kdf = df_or_s._kdf
+            self._ks = df_or_s
+
+    def __getitem__(self, key):
+        from databricks.koalas.frame import DataFrame
+
+        if self._ks is None and (not isinstance(key, tuple) or len(key) != 2):
+            raise TypeError("Use DataFrame.at like .at[row_index, column_name]")
+        if self._ks is not None and len(key) != 1:
```
ueshin: Need to check key is not str?

databricks/koalas/indexing.py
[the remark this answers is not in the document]
ueshin: Could you add tests for such a case?

databricks/koalas/indexing.py
ueshin: I mean what if `kdf.a.at['abc']`, for example?

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,80 @@ def _unfold(key, kseries):
     return rows_sel, cols_sel
 
 
+class AtIndexer(object):
+    """
+    Access a single value for a row/column label pair.
+    Similar to ``loc``, in that both provide label-based lookups. Use ``at`` if you only need to
+    get a single value in a DataFrame or Series.
```
thunterdb: you should also mention the behavior when multiple values are matched (returning an array)

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,80 @@ def _unfold(key, kseries):
...
+    ...                    index=[4, 5, 6], columns=['A', 'B', 'C'])
+    >>> kdf
+        A   B   C
+    4   0   2   3
+    5   0   4   1
+    6  10  20  30
+
+    Get value at specified row/column pair
+
+    >>> kdf.at[4, 'B']
```
thunterdb: can you also put an example with multiple values matching the same index, to show that it returns an array?

databricks/koalas/indexing.py
floscha: I've added it to the docs with 0dc9e08.

databricks/koalas/indexing.py
floscha: Sure, done with 83a77d0.

databricks/koalas/indexing.py
floscha: Good point! I've added the check and an additional unit test (that used to fail previously) with 
415e78a.

databricks/koalas/indexing.py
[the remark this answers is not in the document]
floscha: Sure, done with fb3587d.

databricks/koalas/indexing.py
```diff
@@ -82,6 +82,84 @@ def _unfold(key, kseries):
...
+    5  10  20  30
+
+    Get value at specified row/column pair
+
+    >>> kdf.at[4, 'B']
+    2
+
+    Get array if an index occurs multiple times
+
+    >>> kdf.to_pandas().at[5, 'B']
```
ueshin: `kdf.at[5, 'B']` ?

databricks/koalas/tests/test_indexing.py
```diff
@@ -126,6 +126,45 @@ def pdf(self):
...
+        self.assertEqual(kdf.at[3, 'b'], 6)
+        self.assertEqual(kdf.at[3, 'b'], pdf.at[3, 'b'])
+        np.testing.assert_array_equal(kdf.at[9, 'b'], np.array([0, 0, 0]))
+        np.testing.assert_array_equal(kdf.at[9, 'b'], pdf.at[9, 'b'])
+
+        # Assert .at for Series
+        self.assertEqual(test_series.at['b'], 6)
+        self.assertEqual(test_series.at['b'], pdf.loc[3].at['b'])
+
+        #
```
ueshin: forgot to add a comment?

databricks/koalas/indexing.py
floscha: You're right. Fixed with ebea010.

databricks/koalas/tests/test_indexing.py
floscha: Correct. Guess I was kind of tired yesterday when committing this... Now fixed with ebea010 though.

ueshin:
Thanks! merging.

PR #385, opened 2019-05-25
Run mypy before sphinx test

floscha:
This PR simply modifies the *lint-python* script to run mypy before sphinx. My rationale behind this is that the mypy check fails much more often than sphinx, while the latter takes comparably long to run. Running mypy first thus leads to faster feedback and therefore increased development speed.

### DIFF
--- a/dev/lint-python
+++ b/dev/lint-python
@@ -211,6 +211,32 @@ Skipping pydocstyle checks for now."
     fi
 }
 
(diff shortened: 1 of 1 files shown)


ueshin:
Sounds good.
I'd merge this now.

HyukjinKwon:
+1

PR #386, opened 2019-05-26
DataFrame.pivot_table

### LINKED ISSUE #382: Implement missing functions for 10-minute tutorial with Koalas

garawalid:
+ Initial implementation of `DataFrame.pivot_table()`
+ Ignored parameters : `margins`, `dropna` and `margins_name`
+ Resolves #382

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -39,7 +39,6 @@ from databricks.koalas.internal import _InternalFrame
 from databricks.koalas.missing.frame import _MissingPandasLikeDataFrame
 from databricks.koalas.ml import corr
 
-
 # These regular expression patterns are complied and defined here to avoid to compile the same
 # pattern every time it is used in _repr_ and _repr_html_ in DataFrame.
 # Two patterns basically seek the footer string from Pandas'
@@ -2469,7 +2468,92 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
     def pivot_table(self, values=None, index=None, columns=None,
                     aggfunc='mean', fill_value=None):
+        """
+        Create a spreadsheet-style pivot table as a DataFrame. The levels in
+        the pivot table will be stored in MultiIndex objects (hierarchical
+        indexes) on the index and columns of the result DataFrame.
 
+        Parameters
+        ----------%s
+        values : column to aggregate.
+            They sould be either a list or a string.
+        index : column (string) or list of columns
+            If an array is passed, it must be the same length as the data.
+            The list should contain string.
+        columns : column
+            Columns used in the pivot operation. Only one column is supported and
+            it should be a string.
+        aggfunc : function (string), dict, default mean
+            If dict is passed, the resulting pivot table will have
+            columns concatenated by "_" where the first part is the value
+            of columns and the second part is the column name in values
+            If dict is passed, the key is column to aggregate and value
+            is function or list of functions.
+        fill_value : scalar, default None
+            Value to replace missing values with.
+
+        Returns
+        -------
+        table : DataFrame
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({"A": ["foo", "foo", "foo", "foo", "foo",
+        ...                          "bar", "bar", "bar", "bar"],
+        ...                    "B": ["one", "one", "one", "two", "two",
+        ...                          "one", "one", "two", "two"],
+        ...                    "C": ["small", "large", "large", "small",
+        ...                          "small", "large", "small", "small",
+        ...                          "large"],
+        ...                    "D": [1, 2, 2, 3, 3, 4, 5, 6, 7],
+        ...                    "E": [2, 4, 5, 5, 6, 6, 8, 9, 9]})
+        >>> df
+             A    B      C  D  E
+        0  foo  one  small  1  2
+        1  foo  one  large  2  4
+        2  foo  one  large  2  5
+        3  foo  two  small  3  5
+        4  foo  two  small  3  6
+        5  bar  one  large  4  6
+        6  bar  one  small  5  8
+        7  bar  two  small  6  9
+        8  bar  two  large  7  9
+
+        This first example aggregates values by taking the sum.
+
+        >>> table = df.pivot_table(values='D', index=['A', 'B'],
+        ...                     columns='C', aggfunc='sum')
+        >>> table # doctest: +SKIP
+                 large  small
+        A   B
+        foo one    4.0      1
+            two    NaN      6
+        bar two    7.0      6
+            one    4.0      5
+
+        We can also fill missing values using the `fill_value` parameter.
+
+        >>> table = df.pivot_table(values='D', index=['A', 'B'],
+        ...                     columns='C', aggfunc='sum', fill_value=0)
+        >>> table # doctest: +SKIP
+                 large  small
+        A   B
+        foo one      4      1
+            two      0      6
+        bar two      7      6
+            one      4      5
+
+        We can also calculate multiple types of aggregations for any given
+        value column.
+
+        >>> table = df.pivot_table(values = ['D','E'], index =['C'],
+        ...                     columns="A", aggfunc={'D':'mean','E':'sum'})
+        >>> table # doctest: +SKIP
+               bar_D  bar_E     foo_D  foo_E
+        C
+        small    5.5     17  2.333333     13
+        large    5.5     15  2.000000      9
+        """
         if not isinstance(columns, str):
             raise ValueError("columns should be string.")
 
@@ -2486,7 +2570,6 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                                       " as dict and without index.")
 
         if isinstance(aggfunc, str):
-
             agg_cols = [F.expr('{1}({0}) as {0}'.format(values, aggfunc))]
 
         elif isinstance(aggfunc, dict):
@@ -2498,7 +2581,6 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 raise ValueError("Columns in aggfunc must be the same as values.")
 
         if index is None:
-
             sdf = self._sdf.groupBy().pivot(pivot_col=columns).agg(*agg_cols)
 
         elif isinstance(index, list):
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2475,7 +2475,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         indexes) on the index and columns of the result DataFrame.
 
         Parameters
-        ----------%s
+        ----------
         values : column to aggregate.
             They should be either a list of one column or a string. A list of columns
             is not supported yet.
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -1653,6 +1653,58 @@ def head(self, n=5):
 
         return DataFrame(self._sdf.limit(n), self._metadata.copy())
 
+    def pivot_table(self, values=None, index=None, columns=None,
+                    aggfunc='mean', fill_value=None):
+
+        if not isinstance(columns, str):
+            raise ValueError("columns should be string.")
```
floscha: I know the code base is not 100% consistent in this regard, but error messages should probably start with uppercase letters.

databricks/koalas/frame.py
```diff
@@ -1653,6 +1653,58 @@ def head(self, n=5):
...
+                isinstance(key, str) and isinstance(value, str) for key, value in aggfunc.items())):
+            raise ValueError("aggfunc must be a dict mapping from column name (string) "
+                             "to aggregate functions (string).")
+
+        if isinstance(aggfunc, dict) and index is None:
+            raise NotImplementedError("pivot_table doesn't support aggfuct"
+                                      " as dict and without index.")
+
+        if isinstance(aggfunc, str):
+
```
floscha: This empty line should be removed.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -569,3 +569,45 @@ def test_sample(self):
...
+
+        msg = "values should be string or list of columns."
+        with self.assertRaisesRegex(ValueError, msg):
+            kdf.pivot_table(index=['c'], columns="a", values=5)
+
+        msg = "index should be a None or a list of columns."
+        with self.assertRaisesRegex(ValueError, msg):
+            kdf.pivot_table(index="c", columns="a", values="b")
+
+        pdf = pd.DataFrame({'a': [4, 2, 3, 4, 8, 6],
```
floscha: Could this DataFrame not simply be defined in line 575?

floscha:
The unit tests you've added so far only assert exception handling. It would be great if you could add both docstest and unit tests that assert the actual pivoting functionality. You could use the [pandas examples](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.pivot_table.html) to start with.

garawalid:
@floscha thanks for your comments. I skipped the unit test because the result is not exactly the same and requires changing the design of the metadata. I'll add some examples in the docstring. 

Here is an example : 
```python 
print(kdf.pivot_table(index = ['c'], columns="a",values="b"))
print(pdf.pivot_table(index = ['c'], columns=["a"],values="b"))

print(kdf.pivot_table(index = ['c'], columns="a",values = ['b','e'],aggfunc={'b':'mean','e':'sum'}))
print(pdf.pivot_table(index = ['c'], columns=["a"],values = ['b','e'], aggfunc={'b':'mean','e':'sum'}))
```

```
     2    3    4    6    8
c                         
7  NaN  NaN  NaN  NaN  2.0
9  NaN  2.0  NaN  NaN  NaN
1  NaN  NaN  1.0  NaN  NaN
2  2.0  NaN  NaN  NaN  NaN
4  NaN  NaN  4.0  4.0  NaN
  
a    2    3    4    6    8
c                         
1  NaN  NaN  1.0  NaN  NaN
2  2.0  NaN  NaN  NaN  NaN
4  NaN  NaN  4.0  4.0  NaN
7  NaN  NaN  NaN  NaN  2.0
9  NaN  2.0  NaN  NaN  NaN


   2_b  2_e  3_b  3_e  4_b  4_e  6_b  6_e  8_b  8_e
c                                                  
7  NaN  NaN  NaN  NaN  NaN  NaN  NaN  NaN  2.0  2.0
9  NaN  NaN  2.0  2.0  NaN  NaN  NaN  NaN  NaN  NaN
1  NaN  NaN  NaN  NaN  1.0  1.0  NaN  NaN  NaN  NaN
2  2.0  2.0  NaN  NaN  NaN  NaN  NaN  NaN  NaN  NaN
4  NaN  NaN  NaN  NaN  4.0  4.0  4.0  4.0  NaN  NaN
  
     b                        e                    
a    2    3    4    6    8    2    3    4    6    8
c                                                  
1  NaN  NaN  1.0  NaN  NaN  NaN  NaN  1.0  NaN  NaN
2  2.0  NaN  NaN  NaN  NaN  2.0  NaN  NaN  NaN  NaN
4  NaN  NaN  4.0  4.0  NaN  NaN  NaN  4.0  4.0  NaN
7  NaN  NaN  NaN  NaN  2.0  NaN  NaN  NaN  NaN  2.0
9  NaN  2.0  NaN  NaN  NaN  NaN  2.0  NaN  NaN  NaN

```

databricks/koalas/tests/test_dataframe.py
garawalid: I checked it and it works. Thanks!

databricks/koalas/tests/test_dataframe.py
```diff
@@ -569,3 +569,45 @@ def test_sample(self):
             kdf.sample()
         with self.assertRaises(NotImplementedError):
             kdf.sample(n=1)
+
+    def test_pivot_table(self):
```
thunterdb: you should be able to replicate the dask test in https://github.com/dask/dask/blob/master/dask/dataframe/tests/test_reshape.py#L188

(without the categorical example)

garawalid: Thanks @thunterdb. 
I kept some tests as `Todo`, I'll update them once we implement hierarchical columns and column's name.

thunterdb: Sounds great.

floscha:
I see your point @garawalid and I agree that changing Koalas' way of handling multilevel indices/columns should not be part of this PR. However, pandas' `pivot_table` appears to sort the indices which can also easily be achieved with Koalas by calling `.sort_index()` on the pivoted DataFrame. Maybe you should add this to better conform with the expectations of pandas users. On the downside, `.sort_index()` might cause some computational overhead that I cannot really estimate.

garawalid:
Thanks @floscha for pointing out `sort_index()`, I think we can leave it like this and Pandas users will do it manually if necessary. In fact, we have to keep in mind that the DataFrame can be big, and they won't be able to see the full DataFrame. Also, the cost of sorting becomes an overhead if the DataFrame is big. Meanwhile, we can use `sort_index()` when we compare results in tests

databricks/koalas/tests/test_dataframe.py
garawalid: Good.

databricks/koalas/frame.py
```diff
@@ -1653,6 +1652,141 @@ def head(self, n=5):
...
 
+    def pivot_table(self, values=None, index=None, columns=None,
+                    aggfunc='mean', fill_value=None):
+        """
+        Create a spreadsheet-style pivot table as a DataFrame. The levels in
+        the pivot table will be stored in MultiIndex objects (hierarchical
+        indexes) on the index and columns of the result DataFrame.
+
+        Parameters
+        ----------%s
```
HyukjinKwon: seems `%s` should be removed.

HyukjinKwon:
hey @ganeshchand can we update this one? It's nice to have it.

floscha:
Didn't you mean to ping @garawalid?

HyukjinKwon:
D'oh

databricks/koalas/frame.py
garawalid: I'll take care of it.

garawalid:
@HyukjinKwon thanks for the review. I'll update this PR. I've been quite busy lately, I'll join you ASAP. 

@HyukjinKwon @floscha 
I've been thinking about discarding the case where `values` is a list. And, we can update it later once we have multi level columns. 

What do you think ?

floscha:
I think generally it's fine to only support one value for now and add multi-level support later. However, I would still treat `values` as a list (for type hints and stuff) but raise a `ValueError` for `len(values) > 1`, telling the user that only pivoting for one level is available for now. Then someone can add multi-level support later without changing the signature of the method.

garawalid:
@floscha  thanks for your suggestion. 

I disabled multi level columns for instance. 
@floscha, @HyukjinKwon  would you mind review this PR.

databricks/koalas/frame.py
```diff
@@ -2467,6 +2467,145 @@ def head(self, n=5):
...
```
HyukjinKwon: Let's remove `%s`. Otherwise, I think @floscha's suggestion sounds fine.

databricks/koalas/frame.py
```diff
@@ -2467,6 +2467,145 @@ def head(self, n=5):
...
+        A   B
+        foo one    4.0      1
+            two    NaN      6
+        bar two    7.0      6
+            one    4.0      5
+
+        We can also fill missing values using the `fill_value` parameter.
+
+        >>> table = df.pivot_table(values='D', index=['A', 'B'],
+        ...                     columns='C', aggfunc='sum', fill_value=0)
```
HyukjinKwon: looks indentation should be fixed

databricks/koalas/frame.py
```diff
@@ -2467,6 +2467,145 @@ def head(self, n=5):
...
+        5  bar  one  large  4  6
+        6  bar  one  small  5  8
+        7  bar  two  small  6  9
+        8  bar  two  large  7  9
+
+        This first example aggregates values by taking the sum.
+
+        >>> table = df.pivot_table(values='D', index=['A', 'B'],
+        ...                     columns='C', aggfunc='sum')
+        >>> table # doctest: +SKIP
```
HyukjinKwon: two spaces before inlined comment (per pep8)

HyukjinKwon: Actually, why do we need to skip? I think we can do `# doctest: +NORMALIZE_WHITESPACE` instead.

garawalid:
@HyukjinKwon thanks for your review. Of course, I'll open it.

databricks/koalas/frame.py
garawalid: It works like a charm, thanks! 
Btw, do you use some plug-in in PyCharm to check pep8 style inside the docstring ?

databricks/koalas/frame.py
garawalid: Still not sure if indentation is fixed. @HyukjinKwon would you mind please to check it again.

HyukjinKwon:
Thanks for working on this @garawalid. this is awesome.

Merged

PR #387, opened 2019-05-27
Add DataFrame.merge to frame.rst

floscha:
I've just seen that both #264 and #381 missed adding `DataFrame.merge` to the .rst doc files. Therefore, this PR simply adds the missing entry.


ueshin:
@floscha oh, good catch!

ueshin:
Thanks! merging.

PR #388, opened 2019-05-27
Implement append() method for Koalas DataFrames and Series

floscha:
As requested in #382, this PR implements pandas' `append()` method for Koalas DataFrames and Series.


floscha:
@HyukjinKwon could you (or anyone else) review this, please?

rxin:
ping @HyukjinKwon and @ueshin

HyukjinKwon:
Sorry for late response. I will take a look within tomorrow (KST). It's quite late here now ..

databricks/koalas/frame.py
```diff
@@ -2595,6 +2595,96 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
 
         return DataFrame(joined_table.select(*exprs))
 
+    def append(self, other: 'DataFrame', ignore_index: bool = False,
+               verify_integrity: bool = False, sort: bool = None) -> 'DataFrame':
```
garawalid: I think `sort` should be `False` instead of `None`, even though it's `None` in Pandas API.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -557,6 +557,34 @@ def test_merge_raises(self):
...
+        msg = "The 'sort' parameter is currently not supported"
+        with self.assertRaises(ValueError, msg=msg):
+            kdf.append(kdf, sort=True)
+
+        # Assert using 'verify_integrity' only raises an exception for overlapping indices
+        kdf.append(other_kdf, verify_integrity=True)
+        msg = "Indices have overlapping values"
+        with self.assertRaises(ValueError, msg=msg):
+            kdf.append(kdf, verify_integrity=True)
+
```
garawalid: I'd add these tests. I tested them and they work fine! 
```python 
        pdf = pd.DataFrame([[1, 2, 6], [3, 4, 5]], columns=list('ABC'))
        kdf = ks.from_pandas(pdf)
        other_pdf = pd.DataFrame([[3, 4], [5, 6]], columns=list('BC'), index=[2, 3])
        other_kdf = ks.from_pandas(other_pdf)

        self.assert_eq(kdf.append(other_kdf), pdf.append(other_pdf))
```

databricks/koalas/frame.py
```diff
@@ -2595,6 +2595,96 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+                    .count()) > 0:
+                raise ValueError("Indices have overlapping values")
+
+        # Keep schema before it gets modified by adding columns
+        self_schema_fields = self_sdf._schema.fields
+        for schema_field in other_sdf._schema.fields:
+            col_name, col_type = schema_field.name, schema_field.dataType
+            if col_name not in (index_columns + self._metadata.data_columns):
+                self_sdf = self_sdf.withColumn(col_name, F.lit(None).cast(col_type))
+        for schema_field in self_schema_fields:
```
garawalid: `self_sdf._schema.fields` instead of `self_schema_fields` and you can delete line 2663.

databricks/koalas/frame.py
floscha: You're right. As I understand, the reason pandas uses `None` is mainly to avoid backward compatibility issues that we don't have with Koalas. I changed it to `False` with 79119d1.

databricks/koalas/tests/test_dataframe.py
floscha: What functionality exactly does this test assert that wasn't tested before? The existing tests already cover not having some columns in common.

databricks/koalas/frame.py
floscha: Unfortunately, this does not work (as you can verify by the unit tests failing as a consequence). The reason is that during the for loop starting from line 2664, the columns of the `self_sdf` Spark DataFrame are modified. Therefore we need to store the schema of the original columns in `self_schema_fields` before the loop. I tried to explain this with the comment in 2662 but since it appears to not be clear enough, I made the explanation more verbose with bb4db59.

databricks/koalas/tests/test_dataframe.py
garawalid: You are right, you can ignore it!

floscha:
I just realized that Koalas already implements `concat()` which is the more general form of `append()`. So I'll have a look if `concat()` could possibly be reused.

databricks/koalas/frame.py
garawalid: Sounds good for me! Thanks

HyukjinKwon:
Sounds good!

floscha:
Indeed reusing `koalas.concat()` works like a charm. While the `verify_integrity` parameter could also be used by `concat()`, I'll leave it in `append()` for now to keep this PR clean. Instead, I'll add it (and some other refactorings) to `concat()` in a future PR.

databricks/koalas/frame.py
```diff
@@ -2836,6 +2836,71 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        3  3  4
+        """
+        if isinstance(other, ks.Series):
+            raise ValueError("DataFrames.append() does not support appending Series to DataFrames")
+        if sort:
+            raise ValueError("The 'sort' parameter is currently not supported")
+
+        index_columns = self._metadata.index_columns
+        if len(index_columns) != len(other._metadata.index_columns):
+            raise ValueError("Both DataFrames have to have the same number of index levels")
```
ueshin: Can we skip this check if `ignore_index is True`?

databricks/koalas/frame.py
```diff
@@ -2836,6 +2836,71 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        # TODO Maybe extend to multilevel indices in the future
+        if len(index_columns) != 1 or len(other._metadata.index_columns) != 1:
+            raise ValueError("DataFrames.append() only supports indices with level 1 right now")
```
ueshin: Is the reason why we don't support multilevel index for the following `verify_integrity` check?

ueshin: We can leave it for now anyway.

databricks/koalas/frame.py
floscha: No, because (right now) Koalas cannot concat DataFrames with different index levels like in pandas:

```python
>>> df1 = pd.DataFrame({'A':[1,2,3]},index=[1,2,3])
>>> df2 = pd.DataFrame({'A':[1,2,3]},index=[[1,2,3],[4,5,6]])
>>> df1.append(df2)
        A
1       1
2       2
3       3
(1, 4)  1
(2, 5)  2
(3, 6)  3
```

Koalas' implementation of `concat()` will throw an exception for this case. While we could fix `concat()` to correctly handle this, I believe that this is really an uncommon use case so I'd rather not waste too much energy on this.

databricks/koalas/frame.py
floscha: No, it's mainly because I wanted to start with the more simple use case of single-index DataFrames. However, since the existing implementation of `concat()` already supports multi-index DataFrames, I changed `append()` with 79d0359 to support multiple indices as well (as long as they have the same number of levels).

databricks/koalas/frame.py
ueshin: I mean `ignore_index is True` .. :

```py
>>> import pandas as pd
>>> df1 = pd.DataFrame({'A':[1,2,3]},index=[1,2,3])
>>> df2 = pd.DataFrame({'A':[1,2,3]},index=[[1,2,3],[4,5,6]])
>>> df1.append(df2, ignore_index=True)
   A
0  1
1  2
2  3
3  1
4  2
5  3
```

`ks.concat` supports such case:

```py
>>> kdf1 = ks.DataFrame({'A':[1,2,3]},index=[1,2,3])
>>> kdf2 = ks.DataFrame({'A':[1,2,3]},index=[[1,2,3],[4,5,6]])
>>> ks.concat([kdf1, kdf2], ignore_index=True)
   A
0  1
1  2
2  3
3  1
4  2
5  3
```

ueshin:
Thanks! I'd merge this and I'll submit a follow-up.

PR #389, opened 2019-05-27
Add Series.map

HyukjinKwon:
This PR adds `Series.map`.  If the function is given as an argument, seems we can just reuse `apply`.

```python
>>> import pandas as pd
>>> pdf = pd.DataFrame({'a': [1,2,3,4,5,6]})
>>> pdf.a.apply(lambda x: x+  1)
0    2
1    3
2    4
3    5
4    6
5    7
Name: a, dtype: int64
>>> pdf.a.map(lambda x: x+  1)
0    2
1    3
2    4
3    5
4    6
5    7
Name: a, dtype: int64
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -154,6 +154,91 @@ class Series(_Frame, IndexOpsMixin):
         """ Returns the data type as defined by Spark, as a Spark DataType object."""
         return self.schema.fields[-1].dataType
 
+    # TODO: arg should support Series
+    def map(self, arg):
+        """
+        Map values of Series according to input correspondence.
+
+        Used for substituting each value in a Series with another value,
+        that may be derived from a function, a ``dict``.
+
+        Parameters
+        ----------
+        arg : function or dict
+            Mapping correspondence.
+
+        Returns
+        -------
+        Series
+            Same index as caller.
+
+        See Also
+        --------
+        Series.apply : For applying more complex functions on a Series.
+        DataFrame.applymap : Apply a function elementwise on a whole DataFrame.
+
+        Notes
+        -----
+        When ``arg`` is a dictionary, values in Series that are not in the
+        dictionary (as keys) are converted to ``None``. However, if the
+        dictionary is a ``dict`` subclass that defines ``__missing__`` (i.e.
+        provides a method for default values), then this default is used
+        rather than ``None``.
+
+        Examples
+        --------
+        >>> s = ks.Series(['cat', 'dog', None, 'rabbit'])
+        >>> s
+        0       cat
+        1       dog
+        2      None
+        3    rabbit
+        Name: 0, dtype: object
+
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -154,6 +154,54 @@ def spark_type(self):
...
+
+        Returns
+        -------
+        Series
+            Same index as caller.
+
+        See Also
+        --------
+        Series.apply : For applying more complex functions on a Series.
+        DataFrame.applymap : Apply a function elementwise on a whole DataFrame.
```
HyukjinKwon: I will add this one soon.

databricks/koalas/series.py
```diff
@@ -154,6 +154,54 @@ def spark_type(self):
         """ Returns the data type as defined by Spark, as a Spark DataType object."""
         return self.schema.fields[-1].dataType
 
+    # TODO: arg should support Series and dictionary mapping too.
+    def map(self, arg):
```
thunterdb: can you add the dictionary mapping? I was just needing it yesterday and this is a big ask from @anabranch as well.

HyukjinKwon: Let me take a look

databricks/koalas/series.py
```diff
@@ -154,6 +154,91 @@ def spark_type(self):
...
+            for to_replace, value in arg.items():
+                if is_start:
+                    current = F.when(self._scol == F.lit(to_replace), value)
+                    is_start = False
+                else:
+                    current = current.when(self._scol == F.lit(to_replace), value)
+
+            if hasattr(arg, "__missing__"):
+                tmp_val = arg[np._NoValue]
+                del arg[np._NoValue]  # Remove in case it's set in defaultdict.
```
HyukjinKwon: It's a hack to mimic Pandas's behaviour with defaultdict... but I believe basically the behaviours are matched when the input is a dictionary

thunterdb: Super cool! My only concern is that you are mutating 'arg', but this strikes me as an edge case for now.

databricks/koalas/series.py
```diff
@@ -154,6 +154,92 @@ def spark_type(self):
...
+        2      I am a None
+        3    I am a rabbit
+        Name: 0, dtype: object
+        """
+        if isinstance(arg, dict):
+            is_start = True
+            # In case dictionary is empty.
+            current = F.when(F.lit(False), F.lit(None))
+
+            for to_replace, value in arg.items():
```
thunterdb: I am fine with this implementation for the time being (and it actually allows exotic case with mixed return types), but you should put a warning in the doc that the dictionary is meant to be small, not `dict(zip(range(10000), range(10000)))`, probably even hardcode a limit to 30-100 for the time being with a suggestion to use `applymap` instead.

HyukjinKwon: Yes .... absolutely agree ..

PR #390, opened 2019-05-27
Add DataFrame.applymap

HyukjinKwon:
This PR adds DataFrame.`applymap`.

We can reuse `Series.apply`:

```python
>>> import pandas as pd
>>> pd.DataFrame({'a': [1,2,3], 'b': [1.0, 2.0, 3.0]})
   a    b
0  1  1.0
1  2  2.0
2  3  3.0
>>> pdf = pd.DataFrame({'a': [1,2,3], 'b': [1.0, 2.0, 3.0]})
>>> pdf.applymap(lambda x: x + 1)
   a    b
0  2  2.0
1  3  3.0
2  4  4.0
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -210,7 +210,7 @@ class DataFrame(_Frame):
             See https://docs.python.org/3/library/typing.html. For instance, as below:
 
             >>> def function() -> int:
-            ...    1
+            ...     return 1
 
         Parameters
         ----------
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -199,6 +199,59 @@ class DataFrame(_Frame):
         row.name = None
         return row  # Return first row as a Series
 
+    def applymap(self, func):
+        """
+        Apply a function to a Dataframe elementwise.
+
+        This method applies a function that accepts and returns a scalar
+        to every element of a DataFrame.
+
+        .. note:: unlike pandas, it is required for `func` to specify return type hint.
+
+        Parameters
+        ----------
+        func : callable
+            Python function, returns a single value from a single value.
+
+        Returns
+        -------
+        DataFrame
+            Transformed DataFrame.
+
+        Examples
+        --------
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -197,6 +197,59 @@ def _reduce_for_stat_function(self, sfun):
...
+
+        This method applies a function that accepts and returns a scalar
+        to every element of a DataFrame.
+
+        .. note:: unlike pandas, it is required for `func` to specify return type hint.
+
+        Parameters
+        ----------
+        func : callable
+            Python function, returns a single value from a single value.
```
thunterdb: It is probably worth mentioning that this function must have python 3's style type hints to work, and to refer to the example for what that means (people are not necessarily aware of python 3's typing system).

thunterdb: (and that only the return type hint is necessary, not the inputs)

ueshin: There is already a note in the description above, but do we need it here too, or move it to here?
+1 for adding to refer to the example.

databricks/koalas/frame.py
```diff
@@ -199,6 +199,63 @@ def _reduce_for_stat_function(self, sfun):
...
+        Apply a function to a Dataframe elementwise.
+
+        This method applies a function that accepts and returns a scalar
+        to every element of a DataFrame.
+
+        .. note:: unlike pandas, it is required for `func` to specify return type hint.
+            See https://docs.python.org/3/library/typing.html. For instance, as below:
+
+            >>> def function() -> int:
+            ...    1
```
ueshin: nit: `return 1`?

HyukjinKwon: argh

ChethanUK:
@HyukjinKwon is there any way to increase the number of threads in each Stage?

[https://i.imgur.com/O2BMkVA.png](Number of threads is 1)
Now, Only 1 thread per each stage.

Also, Why is max rec size is 100KB? 
Stage 0 contains a task of very large size (46976 KB). The maximum recommended task size is 100 KB.

PR #392, opened 2019-05-27
Add Series.pipe

HyukjinKwon:
Looks we can just move the code to support `Series.pipe`. One doctest is added for Series.


HyukjinKwon:
Tests passed

ueshin:
Thanks! merging.

PR #397, opened 2019-05-28
DataFrame.cache()

### LINKED ISSUE #333: DataFrame.cache()
This would be a super useful function to have, even though it's Spark specific. In terms of documentation, we should create a Spark specific section in frame.rst.

shril:
This PR proposes DataFrame.cache() API.

Resolves #333


databricks/koalas/frame.py
```diff
@@ -1111,6 +1112,35 @@ def to_koalas(self):
         else:
             return DataFrame(self)
 
+    @contextmanager
+    def cache(self):
+        """
+        Yields and caches the current DataFrame as a Spark DataFrame.
```
HyukjinKwon: I think this should return a Koalas DataFrame with keeping metadata as is

thunterdb: @HyukjinKwon is this implementation good enough? I think you were involved in that comment for pyspark: 
https://github.com/apache/spark/pull/14579#discussion_r74813935

Personally, I think the situation that was discussed in the PR is an optimization, but you have more context than I do.

HyukjinKwon: ah, I think that's fine for now. Users are already able to call `cache` multiple times. It will show an error that:

```
19/05/29 13:25:43 WARN CacheManager: Asked to cache already cached data.
```

shril: @HyukjinKwon, @thunterdb, do you guys want me to implement it like this -

```
def __enter__(self):
        return self

def __exit__(self):
        self.unpersist()
```

Or the current implementation is fine?

shril: > I think this should return a Koalas DataFrame with keeping metadata as is

@HyukjinKwon, I have a follow-up doubt. In this cache function - 
- I have to convert the koalas DataFrame into Spark DataFrame.
- Then I have to cache that Spark DataFrame and converting the cached DataFrame to koalas dataframe.
- Finally, I have to uncache the Spark DataFrame.

Am I correct or there is some mistake at my end?

HyukjinKwon: If we call `cache()`, it should return Koalas dataframe because we're calling it against Koalas DataFrame.

I was roughly thinking like:

```
koalas_df = ...
koalas_df._sdf = koalas_df._sdf.cache()
return self
```

HyukjinKwon: I was thinking like:

```python
        try:
            self._sdf = self._sdf.cache()
            yield self
        finally:
            self._sdf.unpersist()
```

Does this work?

Also, can we make this `cache` API acts like for both normal API call:

```python
df = df.cache()
```

and also context manager:

```
with df.cache() as df:
    ....
```

?

shril: In my current implementation, I am only able to use it with `with` syntax.

shril: @HyukjinKwon, The `try-finally` block won't let us use this
```
df = df.cache()
```
Since it only `yields` the df but not `returns` it.
I have changed my implementation. Please have a look.

shril:
@HyukjinKwon, I have changed the implementation from using `contextmanager` to `__enter__()` and `__exit()__` functions.

The reason for the change of implementation is that `df = df.cache()` was not working in `try-finally` block. The block used to `yield` the `DataFrame` so it could only be used like this - 

```
with df.cache() as cached_df:
    # do_something
```

But in the current implementation, I am returning the `DataFrame`, so we can run further operations on it.
I hope this solves the problem. Please have a look.

Cheers to @thunterdb. The link that you shared in PR #333 helped a lot.

shril:
@HyukjinKwon, @ueshin can you please look into my PR?

databricks/koalas/frame.py
```diff
@@ -1505,6 +1512,38 @@ def to_koalas(self):
...
+
+        >>> with df.cache() as cached_df:
+        ...    print(cached_df.count())
+        ...
+        dogs    4
+        cats    4
+        dtype: int64
+
+        >>> df = df.cache()
+        >>> df.to_dict()
```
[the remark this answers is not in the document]
shril: Thanks @ueshin, I am making the changes.

shril:
Hi @ueshin,
I have made the changes you asked for.
Exposing the `DataFrame` was not a good decision on my part. Thank you for correcting it.

Please review my changes. I am happy to help regarding any further issues.

databricks/koalas/frame.py
```diff
@@ -1509,6 +1509,46 @@ def to_koalas(self):
...
+        1    0.533333
+        2    0.866667
+        3    1.100000
+        dtype: float64
+
+        To uncache the dataframe, use `unpersist` function
+
+        >>> df.unpersist()
+        """
+        return _CachedDataFrame(self._sdf)
```
ueshin: We need to add metadata to manage index?

databricks/koalas/frame.py
```diff
@@ -3514,3 +3554,21 @@ def _reduce_spark_multi(sdf, aggs):
...
+    def __init__(self, sdf):
+        self._cached = sdf.cache()
+        super(_CachedDataFrame, self).__init__(self._cached)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        if self._sdf.is_cached:
+            self._sdf.unpersist()
```
ueshin: `self.unpersist()`?

databricks/koalas/frame.py
```diff
@@ -3514,3 +3554,21 @@ def _reduce_spark_multi(sdf, aggs):
     l2 = list(row)
     assert len(l2) == len(aggs), (row, l2)
     return l2
+
+
+class _CachedDataFrame(DataFrame):
```
ueshin: Could you add a brief doc?

databricks/koalas/frame.py
```diff
@@ -3514,3 +3554,21 @@ def _reduce_spark_multi(sdf, aggs):
...
+        super(_CachedDataFrame, self).__init__(self._cached)
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        if self._sdf.is_cached:
+            self._sdf.unpersist()
+
+    def unpersist(self):
```
ueshin: Could you add a doc?

shril: Adding the doc.

databricks/koalas/frame.py
shril: Adding a doc.

databricks/koalas/frame.py
shril: Added.

databricks/koalas/frame.py
shril: Added.

databricks/koalas/frame.py
shril: Sorry, my bad.
Changed.

databricks/koalas/frame.py
shril: Added metadata in the parameters.

shril:
Hi @ueshin, 
I have made the changes you asked for.
Please have a look.

databricks/koalas/frame.py
ueshin: Oops, sorry, I meant:

```py
    def __exit__(...):
        self.unpersist()
```

shril: My bad, changed.

shril:
cc: @HyukjinKwon, can you please look into this?

HyukjinKwon:
Sorry for late response. yup, I am taking a look now :).

HyukjinKwon:
Merged. Thank you @shril and @ueshin

PR #398, opened 2019-05-29
cache implementation in databricks.koalas namespace

shril:
PR #333


shril:
@HyukjinKwon, can you please help me out here?
The build is failing in some update statement - 

```
The command "sudo apt-get update" failed and exited with 100 during .
```

HyukjinKwon:
I retriggered the job but actually I was thinking about avoiding to add it into name space. Let's get the dataframe one first.

shril:
Hi @HyukjinKwon, the dataframe one is done.
Will we proceed with this one too or will we just go with the dataframe one?

@rxin can we have your insights here?

HyukjinKwon:
Let's get https://github.com/databricks/koalas/pull/397 in first anyway.

shril:
@HyukjinKwon, I am okay with either of them.

shril:
@ueshin Can you have a look here also?

ueshin:
Actually I don't think we still need this if `df.cache()` can work with `with` as @HyukjinKwon suggested.

HyukjinKwon:
Let's close this for now. I think `df.cache()` with context manager is powerful enough.

PR #400, opened 2019-05-29
Add auto-generated ml doc dir into gitignore and lint-python

HyukjinKwon:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	docs/source/reference/ml/
```

I realised that we added `ml` directory in API doc. Let's ignore those as well like `api` directory.


HyukjinKwon:
oh haha @ueshin. I think your approach is better.

PR #403, opened 2019-05-30
Set a simple rule for writing doctests and fix them accordingly

HyukjinKwon:
This PR adds a simple rule for writing doctests.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -169,11 +169,12 @@ There are some explicit notes to keep in mind:
     >>> df.call()
     ...
 
-    >>> df.call()
+    >>> df.call(func)
     ...
 
-    >>> df = df.call().call()
-    >>> df.call()
+    >>> df = df.call(flag=True).call()
+    >>> df.call(func, flag=True)
+    ...
     ```
 
 ## Release Instructions
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -117,6 +117,45 @@ It generates HTMLs under `docs/build/html` directory. Open `docs/build/html/inde
 ## Coding Conventions
 We follow [PEP 8](https://www.python.org/dev/peps/pep-0008/) with one exception: lines can be up to 100 characters in length, not 79.
 
+## Doctest Conventions
+
+When writing doctests, usually the doctests in pandas are converted into Koalas to make sure the same codes work in Koalas, except:
+
+  - For statements for preparation used in doctests that don't fit in two lines, insert a newline to make it a separate example block in API documentation. See the examples below:
+
+    ```python
+    >>> df = ks.DataFrame([[1, 2], [4, 5], [7, 8]],
+    ...                   index=['cobra', 'viper', 'sidewinder'],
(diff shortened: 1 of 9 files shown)


databricks/koalas/frame.py
```diff
@@ -232,13 +232,15 @@ def applymap(self, func):
...
         ...     return len(str(x))
+
         >>> df.applymap(str_len)
            0  1
         0  3  4
         1  5  5
 
         >>> def power(x) -> float:
         ...     return x ** 2
+
```
HyukjinKwon: So .. I was thinking we just loose the condition - inserting a newline for two line preparation is optional ..

rxin:
@HyukjinKwon @ueshin  actually I don’t think it is a rule for separating code that’s more than one line. It’s more a rule for the docs page examples. When there is a blank line, it’s a separate code block in the examples. So I think the real rule is to define logically separate blocks. Each block should be for one thing, e.g. first block for data definition and basic example, second block for using the function with a specific argument, and third block for another argument ...

HyukjinKwon:
I agree that it's a kind of a general guide - to group them logically. The problem I found following this guide was that if we even should separate each for one liners too:

Here's an example of `Series.dtypes`

Screen Shot 2019-05-31 at 9 53 36 AM

I thought that grouping is already being done in Pandas' documentation in general. So, I was thinking about adding frequent exceptions (comparing to Pandas') in our doc only.

Let me add some more words about that grouping in our doc. Does that make sense, @rxin?

CONTRIBUTING.md
```diff
@@ -117,6 +117,65 @@ It generates HTMLs under `docs/build/html` directory. Open `docs/build/html/inde
...
+
+    >>> df.call()
+    ...
+    ```
+
+    ```python
+    >>> def func(x) -> str:
+    ...     s = str(x)
+    ...     return x
+
```
rxin: do we need a new line here? this is just part of the next logical block isn't it?

HyukjinKwon: hmmm ... I was thinking everything else before call `call` is a preparation. Let me scope this to "statements for data preparation" for now.

CONTRIBUTING.md
```diff
@@ -117,6 +117,56 @@ It generates HTMLs under `docs/build/html` directory. Open `docs/build/html/inde
...
+There are some explicit notes to keep in mind:
+
+  - For statements for data preparation used in doctests that don't fit in two lines, insert a newline to make it a separate example block in API documentation. See the examples below:
+
+    ```python
+    >>> df = ks.DataFrame([[1, 2], [4, 5], [7, 8]],
+    ...                   index=['cobra', 'viper', 'sidewinder'],
+    ...                   columns=['max_speed', 'shield'])
+
```
rxin: pandas doesn't separate this, does it? https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.rsub.html#pandas.DataFrame.rsub

HyukjinKwon: Let me get rid of this case too.

HyukjinKwon:
@rxin, how about now?

PR #404, opened 2019-05-30
Inline Datetimelike accessor documentation to Pandas'

HyukjinKwon:
This PR targets to inline Datetimelike accessor documentation to Pandas'

The problem was that it seems like we can't access property's property (`Series.dt.date` for instance).

I mimicked Pandas' approach (class attribute -> property). Plus, it will cache once it's called. 

Another problem was that the table is located in the center. I don't know why there's difference so I just edited CSS to make the tables alined left like Pandas'. I think it's hacky way to fix.

**Pandas:**

Screen Shot 2019-05-30 at 1 48 56 PM

**After this PR:**

Screen Shot 2019-05-30 at 3 44 41 PM

### DIFF
--- a/databricks/koalas/datetimes.py
+++ b/databricks/koalas/datetimes.py
@@ -17,12 +17,16 @@
 """
 Date/Time related functions on Koalas Series
 """
+from typing import TYPE_CHECKING
+
 import pandas as pd
 import pyspark.sql.functions as F
 from pyspark.sql.types import DateType, TimestampType, LongType, StringType
 
-import databricks.koalas as ks
 from databricks.koalas.base import _wrap_accessor_pandas, _wrap_accessor_spark
+
+if TYPE_CHECKING:
+    import databricks.koalas as ks
 
 
 class DatetimeMethods(object):
--- a/databricks/koalas/datetimes.py
+++ b/databricks/koalas/datetimes.py
@@ -27,7 +27,7 @@ from databricks.koalas.base import _wrap_accessor_pandas, _wrap_accessor_spark
 
 class DatetimeMethods(object):
     """Date/Time methods for Koalas Series"""
-    def __init__(self, series: ks.Series):
+    def __init__(self, series: 'ks.Series'):
         if not isinstance(series.spark_type, (DateType, TimestampType)):
             raise ValueError(
                 "Cannot call DatetimeMethods on type {}"
@@ -37,7 +37,7 @@ class DatetimeMethods(object):
 
     # Properties
     @property
-    def date(self) -> ks.Series:
+    def date(self) -> 'ks.Series':
         """
         Returns a Series of python datetime.date objects (namely, the date
         part of Timestamps without timezone information).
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
cc @icexelloss too

databricks/koalas/datetimes.py
```diff
@@ -27,7 +27,7 @@
 
 class DatetimeMethods(object):
     """Date/Time methods for Koalas Series"""
-    def __init__(self, series: ks.Series):
+    def __init__(self, series: 'ks.Series'):
```
HyukjinKwon: This is to avoid circular reference

docs/source/reference/series.rst
```diff
@@ -133,6 +133,52 @@ Reshaping, sorting, transposing
...
+   Series.dt.day
+   Series.dt.dayofweek
+   Series.dt.weekday
+   Series.dt.dayofyear
+   Series.dt.hour
+   Series.dt.minute
+   Series.dt.second
+   Series.dt.millisecond
+   Series.dt.microsecond
+
```
ueshin: I know this is from the previous one, but I'm wondering why there is a blank line here.

ueshin: oh, seems like `strftime` doesn't exist in pandas' document?

icexelloss:
LGTM as well! 

Although I am curious what's the difference between property and CachedAccessor that makes sphinx happy?

icexelloss:
Another point is that I think this is not public pandas API. Maybe we want copy the class out to prevent issues in case it is moved (although probably unlikely). pandas also ported it from xarray

HyukjinKwon:
> Although I am curious what's the difference between property and CachedAccessor that makes sphinx happy?

I am not sure .. :-). I am just guessing we can't go through something happens in runtime at property when generating docs. Property doc can be found but property.property doc can't be found statically (I guess?).

> Another point is that I think this is not public pandas API. Maybe we want copy the class out to prevent issues in case it is moved (although probably unlikely). pandas also ported it from xarray

Nice point. Let's do that when something is found actually broken. Should be fine.

databricks/koalas/series.py
```diff
@@ -24,6 +24,7 @@
 
 import numpy as np
 import pandas as pd
+from pandas.core.accessor import CachedAccessor
```
rxin: Maybe inline the CachedAccessor code? Is it similar to our lazy property? If yes, can we reconcile the two?

HyukjinKwon: Yea, the purpose is similar but one additional stuff it does is that it creates an instance of `CachedAccessor.accessor` (here `DatetimeMethods`) with the owner (`Series`).

There's another difference. When we access it via statically (class method), it will return the class itself whereas our lazy property isn't.

With out lazy property we can do it more freely but can't do above stuff. If you mean to deduplicate, I am not sure if we can. It's basically just get and set but one is class and is function.

Do you want to port this into our code base too? It will be only used for accessors.

rxin: Let's port this into our codebase then. It's ok to have some duplication.

rxin:
I merged this first. Please submit a separate PR to inline the accessor, since it's not a public API in pandas.

PR #412, opened 2019-05-31
Some nunique polish

rxin:


databricks/koalas/frame.py
```diff
@@ -1155,27 +1155,36 @@ def nunique(self, axis: int = 0, dropna: bool = True, approx: bool = False,
             Don’t include NaN in the count.
         approx: bool, default False
             If False, will use the exact algorithm and return the exact number of unique.
-            If True, it uses Spark's approximate algorithm, which is faster in most circumstances.
-            Note: this parameter is specific to Spark and is not found in pandas.
+            If True, it uses the HyperLogLog approximate algorithm, which is significantly faster
+            for large amount of data.
+            Note: this parameter is specific to Koalas and is not found in pandas.
```
floscha: Minor note, but I guess it should be capitalized like `Note: This ...` (I know, this mistake is originally from my PR 🙈)

databricks/koalas/series.py
```diff
@@ -1069,11 +1069,12 @@ def nunique(self, dropna: bool = True, approx: bool = False, rsd: float = 0.05)
             Don’t include NaN in the count.
         approx: bool, default False
             If False, will use the exact algorithm and return the exact number of unique.
-            If True, it uses Spark's approximate algorithm, which is faster in most circumstances.
-            Note: this parameter is specific to Spark and is not found in pandas.
+            If True, it uses the HyperLogLog approximate algorithm, which is significantly faster
+            for large amount of data.
+            Note: this parameter is specific to Koalas and is not found in pandas.
```
floscha: Again, should rather be `Note: This ...`

databricks/koalas/frame.py
```diff
@@ -1155,27 +1155,36 @@ def nunique(self, axis: int = 0, dropna: bool = True, approx: bool = False,
...
+            Just like ``approx`` this parameter is specific to Koalas.
 
         Returns
         -------
         The number of unique values per column as a pandas Series.
 
         Examples
         --------
-        >>> ks.DataFrame({'A': [1, 2, 3], 'B': [np.nan, 3, np.nan]}).nunique()
+        >>> df = ks.DataFrame({'A': [1, 2, 3], 'B': [np.nan, 3, np.nan]})
```
floscha: My original motivation to not reuse `df` was for the single tests to be more self-contained. But this works fine with me as well and is probably more consistent with the other doctests.

databricks/koalas/frame.py
```diff
@@ -1155,27 +1155,36 @@ def nunique(self, axis: int = 0, dropna: bool = True, approx: bool = False,
...
-        >>> ks.DataFrame({'A': [1, 2, 3], 'B': [np.nan, 3, np.nan]}).nunique(dropna=False)
+        >>> df.nunique(dropna=False)
         A    3
         B    2
         Name: 0, dtype: int64
+
+        On big data, we recommend using the approximate algorithm to speed up this function.
+        The result will be very close to the exact unique count.
+
+        >>> df.nunique(approx=True)  # doctest: +SKIP
```
floscha: Why did you skip this doctest? It runs perfectly fine for me. Regardless, you should probably add ``Name: 0, dtype: int64`` as the last line of the output.

databricks/koalas/series.py
```diff
@@ -1086,6 +1087,12 @@ def nunique(self, dropna: bool = True, approx: bool = False, rsd: float = 0.05)
 
         >>> ks.Series([1, 2, 3, np.nan]).nunique(dropna=False)
         4
+
+        On big data, we recommend using the approximate algorithm to speed up this function.
+        The result will be very close to the exact unique count.
+
+        >>> ks.Series([1, 2, 3, np.nan]).nunique(approx=True)  # doctest: +SKIP
```
floscha: Same as above, this doctest runs fine for me and might not need to be skipped. Also, to be consistent with the doctests for `DataFrame.nunique()`, `Series([1, 2, 3, np.nan])` should probably be moved to its own variable at the beginning of the doctests.

databricks/koalas/frame.py
rxin: I worry it won't be deterministic ...

rxin: I removed the skip.

rxin:
@floscha I pushed a new version.

databricks/koalas/frame.py
floscha: I initially thought the same but it appears to use some seed or something. So for the same `rsd`, it always returns the same result.

floscha:
Do you have any rationale for storing the test data for the DataFrame doctests in `df` but defining the data inline for Series? Anyway, this is more of a coding style matter I'd say and other than that the PR looks fine to me.

rxin:
The only reason I changed it in frame was because otherwise it will be more than one line in the approx example :) Agree we should probably change all of them. Going to merge this now.

PR #417, opened 2019-06-03
Fix Sphinx version to `2.0.1`.

ueshin:
Now that Sphinx on PyPI was upgraded to `2.1.0`, but it raises unknown warnings.
I'd fix the Sphinx version to `2.0.1` for now to unblock PRs.


ueshin:
merging.

PR #418, opened 2019-06-03
Keep column name from isna, isnull, notnull, notna

### LINKED ISSUE #406: isna() is not working with UDFs
I am experiencing the following issue with to_timestamp. Here is a reproduction:

```py
@ks.pandas_wraps
def f(x) -> ks.Col[int]:
  return 2 * x

df = ks.DataFrame({"x":[1, None]})
f(df["x"]).isna()
```

It works with pandas but triggers the following error in spark: `<repr(<databricks.koalas.series.Series at 0x7fd2d0777438>) failed: KeyError: '(CAST(clean_fun(x) AS INT) AS `f(x)` IS NULL)'>`

HyukjinKwon:
This PR keep original names from isna, isnull, notnull, notna. Keeping expression name as is can cause analysis exceptions due to nested `` ` `` in the name for instance.

Resolves #406

### DIFF
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -388,6 +388,9 @@ class SeriesTest(ReusedSQLTestCase, SQLTestUtils):
             repr(pser.map(d).rename(0)))
 
     def test_pandas_wraps(self):
+        # This test checks the return column name of `isna()`. Previously it returned the column
+        # name as its internal expression which contains, for instance, '`f(x)`' in the middle of
+        # column name which currently cannot be recognized in PySpark.
         @koalas.pandas_wraps
         def f(x) -> koalas.Col[int]:
             return 2 * x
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Tests passed

databricks/koalas/tests/test_series.py
```diff
@@ -386,3 +386,13 @@ def test_map(self):
         self.assertEqual(
             repr(kser.map(d)),
             repr(pser.map(d).rename(0)))
+
+    def test_pandas_wraps_and_isna(self):
```
rxin: can you add a line of comment explaining what this test is doing?

rxin:
LGTM other than that one comment.

databricks/koalas/tests/test_series.py
HyukjinKwon: yup

PR #419, opened 2019-06-03
[WIP] Backquote column names to prevent special characters in column names

### LINKED ISSUE #217: Fails to read from parquet table with flat schema and fields containing "."
I observed that when loading parquet data from a table that has a mix of normal fields and fields that have `.` in their name:

Sample data:
```python
from pyspark.sql.functions import lit
spark.range(1).withColumn("a.b", lit(1)).withColumn("a_b", lit(2)).coalesce(1).write.parquet("/mnt/ivan/sample.parquet")
```

Code works when accessing "a_b":
```python
df = ks.read_parquet("/mnt/ivan/sample.parquet")
df['a_b']

# Out[9]: 
# 0    2
# Name: a_b, dtype: int32
```

Accessing "a.b" column fails:
```
df = ks.read_parquet("/mnt/ivan/sample.parquet")
df['a.b']

# KeyError: 'a.b'
# ---------------------------------------------------------------------------
# KeyError                                  Traceback (most recent call last)
# <command-3188040896828876> in <module>()
#       1 df = ks.read_parquet("/mnt/ivan/sample.parquet")
# ----> 2 df['a.b']
# 
# /databricks/python/lib/python3.5/site-packages/databricks/koalas/frame.py in __getitem__(self, key)
#     448 
#     449     def __getitem__(self, key):
# --> 450         return self._pd_getitem(key)
#     451 
#     452     def __setitem__(self, key, value):
# 
# /databricks/python/lib/python3.5/site-packages/databricks/koalas/frame.py in _pd_getitem(self, key)
#     427                 return Series(self._sdf.__getitem__(key), self, self._metadata.index_info)
#     428             except AnalysisException:
# --> 429                 raise KeyError(key)
#     430         if np.isscalar(key) or isinstance(key, (tuple, string_types)):
#     431             raise NotImplementedError(key)
# 
# KeyError: 'a.b'
```

Escaping column name with "`" also does not work:
```python
df = ks.read_parquet("/mnt/ivan/sample.parquet")
df['`a.b`']

# Out[11]: <repr(<databricks.koalas.series.Series at 0x7fe1347c53c8>) failed: KeyError: 'a.b'>
```

HyukjinKwon:
This PR proposes to backquote column names to prevent special characters in column names when it's accessed via `__getitem__` (`df[...]`).

Resolves #217

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -286,7 +286,7 @@ class DataFrame(_Frame):
         exprs = []
         num_args = len(signature(sfun).parameters)
         for col in self.columns:
-            col_sdf = self._sdf[col]
+            col_sdf = self._sdf["`%s`" % col]
             col_type = self._sdf.schema[col].dataType
             if isinstance(col_type, BooleanType) and sfun.__name__ not in ('min', 'max'):
                 # Stat functions cannot be used with boolean values by default
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3240,7 +3240,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             raise KeyError("none key")
         if isinstance(key, str):
             try:
-                return Series(self._sdf.__getitem__(key), anchor=self,
+                # Currently, we don't support nested structure properly in Koalas
+                # as it's not supported in pandas.
+                # If Koalas should support it, we should properly resolve nested
+                # access.
+                return Series(self._sdf.__getitem__("`%s`" % key), anchor=self,
                               index=self._metadata.index_map)
             except AnalysisException:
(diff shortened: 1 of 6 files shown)


HyukjinKwon:
Currently, we don;t support nested support yet .. so I think it makes sense to backquote in those accesses.. Today (KST) is a release date. I will get this in if there aren't any objection

HyukjinKwon:
Hm, yea let me double check.

HyukjinKwon:
Hm .. I wonder if if there's only way to fix all places for `sdf[...]`.

HyukjinKwon:
Hm .. let me try another approach.

HyukjinKwon:
I am closing it for now

Issue #280, opened 2019-05-09
mean/kurt/var/std/skew should apply on numeric columns by default


rxin:
See screenshot:

![Screen Shot 2019-05-08 at 11 08 37 PM](https://user-images.githubusercontent.com/323388/57431000-3a893c80-71e6-11e9-8c9b-eec113e1c7ac.png)

nitlev:
Is someone working on this one ?

HyukjinKwon:
Please go ahead.

nitlev:
Submitted PR #422

HyukjinKwon:
Fixed in https://github.com/databricks/koalas/issues/280

PR #422, opened 2019-06-03
Stats on numeric columns

nitlev:
Solves #280 

Ensure stats functions other than min/max/count are applied on numeric/boolean columns only by default.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -180,8 +180,9 @@ class DataFrame(_Frame):
             col_type = self._sdf.schema[col].dataType
 
             numeric_only = numeric_only if numeric_only is not None else True
+            is_numeric_or_boolean = isinstance(col_type, (NumericType, BooleanType))
             min_or_max = sfun.__name__ in ('min', 'max')
-            keep_column = not numeric_only or min_or_max or isinstance(col_type, (NumericType, BooleanType))
+            keep_column = not numeric_only or is_numeric_or_boolean or min_or_max
 
             if keep_column:
                 if isinstance(col_type, BooleanType) and not min_or_max:
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -178,19 +178,20 @@ class DataFrame(_Frame):
         for col in self.columns:
             col_sdf = self._sdf[col]
             col_type = self._sdf.schema[col].dataType
-            if isinstance(col_type, BooleanType) and sfun.__name__ not in ('min', 'max'):
-                # Stat functions cannot be used with boolean values by default
-                # Thus, cast to integer (true to 1 and false to 0)
-                # Exclude the min and max methods though since those work with booleans
-                col_sdf = col_sdf.cast('integer')
-            if num_args == 1:
-                # Only pass in the column if sfun accepts only one arg
-                col_sdf = sfun(col_sdf)
-            else:  # must be 2
-                assert num_args == 2
-                # Pass in both the column and its data type if sfun accepts two args
-                col_sdf = sfun(col_sdf, col_type)
-            exprs.append(col_sdf.alias(col))
+            if isinstance(col_type, NumericType):
+                if isinstance(col_type, BooleanType) and sfun.__name__ not in ('min', 'max'):
+                    # Stat functions cannot be used with boolean values by default
+                    # Thus, cast to integer (true to 1 and false to 0)
+                    # Exclude the min and max methods though since those work with booleans
+                    col_sdf = col_sdf.cast('integer')
+                if num_args == 1:
+                    # Only pass in the column if sfun accepts only one arg
+                    col_sdf = sfun(col_sdf)
+                else:  # must be 2
+                    assert num_args == 2
+                    # Pass in both the column and its data type if sfun accepts two args
+                    col_sdf = sfun(col_sdf, col_type)
+                exprs.append(col_sdf.alias(col))
 
         sdf = self._sdf.select(*exprs)
         pdf = sdf.toPandas()
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -169,8 +169,12 @@ class DataFrame(_Frame):
         Applies sfun to each column and returns a pd.Series where the number of rows equal the
         number of columns.
 
-        :param sfun: either an 1-arg function that takes a Column and returns a Column, or
+        Parameters
+        ----------
+        sfun : either an 1-arg function that takes a Column and returns a Column, or
         a 2-arg function that takes a Column and its DataType and returns a Column.
+        numeric_only : boolean, default False
+            If True, sfun is applied on numeric columns (including booleans) only.
         """
         from inspect import signature
         exprs = []
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1899,7 +1899,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         Single    5
         dtype: int64
         """
-        return self._reduce_for_stat_function(_Frame._count_expr)
+        return self._reduce_for_stat_function(_Frame._count_expr, numeric_only=False)
 
     def drop(self, labels=None, axis=1, columns: Union[str, List[str]] = None):
         """
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -164,7 +164,7 @@ class DataFrame(_Frame):
         else:
             self._metadata = metadata
 
-    def _reduce_for_stat_function(self, sfun, numeric_only=None):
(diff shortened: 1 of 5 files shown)


databricks/koalas/frame.py
```diff
@@ -164,7 +164,7 @@ def _init_from_spark(self, sdf, metadata=None):
         else:
             self._metadata = metadata
 
-    def _reduce_for_stat_function(self, sfun):
+    def _reduce_for_stat_function(self, sfun, numeric_only=None):
```
rxin: should this just default to True rather than None? Then you can remove that line below.

nitlev: I agree that it would make more sense, but the signature for all stats function on pandas have `None` as a default value for `numeric_only`. So I could replicate the conversion to a Boolean value inside all methods that call `_reduce_for_stat_function`, or keep the conversion inside `_reduce_for_stat_function`. I'd prefer keeping it inside to remove duplicate code, but I am open to suggestions :-) wdyt ?

nitlev: @rxin Let me know if you want me to change anything :-)

floscha: I would also argue against `None` as the default, especially since pandas' way of [handling `None`](https://github.com/pandas-dev/pandas/blob/master/pandas/core/frame.py#L7497) seems rather unintuitive to me.

When for example running `pd.DataFrame({'A': [1, 2, 3], 'B': ['a', 'b', 'c']}).max()`, we get
```
A    3
B    c
dtype: object
```
as a result which shows that by default `numeric_only` should be `False` (this can also be verified by looking at the code linked to above).

However, `numeric_only=True` should probably still be used for some aggregations since for example the mean and median of strings are hard to make sense of. In fact, pandas also doesn't output any value for column *B* when calling `.mean()` on the DataFrame above.

nitlev: Seems fair. So if I recap, for some aggregations method, the default value should be `False`, for some it should be `True` (which is fine i guess). Consequently, there is no reasonable default value for `_reduce_for_stat_function`, which means the most reasonable choice here is probably to simply remove the default value, and make sure we pass down a value for `numeric_only` every time we call it.

rxin: I agree - not having a difficult value seems better.

floscha: I'd argue that the default value should be `False` since ignoring non-numerical values should only happen if explicitly stated. `numeric_only=False` however, uses all columns of the DataFrame which I'd say is what the user expects without specifying anything.

Before making any changes to your code, I would probably wait for an opinion from @rxin though.

rxin: Sorry @floscha I don't have a strong opinion here. I think we should just follow your suggestion :)

databricks/koalas/tests/test_stats.py
```diff
@@ -122,3 +122,53 @@ def test_stats_on_boolean_series(self):
 
         self.assertAlmostEqual(ds.var(), s.var())
         self.assertAlmostEqual(ds.std(), s.std())
+
+    def test_some_stats_functions_should_discard_non_numeric_columns_by_default(self):
+        df = pd.DataFrame({'i': [0, 1, 2],
```
floscha: The convention used for most tests is to use `pdf` for pandas DataFrames and `kdf` for Koalas DataFrames. It would be great if you could adapt those to make the code easier to read for other people who'll find `kdf` easier to understand than `ddf`.

The same applies to lines 147-150 and 162-165.

databricks/koalas/tests/test_stats.py
```diff
@@ -122,3 +122,53 @@ def test_stats_on_boolean_series(self):
...
+                           'b': [False, False, True],
+                           's': ['x', 'y', 'z']})
+        ddf = koalas.from_pandas(df)
+
+        # min and max do not discard non-numeric columns by default
+        self.assertEqual(len(ddf.min()), 3)
```
floscha: Regarding all `assertEqual` statements: Maybe it makes more sense to compare the results from the Koalas aggregations with pandas rather than the actual results, such that line 133 could for example be replaced with `self.assertEqual(len(kdf.min()), len(pdf.min()))`. This focuses more on the consistency between both libraries while I believe pandas is pretty well-tested itself so we don't need to write all sorts of tests twice. Maybe @HyukjinKwon also has an opinion about this?

databricks/koalas/series.py
```diff
@@ -1772,7 +1772,11 @@ def describe(self, percentiles: Optional[List[float]] = None) -> 'Series':
 
     # ----------------------------------------------------------------------
 
-    def _reduce_for_stat_function(self, sfun):
+    def _reduce_for_stat_function(self, sfun, numeric_only=None):
+        """
+        :param sfun: the stats function to be used for aggregation
```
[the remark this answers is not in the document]
nitlev: Good catch

databricks/koalas/tests/test_stats.py
nitlev: Ok, I'll change that

nitlev: Although in this particular test file the convention is quite different. Do you want me to change them all to improve consistency across all tests ? Or should I change only the tests this commit adds ?

floscha: You've already changed all variable names, right? 😉 But that's fine with me.

databricks/koalas/frame.py
```diff
@@ -169,8 +169,12 @@ def _reduce_for_stat_function(self, sfun, numeric_only=False):
         Applies sfun to each column and returns a pd.Series where the number of rows equal the
         number of columns.
 
-        :param sfun: either an 1-arg function that takes a Column and returns a Column, or
+        Parameters
+        ----------
+        sfun : either an 1-arg function that takes a Column and returns a Column, or
         a 2-arg function that takes a Column and its DataType and returns a Column.
```
floscha: Line 175 needs to be indented so that it's under `sfun`.

databricks/koalas/tests/test_stats.py
```diff
@@ -26,149 +24,150 @@
...
             self.assertPandasAlmostEqual(res, sol)
 
             # Series
-            a = df.A
-            b = df.B
-            da = ddf.A
-            db = ddf.B
+            a = pdf.A
+            b = pdf.B
+            da = kdf.A
```
floscha: You could also rename `da` and `db` to `ka` and `kb` since adding the `k` suffix for the Koalas version of pandas variables is what you commonly see in this project.

databricks/koalas/tests/test_stats.py
```diff
@@ -26,149 +24,150 @@
...
-        ddf = koalas.from_pandas(df)
+        pdf = pd.DataFrame({'i': [0, 1, 2],
+                            'b': [False, False, True],
+                            's': ['x', 'y', 'z']})
+        kdf = koalas.from_pandas(pdf)
 
         # min and max do not discard non-numeric columns by default
-        self.assertEqual(len(ddf.min()), 3)
-        self.assertEqual(len(ddf.max()), 3)
+        self.assertEqual(len(kdf.min()), len(kdf.min()))
```
floscha: Starting from here to the end of the file, you keep comparing `kdf` with `kdf`. That's probably some copy & paste error but it should, of course, compare `kdf` with `pdf`.

nitlev: Right, not very useful ;-)

databricks/koalas/tests/test_stats.py
nitlev: Yes I did, but I figured I could always go back if you didn't like it :-)

databricks/koalas/tests/test_stats.py
nitlev: Fixing this led me to realise that koalas behavior differs from pandas since in koalas we can apply stats function on non-numeric columns; the returned value is just `NaN`. In pandas, forcing to compute the `mean` on a string column raises an `TypeError`. Do we want to keep pandas behavior, and raise an error ?
Personaly, I'd favor the simplest path for now, and would create another issue for this change later on, if needed. Any strong opinion about this ?

HyukjinKwon: Yea I agree with comparing to Pandas' in general. Handling NaN is pretty annoying problem currently in Koalas.

> forcing to compute the mean on a string column raises

It would be better to match it with Pandas' for now. Creating another issue makes sense to me. We can fix it here too. I'll leave it to you @nitlev

floscha: I also prefer pandas' behavior as it keeps the user from producing unintended results by accident. Maybe it's cleaner to move this to a new PR though and merge this one if you approve @HyukjinKwon.

HyukjinKwon:
Merged. Thanks for thorough review @floscha and reflecting it nicely @nitlev.

Issue #322, opened 2019-05-14
internal refactoring: create an immutable internal dataframe structure


rxin:
Right now DataFrame wraps around two data structures: _metadata and _sdf (a Spark DataFrame). Series wraps around three data structures: _scol (a Spark Column), _kdf (the Koalas DataFrame), and _index_map.

I think it will be easier to reason about if we create a single internal data structure that's immutable (say InternalFrame). The user facing DataFrame and Series are still mutable, but they simply wrap around one immutable data structure. They can point to a new immutable data structure if any inplace change is required.

Aside from just being a cleaner design, I had one motivating use case for this: I wanted to cache the result of a to_pandas() call (so repeatedly calling a DataFrame's `__repr__` won't trigger more than one computation), but given the current mutable structure it is very difficult to reason about correctness. If we have an internal, immutable frame, we can just put the cache in that frame. The cache will always be correct (and does not need to be invalidated) for that frame since the frame is immutable.

rxin:
BTW if Metadata is already immutable, we could also just move _sdf and _scol into the Metadata class, and then rename the class to InternalFrame ...

garawalid:
I agree with you. Especially, caching the result of to_panads() will save computation because we use it in many methods like to_numpy()

thunterdb:
Having a separate immutable and internal data structure sounds good to me.
We will need an option to disable it though because these arrays can be
very large and should not always be kept in memory.

Series and DataFrame are pretty different both in behaviour and API, so
these should be kept separate in my opinion (shape, indexing, slicing, ...
all have different behaviour).

HyukjinKwon:
I am going to retarget this to 0.6.0

rxin:
@HyukjinKwon or @ueshin will one of you have time to work on this?

ueshin:
@rxin I'll take this.

HyukjinKwon:
Actually this is fixed.

rxin:
Is this completely done?

HyukjinKwon:
I think this is completely done now. We could fix followups if there are some bugs. Let me track it in separate tickets.

PR #423, opened 2019-06-04
Introduce an internal immutable frame.

ueshin:
This is part of #322.
This introduces `_InternalFrame` to store immutable structure of `sdf`, `data_column`, `index_map`, and `scol`.
We can still do further refactoring, but I'd leave it in the future PRs to keep each PR simpler.


databricks/koalas/internal.py
```diff
@@ -0,0 +1,469 @@
...
+        """ Return the managed index names. """
+        return [index_name for _, index_name in self.index_map]
+
+    @property
+    def scol(self) -> Optional[spark.Column]:
+        """ Return the managed Spark Column. """
+        return self._scol
+
+    @lazy_property
+    def spark_dataframe(self) -> spark.DataFrame:
```
rxin: spark_df?

rxin:
Would it be useful to define a copy constructor, so we can do

```
new_idf = idf.copy(index_names=...)
```

databricks/koalas/internal.py
ueshin: Renamed: `spark_dataframe` -> `spark_df`, `pandas_dataframe` -> `pandas_df`.

ueshin:
We already have the copy constructor.

ueshin:
Thanks! I will continue further refactoring.

PR #424, opened 2019-06-04
Add scikit-learn to `requirements-dev.txt`.

ueshin:
MLflow in PyPI was upgraded to `1.0.0` and `scikit-learn` is now an "extra" dependency.
This will also happen with conda soon.
This PR adds it to `requirements-dev.txt` as a test dependency.


ueshin:
Hmm, seems like mlflow 1.0.0 has API breaking change. Let me close this for now to see further.

PR #425, opened 2019-06-04
Fix MLflow version to `0.9.1`.

ueshin:
Now that MLflow on PyPI was upgraded to 1.0.0, but it has API breaking changes.
I'd fix the MLflow version to 0.9.1 for now to unblock PRs.


ueshin:
merging.

PR #428, opened 2019-06-04
Add Series.transform

HyukjinKwon:
This seems virtually an alias of `Series.apply` (if it's in Series). I revised the doc a bit so that it can be used in DataFrames too in the future.

Also, `transform` can take a list of functions that return a dataframe.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1760,6 +1760,74 @@ class Series(_Frame, IndexOpsMixin):
         wrapped = ks.pandas_wraps(return_col=return_sig)(apply_each)
         return wrapped(self, *args, **kwds).rename(self.name)
 
+    def transform(self, func, *args, **kwargs):
+        """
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -1760,6 +1760,74 @@ def apply(self, func, args=(), **kwds):
...
+               sqrt       exp
+        0  0.000000  1.000000
+        1  1.000000  2.718282
+        2  1.414214  7.389056
+
+        """
+        if isinstance(func, list):
+            applied = []
+            for f in func:
+                applied.append(self.apply(f).rename(f.__name__))
```
ueshin: Should we pass `args` and `kwargs` here as well?

HyukjinKwon: ah, it's being ignored in pandas' too ...

HyukjinKwon:
let me merge. Thanks @ueshin!

PR #430, opened 2019-06-05
Add basic read function wrappers.

ueshin:
This PR adds a couple of basic read function wrappers of pandas' corresponding to `to_xxx()` wrappers:

- `read_clipboard()`
- `read_excel()`
- `read_html()`

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -404,7 +404,7 @@ def read_excel(io, sheet_name=0, header=0, names=None, index_col=None, usecols=N
     date_parser : function, optional
         Function to use for converting a sequence of string columns to an array of
         datetime instances. The default uses ``dateutil.parser.parser`` to do the
-        conversion. Pandas will try to call `date_parser` in three different ways,
+        conversion. Koalas will try to call `date_parser` in three different ways,
         advancing to the next if an exception occurs: 1) Pass one or more arrays
         (as defined by `parse_dates`) as arguments; 2) concatenate (row-wise) the
         string values from the columns defined by `parse_dates` into a single array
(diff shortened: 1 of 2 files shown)


databricks/koalas/namespace.py
```diff
@@ -279,6 +280,373 @@ def read_parquet(path, columns=None):
...
+
+        If a column or index contains an unparseable date, the entire column or
+        index will be returned unaltered as an object data type. For non-standard
+        datetime parsing, use ``pd.to_datetime`` after ``pd.read_csv``
+
+        Note: A fast-path exists for iso8601-formatted dates.
+    date_parser : function, optional
+        Function to use for converting a sequence of string columns to an array of
+        datetime instances. The default uses ``dateutil.parser.parser`` to do the
+        conversion. Pandas will try to call `date_parser` in three different ways,
```
HyukjinKwon: nit: Pandas -> Koalas

databricks/koalas/namespace.py
```diff
@@ -279,6 +280,373 @@ def read_parquet(path, columns=None):
...
+
+    See Also
+    --------
+    read_csv
+
+    Notes
+    -----
+    Before using this function you should read the `gotchas about the
+    HTML parsing libraries
+    <https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#io-html-gotchas>`__.
```
HyukjinKwon: nit: we can just remove the whole "Notes" (because of the pandas' link) I guess.

ueshin: sounds good.

ueshin:
merging.

naveenag86:
Using Koalas, is it possible to create excel and multiple worksheets? 
and also does it allow some kind of formatting like font size, color fill etc..?

HyukjinKwon:
If it works in pandas, Koalas should. If that doesn't work, feel free to file an issue.

PR #434, opened 2019-06-06
Skip index check when `ignore_index` is True.

ueshin:
This is a follow-up of #388.
We can skip index level check and integrity verification when `ignore_index` is True.

### DIFF
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -585,6 +585,9 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
         with self.assertRaises(ValueError, msg=msg):
             kdf.append(kdf, verify_integrity=True)
 
+        self.assert_eq(kdf.append(kdf, ignore_index=True, verify_integrity=True),
+                       pdf.append(pdf, ignore_index=True, verify_integrity=True))
+
         # Assert appending multi-index DataFrames
         multi_index_pdf = pd.DataFrame([[1, 2], [3, 4]], columns=list('AB'),
                                        index=[[2, 3], [4, 5]])
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -601,6 +601,10 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
         self.assert_eq(multi_index_kdf.append(multi_index_kdf),
                        multi_index_pdf.append(multi_index_pdf))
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_dataframe.py
```diff
@@ -580,23 +580,48 @@ def test_append(self):
...
+                       pdf.append(pdf, ignore_index=True, verify_integrity=True))
+
         # Assert appending multi-index DataFrames
         multi_index_pdf = pd.DataFrame([[1, 2], [3, 4]], columns=list('AB'),
                                        index=[[2, 3], [4, 5]])
         multi_index_kdf = ks.from_pandas(multi_index_pdf)
+        other_multi_index_pdf = pd.DataFrame([[5, 6], [7, 8]], columns=list('AB'),
+                                             index=[[2, 3], [6, 7]])
+        other_multi_index_kdf = ks.from_pandas(other_multi_index_pdf)
+
```
floscha: Since you already define `other_multi_index_pdf` here, maybe you could also call `self.assert_eq(multi_index_kdf.append(other_multi_index_kdf, verify_integrity=True),
                       multi_index_pdf.append(other_multi_index_pdf, verify_integrity=True))` in the assertion below. This would however be more for the sake of consistency and shouldn't make any technical difference.

Otherwise, the PR looks fine to me.

ueshin: Do you mean lines 605-606 below?

floscha: No, I meant replacing lines 601 and 602 with what I wrote.

ueshin: What you wrote is the same as lines 605-606, right?

Lines 601-602 is testing different thing as you did at line 566, whether we can append the dataframes with duplicate indexes if verify_integrity is False (and the result is the same as pandas').
Or do you want to add the same check as line 570, like `self.assert_eq(multi_index_kdf.append(other_multi_index_kdf), multi_index_pdf.append(other_multi_index_pdf))`?

floscha: Oh sorry, my bad. I meant without `verify_integrity=True`. So instead of appending `multi_index_kdf` to itself, you might just do `multi_index_kdf.append(other_multi_index_kdf)` since you've already defined `other_multi_index_kdf` at that point anyway. Sorry again for the confusion.

ueshin: I see, thanks! I added the test (513f339), so I think this is ready to merge now.

ueshin:
Thanks for reviewing! merging.

PR #437, opened 2019-06-07
Add Series[T] and DataFrame[Types[T, ...]] type hint

HyukjinKwon:
This PR proposes to add a generic type hint support for both `Series` and `DataFrame`.

Looks we need a way to specify type hint in order for them to be set when we run Python native functions via, for instance, `apply(func)`. For `DataFrame`, `groupby(..).apply(func)` needs it if we implement it. Consider this example:

```python
def func(pdf) -> DataFrame[...]
    return pdf

df.groupby.apply(func)
```

We already have a way in case of `Series` (previously `Col`). I renamed and refactored it to `Series`.

The new type hints are used as below:

```python
def func(...) -> Series[np.float]:
    ...

def func(...) -> DataFrame[Types[np.float, int, str]]:
    ...
```

There are some backgrounds about why `DataFrame` happened to need another `Types`. Seems variadic generic type hint isn't supported yet: https://github.com/python/typing/issues/193

**Note:** there is an internal API called `_VariadicGenericAlias` in Python 3.7. Python 3.5 and 3.6 uses a different way since it's internal. If `DataFrame[Types[np.float, int, str]]` is really ugly and it should be `DataFrame[np.float, int, str]`, I can try to reuse those internal stuff that works from 3.5, 3.6 3.7 - the code amount will be considerable. UPDATE: I took a look and seems not possible.

So, I renamed `Tuple` (existing variadic generic type) to `Types` and let `DataFrame` takes this. Seems inheritance doesn't work and I couldn't find other workarounds.

Another point is that:

```python
def func(...) -> DataFrame[Types[np.float, int, str]]
    ...
```

Seems we cannot specify field names. I currently gave some default names `c0, c1, ... cn`.

Please let me know if you guys have a different or better idea than this.

### DIFF
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -39,11 +39,11 @@ from databricks.koalas.namespace import *
 from databricks.koalas.frame import DataFrame
 from databricks.koalas.indexes import Index, MultiIndex
 from databricks.koalas.series import Series
-from databricks.koalas.typedef import Col, pandas_wraps
+from databricks.koalas.typedef import pandas_wraps, Types
 from databricks.koalas.sql import sql
 
 __all__ = ['read_csv', 'read_parquet', 'to_datetime', 'from_pandas',
-           'get_dummies', 'DataFrame', 'Series', 'Index', 'MultiIndex', 'Col', 'pandas_wraps',
+           'get_dummies', 'DataFrame', 'Series', 'Index', 'MultiIndex', 'Types', 'pandas_wraps',
            'sql', 'range', 'concat']
 
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -20,7 +20,7 @@ A wrapper class for Spark DataFrame to behave similar to pandas DataFrame.
 import re
 import warnings
 from functools import partial, reduce
-from typing import Any, Optional, List, Tuple, Union
+from typing import Any, Optional, List, Tuple, Union, TypeVar, Generic
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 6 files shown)


rxin:
Seems fine. I don't understand Python that well :)

HyukjinKwon:
cc @icexelloss too

HyukjinKwon:
Tests passed

rxin:
@floscha any opinion here?

floscha:
I think it's fine for most parts.

It'd be great though if you could add your comment "Seems we cannot specify field names. I currently gave some default names `c0, c1, ... cn`." to line 56 of *typedef.py* so people can understand this behavior from reading the code only, rather than the discussion in this PR.

If we cannot get rid of `Types`, maybe it's semantically nicer to rename it to `Columns` so that we could write `DataFrame[Columns[np.float, int, str]]` which in natural languages would read as a DataFrame of columns of type float, int, and str.

HyukjinKwon:
Thanks for input @floscha. One nit about naming `Columns` are. It sounds like that includes column name too .. it's actually `DataFrame.dtypes` and `Types` took after from "data types`  in the doc of `dtypes`.

HyukjinKwon:
I'm gonna take another look and see if it's really really impossible to get rid of `Types` one more time - it will take awhile but do it within today.

I will merge this one after that.

HyukjinKwon:
Please take a look for https://github.com/databricks/koalas/pull/453 when you guys find some time.

mrahtz:
@HyukjinKwon I noticed you referenced [python/typing#193](https://github.com/python/typing/issues/193) on variadic generics in this thread. Heads up that we've been working on a draft of a PEP for this in[ PEP 646](https://www.python.org/dev/peps/pep-0646/). If this is something you still care about, take a read and let us know any feedback in [this thread](https://mail.python.org/archives/list/typing-sig@python.org/thread/WFV5K2LK3LFQSO63X2KUOCK3VVLAQ374/) in typing-sig. Thanks!

HyukjinKwon:
That's super awesome, @mrahtz!

rxin:
That's exciting @mrahtz!

PR #440, opened 2019-06-08
Make reference doc header consistent

rxin:
I changed all top level headers to use `-` rather than a mixture of `-` and `~`, which was error prone and had a few mistakes. As part of this, I also up-leveled Series.dt and Series.str, to make them show up in the table of contents on the left side so it is easier to find.

### DIFF
--- a/docs/source/reference/series.rst
+++ b/docs/source/reference/series.rst
@@ -127,7 +127,7 @@ Missing data handling
    Series.fillna
 
 Reshaping, sorting, transposing
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+-------------------------------
 .. autosummary::
    :toctree: api/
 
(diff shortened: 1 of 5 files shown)


rxin:
Going to merge this in because my next PR depends on this.

cc @HyukjinKwon @ueshin fyi

HyukjinKwon:
lgtm

docs/source/reference/series.rst
```diff
@@ -157,15 +157,15 @@ String    :ref:`str <api.series.str>`
 
 .. _api.series.dt:
 
-Datetimelike Properties
-~~~~~~~~~~~~~~~~~~~~~~~
+Date Time Handling
```
HyukjinKwon: This also was from Pandas (https://pandas.pydata.org/pandas-docs/stable/reference/series.html#datetimelike-properties)

PR #441, opened 2019-06-08
Cross link top level I/O methods

rxin:
Link the read and write methods with each other.

### DIFF
--- a/docs/source/reference/io.rst
+++ b/docs/source/reference/io.rst
@@ -13,8 +13,8 @@ Data Generator
 
    range
 
-Flat File
----------
+Flat File / CSV
+---------------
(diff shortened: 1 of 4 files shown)


docs/source/reference/io.rst
```diff
@@ -13,8 +13,8 @@ Data Generator
 
    range
 
-Flat File
----------
+Flat File / CSV
```
HyukjinKwon: @rxin, this title was from Pandas tho (https://pandas.pydata.org/pandas-docs/stable/reference/io.html#flat-file)

rxin:
Yea but if we can do a better job than pandas we should do that.

PR #448, opened 2019-06-08
ks.sql()

### LINKED ISSUE #285: Allow querying DataFrame directly in sql method

rxin:
This builds on earlier sql work by @thunterdb and @floscha. I updated the following based on Tim's PR https://github.com/databricks/koalas/pull/360:

- Increased test coverage
- Fixed a bug with string conversion
- Added license header
- Removed pymysql dependency and inlined the string escape function
- Changed the function to instead expose indexes as top level columns, rather than hiding it. I think that makes more sense in general because it's very likely I'd like to query the indexes.

Resolves #285

### DIFF
--- a/databricks/koalas/sql.py
+++ b/databricks/koalas/sql.py
@@ -144,6 +144,8 @@ def _get_local_scope():
     try:
         return inspect.stack()[2][0].f_locals
     except Exception as e:
+        # TODO (rxin, thunterdb): use a more narrow scope exception.
+        # See https://github.com/databricks/koalas/pull/448
         return {}
 
 
@@ -157,6 +159,8 @@ def _get_ipython_scope():
         shell = get_ipython()
         return shell.user_ns
     except Exception as e:
+        # TODO (rxin, thunterdb): use a more narrow scope exception.
+        # See https://github.com/databricks/koalas/pull/448
         return None
 
 
--- a/databricks/koalas/sql.py
+++ b/databricks/koalas/sql.py
@@ -1,6 +1,21 @@
+#
+# Copyright (C) 2019 Databricks, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
 import _string
-from typing import Dict, Any, Set, Optional
-import pymysql
+from typing import Dict, Any, Optional
 import inspect
 import pandas as pd
 
@@ -22,6 +37,14 @@ def sql(query: str, globals=None, locals=None, **kwargs) -> DataFrame:
     """
     Execute a SQL query and return the result as a Koalas DataFrame.
 
+    This function also supports embedding Python variables (locals, globals, and parameters)
+    in the SQL statement by wrapping them in curly braces. See examples section for details.
+
+    In addition to the locals, globals and parameters, the function will also attempt
+    to determine if the program currently runs in an IPython (or Jupyter) environment
+    and to import the variables from this environment. The variables have the same
+    precedence as globals.
+
     Parameters
     ----------
     query : str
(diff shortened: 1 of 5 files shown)


databricks/koalas/sql.py
```diff
@@ -0,0 +1,286 @@
...
+from databricks.koalas.series import Series
+
+
+__all__ = ["sql"]
+
+from builtins import globals as builtin_globals
+from builtins import locals as builtin_locals
+
+
+def sql(query: str, globals=None, locals=None, **kwargs) -> DataFrame:
```
floscha: `DataFrame` refers to pyspark.sql.DataFrame here, right? I would've expected the return type to be a Koalas DataFrame though.

databricks/koalas/sql.py
```diff
@@ -0,0 +1,286 @@
...
+    to determine if the program currently runs in an IPython (or Jupyter) environment
+    and to import the variables from this environment. The variables have the same
+    precedence as globals.
+
+    Parameters
+    ----------
+    query : str
+        the SQL query
+
+    globals : the dictionary of global variables, if explicitly set by the user
```
floscha: To follow the convention, this line should probably contain the type (in this case dict) only and the explanation indented on the next line. Same for `locals` and `kwargs`.

databricks/koalas/sql.py
```diff
@@ -0,0 +1,286 @@
...
+
+    locals : the dictionary of local variables, if explicitly set by the user
+
+    kwargs : other variables that the user may want to set manually.
+
+    Returns
+    -------
+    DataFrame
```
floscha: Maybe hint that the return type is a Koalas DataFrame(?) rather than a pandas or SparkSQL DataFrame.

databricks/koalas/sql.py
```diff
@@ -0,0 +1,286 @@
...
+    # Highest order of precedence is the locals
+    _dict.update(kwargs)
+    return SQLProcessor(_dict, query, default_session()).execute()
+
+
+def _get_local_scope():
+    # Get 2 scopes above (_get_local_scope -> sql -> ...) to capture the vars there.
+    try:
+        return inspect.stack()[2][0].f_locals
+    except Exception as e:
```
floscha: What exception(s) can occur here? And it would probably be better to catch the actual exception rather than the generic `Exception`.

databricks/koalas/sql.py
```diff
@@ -0,0 +1,286 @@
...
+def _get_ipython_scope():
+    """
+    Tries to extract the dictionary of variables if the program is running
+    in an IPython notebook environment.
+    """
+    try:
+        from IPython import get_ipython
+        shell = get_ipython()
+        return shell.user_ns
+    except Exception as e:
```
floscha: Same as above.

databricks/koalas/sql.py
rxin: This is a koalas DataFrame.

databricks/koalas/sql.py
rxin: This was from Tim's code that I didn't modify ... I'm not sure what he was expecting.

databricks/koalas/sql.py
rxin: done (will push soon with some other changes)

databricks/koalas/sql.py
rxin: done (will push soon with some other changes)

databricks/koalas/sql.py
rxin: See the import on line 26

databricks/koalas/sql.py
rxin: @thunterdb were you expecting a more specific exception?

databricks/koalas/sql.py
rxin: @thunterdb were you expecting a more specific exception?

databricks/koalas/sql.py
floscha: My bad. Didn't see that you imported `pyspark.sql.DataFrame` as `SDataFrame`

rxin:
@floscha any other feedback other than the exception handling part? If not, I will merge this and fix the exception as a follow up when Tim is back.

rxin:
Thanks. I added a TODO just in case we forget. Merging now.

databricks/koalas/sql.py
thunterdb: @rxin indeed this was a shortcut to catch all the cases: (the stack is not deep enough, the current stack contains continuations, ...) I do not think it is going to cause issues for the time being, but we should certainly refine the error if there is user confusion coming from this part.

databricks/koalas/sql.py
thunterdb: @rxin this is also a catch all for:
 - import errors
 - ipython not initialized or not running
Because it tries to catpure the variables in the scope without checking really the intent of the user (are we running in jupyter?), it is rather heavy-handed.

There is probably a better way by looking at the internals of how python implements string interpolation (and can deal with expressions), but I could not find something simple. I believe that the current implementation is good enough, but let me know if you have some thoughts on this.

HyukjinKwon:
BTW, I think we can use the same code path when we happen to implement `to_sql`  (https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.DataFrame.to_sql.html)

Issue #438, opened 2019-06-08
index changed after merge


yuwillrun:
```python
left_kdf = ks.DataFrame({'A': [1, 2]}, index = ['zero', 'one'])
left_kdf
         A
zero   1
one    2

right_kdf = ks.DataFrame({'B': ['x', 'y']}, index=['one', 'two'])
right_kdf
         B
one    x
two    y

left_kdf.merge(right_kdf, left_index=True, right_index=True)
       A     B
0     2     x
```

Index is changed. Expected answer is 
```python
left_kdf.merge(right_kdf, left_index=True, right_index=True)
       A     B
one     2     x

```

yuwillrun:
@ueshin @thunterdb. Do we need fix it?

rxin:
cc @floscha thoughts?

floscha:
Looks like something worth fixing to me. I can create a PR for this later today.

yuwillrun:
> Looks like something worth fixing to me. I can create a PR for this later today.
sorry, i see  a `# TODO: in some case, we can keep indexes.` just now. If you fix this, i can add some function like `join`,`reindex` based on `merge` easily.

yuwillrun:
> Looks like something worth fixing to me. I can create a PR for this later today.

Hello floscha. How long will you take it to finish ?

floscha:
I'm almost done with fixing it. Hopefully I can find some time in the next couple of hours to finish it up.

yuwillrun:
> I'm almost done with fixing it. Hopefully I can find some time in the next couple of hours to finish it up.

Hello, floscha. When you are done, please tell me. Thanks~

floscha:
I've now created a PR with the fix, @yuwillrun. Unfortunately, it took me a bit longer to cover some special corner cases.

yuwillrun:
It seems there is still some problem with your PR.

floscha:
There was another corner case I was missing but it should be working now.

PR #450, opened 2019-06-09
DataFrame.join

yuwillrun:
First try to implement DataFrame.join function

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2915,21 +2915,14 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             joins index-on-index. If multiple values given, the other DataFrame must have a
             MultiIndex. Can pass an array as the join key if it is not already contained in the
             calling DataFrame. Like an Excel VLOOKUP operation.
-        how: {‘left’, ‘right’, ‘outer’, ‘inner’}, default ‘inner’
-            How to handle the operation of the two objects.
-            left: left: use calling frame’s index (or column if on is specified)
-            right: right: use other’s index.
-            outer: outer: form union of calling frame’s index (or column if on is specified) with
-                other’s index, and sort it. lexicographically.
-            inner: inner: form intersection of calling frame’s index (or column if on is specified)
-                with other’s index, preserving the order of the calling’s one.
-        lsuffix: str, default ‘‘
-            Suffix to use from left frame’s overlapping columns
-        rsuffix: str, default ‘‘
-            Suffix to use from right frame’s overlapping columns.
-        sort : bool, default False
-            Order result DataFrame lexicographically by the join key. If False, the order of the
-            join key depends on the join type (how keyword).
+        how: How to handle the operation of the two objects.
+            {‘left’, ‘right’, ‘outer’, ‘inner’}, default‘inner’
+            left: use calling frame’s index (or column if on is specified).
+            right: use other’s index.
+            outer: form union of calling frame’s index (or column if on is specified) with
+            other’s index, and sort it. lexicographically.
+            inner: form intersection of calling frame’s index (or column if on is specified)
+            with other’s index, preserving the order of the calling’s one.
 
         Returns
         -------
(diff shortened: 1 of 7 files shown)


databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
 
         return DataFrame(joined_table.select(*exprs))
 
+    # TODO: implenment `on` parameter. See Pandas's DataFrame.join.
+    def join(left, other: 'DataFrame',
```
floscha: I'd suggest going with either `self` and `other` or `left` and `right` here.

databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        Parameters
+        ----------
+        other: DataFrame, Series, or list of DataFrame.
+        on: str, list of str, or array-like, optional
+            Column or index level name(s) in the caller to join on the index in other, otherwise
+            joins index-on-index. If multiple values given, the other DataFrame must have a
+            MultiIndex. Can pass an array as the join key if it is not already contained in the
+            calling DataFrame. Like an Excel VLOOKUP operation.
+        how: How to handle the operation of the two objects.
+            {‘left’, ‘right’, ‘outer’, ‘inner’}, default‘inner’
```
floscha: `{‘left’, ‘right’, ‘outer’, ‘inner’}, default ‘inner’` should be on the same line as `how:` and `How to handle the operation of the two objects.` on the line below.

databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        DataFrame.merge: For column(s)-on-columns(s) operations.
+
+        Notes
+        -----
+        Parameters on, lsuffix, and rsuffix are not supported when passing a list of DataFrame
+        objects.
+
+        Examples
+        --------
+        >>> kdf1 = ks.DataFrame({'key': ['K0', 'K1', 'K2', 'K3', 'K4', 'K5'],
```
floscha: Couldn't the example DataFrames be a bit shorter, so that they're easier to read while showcasing the same functionality?

databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+            Column or index level name(s) in the caller to join on the index in other, otherwise
+            joins index-on-index. If multiple values given, the other DataFrame must have a
+            MultiIndex. Can pass an array as the join key if it is not already contained in the
+            calling DataFrame. Like an Excel VLOOKUP operation.
+        how: How to handle the operation of the two objects.
+            {‘left’, ‘right’, ‘outer’, ‘inner’}, default‘inner’
+            left: use calling frame’s index (or column if on is specified).
+            right: use other’s index.
+            outer: form union of calling frame’s index (or column if on is specified) with
+            other’s index, and sort it. lexicographically.
```
floscha: This should be indented one more level for better readability. Same for line 2925.

databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        5  K5  A5
+        >>> kdf2
+          key   B
+        0  K0  B0
+        1  K1  B1
+        2  K2  B2
+
+        Join DataFrames using their indexes.
+
+        >>> join_kdf = kdf1.join(kdf2, lsuffix='_caller', rsuffix='_other')
```
floscha: Why not simply use `lsuffix='_left', rsuffix='_right'`?

databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        right_keys = other._metadata.index_columns
+
+        if left_keys != right_keys:
+            raise ValueError(
+                'Index type and names should be same in the objects to join. '
+                'You passed different indices '
+                '{left_keys} and {right_keys}'.format(
+                    left_keys=left_keys, right_keys=right_keys))
+
+        left_table = left._sdf.alias('left_table')
```
floscha: The following code looks a lot like the `merge()` method. Don't you think we could reuse that one?

databricks/koalas/frame.py
```diff
@@ -2897,6 +2897,157 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        >>> join_kdf.sort_values(by=join_kdf.columns)
+          key   A     B
+        0  K0  A0    B0
+        1  K1  A1    B1
+        2  K2  A2    B2
+        3  K3  A3  None
+        4  K4  A4  None
+        5  K5  A5  None
+        """
+        if on:
```
floscha: The TODO in line 2900 says that `on` is not implemented. Why are you using it here though? I think you should instead raise a `ValueError` unless you plan to implement `on` in this PR. Then you should, of course, add the respective unit tests as well.

databricks/koalas/frame.py
yuwillrun: Yes,you remind me. It seems like `merg`e.Maybe i can use `merge` instead when https://github.com/databricks/koalas/issues/438 `merge` index issue fix done.Thanks for your kind guidance.

PR #453, opened 2019-06-10
Add Series[T] and DataFrame[T, ...] type hint

HyukjinKwon:
This PR proposes an alternative take for https://github.com/databricks/koalas/pull/437. The main diff is that:

- In python 3.7, hacks `__class_getitem__` from `Generic` at `DataFrame`. Here, it just wraps the given types into a tuple type, which is existing variadic generic.

- In python 3.6 and python 3.5, similarly it wraps by a tuple type too; however, the logic is defined in metaclass. So, it should be wrapped in a different way from python 3.7's

<br/>
<br/>
<br/>

Looks we need a way to specify type hint in order for them to be set when we run Python native functions via, for instance, `apply(func)`. For `DataFrame`, `groupby(..).apply(func)` needs it if we implement it. Consider this example:

```python
def func(pdf) -> DataFrame[...]
    return pdf

df.groupby.apply(func)
```

We already have a way in case of `Series` (previously `Col`). I renamed and refactored it to `Series`.

The new type hints are used as below:

```python
def func(...) -> Series[np.float]:
    ...

def func(...) -> DataFrame[np.float, int, str]:
    ...
```

Note that:

```python
def func(...) -> DataFrame[np.float, int, str]
    ...
```

Seems we cannot specify field names. I currently gave some default names `c0, c1, ... cn`.

### DIFF
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -39,7 +39,6 @@ from databricks.koalas.frame import DataFrame
 from databricks.koalas.indexes import Index, MultiIndex
 from databricks.koalas.series import Series
 from databricks.koalas.typedef import pandas_wraps
-from databricks.koalas.sql import sql
 
 __all__ = ['read_csv', 'read_parquet', 'to_datetime', 'from_pandas',
            'get_dummies', 'DataFrame', 'Series', 'Index', 'MultiIndex', 'pandas_wraps',
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -38,10 +38,11 @@ assert_pyspark_version()
 from databricks.koalas.frame import DataFrame
 from databricks.koalas.indexes import Index, MultiIndex
 from databricks.koalas.series import Series
-from databricks.koalas.typedef import Col, pandas_wraps
+from databricks.koalas.typedef import pandas_wraps
+from databricks.koalas.sql import sql
 
 __all__ = ['read_csv', 'read_parquet', 'to_datetime', 'from_pandas',
-           'get_dummies', 'DataFrame', 'Series', 'Index', 'MultiIndex', 'Col', 'pandas_wraps',
+           'get_dummies', 'DataFrame', 'Series', 'Index', 'MultiIndex', 'pandas_wraps',
            'sql', 'range', 'concat', 'melt']
 
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -20,7 +20,8 @@ A wrapper class for Spark DataFrame to behave similar to pandas DataFrame.
 import re
 import warnings
 from functools import partial, reduce
-from typing import Any, Optional, List, Tuple, Union
+import sys
+from typing import Any, Optional, List, Tuple, Union, Generic, TypeVar
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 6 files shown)


HyukjinKwon:
Okay, @rxin, @floscha, @icexelloss, @tahasyeddb, @ueshin, I managed to do it with a bunch of hacks.

The downside of the current PR is:

  - We should add a hack for each version of Python until variadic generic is officially supported. Note that at some points it might be impossible for us to hack anymore
  - This kind of hack might tie up with possible corner holes (like namedtuple hack in pyspark) forever.

Nevertheless, Koalas can support neat return hint like:

```
def test(...) -> ks.DataFrame[int, str]:
    pass
```

Which way do you guys prefer? I prefer this way experimentally considering Koalas is still premature.

rxin:
The thing is for vast majority of functions this won't apply right? Because the output type depends on the input type, which is not known statically.

HyukjinKwon:
Yup, in case of DataFrame this only applies to APIs that should be implemented via Grouped Map Pandas UDF for now where the types should be specified.

In case of Series, it applies to other APIs that should udf Pandas UDFs, for instance, `[DataFrame|Series].[transform|apply]`, and `pandas_wraps`.

Additionally people might use this to note the return types of a function, for instance,

```python
def read_csv() -> DataFrame[str, int]:
    ks.DataFrame(spark.schema("a strint, b int").csv())
```

For this case, it's just something optional and pretty.

HyukjinKwon:
BTW, we're encouraging to use type hints. Providing a proper way to specify types for our DataFrame and Series might be a good idea.

thunterdb:
@HyukjinKwon thank you for fixing this, I wanted to unify `Col` and `Series` but could not find a way to do that properly. A couple of thoughts:

Do not attempt to recreate a full annotation system like in scala, because in practice it is not that useful and there are workarounds:

workarounds: python is flexible enough that I think you can do something like:

```py
MyReturnType = StructType([StructField(...), ...])

def f() -> DataFrame[MyReturnType]: raise
```
(not tested)

not that useful: in practice, people want to be able to type functions and Series, not DataFrames. DataFrames in ETL keep having columns being added and deleted. All the business logic is done by taking a Series and transforming it into another Series, or a structure.

Regarding variadic types. I think that we can resolve them, based on the inputs. If you write something like this:

```py
T = type()
def f(x: Series[T]) -> Series[T]: pass
```
You should be able to recover the annotations of `x` and then infer the final type. This is main case I can think of. This improvement is good for another PR.

HyukjinKwon:
```python
MyReturnType = StructType([StructField(...), ...])

def f() -> DataFrame[MyReturnType]: raise
```

I wonder if this is possible ..I quickly tested:

```
      File "/Applications/PyCharm.app/Contents/helpers/pycharm/docrunner.py", line 140, in __run
        compileflags, 1), test.globs)
      File "<doctest _infer_return_type[2]>", line 1, in <module>
        def func() -> ks.Series[1]:
      File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 251, in inner
        return func(*args, **kwds)
      File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 824, in __class_getitem__
        params = tuple(_type_check(p, msg) for p in params)
      File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 824, in <genexpr>
        params = tuple(_type_check(p, msg) for p in params)
      File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/typing.py", line 139, in _type_check
        raise TypeError(f"{msg} Got {arg!r:.100}.")
    TypeError: Parameters to generic types must be types. Got StructType(List(StructField(a,StringType,true))).
```

Let me take another look and see if I can do similar stuff for that but from a cursory look, I am pretty sure that we will need another hack like this one ...

```python
DataFrame[np.float, int, str]
```

This annotation isn't completely new because `Tuple[str, int, ...]` is already supported in this way which is pretty similar instance with DataFrame. Considering we tried to represent Row as a tuple before in Spark side (https://github.com/apache/spark/blob/eec1a3c2862955bf2620d4cc5116fbd86e29952e/python/pyspark/sql/types.py#L1405), this wouldn't be that crazy.

FWIW, I found similar attempt here `https://github.com/python/typing/issues/193#issuecomment-479484916`

HyukjinKwon:
ping, what do you guys think about this?

databricks/koalas/__init__.py
```diff
@@ -38,10 +38,11 @@ def assert_pyspark_version():
 from databricks.koalas.frame import DataFrame
 from databricks.koalas.indexes import Index, MultiIndex
 from databricks.koalas.series import Series
-from databricks.koalas.typedef import Col, pandas_wraps
+from databricks.koalas.typedef import pandas_wraps
+from databricks.koalas.sql import sql
```
ueshin: `sql` module is imported later.

PR #454, opened 2019-06-10
Clean up after test_table.

ueshin:
This is a follow-up of #449.
The test `DataFrameSparkIOTest.test_table()` needs a clean-up.


ueshin:
merging.

PR #455, opened 2019-06-10
Set numpydoc to 0.8

### LINKED ISSUE #439: Remove "self" from API docs
Take a look at this page: https://koalas.readthedocs.io/en/latest/reference/series.html

Almost all methods have "self" as the first parameter. We should remove them. Perhaps take a look at how pandas does it.

HyukjinKwon:
This PR sets the numpydoc version to 0.8. Seems like NumPy 0.9.x has a bug about documentation generation.

It took me a while .. so I would like to leave three notes I learnt while debugging this:

- readthedoc caches installed package and we don't specify the version. So, once 0.8.0 is installed, it can be unable to reproduce in the same code base.

- readthedoc uses different command to build sphinx (not like ours). We use `make clean` but sphinx uses `python sphinx-build`:
    ```
    python /home/docs/checkouts/readthedocs.org/user_builds/koalas/envs/stable/bin/sphinx-build -T -E -b readthedocs -d _build/doctrees-readthedocs -D language=en . _build/html
    ```

    this ignores our `-W` option and `PYTHONPATH` in Makefile. There might be some more diff.

- v0.2.0 doc is fine (https://koalas.readthedocs.io/en/v0.2.0/reference/frame.html). Seems like numpy doc fails on a certain condition.

**Before:**

```
Requirement already satisfied: numpydoc in /home/docs/checkouts/readthedocs.org/user_builds/koalas/envs/latest/lib/python3.6/site-packages (from -r requirements-dev.txt (line 12)) (0.9.1)
```

https://koalas.readthedocs.io/en/latest/reference/frame.html

**After:**

```
Requirement already satisfied: numpydoc in /home/docs/checkouts/readthedocs.org/user_builds/koalas/envs/stable/lib/python3.6/site-packages (from -r requirements-dev.txt (line 12)) (0.8.0)
```

https://koalas.readthedocs.io/en/stable/reference/frame.html

Resolves #439

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,7 +9,7 @@ mlflow==0.9.1
 # Documentation build
 sphinx==2.0.1
 nbsphinx
-numpydoc<=0.8
+numpydoc==0.8
 
 # Linter
 mypy
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,7 +9,7 @@ mlflow==0.9.1
 # Documentation build
 sphinx==2.0.1
 nbsphinx
-numpydoc
+numpydoc<=0.8
 
 # Linter
 mypy


HyukjinKwon:
I am 99% sure this is the cause. Let me just merge once the build passes. cc @rxin

requirements-dev.txt
```diff
@@ -9,7 +9,7 @@ mlflow==0.9.1
 # Documentation build
 sphinx==2.0.1
 nbsphinx
-numpydoc
+numpydoc<=0.8
```
HyukjinKwon: BTW, this version was set before but unset due to Python 3.5 support .. I should check Travis CI picks Python 3.5 correctly..

rxin: what if you just pin it to 0.8?

HyukjinKwon: it upgrades Python version from 3.5 to 3.6. So it setting to 0.8.0 was removed.. which was .. here https://github.com/databricks/koalas/commit/47a36605904771f545ef95091c5483743761294b

HyukjinKwon: Seems like some numpydoc version higher than 0.8.0 on conda doesn't support Python 3.5.

If we pin both, it fails. I tested this condition before when we do mlflow.

HyukjinKwon: ah but we now use pip for python 3.5 installation. seems we can set the version

PR #452, opened 2019-06-09
ks.read_delta / DataFrame.to_delta

rxin:
Still need to figure out how to do testing (primarily how to load the dependency in).

Resolves #444

PR #456, opened 2019-06-11
ks.read_delta / DataFrame.to_delta

### LINKED ISSUE #444: Implement read_delta and DataFrame.to_delta
Would be cool to have Delta support built-in so it's easy for users to read/write Delta Lake tables.

Once we implement this, we should update read_csv / write_csv docs to suggest users to use delta or parquet instead of csv.

HyukjinKwon:
This PR takes over #452 so that they can be tested in doctests and unittests.

Closes #452
Resolves #444

### DIFF
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
 import pytest
 import numpy
 import tempfile
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -35,6 +35,7 @@ if LooseVersion(__version__) >= LooseVersion("2.4.2"):
 else:
     session = utils.default_session()
 
+
 @pytest.fixture(autouse=True)
 def add_ks(doctest_namespace):
     doctest_namespace['ks'] = koalas
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -28,11 +28,12 @@ from databricks.koalas import utils
 
 
 # Initialize Spark session that should be used in doctests or unittests.
-# Detla requires Spark 2.4.2+. See
+# Delta requires Spark 2.4.2+. See
 # https://github.com/delta-io/delta#compatibility-with-apache-spark-versions.
 if LooseVersion(__version__) >= LooseVersion("2.4.2"):
     session = utils.default_session({"spark.jars.packages": "io.delta:delta-core_2.11:0.1.0"})
-
+else:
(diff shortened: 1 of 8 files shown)


HyukjinKwon:
cc @rxin

databricks/conftest.py
```diff
@@ -15,8 +15,17 @@
...
+import atexit
+import shutil
+import uuid
 
 from databricks import koalas
+from databricks.koalas import utils
+
+
+# Initialize Spark session that should be used in doctests or unittests.
+session = utils.default_session({"spark.jars.packages": "io.delta:delta-core_2.12:0.1.0"})
```
ueshin: `_2.11` instead of `_2.12` for Spark 2.4.3 since Spark 2.4.3 is compiled with Scala 2.11 by default?

ueshin: As for Spark < 2.4.2, it might not work properly according to the doc https://github.com/delta-io/delta#compatibility-with-apache-spark-versions.

HyukjinKwon: ohh let me fix it soon.

HyukjinKwon:
K .. passed.

rxin:
Looks great.

@ueshin should take a look at this.

databricks/conftest.py
```diff
@@ -27,3 +42,18 @@ def add_ks(doctest_namespace):
...
+    path = tempfile.mkdtemp()
+    atexit.register(lambda: shutil.rmtree(path, ignore_errors=True))
+    doctest_namespace['path'] = path
+
+
+@pytest.fixture(autouse=True)
+def add_db(doctest_namespace):
+    db_name = str(uuid.uuid4()).replace("-", "")
+    utils.default_session().sql("CREATE DATABASE %s" % db_name)
+    atexit.register(lambda: session.sql("DROP DATABASE IF EXISTS %s CASCADE" % db_name))
```
ueshin: `utils.default_session()` instead of `session`?
Seems like we got errors in Python 3.5 Job:

```
Error in atexit._run_exitfuncs:
  File "/home/travis/build/databricks/koalas/databricks/conftest.py", line 58, in <lambda>
    atexit.register(lambda: session.sql("DROP DATABASE IF EXISTS %s CASCADE" % db_name))
NameError: name 'session' is not defined
```

HyukjinKwon: oh nice catch!

HyukjinKwon:
argh

HyukjinKwon:
Thanks, @rxin and @ueshin

PR #462, opened 2019-06-12
Make tests runnable with Spark 3.0 (changing Detla Scala version for Spark 3.0.0+)

HyukjinKwon:
I use Spark upstream (3.0.0) in dev to pre-test and find any issue. This PR makes Koalas testing working with Spark upstream (3.0.0).

Currently, it fails even when I test other APIs like:

```
 ./dev/pytest -k read_parquet --doctest-modules databricks
```


HyukjinKwon:
I manually tested

PR #469, opened 2019-06-13
Add DataFrame.T and DataFrame.transpose

### LINKED ISSUE #467: Missing T property
As mentioned in #382, Koalas is currently missing the `DataFrame.T` property. Since I personally use that property a lot when working with pandas, I'd be happy if it was available in Koalas as well. However, because of its distributed nature, simply turning rows into columns does not easily work with Spark. After thinking about how to implement `T` with Spark for a while now, I couldn't find any solution that is feasible for large datasets. Does anyone have an idea how to actually implement this or should we just raise a `NotImplementedError` that explains why `T` isn't supported by Koalas?

HyukjinKwon:
This PR adds DataFrame.T and DataFrame.transpose.

This will be an expensive operation ... but still work.

Resolves #467

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1270,6 +1270,185 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         return validate_arguments_and_invoke_function(
             kdf.to_pandas(), self.to_latex, pd.DataFrame.to_latex, args)
 
+    # TODO: enable doctests once we drop Spark 2.3.x (due to type coercion logic
+    #  when creating arrays)
+    def transpose(self, limit=1000):
+        """
+        Transpose index and columns.
+
+        Reflect the DataFrame over its main diagonal by writing rows as columns
+        and vice-versa. The property :attr:`.T` is an accessor to the method
+        :meth:`transpose`.
+
+        .. note:: This method is based on an expensive operation due to the nature
+            of big data. Internally it needs to generate each row for each value, and
+            then group twice - it is a huge operation. To prevent misusage, this method
+            has the default limit of input length, 1000 and raises a ValueError.
+
+                >>> ks.DataFrame({'a': range(1001)}).transpose()  # doctest: +NORMALIZE_WHITESPACE
+                Traceback (most recent call last):
+                  ...
+                ValueError: Current DataFrame has more then the given limit 1000 rows. Note that,
+                before changing the given 'limit', this operation is considerably expensive.
+
+        Parameters
+        ----------
+        limit : int
+            This parameter sets the limit of the current DataFrame. Set `None` to unlimit
+            the input length. When the limit is set, it is executed by the shortcut by collecting
+            the data into driver side, and then using pandas API. If the limit is unset,
+            the operation is executed by PySpark.
+
+        Returns
+        -------
+        DataFrame
+            The transposed DataFrame.
+
+        Notes
+        -----
+        Transposing a DataFrame with mixed dtypes will result in a homogeneous
+        DataFrame with the coerced dtype. For instance, if int and float have
+        to be placed in same column, it becomes float. If type coercion is not
+        possible, it fails.
+
+        Also, note that the values in index should be unique because they become
+        unique column names.
+
+        In addition, if Spark 2.3 is used, the types should always be exactly same.
+
+        Examples
+        --------
+        **Square DataFrame with homogeneous dtype**
+
+        >>> d1 = {'col1': [1, 2], 'col2': [3, 4]}
+        >>> df1 = ks.DataFrame(data=d1, columns=['col1', 'col2'])
+        >>> df1
+           col1  col2
+        0     1     3
+        1     2     4
+
+        >>> df1_transposed = df1.T.sort_index()  # doctest: +SKIP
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
cc @rxin, @thunterdb,  @ueshin, @floscha, @garawalid

HyukjinKwon:
What do you guys think about this?

ueshin:
The implementation itself LGTM, but I'd say this is too dangerous for big data.
At least, how about adding the limitation to support number of rows, or adding an argument or a config to limit the rows for users to configure? Maybe we need a note in documents as well.

HyukjinKwon:
Let me at least leave a note like some other APIs we have (e.g., to_pandas)

HyukjinKwon:
Tests passed

HyukjinKwon:
@rxin, @shril, @floscha, @ueshin, if you guys don't mind, I will add the limitation logic later.
To check the length, the only idea I have it is to check the length of each partition.

To do this, we should call `mapPartitions` which is expensive. I think it's a bit odd that we add an expensive operation to check the length.

I am checking the possibility to have an API like `mapPartitionsInArrow`. Maybe, I will integrate it later if you guys don't mind.

HyukjinKwon:
Can I merge this one first, @rxin if you dont mind? or you want to add the length checking together?

rxin:
I'd add a simple length check, e.g. we should not have more than 1000 rows.

HyukjinKwon:
@rxin, sorry but could I ask kind of a simple length check do you mean?

- `mapPartitions` : it needs to serialize and deserialize whole data row by row.
- `take(1000)`: it could lead execute twice the input dataframe.
- Other ways I missed ... ?

Let me use `take(1000)` way first then. Let me know if you meant another way.

rxin:
take 1001, and fail if there are 1001 records returned

HyukjinKwon:
done!

rxin:
ca we just take the first 1001 rows and call pandas' pivot and turn it back into a koalas dataframe? Only going through data once.

HyukjinKwon:
yup.

floscha:
So when wanting to process more than 1000 rows, the user would simply run `ks.DataFrame(kdf.to_pandas().T)` as a workaround, I guess? I can imagine cases where one would, for example, transpose a 2000x5 DataFrame (i.e. 10000 values) which could still be processed with pandas on a single machine. Should we maybe mention this somewhere in the docs/comments?

rxin:
Yea - alternatively allow it to take a flag on max rows. I just worry about users waiting for a while and then everything crashes.

floscha:
I think a flag similar to [options](https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html) in pandas sounds like a good idea, especially since pandas users should already be familiar with this. Wouldn't it then make sense to consequentially have the same for other methods like `to_pandas()` or `to_numpy()` as well, to guard users from shooting themselves in the foot?

rxin:
Great idea. We might want different thresholds though because I can see
getting a million rows into a pandas DataFrame, but creating a million
columns will kill Sparks driver.

HyukjinKwon:
> Should we maybe mention this somewhere in the docs/comments?

I think I mentioned this about how it works..

floscha:
I think the comments are good for most parts. The only issue I see right now is that the `T` property cannot be called with the limit argument. So if a user gets the current `ValueError`, it might not be obvious to them that they must switch from `.T` to `.transpose(limit=n)`. Maybe you could thus change to error message to something like `... rows. Please use df.transpose(limit=<maximum number of rows>) to retrieve more than %s rows. Note that, ...` to give the user a better hint.

Also, if we wanted to add some pandas-like options, it would probably be best to move this (including further discussions) to another PR to keep this one clean, right?

databricks/koalas/frame.py
```diff
@@ -1270,6 +1270,185 @@ def to_latex(self, buf=None, columns=None, col_space=None, header=True, index=Tr
         return validate_arguments_and_invoke_function(
             kdf.to_pandas(), self.to_latex, pd.DataFrame.to_latex, args)
 
+    # TODO: enable doctests once we drop Spark 2.3.x (due to type coercion logic
+    #  when creating arrays)
+    def transpose(self, limit=1000):
```
floscha: Maybe you could add an `Optional[int]` type hint here.

databricks/koalas/frame.py
```diff
@@ -1270,6 +1270,185 @@ def to_latex(self, buf=None, columns=None, col_space=None, header=True, index=Tr
...
+
+                >>> ks.DataFrame({'a': range(1001)}).transpose()  # doctest: +NORMALIZE_WHITESPACE
+                Traceback (most recent call last):
+                  ...
+                ValueError: Current DataFrame has more then the given limit 1000 rows. Note that,
+                before changing the given 'limit', this operation is considerably expensive.
+
+        Parameters
+        ----------
+        limit : int
```
floscha: I guess you could also mention the default of 1000 here.

HyukjinKwon:
Makes sense. Thanks for review @floscha

HyukjinKwon:
I think I addressed all comments. Let me merge this one. let me know if you guys have more comments.

PR #473, opened 2019-06-14
DataFrame.join()

yuwillrun:


yuwillrun:
@ueshin need to code review.Thanks~~

databricks/koalas/frame.py
```diff
@@ -3222,6 +3222,113 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
         else:
             return DataFrame(selected_columns)
 
+    def join(left, right: 'DataFrame', on: Optional[Union[str, List[str]]] = None,
```
HyukjinKwon: Seems like a mistake. left -> self. I think it's a Python convention.

databricks/koalas/frame.py
```diff
@@ -3222,6 +3222,113 @@ def merge(left, right: 'DataFrame', how: str = 'inner',
...
+        DataFrame.merge: For column(s)-on-columns(s) operations.
+
+        Notes
+        -----
+        Parameters on, lsuffix, and rsuffix are not supported when passing a list of DataFrame
+        objects.
+
+        Examples
+        --------
+        >>> kdf1 = ks.DataFrame({'key': ['K0', 'K1', 'K2', 'K3'],
```
HyukjinKwon: Out of curiosity, why did we use a dataset different from Pandas? (https://github.com/pandas-dev/pandas/blob/v0.24.2/pandas/core/frame.py#L6694-L6815)

HyukjinKwon:
Looks good to me otherwise

databricks/koalas/frame.py
ueshin: oh, I used `left` and `right` when I updated `merge` function before. I'll submit the PR to fix `merge` function soon.

databricks/koalas/frame.py
yuwillrun: In another PR, i use pandas example, but be told the example is too complex during code review.So in this PR, i simplify the redundant rows.Which way do you think is better，same pandas or simplify examlple ?

databricks/koalas/frame.py
ueshin: FYI: the PR #479 was merged.

yuwillrun:
> Looks good to me otherwise

Fix already.

yuwillrun:
@ueshin Need to merge.~~

ueshin:
Thanks! merging.

PR #474, opened 2019-06-17
Add DataFrame.melt and koalas.melt

HyukjinKwon:
This PR proposes to add:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.melt.html?highlight=melt#pandas.DataFrame.melt
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.melt.html?highlight=melt#pandas.melt

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3958,6 +3958,118 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         else:
             return DataFrame(internal)
 
+    def melt(self, id_vars=None, value_vars=None, var_name='variable',
+             value_name='value'):
+        """
+        Unpivot a DataFrame from wide format to long format, optionally
+        leaving identifier variables set.
+
+        This function is useful to massage a DataFrame into a format where one
+        or more columns are identifier variables (`id_vars`), while all other
+        columns, considered measured variables (`value_vars`), are "unpivoted" to
+        the row axis, leaving just two non-identifier columns, 'variable' and
+        'value'.
+
+        Parameters
+        ----------
+        frame : DataFrame
+        id_vars : tuple, list, or ndarray, optional
+            Column(s) to use as identifier variables.
+        value_vars : tuple, list, or ndarray, optional
+            Column(s) to unpivot. If not specified, uses all columns that
+            are not set as `id_vars`.
+        var_name : scalar, default 'variable'
+            Name to use for the 'variable' column.
+        value_name : scalar, default 'value'
+            Name to use for the 'value' column.
+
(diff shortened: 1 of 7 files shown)


databricks/koalas/frame.py
```diff
@@ -3742,6 +3742,127 @@ def drop_duplicates(self, subset=None, inplace=False):
...
+            F.struct(*(
+                [F.lit(column).alias(var_name)] +
+                [F.col(id_var) for id_var in id_vars] +
+                [F.col(column).alias(value_name)])
+            ) for column in data_columns if column in value_vars]))
+
+        # New index column is always single index.
+        internal_index_column = "__index_level_0__"
+
+        columns = ([F.monotonically_increasing_id().alias(internal_index_column)] +
```
HyukjinKwon: one thing different from Pandas is that we use `monotonically_increasing_id` instead of sequential id

databricks/koalas/frame.py
```diff
@@ -3742,6 +3742,127 @@ def drop_duplicates(self, subset=None, inplace=False):
...
+        if len(value_vars) == 0:
+            value_vars = data_columns
+
+        data_columns = [data_column for data_column in data_columns if data_column not in id_vars]
+        sdf = self._sdf
+
+        pairs = F.explode(F.array(*[
+            F.struct(*(
+                [F.lit(column).alias(var_name)] +
+                [F.col(id_var) for id_var in id_vars] +
```
ueshin: Seems like we don't use this?

databricks/koalas/frame.py
HyukjinKwon: Hmm .. should we just don't set the index? I am not sure which way is better. Pandas' looks setting range index for this case

ueshin: So far, we just don't set the index if pandas resets and sets range index.
Maybe we need a discussion about such a case, whether we should use `monotonically_increasing_id` or leave it as-is.

HyukjinKwon: Hm .. okie. let me remove it here for now.

docs/source/reference/frame.rst
```diff
@@ -140,6 +140,7 @@ Reshaping, sorting, transposing
    DataFrame.sort_values
    DataFrame.nlargest
    DataFrame.nsmallest
+   DataFrame.melt
```
ueshin: Shall we add to `general_functions.rst` as well?

ueshin:
I'd merge this since seems like the travis builds were duplicated for the latest commit ace79b7 for some reason, and the one passed.

PR #475, opened 2019-06-17
Remove setters for old metadata.

### LINKED ISSUE #468: Misleading index argument in DataFrame constructor
When creating a new DataFrame with Koalas, the class' constructor accepts an `index` argument just like pandas. Unlike for pandas, however, `index` can't be a list of scalars but only a native `Metadata` object which is really confusing for pandas users. While I understand the reason `Metadata` exists, I'd suggest renaming the existing `index` argument to `metadata` and instead have an `index` argument which accepts a list of scalars and automatically infers `data_columns` like when calling the constructor with `metadata=None`.

ueshin:
This is part of #322.
This PR removes setters for old metadata structures and refine constructors for `DataFrame` and `Series` to avoid using old metadata structures.
Closes #468.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3735,8 +3735,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             raise KeyError("none key")
         if isinstance(key, str):
             try:
-                return Series(self._sdf.__getitem__(key), anchor=self,
-                              index=self._metadata.index_map)
+                return Series(self._internal.copy(scol=self._sdf.__getitem__(key)), anchor=self)
             except AnalysisException:
                 raise KeyError(key)
         if np.isscalar(key) or isinstance(key, (tuple, str)):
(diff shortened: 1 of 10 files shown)


databricks/koalas/frame.py
```diff
@@ -239,14 +234,11 @@ def __init__(self, data=None, index=None, columns=None, dtype=None, copy=False):
...
+            assert index is None
             assert columns is None
             assert dtype is None
             assert not copy
-            if index is None:
-                super(DataFrame, self).__init__(_InternalFrame(data))
-            else:
-                super(DataFrame, self).__init__(_InternalFrame(
-                    data, data_columns=index.data_columns, index_map=index.index_map))
+            super(DataFrame, self).__init__(_InternalFrame(data))
```
floscha: Not sure if it's too rare to be a use case, but why not allow passing `index` and `column` to the `_InternalFrame` constructor? `index` should then be a pandas-ish index (i.e. list of scalars) that implicitly needs to be converted to an index map.

ueshin: I'm not sure about the usage of `column` argument, but as for `index`, I think passing pandas-ish index is difficult for Spark backend.

Btw, the `_InternalFrame`'s constructor shouldn't do complex tasks, so we need to prepare it in `DataFrame`'s constructor or prior.

PR #479, opened 2019-06-18
Fix a merge function argument name to use `self` instead of `left`.

ueshin:
This is a follow-up of #381.
We should use `self` for the first argument to instance methods.


ueshin:
merging.

PR #494, opened 2019-06-21
Add an infrastructure for usage logging.

ueshin:


.travis.yml
```diff
@@ -29,6 +29,12 @@ matrix:
         - SPARK_VERSION=2.4.3
         - PANDAS_VERSION=0.24.2
         - PYARROW_VERSION=0.13.0
+    - python: "3.6"
+      env:
```
HyukjinKwon: @ueshin, BTW, is this matrix was added to test `KOALAS_USAGE_LOGGER='databricks.koalas.usage_logging.usage_logger`? I thought the matrix was good enough and we could add `KOALAS_USAGE_LOGGER` to one of three.

ueshin: Maybe we can use `python: "3.6"` env?

Issue #394, opened 2019-05-27
DataFrame.replace


thunterdb:
This is a very convenient function and I have heard multiple users asking for it. This is essentially a more elaborate API on top of DataFrame.map, which is in the process of being merged in #389 . 

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.replace.html

For the time being, I recommend implementing the following options:
 - to_replace=None
 - value=None

The following options are not as useful in my experience and can be added later:
 - inplace=False
 - regex=False

The following options are more difficult to support with spark, so we can leave them aside for now:
- limit=None
- method='pad'

HyukjinKwon:
FYI, I ended up with using CASE WHEN due to behaviour difference in Spark's replace .....

akhilanandbv003:
Can I give it a try?

HyukjinKwon:
Please go ahead!

parekhaagam:
I would like to contribute @akhilanandbv003  do u need any help?

akhilanandbv003:
@parekhaagam I have been busy with some other stuff, haven't made any progress in terms of this issue. You can go ahead and work on it.

parekhaagam:
Thank you @akhilanandbv003.  I will work on it and will update soon

shril:
I have opened a PR for this. @thunterdb please have a look.

shril:
@thunterdb, you can close this issue.

PR #495, opened 2019-06-22
DataFrame.replace

shril:
PR for #394

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2416,13 +2416,10 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         if regex is not False:
             raise NotImplementedError("replace currently doesn't supports regex")
 
-        if not isinstance(value, (float, int, str, dict, list)):
-            raise TypeError("Unsupported type %s" % type(value))
-
         if not isinstance(value, (int, float, str, list, dict)):
-            raise TypeError("Unsupported type %s" % type(value))
+            raise TypeError("Unsupported type {}".format(type(value)))
         if not isinstance(to_replace, (int, float, str, list, dict)):
-            raise TypeError("Unsupported type %s" % type(to_replace))
+            raise TypeError("Unsupported type {}".format(type(to_replace)))
 
         if isinstance(value, (list)) and isinstance(to_replace, (list)):
             if len(value) != len(to_replace):
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2350,6 +2350,92 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         else:
             return DataFrame(internal)
 
+    def replace(self, to_replace=None, value=None, subset=None,
+                inplace=False, limit=None, regex=False, method='pad') \
+                -> 'DataFrame':
+        """
+        Returns a new DataFrame replacing a value with another value.
+
+        Parameters
+        ----------
+        to_replace : int, float, string, or list
+            Value to be replaced. If the value is a dict, then value is ignored and
+            to_replace must be a mapping from column name (string) to replacement value.
+            The value to be replaced must be an int, float, or string.
+        value : int, float, string, or list
+            Value to use to replace holes. The replacement value must be an int, float,
+            or string. If value is a list, value should be of the same length with to_replace.
+        subset : sting, list
+            Optional list of column names to consider. Columns specified in subset that
+            do not have matching data type are ignored. For example, if value is a string,
+            and subset contains a non-string column, then the non-string column is simply ignored.
+        inplace : boolean, default False
+            Fill in place (do not create a new object)
+
+        Returns
+        -------
+        DataFrame
+            Object after replacement.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({"name": ['Ironman', 'Captain America', 'Thor', 'Hulk'],
+        ...                    "weapon": ['Mark-45', 'Shield', 'Mjolnir', 'Smash']})
+        >>> df
+                      name   weapon
+        0          Ironman  Mark-45
+        1  Captain America   Shield
+        2             Thor  Mjolnir
+        3             Hulk    Smash
+
+        >>> df.replace('Ironman', 'War-Machine')
+                      name   weapon
+        0      War-Machine  Mark-45
+        1  Captain America   Shield
+        2             Thor  Mjolnir
+        3             Hulk    Smash
+
+        >>> df.replace(['Ironman', 'Captain America'], ['Rescue', 'Hawkeye'], inplace=True)
+        >>> df
+              name   weapon
+        0   Rescue  Mark-45
+        1  Hawkeye   Shield
+        2     Thor  Mjolnir
+        3     Hulk    Smash
+
+        >>> df.replace('Mjolnir', 'Stormbuster', subset='weapon')
+              name       weapon
+        0   Rescue      Mark-45
+        1  Hawkeye       Shield
+        2     Thor  Stormbuster
+        3     Hulk        Smash
+        """
+        if method != 'pad':
+            raise NotImplementedError("replace currently works only for method='pad")
+        if limit is not None:
+            raise NotImplementedError("replace currently works only when limit=None")
+        if regex is not False:
+            raise NotImplementedError("replace currently doesn't supports regex")
+
+        if not isinstance(value, (float, int, str, dict, list)):
+            raise TypeError("Unsupported type %s" % type(value))
+
+        if not isinstance(value, (int, float, str, list, dict)):
+            raise TypeError("Unsupported type %s" % type(value))
+        if not isinstance(to_replace, (int, float, str, list, dict)):
+            raise TypeError("Unsupported type %s" % type(to_replace))
+
+        if isinstance(value, (list)):
+            if len(value) != len(to_replace):
+                raise ValueError('Length of to_replace and value must be same')
+
+        sdf = self._sdf.na.replace(to_replace, value, subset)
+        internal = self._internal.copy(sdf=sdf)
+        if inplace:
+            self._internal = internal
+        else:
+            return DataFrame(internal)
+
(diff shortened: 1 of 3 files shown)


shril:
@HyukjinKwon, I am getting a very weird error here.
I would be obliged if you can help me here.

databricks/koalas/frame.py
```diff
@@ -2350,6 +2350,88 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+        """
+        if method != 'pad':
+            raise NotImplementedError("replace currently works only for method='pad")
+        if limit is not None:
+            raise NotImplementedError("replace currently works only when limit=None")
+        if regex is not False:
+            raise NotImplementedError("replace currently doesn't supports regex")
+
+        if not isinstance(value, (int, float, str, list, dict)):
+            raise TypeError("Unsupported type {}".format(type(value)))
```
HyukjinKwon: I think error happened here ... It's `None` by default and `None` is not an instance of `(int, float, str, list, dict)`. So, it looks it raised an exception.

HyukjinKwon:
```
        if not isinstance(value, (int, float, str, list, dict)):
>           raise TypeError("Unsupported type {}".format(type(value)))
E           TypeError: Unsupported type <class 'NoneType'>
```

hsubbaraj:
@shril  @HyukjinKwon  Hi, I've been looking at this problem, and I have a couple comments/questions!

1. This doesn't have the same functionality as pandas replace, due to some of the differences that pyspark replace() has. pyspark's replace() doesn't handle non-mapping dictionaries as input and doesn't handle nested dictionary cases. Here are 2 examples of this (from the pandas replace() documentation)
`>>> df = pd.DataFrame({'A': [0, 1, 2, 3, 4],
...                    'B': [5, 6, 7, 8, 9],
...                    'C': ['a', 'b', 'c', 'd', 'e']})
>>> df.replace({'A': 0, 'B': 5}, 100)
     A    B  C
0  100  100  a
1    1    6  b
2    2    7  c
3    3    8  d
4    4    9  e

>>> df.replace({'A': {0: 100, 4: 400}})
     A  B  C
0  100  5  a
1    1  6  b
2    2  7  c
3    3  8  d
4  400  9  e`

2. When working with DataFrames that have numbers, we have to be careful about changing the index. Since the internal representation is a Spark Dataframe that has a column for the index : 
Screen Shot 2019-06-24 at 11 32 24 AM

if we just call pyspark's replace(), values in that index will also be replaced. 
Screen Shot 2019-06-24 at 11 35 04 AM

shril:
Interesting observation @hsubbaraj, I am going to test the same on my local implementation of `DataFrame.replace` for `koalas` and let you know the results.
Thank you for bringing this to our notice.

shril:
@hsubbaraj, your hypothesis is correct. Although I have also given a parameter called `subset` to protect the integrity of `index` column. This parameter is not there in the `pandas` API but keeping in view of the above problem I have provided it in `koalas`.

@HyukjinKwon what are your thoughts here?
```
>>> import pandas as pd

>>> kdf = ks.DataFrame({'A': [0, 1, 2, 3, 4],  'B': [5, 6, 7, 8, 9], 'C': ['a', 'b', 'c', 'd', 'e']})

>>> pdf = pd.DataFrame({'A': [0, 1, 2, 3, 4],  'B': [5, 6, 7, 8, 9], 'C': ['a', 'b', 'c', 'd', 'e']})

>>> pdf
   A  B  C
0  0  5  a
1  1  6  b
2  2  7  c
3  3  8  d
4  4  9  e
>>> kdf
   A  B  C                                                                      
0  0  5  a
1  1  6  b
2  2  7  c
3  3  8  d
4  4  9  e

>>> pdf.replace([0, 1, 2, 3], [4, 3, 2, 1])
   A  B  C
0  4  5  a
1  3  6  b
2  2  7  c
3  1  8  d
4  4  9  e

>>> kdf.replace([0, 1, 2, 3], [4, 3, 2, 1])
   A  B  C
4  4  5  a
3  3  6  b
2  2  7  c
1  1  8  d
4  4  9  e

>>> kdf.replace([0, 1, 2, 3], [4, 3, 2, 1], subset='index')
   A  B  C
0  0  5  a
1  1  6  b
2  2  7  c
3  3  8  d
4  4  9  e
```

HyukjinKwon:
@shril, per https://github.com/databricks/koalas/pull/495#issuecomment-505129634, can we support what we can easily do and document the other limitation for the future work later?

shril:
@HyukjinKwon, if we go by the [Koalas Design Principles](https://github.com/databricks/koalas#unify-small-data-pandas-api-and-big-data-spark-api-but-pandas-first) `DataFrame.replace` should be implemented as the alias of the `pandas` function.

Although this case is an exception of its own type. The `subset` parameter is provided in the `Spark` API for a reason. Therefore, I have included it in my own implementation. 
Anyway, I have a better idea in my mind. Give me a day's time and I will push new changes.

shril:
@HyukjinKwon, I have made the required changes and also tackled the edge cases. I have also added the notes regarding the subtle difference between the `koalas` and `pandas` API. Please have a look.

Getting this error in build - 
```

The command "if [[ $TRAVIS_PYTHON_VERSION == "3.5" ]]; then pip install -r requirements-dev.txt && pip list; else conda install -c conda-forge --yes --file requirements-dev.txt && conda list; fi" failed and exited with 1 during .
```

Can you please re-run the job?

HyukjinKwon:
I am not sure why it suddenly fails yet. I just roughly suspect it's failed due to conda server side issue.

HyukjinKwon:
Let's wait for a bit ...

HyukjinKwon:
@shril, can you sync with the master branch? a96c01f fixed the build

shril:
@HyukjinKwon I believe things are fine now. You can review my changes.

shril:
cc: @HyukjinKwon, can you please look into this.

databricks/koalas/frame.py
```diff
@@ -2350,6 +2350,134 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+        Parameters
+        ----------
+        to_replace : int, float, string, or list
+            Value to be replaced. If the value is a dict, then value is ignored and
+            to_replace must be a mapping from column name (string) to replacement value.
+            The value to be replaced must be an int, float, or string.
+        value : int, float, string, or list
+            Value to use to replace holes. The replacement value must be an int, float,
+            or string. If value is a list, value should be of the same length with to_replace.
+        subset : sting, list
```
HyukjinKwon: typo .. `sting` -> `string`

databricks/koalas/frame.py
```diff
@@ -2350,6 +2350,134 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+            raise NotImplementedError("replace currently works only when limit=None")
+        if regex is not False:
+            raise NotImplementedError("replace currently doesn't supports regex")
+
+        if value is not None and not isinstance(value, (int, float, str, list, dict)):
+            raise TypeError("Unsupported type {}".format(type(value)))
+        if to_replace is not None and not isinstance(to_replace, (int, float, str, list, dict)):
+            raise TypeError("Unsupported type {}".format(type(to_replace)))
+
+        if isinstance(value, (list)) and isinstance(to_replace, (list)):
```
HyukjinKwon: nit .. `(list)` -> `list`.

databricks/koalas/frame.py
```diff
@@ -2350,6 +2350,134 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+        1    1    6  b
+        2    2    7  c
+        3    3    8  d
+        4    4    9  e
+
+        Notes
+        -----
+        One difference between this implementation and pandas is that it is necessary
+        to specify the column name when you are passing dictionary in `to_replace`
+        parameter. Calling `replace` like `df.replace({0: 10, 1: 100})` will
```
HyukjinKwon: Maybe it's better to say... 

``Calling `replace` on its index such as `df.replace ...` ``

databricks/koalas/frame.py
```diff
@@ -2350,6 +2350,134 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+
+        Returns
+        -------
+        DataFrame
+            Object after replacement.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({"name": ['Ironman', 'Captain America', 'Thor', 'Hulk'],
+        ...                    "weapon": ['Mark-45', 'Shield', 'Mjolnir', 'Smash']})
```
HyukjinKwon: Ah, also I think we should specify `column` explicitly so that it's guaranteed to work in Python 3.5.x too.

databricks/koalas/frame.py
```diff
@@ -2350,6 +2350,134 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+            raise TypeError("Unsupported type {}".format(type(value)))
+        if to_replace is not None and not isinstance(to_replace, (int, float, str, list, dict)):
+            raise TypeError("Unsupported type {}".format(type(to_replace)))
+
+        if isinstance(value, (list)) and isinstance(to_replace, (list)):
+            if len(value) != len(to_replace):
+                raise ValueError('Length of to_replace and value must be same')
+
+        sdf = self._sdf.select(self._internal.data_columns)
+        if isinstance(to_replace, dict):
```
hsubbaraj:
```suggestion
        if isinstance(to_replace, dict) and value is None and \
                (not any(isinstance(i, dict) for i in to_replace.values())):
            sdf = sdf.replace(to_replace, value, subset)
        elif isinstance(to_replace, dict):
```

hsubbaraj: This addition handles the mapping case:
`kdf.replace({0: 10, 1: 100})`

ueshin:
I'd merge this and will submit a follow-up PR to address @HyukjinKwon's comments.
@hsubbaraj Could you submit a PR to address your comment? Thanks!

PR #496, opened 2019-06-23
Remaining Koalas Series.str functions

hjoo:
As discussed in #281, this PR completes most of the remaining Series.str functions for manipulating strings in Koalas Series.

Functions not implemented:
* `cat()` - returns scalar string, not a Koalas series
* `decode()`, `encode()` - BinaryType not supported in pyspark < 2.4
* `extract()`, `extractall()` - returns dataframe of varying dimensions
* `partition()`, `rpartition()`- returns dataframe of varying dimensions
* `get_dummies()` - returns dataframe of varying dimensions

### DIFF
--- a/databricks/koalas/strings.py
+++ b/databricks/koalas/strings.py
@@ -827,7 +827,7 @@ class StringMethods(object):
 
     def contains(self, pat, case=True, flags=0, na=None, regex=True) -> 'ks.Series':
         """
-        Testcif pattern or regex is contained within a string of a Series.
+        Test if pattern or regex is contained within a string of a Series.
 
         Return boolean Series based on whether a given pattern or regex is
         contained within a string of a Series.
--- a/databricks/koalas/strings.py
+++ b/databricks/koalas/strings.py
@@ -21,7 +21,7 @@ from typing import TYPE_CHECKING
 
 import numpy as np
 
-from pyspark.sql.types import StringType, BinaryType, BooleanType
+from pyspark.sql.types import StringType, BinaryType, BooleanType, IntegerType, ArrayType, LongType
 
 from databricks.koalas.base import _wrap_accessor_pandas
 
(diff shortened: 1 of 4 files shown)


databricks/koalas/tests/test_series_string.py
```diff
@@ -130,3 +137,182 @@ def test_string_isnumeric(self):
...
+        self.check_func(lambda x: x.str.find('a', start=0, end=1))
+
+    def test_string_findall(self):
+        self.check_func(lambda x: x.str.findall('es|as'))
+        self.check_func(lambda x: x.str.findall('wh.*', flags=re.IGNORECASE))
+
+    def test_string_index(self):
+        pds = pd.Series(['tea', 'eat'])
+        self.check_func_on_series(lambda x: x.str.index('ea'), pds)
+        with self.assertRaises(ValueError):
```
HyukjinKwon: I think it's fine to catch just `Exception`. Seems like lower Spark version throws a different exception.

databricks/koalas/strings.py
```diff
@@ -350,3 +360,717 @@ def isdecimal(self) -> 'ks.Series':
...
+        regex : bool, default True
+            If True, assumes the pat is a regular expression.
+            If False, treats the pat as a literal string.
+
+
+        Returns
+        -------
+        Series of boolean values
+            A Series of boolean values indicating whether the given pattern is
+            contained within the string of each element of the Series.
```
HyukjinKwon: while we're here, can we copy and paste Pandas doctests too? e.g., https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.contains.html?highlight=contains#pandas.Series.str.contains

The doctests don't need to be exactly same if it doesn't work currently.

HyukjinKwon: I don't mind if you want to do it separately in another PR.

hjoo: Added doctests for all the methods doctest-ed in Pandas and some more.

hjoo:
Thanks @HyukjinKwon !

HyukjinKwon:
Thanks, @hjoo.

PR #481, opened 2019-06-18
DataFrame.update()

yuwillrun:

PR #498, opened 2019-06-24
DataFrame.update()

yuwillrun:
synac base master

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3591,8 +3591,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             old_col = update_sdf[column_name]
             new_col = update_sdf[column_name + '_new']
             if errors == 'raise':
-                null_sdf = update_sdf.filter(old_col.isNull() & new_col.isNull())
-                if null_sdf:
+                null_row_num = update_sdf.filter(old_col.isNull() & new_col.isNull()).count()
+                if null_row_num > 0:
                     raise ValueError("Data overlaps.")
 
             if overwrite:
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3494,9 +3494,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         from databricks.koalas.namespace import concat
         return concat([self, other], ignore_index=ignore_index)
 
-    # TODO: add 'filter_func' parameter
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -3494,6 +3494,116 @@ def append(self, other: 'DataFrame', ignore_index: bool = False,
...
+
+        update_columns = list(set(self._internal.data_columns)
+                              .intersection(set(other._internal.data_columns)))
+        update_sdf = self.join(other[update_columns], rsuffix='_new')._sdf
+
+        for column_name in update_columns:
+            old_col = update_sdf[column_name]
+            new_col = update_sdf[column_name + '_new']
+            if errors == 'raise':
+                null_sdf = update_sdf.filter(old_col.isNull() & new_col.isNull())
```
HyukjinKwon: @yuwillrun, seems https://github.com/databricks/koalas/pull/481/files#r296667059 comment was missed.

`update_sdf.filter(old_col.isNull() & new_col.isNull())` always return a dataframe. So, it will always raise an exception.

If it's tricky to do, we can remove `errors` for now and do it later in another PR.

yuwillrun: Sorry, i will check it now.

yuwillrun: Fixed already.

databricks/koalas/frame.py
```diff
@@ -3494,6 +3494,116 @@ def append(self, other: 'DataFrame', ignore_index: bool = False,
...
+
+        update_columns = list(set(self._internal.data_columns)
+                              .intersection(set(other._internal.data_columns)))
+        update_sdf = self.join(other[update_columns], rsuffix='_new')._sdf
+
+        for column_name in update_columns:
+            old_col = update_sdf[column_name]
+            new_col = update_sdf[column_name + '_new']
+            if errors == 'raise':
+                null_row_num = update_sdf.filter(old_col.isNull() & new_col.isNull()).count()
```
HyukjinKwon: The problem here is that if `update_sdf` is pretty large, `count` is non-trivial operation. It needs to scan whole data again. Let's just don't support `errors` for now.

yuwillrun: OK, I'll remove `errors` parameter now.

HyukjinKwon:
Thanks, @yuwillrun

PR #499, opened 2019-06-24
Fix for frame.replace()

hsubbaraj:
This is a PR for #394 that also fixes the issue raised in #495! @thunterdb @HyukjinKwon @shril can you take a look?


HyukjinKwon:
@hsubbaraj, if you're going to take #495 over, can you pick up https://github.com/databricks/koalas/pull/495's commit into your PR here so that I can push two commits authored by you and @shril?

HyukjinKwon:
Let's do it at #495. Seems close to merge.

PR #501, opened 2019-06-25
DataFrame.cummax()

RainFung:
implement DataFrame.cummax() base on DataFrame.cumsum()

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3495,6 +3495,98 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         from databricks.koalas.namespace import concat
         return concat([self, other], ignore_index=ignore_index)
 
+    # TODO: add 'filter_func' and 'errors' parameter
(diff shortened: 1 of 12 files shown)


databricks/koalas/frame.py
```diff
@@ -3987,6 +3988,71 @@ def describe(self, percentiles: Optional[List[float]] = None) -> 'DataFrame':
...
+
+        >>> df.cummax()
+             A    B
+        0  2.0  1.0
+        1  3.0  NaN
+        2  3.0  1.0
+        """
+        index_columns = self._internal.index_columns
+        data_columns = self._internal.data_columns
+        window = Window.orderBy(index_columns).rangeBetween(Window.unboundedPreceding, 0)
```
HyukjinKwon: Seems this code is repeated in your PRs. can we deduplicate? for instance define one function and reuse it.

HyukjinKwon:
Im going to take this over at https://github.com/databricks/koalas/pull/521

PR #510, opened 2019-06-26
Add DataFrame.select_dtypes

HyukjinKwon:
This PR adds `DataFrame.select_dtypes`

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2834,7 +2834,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         4   True  1.0  a
         5  False  2.0  b
 
-        Spark SQL DDL type definition can be used as strings as well.
+        Spark SQL DDL type strings can be used as well.
 
         >>> df.select_dtypes(exclude=['string'])
            a      b    c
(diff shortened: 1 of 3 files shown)


rxin:
I feel the object type is pretty stupid in pandas. I think we should support selecting using spark's type system as well. Same thing applies to dtypes/dtype. I'd have dtype return "str" rather than "object" for string columns.

HyukjinKwon:
k. let me take a look

HyukjinKwon:
Should be ready for a look.

databricks/koalas/frame.py
```diff
@@ -2746,6 +2747,167 @@ def dtypes(self):
...
+        -----
+        * To select datetimes, use ``np.datetime64``, ``'datetime'`` or
+          ``'datetime64'``
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'a': [1, 2] * 3,
+        ...                    'b': [True, False] * 3,
+        ...                    'c': [1.0, 2.0] * 3,
+        ...                    'd': ['a', 'b'] * 3})
```
ueshin: We need `columns=[...]`?

PR #513, opened 2019-06-27
Add reason for unsupported functions

HyukjinKwon:
This PR proposes to add a reason for unsupported functions.

While I am here,
  - `values` at Series, Index, DataFrame were moved into missing package with the reason.
  - `to_pickle` at DataFrame

### DIFF
--- a/databricks/koalas/missing/frame.py
+++ b/databricks/koalas/missing/frame.py
@@ -165,4 +165,5 @@ class _MissingPandasLikeDataFrame(object):
 
     to_pickle = unsupported_function(
         'to_pickle',
-        reason="Pickled data is not designed for a long term.")
+        reason="For storage, we encourage you to use Delta or Parquet, instead of Python pickle "
+               "format.")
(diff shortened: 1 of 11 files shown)


rxin:
I actually don't think we should have this, because this would encourage very bad behavior (pickle is just not a great way to store data in the long term).

Can we add a way to throw an error message saying this is not supported because it doesn't make sense, similar to the ones we do for deprecated functions?

HyukjinKwon:
Yea, actually it's not supposed to be used for a long term. Let me fix it like that.

HyukjinKwon:
cc @ueshin too. I believe this is ready for a look.

databricks/koalas/missing/frame.py
```diff
@@ -158,3 +157,12 @@ class _MissingPandasLikeDataFrame(object):
...
     to_panel = unsupported_function('to_panel', deprecated=True)
+
+    # Functions and properties we won't support.
+    values = unsupported_property(
+        'values',
+        reason="If you want to collect your data as an NumPy array, use 'to_numpy()' instead.")
+
+    to_pickle = unsupported_function(
+        'to_pickle',
+        reason="Pickled data is not designed for a long term.")
```
rxin: I'd change this to

"For storage, we encourage you to use Delta or Parquet, instead of Python pickle format."

rxin:
Looks good. Just one comment on the message.

PR #515, opened 2019-07-01
Add a reason why memory_usage is unimplemented

HyukjinKwon:
This PR adds a reason why memory_usage is unimplemented

### DIFF
--- a/databricks/koalas/missing/common.py
+++ b/databricks/koalas/missing/common.py
@@ -1,0 +1,26 @@
+#
+# Copyright (C) 2019 Databricks, Inc.
+#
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
@rxin and @ueshin WDYT?

rxin:
Maybe say something like "Unlike pandas, most DataFrames are not materialized in memory in Spark (and Koalas), and as a result memory_usage() does not what what you intend it to do. Use Spark's web UI to monitor disk and memory usage of your application."

Can we define this once and reference it in all the instances of the function?

HyukjinKwon:
Yea let me fix it like that

PR #520, opened 2019-07-03
Add a note in rank that all data will be moved into single node

HyukjinKwon:
Add a note in rank that all data will be moved into single node


HyukjinKwon:
Today is a release date. I will merge and fix it later. @rxin let me know if the current note looks weird.

PR #504, opened 2019-06-26
DataFrame.cummin()

RainFung:
first cummin version base cummax

PR #500, opened 2019-06-25
DataFrame.cumsum()

RainFung:

PR #521, opened 2019-07-03
Add DataFrame.cummin(), cummax() and cumsum()

HyukjinKwon:
This PR takes over #500 #501, #504.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4595,9 +4595,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         for column_name in data_columns:
             # It defines another column that holds true or false for nulls first.
             is_null_column = "%s_isnull" % column_name
-            sdf = sdf.withColumn(
-                is_null_column,
-                F.when(sdf[column_name].isNull(), F.lit(True)).otherwise(F.lit(False)))
+            sdf = sdf.withColumn(is_null_column, sdf[column_name].isNull())
 
             if skipna:
                 # There is a behavior difference between pandas and PySpark. In case of cummax,
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -29,6 +29,7 @@ from pandas.api.types import is_list_like, is_dict_like
 from pandas.core.dtypes.common import infer_dtype_from_object
 from pandas.core.dtypes.inference import is_sequence
 from pyspark import sql as spark
+from pyspark.sql.window import Window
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -4581,6 +4582,246 @@ def describe(self, percentiles: Optional[List[float]] = None) -> 'DataFrame':
...
+
+        index_columns = self._internal.index_columns
+        data_columns = self._internal.data_columns
+        window = Window.orderBy(
+            index_columns).rowsBetween(Window.unboundedPreceding, Window.currentRow)
+        sdf = self._sdf
+
+        for column_name in data_columns:
+            # It defines another column that holds true or false for nulls first.
+            is_null_column = "%s_isnull" % column_name
```
HyukjinKwon: Actually .. I think those expressions should go to `Series.cum*` .. I will do it in the next release.

HyukjinKwon:
Tests passed.

databricks/koalas/frame.py
```diff
@@ -4581,6 +4582,246 @@ def describe(self, percentiles: Optional[List[float]] = None) -> 'DataFrame':
...
+            sdf = sdf.withColumn(
+                is_null_column,
+                F.when(sdf[column_name].isNull(), F.lit(True)).otherwise(F.lit(False)))
```
ueshin: nit: only `sdf[column_name].isNull()` should work instead of `F.when(...).otherwise(...)`?

HyukjinKwon: D'oh, yes.

RainFung:
It seems you can merge some similar docstring.
They are same except function name.

```
Return cumulative minimum over a DataFrame axis.
        Returns a DataFrame of the same size containing the cumulative minimum.
        .. note:: the current implementation of cummin uses Spark's Window without
            specifying partition specification. This leads to move all data into
            single partition in single machine and could cause serious
            performance degradation. Avoid this method against very large dataset.
        Parameters
        ----------
        skipna : boolean, default True
            Exclude NA/null values. If an entire row/column is NA, the result will be NA.
        Returns
        -------
        DataFrame
        See Also
        --------
        DataFrame.min : Return the minimum over DataFrame axis.
        DataFrame.cummax : Return cumulative maximum over DataFrame axis.
        DataFrame.cummin : Return cumulative minimum over DataFrame axis.
        DataFrame.cumsum : Return cumulative sum over DataFrame axis.
```

HyukjinKwon:
Please go ahead for a PR.

PR #527, opened 2019-07-03
Updated frame.replace()

hsubbaraj:
Fix for replace to add support for the mapping case as referenced in #495 :

`kdf.replace({0: 10, 1: 100})`

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -25,6 +25,7 @@ from typing import Any, Optional, List, Tuple, Union, Generic, TypeVar
 
 import numpy as np
 import pandas as pd
+import pyspark.sql.functions as F
 from pandas.api.types import is_list_like, is_dict_like
 from pandas.core.dtypes.common import infer_dtype_from_object
 from pandas.core.dtypes.inference import is_sequence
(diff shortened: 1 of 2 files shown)


hsubbaraj:
@ueshin can you take a look? Thanks!

databricks/koalas/tests/test_dataframe.py
```diff
@@ -782,6 +782,27 @@ def test_replace(self):
...
         )
 
+        pdf = pd.DataFrame({'A': [0, 1, 2, 3, 4],
+                            'B': [5, 6, 7, 8, 9],
+                            'C': ['a', 'b', 'c', 'd', 'e']})
+
+        kdf = ks.from_pandas(pdf)
+
+        assert(repr(kdf.replace([0, 1, 2, 3], 4)) ==
+               repr(pdf.replace([0, 1, 2, 3], 4)))
```
ueshin: Shall we use `self.assert_eq()`?

databricks/koalas/frame.py
```diff
@@ -25,6 +25,7 @@
 
 import numpy as np
 import pandas as pd
+import pyspark.sql.functions as F
```
ueshin: nit: an unnecessary change?

hsubbaraj: Yes, I didn't notice the import on a different line!

databricks/koalas/tests/test_dataframe.py
hsubbaraj: Sure, I will make the change!

hsubbaraj:
@ueshin should be good to go, can you take a look? Thanks :)

ueshin:
Thanks! merging.

PR #528, opened 2019-07-03
Fix build with pandas 0.23.

### LINKED ISSUE #525: Import Error for "import databricks.koalas" -  ImportError: cannot import name 'infer_dtype_from_object'
Getting Import error while importing koalas in databricks notebook. It was working fine until yesterday.

Running - 

Error -

ImportError: cannot import name 'infer_dtype_from_object'

ueshin:
Currently only the travis build with Python 3.5 is supposed to use the older versions of pandas and pyarrow, but the 3.5 build doesn't work with the older versions because the libraries are not pinned and the newest versions are used since the 3.5 build uses `pip` instead of `conda`.
As as result, we missed the failures with pandas 0.23.
Resolves #525.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,6 @@ install:
   - |
     if [[ $TRAVIS_PYTHON_VERSION == "3.5" ]]; then
       pip install -r requirements-dev.txt && \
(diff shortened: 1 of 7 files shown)


ueshin:
We need to pin library versions with pip command. I'm not sure how to do it yet.

PR #530, opened 2019-07-04
Dataframe.rank()

RainFung:
add rank parameter for koalas.


HyukjinKwon:
Nice. is it possible to fix `Series.rank` and use it instead in `DataFrame.rank`? maybe by iterating each column?

RainFung:
```python
            if method == 'first':
                window = Window.orderBy(index_columns + [column_name],
                                        ascending=[ascending, ascending])\
                    .rowsBetween(Window.unboundedPreceding, Window.currentRow)
                sdf = sdf.withColumn(column_name, F.row_number().over(window))
            elif method == 'dense_rank':
                window = Window.orderBy(column_name, ascending=ascending)\
                    .rowsBetween(Window.unboundedPreceding, Window.currentRow)
                sdf = sdf.withColumn(column_name, F.dense_rank().over(window))
            else:
>               window = Window.orderBy(column_name, ascending=ascending)\
                    .rowsBetween(Window.unboundedPreceding, Window.currentRow)
E               TypeError: orderBy() got an unexpected keyword argument 'ascending'

databricks/koalas/frame.py:5070: TypeError
=============================================
```

Can you help me fix the problem?

RainFung:
@HyukjinKwon  Can you help me add how to implement `ascending` parameter in window function?

RainFung:
https://issues.apache.org/jira/browse/SPARK-18818

HyukjinKwon:
@RainFung, we don't neccesrailly need `ascending` in orderBy. You can `Window.orderBy(F.col(column).desc())` or `Window.orderBy(self[column].asc())`.

RainFung:
@ueshin 
Can u cc this build errored ? Thanks.
`The command "sudo apt-get update" failed and exited with 100 during .`

HyukjinKwon:
Tests were fixed. @RainFung, can you resolve the conflcits? Let me review quickly.

databricks/koalas/frame.py
```diff
@@ -4908,6 +4909,155 @@ def drop_duplicates(self, subset=None, inplace=False):
...
+            asc_func = spark.functions.asc
+        else:
+            asc_func = spark.functions.desc
+
+        index_column = self._internal.index_columns[0]
+        data_columns = self._internal.data_columns
+        sdf = self._sdf
+
+        for column_name in data_columns:
+            if na_option == 'top':
```
HyukjinKwon: @RainFung, is it possible to share the same codes with Series.rank as we did for cummin, commax and cumsum?

databricks/koalas/frame.py
```diff
@@ -4908,6 +4909,155 @@ def drop_duplicates(self, subset=None, inplace=False):
...
+            if method == 'first':
+                window = Window.orderBy(asc_func(column_name), asc_func(index_column))\
+                    .rowsBetween(Window.unboundedPreceding, Window.currentRow)
+                sdf = sdf.withColumn(column_name, F.row_number().over(window))
+            elif method == 'dense':
+                window = Window.orderBy(asc_func(column_name))\
+                    .rowsBetween(Window.unboundedPreceding, Window.currentRow)
+                sdf = sdf.withColumn(column_name, F.dense_rank().over(window))
+            else:
+                if method == 'average':
```
HyukjinKwon: Hey, I meant I was wondering if we can deduplicate this code paths:

https://github.com/databricks/koalas/blob/1dbf5dea29a0bb171ddad4aff802b57e6658cafe/databricks/koalas/series.py#L2239-L2252

So that Series and DataFrame both have the same `rank` functionalities.

databricks/koalas/frame.py
```diff
@@ -4908,6 +4909,155 @@ def drop_duplicates(self, subset=None, inplace=False):
...
+            asc_func = spark.functions.desc
+
+        index_column = self._internal.index_columns[0]
+        data_columns = self._internal.data_columns
+        sdf = self._sdf
+
+        for column_name in data_columns:
+            if na_option == 'top':
+                if ascending:
+                    fill_value = sdf.agg({column_name: 'min'}).collect()[0][0] - 1
```
HyukjinKwon: Let's avoid this way .. it will be pretty expensive and doesn't lazily execute them.. `.collect()` will actually trigger the computation.

Seems difficult to implement for now. We can simply throw an error that says we don't support it for now.

databricks/koalas/frame.py
```diff
@@ -4908,6 +4909,155 @@ def drop_duplicates(self, subset=None, inplace=False):
...
+                    fill_value = sdf.agg({column_name: 'max'}).collect()[0][0] + 1
+                sdf = sdf.fillna({column_name: fill_value})
+            elif na_option == 'bottom':
+                if ascending:
+                    fill_value = sdf.agg({column_name: 'max'}).collect()[0][0] + 1
+                else:
+                    fill_value = sdf.agg({column_name: 'min'}).collect()[0][0] - 1
+                sdf = sdf.fillna({column_name: fill_value})
+            else:
+                null_number = sdf.filter(F.col(column_name).isNull()).count()
```
HyukjinKwon: This `count()` too. We should avoid to execute an action within an API.

RainFung:
Is it necessary to modify the some spark origin code directly in order to support koalas better ?

HyukjinKwon:
Hey, https://github.com/databricks/koalas/pull/530#discussion_r301013063 isn't resolved yet ... 

We can fix Spark if it's a bug or it makes sense within Spark; otherwise, we shouldn't. I don't know how to handle those cases yet. So let's don't fix such cases for now. We can avoid to support such cases here and then revisit later.

RainFung:
OK. I'll modify `series.rank` code today. I hope you can code review it tomorrow.

RainFung:
Can't pass test_series.py? Can you help me check it ?I think left is same as right
```python
Failure
  File "/Users/fengyurun1/anaconda3/envs/koalas-dev-env/lib/python3.6/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/Users/fengyurun1/anaconda3/envs/koalas-dev-env/lib/python3.6/unittest/case.py", line 605, in run
    testMethod()
  File "/Users/fengyurun1/Desktop/koalas/databricks/koalas/tests/test_series.py", line 462, in test_rank
    self.assert_eq(pser.rank(na_option='keep'), kser.rank(na_option='keep').sort_index())
  File "/Users/fengyurun1/Desktop/koalas/databricks/koalas/testing/utils.py", line 246, in assert_eq
    self.assertPandasEqual(lpdf, rpdf)
  File "/Users/fengyurun1/Desktop/koalas/databricks/koalas/testing/utils.py", line 181, in assertPandasEqual
    self.assertTrue((left == right).all(), msg=msg)
  File "/Users/fengyurun1/anaconda3/envs/koalas-dev-env/lib/python3.6/unittest/case.py", line 682, in assertTrue
    raise self.failureException(msg)
AssertionError: False is not true : Series are not equal: 

Left:
0    1.5
1    NaN
2    3.0
3    4.0
4    1.5
Name: x, dtype: float64
float64

Right:
0    1.5
1    NaN
2    3.0
3    4.0
4    1.5
Name: x, dtype: float64
float64



Ran 1 test in 9.317s

FAILED (failures=1)
sys:1: ResourceWarning: unclosed file <_io.BufferedWriter name=5>
```

HyukjinKwon:
In that case, you can do:

```python
self.assertEqual(repr(pser.rank(na_option='keep')), repr(kser.rank(na_option='keep').sort_index()))
```

RainFung:
Create new PR which update series.rank() 
https://github.com/databricks/koalas/pull/546

HyukjinKwon:
Let's track it there

PR #532, opened 2019-07-05
Add DataFrame.pivot

### LINKED ISSUE #492: Support for pivot
Customer requested that we implement pivot. They are migrating to Spark. Their legacy platform involves a lot of pandas code. They said pivot is an important method in their ETL scripts but they haven't provided a use case yet. Will push them a bit more but let me know if you need anymore context.

HyukjinKwon:
This PR adds `DataFrame.pivot`.

This is basically same as `pivot_table(aggfunc='first', ...)`.
There are two differences between pandas' `pivot` and `pivot` proposed here to Koalas.

- `values` is required due to support multi index as columns unlike pandas'. This is followup work.

    ```python
    >>> import as pd
    >>> df = pd.DataFrame({'foo': ['one', 'one', 'one', 'two', 'two',
    ...                            'two'],
    ...                    'bar': ['A', 'B', 'C', 'A', 'B', 'C'],
    ...                     'baz': [1, 2, 3, 4, 5, 6],
    ...                    'zoo': ['x', 'y', 'z', 'q', 'w', 't']})
    >>> df.pivot(index="foo", columns="bar")
        baz       zoo
    bar   A  B  C   A  B  C
    foo
    one   1  2  3   x  y  z
    two   4  5  6   q  w  t
    >>> df.pivot(index="foo", columns="bar").columns
    MultiIndex(levels=[['baz', 'zoo'], ['A', 'B', 'C']],
               labels=[[0, 0, 0, 1, 1, 1], [0, 1, 2, 0, 1, 2]],
               names=[None, 'bar'])
    ```

- pandas' `pivot` raises an value error when duplicated values are found during pivot whereas `pivot` here does not as, up to my knowledge, it is preferred to permissively execute over failing fast when processing large data

    ```python
    >>> import pandas as pd
    >>> df = pd.DataFrame({"foo": ['one', 'one', 'two', 'two'],
    ...                    "bar": ['A', 'A', 'B', 'C'],
    ...                    "baz": [1, 2, 3, 4]}, columns=['foo', 'bar', 'baz'])
    >>> df.pivot(index='foo', columns='bar', values='baz')
        ...
    ValueError: Index contains duplicate entries, cannot reshape
    ```

Resolves #492

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3036,6 +3036,114 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
             return DataFrame(sdf.withColumn(columns, F.lit(index_values))).set_index(columns)
 
+    def pivot(self, index=None, columns=None, values=None):
+        """
+        Return reshaped DataFrame organized by given index / column values.
+
+        Reshape data (produce a "pivot" table) based on column values. Uses
+        unique values from specified `index` / `columns` to form axes of the
+        resulting DataFrame. This function does not support data
+        aggregation.
+
+        Parameters
+        ----------
+        index : string, optional
+            Column to use to make new frame's index. If None, uses
+            existing index.
+        columns : string
+            Column to use to make new frame's columns.
+        values : string, object or a list of the previous
+            Column(s) to use for populating new frame's values.
+        Returns
+        -------
+        DataFrame
+            Returns reshaped DataFrame.
+
+        See Also
+        --------
+        DataFrame.pivot_table : Generalization of pivot that can handle
+            duplicate values for one index/column pair.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'foo': ['one', 'one', 'one', 'two', 'two',
+        ...                            'two'],
+        ...                    'bar': ['A', 'B', 'C', 'A', 'B', 'C'],
+        ...                    'baz': [1, 2, 3, 4, 5, 6],
+        ...                    'zoo': ['x', 'y', 'z', 'q', 'w', 't']},
+        ...                   columns=['foo', 'bar', 'baz', 'zoo'])
+        >>> df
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -3036,6 +3036,114 @@ def pivot_table(self, values=None, index=None, columns=None,
...
+        ... # doctest: +NORMALIZE_WHITESPACE
+             A    B    C
+        0  1.0  NaN  NaN
+        1  NaN  2.0  NaN
+        2  NaN  NaN  3.0
+        3  4.0  NaN  NaN
+        4  NaN  5.0  NaN
+        5  NaN  NaN  6.0
+
+        Notice that, unlike pandas raises an ValueError when duplicated values are found,
```
HyukjinKwon: cc @rxin and @ueshin

HyukjinKwon:
I am merging this one but @rxin and @ueshin, please take a look at https://github.com/databricks/koalas/pull/532#discussion_r300520548 and let me know if you guys think differently.

databricks/koalas/frame.py
rxin: is there an example for this?

HyukjinKwon: Yea, the example is below. There are two `A` and `one`s. In this case, pandas throws an error but here I allowed it partly to prevent execution failure in the middle, and partly because it was easier to implement.

HyukjinKwon: And it only results out one first value of `A` and `one` currently in Koalas.

rxin: OK I'd actually fail it here similar to pandas. The reason is that in most cases people should really be running some aggregation with pivot anyway.

HyukjinKwon: Okie. Let me take a look.

Issue #502, opened 2019-06-25
plot: ufunc 'multiply' did not contain a loop with signature matching types dtype('<U32') dtype('<U32') dtype('<U32')


rxin:
I loaded the SF fire department dataset, and ... [fire-small.txt](https://github.com/databricks/koalas/files/3329631/fire-small.txt)

```
import databricks.koalas as ks
data = ks.read_csv("fire-small.txt", header=0, usecols=range(40))
data['Call Type'].plot.hist()
```

Error

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<command-3469248> in <module>()
----> 1 df['Call Type'].plot.hist()

/local_disk0/pythonVirtualEnvDirs/virtualEnv-b7eb6ad9-1ecb-4cf5-bb84-76d207603fc1/lib/python3.5/site-packages/databricks/koalas/plot.py in hist(self, bins, **kwds)
    540         axes : :class:`matplotlib.axes.Axes` or numpy.ndarray of them
    541         """
--> 542         return self(kind='hist', bins=bins, **kwds)
    543 
    544     def kde(self, bw_method=None, ind=None, **kwds):

/local_disk0/pythonVirtualEnvDirs/virtualEnv-b7eb6ad9-1ecb-4cf5-bb84-76d207603fc1/lib/python3.5/site-packages/databricks/koalas/plot.py in __call__(self, kind, ax, figsize, use_index, title, grid, legend, style, logx, logy, loglog, xticks, yticks, xlim, ylim, rot, fontsize, colormap, table, yerr, xerr, label, secondary_y, **kwds)
    466                            colormap=colormap, table=table, yerr=yerr,
    467                            xerr=xerr, label=label, secondary_y=secondary_y,
--> 468                            **kwds)
    469     __call__.__doc__ = plot_series.__doc__
    470 

/local_disk0/pythonVirtualEnvDirs/virtualEnv-b7eb6ad9-1ecb-4cf5-bb84-76d207603fc1/lib/python3.5/site-packages/databricks/koalas/plot.py in plot_series(data, kind, ax, figsize, use_index, title, grid, legend, style, logx, logy, loglog, xticks, yticks, xlim, ylim, rot, fontsize, colormap, table, yerr, xerr, label, secondary_y, **kwds)
    421                  yerr=yerr, xerr=xerr,
    422                  label=label, secondary_y=secondary_y,
--> 423                  **kwds)
    424 
    425 

[...]

PR #536, opened 2019-07-07
raise error when hist on non-numeric cols

dvgodoy:
This PR fixes bug #502 , raising the proper TypeError when `hist()` is called on a non-numeric column (just like pandas does).
I also had to fix another (AFAIK unreported) bug in `select_dtypes` in the process, as it was not properly appending columns to the list.


databricks/koalas/frame.py
```diff
@@ -3245,7 +3245,7 @@ def select_dtypes(self, include=None, exclude=None):
                     self._sdf.schema[col].dataType in exclude_spark_type)
 
             if should_include:
-                columns += col
+                columns.append(col)
```
HyukjinKwon: Oh, nice catch.

databricks/koalas/plot.py
```diff
@@ -376,6 +378,25 @@ def calc_histogram(self, bins):
 
 class KoalasHistPlot(HistPlot):
     def _args_adjust(self):
+        from databricks.koalas.series import Series
+
+        data = self.data
+        if isinstance(data, Series):
+            data = data.to_frame()
+
+        numeric_data = data.select_dtypes(include=['byte', 'decimal', 'integer', 'float',
```
HyukjinKwon: I think we can simply check `dtypes`.

databricks/koalas/plot.py
```diff
@@ -376,6 +378,25 @@ def calc_histogram(self, bins):
...
+
+        try:
+            is_empty = numeric_data.empty
+        except AttributeError:
+            is_empty = not len(numeric_data)
+
+        # no empty frames or series allowed
+        if is_empty:
+            raise TypeError('Empty {0!r}: no numeric data to '
+                            'plot'.format(numeric_data.__class__.__name__))
```
HyukjinKwon: Can we add a simple test to prevent the mistake like the one in select_dtypes?

dvgodoy:
@HyukjinKwon 
I've just added tests to it - sorry for taking so long, but I went on holidays :-)

HyukjinKwon:
Sorry for late input. I rebased and opened a PR at https://github.com/databricks/koalas/pull/766.
I will credit to you.

PR #540, opened 2019-07-08
Add Series.median()

HyukjinKwon:
This PR adds Series.median()

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1944,6 +1944,36 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         """
         return self._reduce_for_stat_function(_Frame._count_expr)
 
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -1944,6 +1944,36 @@ def count(self):
...
+        median : scalar
+
+        Examples
+        -------
+        >>> ks.Series([24., 21., 25., 33., 26.]).median()
+        25.0
+
+        """
+        if not isinstance(accuracy, int):
+            raise ValueError("actually must be an integer; however, got [%s]" % type(accuracy))
```
harupy: Is **"actually"** a typo?

HyukjinKwon: ah, yea. thanks for catching this.

PR #546, opened 2019-07-10
DataFrame.rank() and Series.rank()

RainFung:


databricks/koalas/tests/test_series.py
```diff
@@ -477,3 +477,29 @@ def test_cumsum(self):
     def test_median(self):
         with self.assertRaisesRegex(ValueError, "accuracy must be an integer; however"):
             koalas.Series([24., 21., 25., 33., 26.]).median(accuracy="a")
+
+    def test_rank(self):
+        pser = pd.Series([1, np.nan, 2, 3, 1], name='x')
+        kser = koalas.from_pandas(pser)
+        self.assert_eq(repr(pser.rank(na_option='keep')),
```
HyukjinKwon: Let's use `assertEqual` instead of `assert_eq`. `assert_eq` compares pandas dataframes too but we only need to compare the strings here.

databricks/koalas/frame.py
```diff
@@ -5306,6 +5307,148 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
...
+            asc_func = spark.functions.desc
+
+        index_column = self._internal.index_columns[0]
+        data_columns = self._internal.data_columns
+        sdf = self._sdf
+
+        for column_name in data_columns:
+            if na_option == 'top':
+                if not ascending:
+                    fill_value = sys.maxsize
```
HyukjinKwon: It's actually kind of dangerous assumption. The execution happens in JVM and the `maxsize` can be different. I think it can also matter if the types are `np.int32` or Spark's`IntegerType` in JVM. Can we avoid to use `maxsize`?

databricks/koalas/frame.py
```diff
@@ -5306,6 +5307,148 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
...
+                    .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+                sdf = sdf.withColumn(column_name, stat_func(F.col('rank')).over(window))
+
+            if na_option == 'keep':
+                sdf = sdf.withColumn(column_name,
+                                     F.when(sdf[column_name + '_tmp'].isNull(), F.lit(None))
+                                     .otherwise(sdf[column_name]))
+
+        return DataFrame(self._internal.copy(sdf=sdf.select(self._internal.columns)))\
+            .astype(np.float64)
```
HyukjinKwon: Why do we need to cast it, out of curiosity?

databricks/koalas/frame.py
```diff
@@ -5306,6 +5307,148 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
 
         return DataFrame(exploded_df)
 
+    # TODO: add axis, numeric_only, pct parameter
+    def rank(self, method='average', na_option='keep', ascending=True):
+        """
+        Compute numerical data ranks (1 through n) along axis. Equal values are
+        assigned a rank that is the average of the ranks of those values.
+
```
HyukjinKwon: Let's keep this note:

```
        .. note:: the current implementation of rank uses Spark's Window without
            specifying partition specification. This leads to move all data into
            single partition in single machine and could cause serious
            performance degradation. Avoid this method against very large dataset.
```

databricks/koalas/frame.py
```diff
@@ -5306,6 +5307,148 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
...
+            elif na_option == 'bottom':
+                if ascending:
+                    fill_value = sys.maxsize
+            elif na_option == 'keep':
+                if ascending:
+                    fill_value = sys.maxsize
+                else:
+                    fill_value = -sys.maxsize
+
+            if 'fill_value' in locals():
```
HyukjinKwon: We can just define `fill_value = None` above and check `fill_value is None`. Using `locals()` is dodgy because `locals()`'s return values can be changed given what vars we define within this func.

databricks/koalas/frame.py
```diff
@@ -5306,6 +5307,148 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
...
+                window = Window.orderBy(asc_func(column_name))\
+                    .rowsBetween(Window.unboundedPreceding, Window.currentRow)
+                sdf = sdf.withColumn(column_name, F.dense_rank().over(window))
+            else:
+                if method == 'average':
+                    stat_func = F.mean
+                elif method == 'min':
+                    stat_func = F.min
+                elif method == 'max':
+                    stat_func = F.max
```
HyukjinKwon: We can throw an error:

```python
else:
   raise ValueError(...)
```

databricks/koalas/frame.py
HyukjinKwon: Can we keep the index? Looks index will disappear after this API call.

databricks/koalas/frame.py
```diff
@@ -5306,6 +5307,148 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
...
+        3  3  1
+
+        >>> df.rank().sort_index()
+             A    B
+        0  1.0  4.0
+        1  2.5  3.0
+        2  2.5  2.0
+        3  4.0  1.0
+
+        >>> df.rank(method='min').sort_index()
```
HyukjinKwon: Let's add some short description for notable doctests like this. For instance, 

```
If method is set to 'min', blah blah blah ...

>>> df.rank(method='min').sort_index()
```

databricks/koalas/frame.py
RainFung: Is there any good way to fill a maximum value here instead?

databricks/koalas/frame.py
RainFung: some spark rank function will return int but pandas.rank() return float all the time.

databricks/koalas/frame.py
RainFung: Good suggestion.

databricks/koalas/frame.py
RainFung: But i throw error before
```
        if method not in ['average', 'min', 'max', 'first', 'dense']:
            msg = "method must be one of 'average', 'min', 'max', 'first', 'dense'"
            raise ValueError(msg)
```

databricks/koalas/frame.py
RainFung: self._internal.columns seems include index name. Why will index disappear? Can you guide me some?

databricks/koalas/frame.py
HyukjinKwon: oh, okie.

databricks/koalas/frame.py
HyukjinKwon: I misread. You're correct. This is a good trick :-)

databricks/koalas/frame.py
HyukjinKwon: I am not sure yet but I think we should avoid to use the "maximum value" concept. Let's think about another way to work around this.

RainFung: I think filling value is the best way. Can you help me figure how to set maximum to aviod types error?

HyukjinKwon: Can we avoid to add `na_option` parameter for now?

HyukjinKwon:
Three comments left:

https://github.com/databricks/koalas/pull/546#discussion_r301866667
https://github.com/databricks/koalas/pull/546#discussion_r301867383
https://github.com/databricks/koalas/pull/546#discussion_r301878205

Looks pretty good otherwise.

databricks/koalas/frame.py
RainFung: na_option removed

RainFung:
Fix all comment already. But meet unknown Travis CI errors.

HyukjinKwon:
I re-triggered. LGTM - I will double check later and clean up if needed.

RainFung:
Pass it. Can you merge?

HyukjinKwon:
Thanks, @RainFung. This will be included in the next release (in next week).

RainFung:
Thanks your kind and  professional guide~

PR #558, opened 2019-07-14
DataFrame.shift()

RainFung:


RainFung:
The `fill_value` parameter is added in `pandas 0.24`. So i cant the `pandas 0.23` Travis CI Test. How can i solve this problem..

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1040,3 +1040,15 @@ def test_rank(self):
...
             kdf.rank(method='nothing')
+
+    def test_shift(self):
+        pdf = pd.DataFrame({'Col1': [10, 20, 15, 30, 45],
+                            'Col2': [13, 23, 18, 33, 48],
+                            'Col3': [17, 27, 22, 37, 52]})
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(pdf.shift(3), kdf.shift(3).sort_index())
+        self.assert_eq(pdf.shift(periods=3, fill_value=0),
+                       kdf.shift(periods=3, fill_value=0).sort_index())
```
HyukjinKwon: @RainFung, you can take one of three usually in this case.

1. Check pandas version via `distutils.version.LooseVersion` and skip the test conditionally with `unittest.skipIf`.
2. Creates a string or pre-defined dataframe and compare with `kdf` here. For instance, you can do:
    ```python
    # Pandas 0.23.x does not have 'fill_value' parameter. Therefore, here we do a manual comparison.
    pdf = pd.DataFrame({'Col1': [0, 0, 0, 10, 20],
                        'Col2': [0, 0, 0, 13, 23],
                        'Col3': [0, 0, 0, 17, 27]})
    self.assert_eq(pdf, kdf.shift(periods=3, fill_value=0).sort_index())
    ```

3. Simply don't add this tests if it looks obviously fine to skip.

databricks/koalas/frame.py
```diff
@@ -1874,6 +1874,63 @@ def notnull(self):
...
+        3    10    13    17
+        4    20    23    27
+        """
+        if not isinstance(periods, int):
+            raise ValueError('Periods parameter only support int')
+
+        index = self._internal.index_columns[0]
+        data_columns = self._internal.data_columns
+        sdf = self._sdf
+        window = Window.orderBy(index)
```
HyukjinKwon: @RainFung, whenever we use Window, let's add the note https://github.com/databricks/koalas/blob/master/databricks/koalas/frame.py#L5319-L5322

databricks/koalas/frame.py
```diff
@@ -1874,6 +1874,67 @@ def notnull(self):
...
+    # TODO: add frep and axis parameter
+    def shift(self, periods=1, fill_value=None):
+        """
+        Shift index by desired number of periods.
+
+        When 'freq' is not passed, shift the index without realigning the data.
+        If 'freq' is passed (in this case, the index must be date or datetime, or it will raise
+        a NotImplementedError), the index will be increased using the periods and the 'freq'.
+
+        .. note:: the current implementation of rank uses Spark's Window without
```
HyukjinKwon: `rank` -> `shift`

RainFung:
fixed

HyukjinKwon:
@RainFung, let's add it to frame.rst for documentation.

HyukjinKwon:
Looks good. @RainFung, can you resolve conflicts?

HyukjinKwon:
I resolved the conflicts and pushed it at 86ca288640154d86ab25c607128ea024813353e5

PR #559, opened 2019-07-14
DataFrame.diff()

RainFung:


databricks/koalas/frame.py
```diff
@@ -1874,6 +1874,80 @@ def notnull(self):
...
+        3 -1.0 -2.0  -9.0
+        4 -1.0 -3.0 -11.0
+        5  NaN  NaN   NaN
+        """
+        if not isinstance(periods, int):
+            raise ValueError('periods only support int type')
+
+        data_columns = self._internal.data_columns
+        sdf = self._sdf
+        window = Window.partitionBy().orderBy(self._internal.index_columns[0])
```
HyukjinKwon: This will cause the computation whole partition in a single note. We should add the similar note to warn to users like we this for cummax, cumsum.

databricks/koalas/frame.py
```diff
@@ -1882,6 +1882,11 @@ def diff(self, periods=1):
         Calculates the difference of a DataFrame element compared with another element in the
         DataFrame(default is the element in the same column of the previous row).
 
+        .. note:: the current implementation of rank uses Spark's Window without
```
HyukjinKwon: `rank` -> `diff`

RainFung:
fix already

HyukjinKwon:
LGTM @RainFung can you resolve conflicts?

databricks/koalas/frame.py
```diff
@@ -1874,6 +1874,85 @@ def notnull(self):
...
+        3 -1.0 -2.0  -9.0
+        4 -1.0 -3.0 -11.0
+        5  NaN  NaN   NaN
+        """
+        if not isinstance(periods, int):
+            raise ValueError('periods only support int type')
+
+        data_columns = self._internal.data_columns
+        sdf = self._sdf
+        window = Window.partitionBy().orderBy(self._internal.index_columns[0])
```
HyukjinKwon: We can remove `partitionBy()`.

HyukjinKwon: Let's check index column

```
        if len(self._internal.index_columns) == 0:
            raise ValueError("Index must be set.")

        index_columns = self._internal.index_columns
```

HyukjinKwon: can we specify `rowsBetween` based on `period` as well to reduce the size of window?

HyukjinKwon:
I merged your commit at c7269ca42a0e185f0b131bc93ae3273d88565953

PR #563, opened 2019-07-16
Add Series.shift() (with minor cleanups)

HyukjinKwon:
This PR add shift() in Series too.


docs/source/reference/frame.rst
```diff
@@ -174,6 +174,16 @@ Combining / joining / merging
...
 
+Time series-related
+-------------------
+.. autosummary::
+   :toctree: api/
+
+   DataFrame.shift
+
+.. autosummary::
+   :toctree: api/
```
ueshin: Do we need this?

databricks/koalas/base.py
```diff
@@ -520,3 +520,66 @@ def any(self, axis: Union[int, str] = 0) -> bool:
...
+
+        Returns
+        -------
+        Copy of input Series/Index, shifted.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'Col1': [10, 20, 15, 30, 45],
+        ...                    'Col2': [13, 23, 18, 33, 48],
+        ...                    'Col3': [17, 27, 22, 37, 52]})
```
ueshin: This equivalent example in `DataFrame` needs `columns=[...]`. Could you fix it too?

HyukjinKwon:
Thanks, @ueshin!

databricks/koalas/base.py
ueshin: Could you also add it to https://github.com/databricks/koalas/blob/a248cc6788e294aebf03d9344429872512cd1db4/databricks/koalas/frame.py#L1901-L1904 as well?

HyukjinKwon: Yup!

PR #565, opened 2019-07-16
add method and limit parameter for DataFrame.fillna()

RainFung:
update DataFrame.fillna()

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2639,7 +2639,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             without specifying partition specification. This leads to move all data into
             single partition in single machine and could cause serious
             performance degradation. Avoid this method against very large dataset.
-        
+
         Parameters
         ----------
         value : scalar, dict, Series
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2631,10 +2631,15 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         else:
             raise NotImplementedError("dropna currently only works for axis=0 or axis='index'")
 
-    # TODO: add 'pad' for method parameter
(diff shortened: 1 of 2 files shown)


RainFung:
ok，i will finish it tomorrow

RainFung:
Can you code review and merge it first? Because I prefer to create new function in new branch.

databricks/koalas/frame.py
```diff
@@ -2690,19 +2715,45 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+                elif method in ['bfill', 'backfill']:
+                    func = F.first
+                    begin = Window.currentRow + 1
+                    if limit is not None:
+                        end = Window.currentRow + limit
+                    else:
+                        end = Window.unboundedFollowing
+                else:
+                    raise ValueError('Expecting pad, ffill, backfill or bfill.')
+                window = Window.orderBy(self._internal.index_columns[0]).rowsBetween(begin, end)
```
ueshin: Can we support multi-index or no index?
If we can, let's support while we're here, otherwise let's check the index size?

HyukjinKwon: Yea, I think supporting multi index isn't that difficult. @RainFung, can you try to fix `self._internal.index_columns[0]` -> `self._internal.index_columns` and add a test with multi index? If that doesn't work, I will make a followup to fix.

HyukjinKwon:
> Can you code review and merge this PR first? Because I prefer to create new function in new branch.

Okay, that's fine but can you address https://github.com/databricks/koalas/pull/565#discussion_r304169899?

databricks/koalas/frame.py
RainFung: fixed

RainFung:
Fixed

ueshin:
Thanks! merging.
@RainFung Could you submit the next PR to create the new functions?

PR #569, opened 2019-07-17
DataFrame.duplicated()

RainFung:


databricks/koalas/frame.py
```diff
@@ -2153,6 +2153,83 @@ def round(self, decimals=0):
...
+        Name: x, dtype: bool
+        """
+        if subset is None:
+            subset = self._internal.data_columns
+        else:
+            diff = set(subset).difference(set(self._internal.data_columns))
+            if len(diff) > 0:
+                raise KeyError(','.join(diff))
+        sdf = self._sdf
+        index = self._internal.index_columns[0]
```
ueshin: Can we support multi-index or no index?
If we can, let's address here, otherwise let's check the index length beforehand.

databricks/koalas/frame.py
```diff
@@ -2153,6 +2153,83 @@ def round(self, decimals=0):
...
+            tmp = sdf.groupby(subset).agg(F.max(index).alias(index)).withColumn('x', F.lit(False))
+            sdf = sdf.join(tmp, [index] + subset, 'left')
+        elif not keep:
+            tmp = sdf.groupby(subset).agg((F.count('*') > 1).alias('x'))
+            sdf = sdf.join(tmp, subset, 'left')
+        else:
+            raise ValueError("keep only support 'first', 'last' and False")
+
+        sdf = sdf.fillna(True, subset='x').select([index] + ['x'])
+        internal = _InternalFrame(sdf=sdf, data_columns=['x'], index_map=[(index, None)])
```
ueshin: How about naming `duplicated` or something instead of `x` to be more meaningful?

ueshin: As for the index name, I think we can preserve the name.

```py
>>> pd.DataFrame({'a': [1, 1, 1, 3], 'b': [1, 1, 1, 4], 'c': [1, 1, 1, 5]}).set_index('a').duplicated()
a
1    False
1     True
1     True
3    False
dtype: bool
```

whereas:

```py
>>> ks.DataFrame({'a': [1, 1, 1, 3], 'b': [1, 1, 1, 4], 'c': [1, 1, 1, 5]}).set_index('a').duplicated()
1    False
1    False
1    False
3    False
Name: x, dtype: bool
```

databricks/koalas/frame.py
```diff
@@ -2153,6 +2153,83 @@ def round(self, decimals=0):
...
+        kdf = DataFrame(internal)
+        return kdf[kdf.columns[0]]
```
ueshin: Using `series._col` might be better, just for consistency.

databricks/koalas/frame.py
RainFung: Because ks.Series set name 'x' automatic, i set 'x' here.

databricks/koalas/frame.py
RainFung: Hello. I don't know when is there no index.  It create __index_level_0__  index anyway .Can you give me example? Thanks.

databricks/koalas/frame.py
RainFung: We can't support index now.

databricks/koalas/frame.py
```diff
@@ -2142,6 +2141,93 @@ def round(self, decimals=0):
...
+        Mark all duplicates as ``True``.
+
+        >>> df.duplicated(keep=False).sort_index()
+        0     True
+        1     True
+        2     True
+        3    False
+        Name: duplicated, dtype: bool
+        """
+        if len(self._internal.index_names) > 1:
```
ueshin: `len(self._internal.index_names) != 1` instead of `>`?

databricks/koalas/frame.py
ueshin: The index name can be preserved with `self._internal.index_names[0]`?

RainFung: Hello, ueshin. We can't support index now, so I think there is no need to preserved name. Can we just name it to 'x' stead?

ueshin: ah, I see what you thought, but currently we DO try to support a single-index.
If we create a DataFrame using a pandas-like constructor or from pandas, then we implicitly have a single-index.
On the other hand, we can explicitly set index by `set_index`, `reindex` or others.
We need to make both cases work here.

HyukjinKwon:
gentle ping @RainFung to update.

RainFung:
I'll update it today.

RainFung:
```python
>>> pd.DataFrame({'a': [1, 1, 1, 3], 'b': [1, 1, 1, 4], 'c': [1, 1, 1, 5]}).set_index('a').duplicated()
a
1    False
1     True
1     True
3    False
dtype: bool
``` 
@ueshin. Hello usshin. Like this example, if we set 'a' index in spark, how do i know which is first duplicated row in the three same 1 value in 'a' index?

databricks/koalas/frame.py
```diff
@@ -2142,6 +2141,93 @@ def round(self, decimals=0):
...
+    def duplicated(self, subset=None, keep='first'):
+        """
+        Return boolean Series denoting duplicate rows, optionally only considering certain columns.
+
+        Parameters
+        ----------
+        subset : column label or sequence of labels, optional
+            Only consider certain columns for identifying duplicates,
+            by default use all of the columns
+        keep : {'first', 'last', False}, default 'first'
```
HyukjinKwon: Let's remove this option for now.

databricks/koalas/frame.py
```diff
@@ -2142,6 +2141,93 @@ def round(self, decimals=0):
...
+            if len(diff) > 0:
+                raise KeyError(','.join(diff))
+
+        sdf = self._sdf
+        index = self._internal.index_columns[0]
+
+        if keep == 'first':
+            tmp = sdf.groupby(subset).agg(F.min(index).alias(index)).withColumn('duplicated',
+                                                                                F.lit(False))
+            sdf = sdf.join(tmp, [index] + subset, 'left')
```
HyukjinKwon: @RainFung, why do we need to join?  Can't we just do it by Window function with partitioning columns?

databricks/koalas/frame.py
```diff
@@ -2142,6 +2141,93 @@ def round(self, decimals=0):
...
+        1  1  1  1
+        2  1  1  1
+        3  3  4  5
+
+        >>> df.duplicated().sort_index()
+        0    False
+        1     True
+        2     True
+        3    False
+        Name: duplicated, dtype: bool
```
HyukjinKwon: you can make the return column name as `self.name`. So it keeps the original name

databricks/koalas/frame.py
```diff
@@ -2142,6 +2141,93 @@ def round(self, decimals=0):
...
+        elif not keep:
+            tmp = sdf.groupby(subset).agg((F.count('*') > 1).alias('duplicated'))
+            sdf = sdf.join(tmp, subset, 'left')
+        else:
+            raise ValueError("keep only support 'first', 'last' and False")
+
+        sdf = sdf.fillna(True, subset='duplicated').select([index] + ['duplicated'])
+        internal = _InternalFrame(sdf=sdf, data_columns=['duplicated'], index_map=[(index, None)])
+        kdf = DataFrame(internal)
+        from databricks.koalas.series import _col
```
HyukjinKwon: Import this on the top at this method.

HyukjinKwon:
https://github.com/databricks/koalas/pull/569#issuecomment-518004672, sorry if I missed the context. Do you need it to implement `keep`? If you don't know, we can don't implement that option for now.

RainFung:
Fixed

databricks/koalas/frame.py
```diff
@@ -2266,6 +2265,98 @@ def round(self, decimals=0):
...
+        >>> df.duplicated(keep=False).sort_index()
+        0     True
+        1     True
+        2     True
+        3    False
+        Name: duplicated, dtype: bool
+        """
+        from databricks.koalas.series import _col
+        if len(self._internal.index_names) > 1:
+            raise ValueError("Now we don't support Index Now.")
```
HyukjinKwon: `Now we don't support Index Now` it sounds weird. Do you mean this method does not support multi-index for now?

databricks/koalas/frame.py
```diff
@@ -2266,6 +2265,98 @@ def round(self, decimals=0):
...
+            diff = set(subset).difference(set(self._internal.data_columns))
+            if len(diff) > 0:
+                raise KeyError(','.join(diff))
+
+        sdf = self._sdf
+        index = self._internal.index_columns[0]
+        if self._internal.index_names[0] is not None:
+            name = self._internal.index_names[0]
+        else:
+            name = 'duplicated'
```
HyukjinKwon: Can you name it to `"0"`? That's the default name we get from `ks.Series`.

databricks/koalas/frame.py
```diff
@@ -2266,6 +2265,98 @@ def round(self, decimals=0):
...
+                ord_func = spark.functions.desc
+            window = Window.partitionBy(subset).orderBy(ord_func(index)).rowsBetween(
+                Window.unboundedPreceding, Window.currentRow)
+            sdf = sdf.withColumn(name, F.row_number().over(window) > 1)
+        elif not keep:
+            window = Window.partitionBy(subset).orderBy(F.col(index).desc())\
+                .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+            sdf = sdf.withColumn(name, F.count(F.col(index)).over(window) > 1)
+        else:
+            raise ValueError("keep only support 'first', 'last' and False")
```
HyukjinKwon: ''keep' currently only supports 'first', 'last' and False.'

databricks/koalas/frame.py
```diff
@@ -50,7 +50,6 @@
 from databricks.koalas.utils import scol_for
 from databricks.koalas.typedef import as_spark_type
 
-
```
HyukjinKwon: @RainFung, let's avoid to add unrelated changes next time. It makes backporting or reverting more difficult in the future.

databricks/koalas/frame.py
```diff
@@ -2266,6 +2265,98 @@ def round(self, decimals=0):
...
+        from databricks.koalas.series import _col
+        if len(self._internal.index_names) > 1:
+            raise ValueError("Now we don't support Index Now.")
+
+        if subset is None:
+            subset = self._internal.data_columns
+        else:
+            diff = set(subset).difference(set(self._internal.data_columns))
+            if len(diff) > 0:
+                raise KeyError(','.join(diff))
```
HyukjinKwon: `','.join(diff)` -> `', '.join(diff)`

databricks/koalas/frame.py
```diff
@@ -2266,6 +2265,98 @@ def round(self, decimals=0):
...
+        2     True
+        3    False
+        Name: duplicated, dtype: bool
+        """
+        from databricks.koalas.series import _col
+        if len(self._internal.index_names) > 1:
+            raise ValueError("Now we don't support Index Now.")
+
+        if subset is None:
+            subset = self._internal.data_columns
```
HyukjinKwon: can you just name `subset` to `group_key` or `group_columns`? subset seems confusing a bit.

HyukjinKwon:
Hey~ @RainFung, don't fold and mark my comments as resolved when they are unresolved ..

RainFung:
> Hey~ @RainFung, don't fold and mark my comments as resolved when they are unresolved ..

Sorry, i'm fixing comments one by one.

HyukjinKwon:
ah, I see. Thanks.

databricks/koalas/frame.py
```diff
@@ -2266,6 +2265,98 @@ def round(self, decimals=0):
...
+                Window.unboundedPreceding, Window.currentRow)
+            sdf = sdf.withColumn(name, F.row_number().over(window) > 1)
+        elif not keep:
+            window = Window.partitionBy(subset).orderBy(F.col(index).desc())\
+                .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+            sdf = sdf.withColumn(name, F.count(F.col(index)).over(window) > 1)
+        else:
+            raise ValueError("keep only support 'first', 'last' and False")
+
+        internal = _InternalFrame(sdf=sdf.select([index] + [name]), data_columns=[name],
```
HyukjinKwon: `.select([index] + [name])` -> `.select(index, name)`

HyukjinKwon:
Looks good otherwise.

RainFung:
Fixed

RainFung:
codecov seems not work

databricks/koalas/frame.py
HyukjinKwon: @RainFung, please don't resolve a comment when it's unresolved.

HyukjinKwon:
Thanks @RainFung for working on this.

PR #570, opened 2019-07-18
DataFrame.bfill() base on DataFrame.fillna()

RainFung:


HyukjinKwon:
@RainFung, can you resolve the conflicts please?

RainFung:
ok

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1106,7 +1106,7 @@ def test_diff(self):
         msg = "should be an int"
         with self.assertRaisesRegex(ValueError, msg):
             kdf.diff(1.5)
-
+
```
HyukjinKwon: Seems like we should remove this.

RainFung:
fixed

Issue #405, opened 2019-05-30
Series.fillna does not anchor


thunterdb:
Example to reproduce:

```py
df = ks.DataFrame({"x":[1, None]})
df["x"] = df["x"].fillna(2)
df
```
leads to:
`AssertionError: Cannot combine column argument because it comes from a different dataframe`

HyukjinKwon:
This is actually bigger problem than I thought. `fillna` requires to make another DataFrame because the corresponding API is existent in Spark DataFrame (not column).

HyukjinKwon:
Hm .. we should maybe try to write all the series APIs only via expressions but I wonder how far it's possible. For instance, `unique` returns the different number of values.

garawalid:
@HyukjinKwon, I agree. I've been investigating the issue, and I couldn't find a simple solution without changing the series API.

HyukjinKwon:
Let me leave this as won't fix at this moment. There is no way to use the same DataFrame by the limitation of Spark API.

PR #575, opened 2019-07-19
Avoid to use to_dataframe in Series.rank to keep anchor

RainFung:
This PR proposes to avoid to use to_dataframe in Series.rank to keep anchor. Currently. `Series.rank` has https://github.com/databricks/koalas/issues/405 issue as well because it converts Spark column into DataFrame within Series API.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1210,7 +1210,6 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         5    6.0
         Name: x, dtype: float64
         """
-
         kseries = _col(self.to_dataframe().fillna(value=value, axis=axis, inplace=False))
         if inplace:
             self._internal = kseries._internal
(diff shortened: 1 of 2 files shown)


gatorsmile:
cc @HyukjinKwon @ueshin

databricks/koalas/series.py
```diff
@@ -2412,7 +2411,43 @@ def rank(self, method='average', ascending=True):
...
+                stat_func = F.min
+            elif method == 'max':
+                stat_func = F.max
+            window1 = Window.orderBy(asc_func(column_name))\
+                .rowsBetween(Window.unboundedPreceding, Window.currentRow)
+            window2 = Window.partitionBy(column_name)\
+                .rowsBetween(Window.unboundedPreceding, Window.unboundedFollowing)
+            scol = stat_func(F.row_number().over(window1)).over(window2)
+        return Series(self._kdf._internal.copy(scol=scol), anchor=self._kdf).rename(column_name)\
+            .astype(np.float64)
```
HyukjinKwon: Ditto, can we deduplicate the codes with `DataFrame.rank`? We can call each column, for instnace,

```scala
        applied = []
        for column in self._internal.data_columns:
            applied.append(self[column].rank(...))

        sdf = self._sdf.select(
            self._internal.index_scols + [c._scol for c in applied])
        internal = self._internal.copy(sdf=sdf, data_columns=[c.name for c in applied])
        return DataFrame(internal)
```
```

HyukjinKwon:
It's great that we avoid `to_dataframe` and we keeps the anchor :-).

HyukjinKwon:
@RainFung, let's be explicit about PR title and description next time. I thought you were proposing Series.rank API. I will fix it for this time.

PR #577, opened 2019-07-19
Set the upper-bound of pandas version.

ueshin:
Pandas 0.25 was released (https://pypi.org/project/pandas/).
We tried to test with pandas 0.25, but there are failures: https://travis-ci.com/databricks/koalas/jobs/217695045.


ueshin:
Seems like the fix for supporting pandas 0.25 is not trivial for me.
Let me set the upper-bound of the pandas version for now.
Track for pandas 0.25 at #578.

ueshin:
Thanks! merging.

PR #581, opened 2019-07-22
Add Index.rename() with related bug fixes

HyukjinKwon:
This PR mainly proposes to add `Index.rename` but fixes multiple related bugs:

1. Most of APIs in `Index` don't work because it doesn't have `alias` API. Now we have `rename` API in Index so `alias` occurrences are replaced to `rename`.
2. `to_pandas` and `repr` in `Index` don't reflect its' `_scol` but always return the initial column as is. This is because `_scol` is lost when we select `self._kdf[[]]` within `to_pandas`. It is fixed together.
3. `spark_type` in `Index` has a mistake:

    ```diff
    - return self.to_series().dataType
    + return self.to_series().spark_type
    ```
4. `rename` in  `Series` does not copy when `index` is not given. Now it copies when `inplace` is `False`.


databricks/koalas/base.py
```diff
@@ -676,4 +676,4 @@ def shift(self, periods=1, fill_value=None):
             shifted_col.isNull() | F.isnan(shifted_col), fill_value
         ).otherwise(shifted_col)
 
-        return self._with_new_scol(col).alias(self.name)
+        return self._with_new_scol(col).rename(self.name)
```
HyukjinKwon: Seems like this is not enough to make this working with Index. I am going to investigate and make another fix to make all of those work with Index as well.

HyukjinKwon:
Tests passed

HyukjinKwon:
Spark 2.3.3 doesn't seem supporting nested types in Arrow, which makes the test failures .. let me check it tomorrow and update.

databricks/koalas/indexes.py
```diff
@@ -124,6 +129,56 @@ def names(self, names: List[str]) -> None:
...
+        >>> df  # doctest: +NORMALIZE_WHITESPACE
+           b
+        e
+        A  A
+        C  B
+        """
+        index_columns = self._kdf._internal.index_columns
+        assert len(index_columns) == 1
+
+        sdf = self._kdf._sdf.select([self._scol] + self._kdf._internal.data_columns)
```
ueshin: `data_scols` instead of `data_columns`?

databricks/koalas/series.py
```diff
@@ -827,8 +827,9 @@ def rename(self, index=None, **kwargs):
         Name: my_name, dtype: int64
         """
         if index is None:
-            return self
-        scol = self._scol.alias(index)
+            scol = self._scol
+        else:
+            scol = self._scol.alias(index)
```
ueshin: nit: `rename` instead of `alias`?

HyukjinKwon: It's a Spark column :-) ..

HyukjinKwon:
Seems like somehow JDK 11 is being picked in Travis ... let me update soon.

databricks/koalas/series.py
ueshin: oh :)

ueshin:
Thanks! merging.

PR #582, opened 2019-07-22
Try conda 4.7.

### LINKED ISSUE #508: Upgrade conda to 4.7 in travis-ci.
When upgrading `conda` to `4.7.5`, an unknown error occurs.
See: https://travis-ci.com/databricks/koalas/jobs/210984050

ueshin:
Seems like conda is trying to deploy 4.7.10 as a default.
It's time to revisit the conda 4.7 build.
Resolves #508.


ueshin:
The behavior related to pinned packages seems changed.
Even if we pin a package, we can still install the explicitly specified version of the package.

PR #589, opened 2019-07-24
Add DataFrame.filter()

HyukjinKwon:
This PR adds DataFrame.filter()


databricks/koalas/frame.py
```diff
@@ -5943,6 +5943,122 @@ def rank(self, method='average', ascending=True):
...
+
+        if items is not None:
+            if is_list_like(items):
+                items = list(items)
+            else:
+                raise ValueError("items should be a list-like object.")
+            if axis in ('index', 0):
+                # TODO: support multi-index here
+                if len(index_scols) > 1:
+                    raise ValueError("Multiindex is not supported yet.")
```
ueshin: We should also check no-index case.

PR #590, opened 2019-07-25
Add basic multi-level column support.

ueshin:


databricks/koalas/frame.py
```diff
@@ -6148,7 +6186,13 @@ def __getattr__(self, key: str) -> Any:
                 return property_or_func.fget(self)  # type: ignore
             else:
                 return partial(property_or_func, self)
-        if key not in self.columns:
+
+        if self._internal.column_index is not None:
+            try:
+                return self._get_from_multilevel_column(key)
+            except KeyError:
```
HyukjinKwon: Hmmm .. I didn't get why we ignore `KeyError` here. Seems like it's going to access the attr by it's internal name:

```python
>>> getattr(ks.DataFrame(pdf), "('x', 'a', '1')")
0    1
1    2
3    3
Name: ('x', 'a', '1'), dtype: int64
```

but not via:

```python
>>> kdf["('x', 'a', '1')"]
  File "/.../koalas/databricks/koalas/frame.py", line 6160, in __getitem__
    return self._pd_getitem(key)
  File "/.../koalas/databricks/koalas/frame.py", line 6102, in _pd_getitem
    return self._get_from_multilevel_column(key)
  File "/.../koalas/databricks/koalas/frame.py", line 6079, in _get_from_multilevel_column
    raise KeyError(key)
KeyError: "('x', 'a', '1')"
```

PR #595, opened 2019-07-26
Fix merge for how='right' or how='outer'.

### LINKED ISSUE #551: Unexpected behavior in ks.merge how='outer'
Hi,

I have been trying to perform a outer join between two koalas dataframes on basis of a column, but am getting very unexpected results. Below is an example,

`left_kdf = ks.DataFrame({'A': [1, 2]})`
`right_kdf = ks.DataFrame({'B': ['x', 'y', 'z'], 'A':[3, 4, 1]})`
`left_kdf.merge(right_kdf, on=["A"], how='outer')`

The output that I get is 

image

However, if this is a SQL style join as said in the documentation, the output should be something like,
 
image

The None part (in Column B) is discussed in [#263](https://github.com/databricks/koalas/issues/263)

What concerns me are the missing values in column A.

ueshin:
Resolves #551.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4659,7 +4659,13 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             scol = scol_for(left_table, col)
             if col in duplicate_columns:
                 if col in left_keys and col in right_keys:
-                    pass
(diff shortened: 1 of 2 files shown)


ueshin:
cc @HyukjinKwon @shreyanshu

PR #597, opened 2019-07-26
Fix multi-index column access.

ueshin:
Fix complex multi-index column access for something like:

```py
>>> pdf
   a  c  e              i
             g
      d  f
   b              h
0  1  a  x  10  100  1000
1  2  b  y  20  200  2000
2  3  c  z  30  300  3000

>>> pdf['a']
   b
0  1
1  2
2  3

>>> pdf['c']
   d

0  a
1  b
2  c

>>> pdf['e']
       g
   f

0  x  10
1  y  20
2  z  30

>>> pdf['']
     h
0  100
1  200
2  300

>>> pdf['i']
0    1000
1    2000
2    3000
Name: i, dtype: int64
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -6172,17 +6172,21 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             sdf = self._sdf.select(self._internal.index_scols +
                                    [self._internal.scol_for(col).alias(idx[0])
                                     for col, idx in columns])
-            return DataFrame(self._internal.copy(
+            kdf = DataFrame(self._internal.copy(
                 sdf=sdf,
                 data_columns=[idx[0] for _, idx in columns],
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -6172,25 +6172,29 @@ def _get_from_multilevel_column(self, key):
...
                                    [self._internal.scol_for(col) for col, _ in columns])
-            return DataFrame(self._internal.copy(
+            kdf = DataFrame(self._internal.copy(
                 sdf=sdf,
                 data_columns=[col for col, _ in columns],
                 column_index=[idx for _, idx in columns]))
+        if all([idx[0] == '' for _, idx in columns]):
+            return kdf._pd_getitem('')
+        else:
+            return kdf
```
ueshin: Actual fix is here.

PR #601, opened 2019-07-28
Fix tests for stats.

ueshin:
I might miss the context, but the tests for stats seem strange?
They should be as this PR?


ueshin:
cc @nitlev @HyukjinKwon

PR #605, opened 2019-07-29
Add axis=columns to count, var, std, max, sum, min, kurtosis, skew and mean in DataFrame

HyukjinKwon:
This PR proposes to suppose axis=columns or 1 to count, var, std, max, sum, min, kurtosis, skew and mean in DataFrame. It was implemented via using pandas UDF.

Note that, to infer the return type, it executes the first 1000 and get the return type from it.

  - If the results is less then 1000, it just uses pandas API as a shortcut
  - Otherwise, it is executed via pandas UDF.

Resolves #599

PR #607, opened 2019-07-31
Add axis=columns to cumsum, cummin, cummax and cumprod in DataFrame

HyukjinKwon:
This PR proposes to add axis=columns to cumsum, cummin, cummax and cumprod in DataFrame.
The approach is similar with https://github.com/databricks/koalas/pull/605 (and therefore this PR tries to deduplicate it).

*This requires Spark 3.0+**

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -36,7 +36,8 @@ from pandas.core.dtypes.inference import is_sequence
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Column
 from pyspark.sql.types import (BooleanType, ByteType, DecimalType, DoubleType, FloatType,
-                               IntegerType, LongType, NumericType, ShortType, StructType)
+                               IntegerType, LongType, NumericType, ShortType, StructType,
+                               StructField)
 from pyspark.sql.utils import AnalysisException
 from pyspark.sql.window import Window
 from pyspark.sql.functions import pandas_udf
@@ -324,7 +325,6 @@ class DataFrame(_Frame, Generic[T]):
             If True, sfun is applied on numeric columns (including booleans) only.
         """
         from inspect import signature
-        from databricks.koalas import Series
 
         if axis in ('index', 0, None):
             exprs = []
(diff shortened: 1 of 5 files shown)


databricks/koalas/tests/test_dataframe.py
```diff
@@ -1079,27 +1082,79 @@ def test_cummin(self):
         self.assert_eq(pdf.cummin(), kdf.cummin())
         self.assert_eq(pdf.cummin(skipna=False), kdf.cummin(skipna=False))
 
+    @unittest.skipIf(
+        LooseVersion(pyspark_version) < LooseVersion('3.0'),
+        "columns axis is supported in Spark 3.0+.")
```
HyukjinKwon: Ah .. I forgot that struct as return type was only added lately as of https://github.com/apache/spark/commit/ddc2052ebd247aa2a8dad34fd5c1cd345fa45118#diff-c1cf83efe5a5b1f2f1f770589988e997 and https://issues.apache.org/jira/browse/SPARK-23836.

I tested all of them against Spark 3.0.

HyukjinKwon:
Tests passed.

databricks/koalas/frame.py
```diff
@@ -361,23 +361,63 @@ def _reduce_for_stat_function(self, sfun, name, axis=None, numeric_only=False):
...
 
-            @pandas_udf(returnType=as_spark_type(pser.dtype.type))
-            def calculate_columns_axis(*cols):
-                return getattr(pd.concat(cols, axis=1), name)(axis=axis, numeric_only=numeric_only)
+        else:
+            raise ValueError("No axis named %s for object type %s." % (axis, type(axis)))
 
-            df = self._sdf.select(calculate_columns_axis(*self._internal.data_scols).alias("0"))
-            return DataFrame(df)["0"]
+    def _compute_columns_axis(self, name, axis, *args, **kwargs):
```
ueshin: Shall we check the pyspark version and raise an exception if the version is old?

HyukjinKwon:
Yes .. I agree with that part too. It's too early to merge this. Maybe I should avoid to document those and raise different error messages. Let me fix it soon.

HyukjinKwon:
Actually, let me reopen this after Spark 3.0 comes out.

PR #576, opened 2019-07-19
Series.fillna()

RainFung:
remove to_dataframe() in fillna

PR #618, opened 2019-08-02
fillna function rewrite

RainFung:
create a new PR base on https://github.com/databricks/koalas/pull/576

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1248,6 +1248,11 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         4       c
         Name: x, dtype: object
         """
+        if self.isnull().sum() == 0:
+            if inplace:
+                return
+            else:
+                return self
         if axis is None:
             axis = 0
         if not (axis == 0 or axis == "index"):
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -30,7 +30,7 @@ from pandas.core.accessor import CachedAccessor
 
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Column
-from pyspark.sql.types import BooleanType, StructType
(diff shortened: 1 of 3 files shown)


RainFung:
@ueshin `If the integer series doesn't have null values, it preserves the integer type.`
```python
>>> pd.Series([1,2,3,4]).fillna(method='ffill')
0    1
1    2
2    3
3    4
dtype: int64
```

Can you help me solve this problem?

HyukjinKwon:
cc @ueshin

databricks/koalas/series.py
```diff
@@ -1210,8 +1227,72 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+                    end = Window.unboundedFollowing
+            else:
+                raise ValueError('Expecting pad, ffill, backfill or bfill.')
+            window = Window.orderBy(self._internal.index_scols).rowsBetween(begin, end)
+            scol = F.when(scol.isNull(), func(scol, True).over(window)).otherwise(scol)
+
+        kseries = Series(self._kdf._internal.copy(scol=scol), anchor=self._kdf)\
+            .rename(column_name)
+        if isinstance(spark_type, NumericType):
+            kseries = kseries.astype(np.float64)
```
ueshin: I'm wondering why we need this cast?

databricks/koalas/series.py
```diff
@@ -1210,8 +1227,72 @@ def fillna(self, value=None, axis=None, inplace=False):
...
+        3       c
+        4       c
+        Name: x, dtype: object
         """
-        kseries = _col(self.to_dataframe().fillna(value=value, axis=axis, inplace=False))
+        if self.isnull().sum() == 0:
+            if inplace:
+                return
+            else:
+                return self
```
ueshin: Maybe we should return a copy to avoid the effect of the following inplace-like operations?

RainFung:
Fixed

PR #623, opened 2019-08-05
Add size in Index

HyukjinKwon:
This PR adds size property in Index.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -75,6 +75,21 @@ class Index(IndexOpsMixin):
         """
         return Index(self._kdf, scol)
 
+    @property
+    def size(self) -> int:
+        """
+        Return an int representing the number of elements in this object.
+
+        Examples
+        --------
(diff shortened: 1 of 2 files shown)


databricks/koalas/indexes.py
```diff
@@ -261,6 +276,21 @@ def any(self, *args, **kwargs):
     def all(self, *args, **kwargs):
         raise TypeError("cannot perform all with this index type: MultiIndex")
 
+    @property
+    def size(self) -> int:
```
HyukjinKwon: DataFrame and Series have it in `general.py`. So, it we add it to `base.py`, then it's overlapped with Series's .. so I had to add it in each.

ueshin: I don't think we need this here except for the example in the doctest, need only in `Index` since `MultiIndex` extends `Index`.

HyukjinKwon: Oops, yes. Argh, I rushed.

Issue #634, opened 2019-08-10
sort_index is not working with MultiIndex


garawalid:
`DataFrame.sort_index()` is not working when the DataFrame uses MultiIndex.  

Reproduce the error: 
```python
import pandas as pd
import databricks.koalas as ks

pdf = pd.DataFrame({
    ('x', 'a', '1'): [1, 2, 3],
    ('x', 'b', '2'): [4, 5, 6],
    ('y.z', 'c.d', '3'): [7, 8, 9],
    ('x', 'b', '4'): [10, 11, 12],
}, index=[0, 1, 3])

kdf = ks.from_pandas(pdf)

kdf.sort_index()
```

```
Traceback (most recent call last):
  File "/home/xxx/databricks/koalas/tests/bug.py", line 15, in <module>
    kdf.sort_index()
  File "/home/xxx/databricks/koalas/frame.py", line 4295, in sort_index
    inplace=inplace, na_position=na_position)
  File "/home/xxx/databricks/koalas/frame.py", line 4191, in sort_values
    for colname, asc in zip(by, ascending)]
  File "/home/xxx/databricks/koalas/frame.py", line 4191, in <listcomp>
    for colname, asc in zip(by, ascending)]
  File "/home/xxx/databricks/koalas/frame.py", line 6343, in __getitem__
    return self._pd_getitem(key)
  File "/home/xxx/databricks/koalas/frame.py", line 6280, in _pd_getitem
    return self._get_from_multiindex_column((key,))
  File "/home/xxx/databricks/koalas/frame.py", line 6235, in _get_from_multiindex_column
    raise KeyError(k)
KeyError: '__index_level_0__'
```

garawalid:
cc @ueshin

ueshin:
Thanks for the report. I'll work on this.

NA-Dev:
I am sorting a dataframe with multiIndex where one is a string, and one is a date. The sorting is not working properly, as I can see dates are still out of order. Python 3.9 Pandas 1.4.2

![Screenshot 2022-08-16 124124](https://user-images.githubusercontent.com/31260131/184944207-14184056-9eae-46dc-8590-f7a0f8030eaa.png)

itholic:
@NA-Dev Would you mind filing a ticket to [Apache Spark JIRA](https://issues.apache.org/jira/projects/SPARK/issues) ??

[...]

PR #635, opened 2019-08-11
Pivot table multiindex

### LINKED ISSUE #511: Support multi-level columns in pivot_table
This issue tracks `DataFrame.pivot_table` (#386)
The actual implementation can handle `values` as a list but it's disabled right now with this condition :
```python
if isinstance(values, list) and len(values) > 1:
    raise NotImplementedError('Values as list of columns is not implemented yet.')
``` 
We need to implement multi-level columns before updating `pivot_table`

garawalid:
Resolves #511. 
In the test, the `kdf` is converted to Pandas DataFrame in order to use `sort_index()`. I'll update the test once #634 resolved.


databricks/koalas/frame.py
```diff
@@ -3521,28 +3520,44 @@ def pivot_table(self, values=None, index=None, columns=None,
...
+            raise ValueError("Columns should be string.")
 
         if not isinstance(values, str) and not isinstance(values, list):
-            raise ValueError('values should be string or list of one column.')
+            raise ValueError('Values should be string or list of one column.')
 
         if not isinstance(aggfunc, str) and (not isinstance(aggfunc, dict) or not all(
                 isinstance(key, str) and isinstance(value, str) for key, value in aggfunc.items())):
-            raise ValueError("aggfunc must be a dict mapping from column name (string) "
+            raise ValueError("Aggfunc must be a dict mapping from column name (string) "
```
HyukjinKwon: I thought this refers `aggfunc` in parameters.

databricks/koalas/frame.py
```diff
@@ -3521,28 +3520,44 @@ def pivot_table(self, values=None, index=None, columns=None,
...
                              "to aggregate functions (string).")
 
         if isinstance(aggfunc, dict) and index is None:
-            raise NotImplementedError("pivot_table doesn't support aggfunc"
+            raise NotImplementedError("Pivot_table doesn't support aggfunc"
```
HyukjinKwon: `pivot_table` as was might make more sense ..

databricks/koalas/frame.py
```diff
@@ -3557,20 +3572,29 @@ def pivot_table(self, values=None, index=None, columns=None,
         elif isinstance(index, list):
             sdf = self._sdf.groupBy(index).pivot(pivot_col=columns).agg(*agg_cols)
         else:
-            raise ValueError("index should be a None or a list of columns.")
+            raise ValueError("Index should be a None or a list of columns.")
```
HyukjinKwon: ditto..

databricks/koalas/frame.py
```diff
@@ -3557,20 +3572,29 @@ def pivot_table(self, values=None, index=None, columns=None,
...
             sdf = sdf.fillna(fill_value)
 
         if index is not None:
-            return DataFrame(sdf).set_index(index)
-        else:
             if isinstance(values, list):
-                index_values = values[-1]
+                if len(values) == 2:
+                    data_columns = [column for column in sdf.columns if column not in index]
+                    data_columns.sort(key=lambda x: x.split('_')[1])
```
HyukjinKwon: nit: `x.split('_')` -> `x.split('_', 1)`

HyukjinKwon: I think this won't work when the function has `_` in its name. can we do `x.split('_', 1)[1]`

databricks/koalas/frame.py
```diff
@@ -3557,20 +3572,29 @@ def pivot_table(self, values=None, index=None, columns=None,
...
 
         if fill_value is not None and isinstance(fill_value, (int, float)):
             sdf = sdf.fillna(fill_value)
 
         if index is not None:
-            return DataFrame(sdf).set_index(index)
-        else:
             if isinstance(values, list):
-                index_values = values[-1]
+                if len(values) == 2:
```
HyukjinKwon: Can we add some comments why we should split by `_`? It seems like a bit difficult to understand if you don't know what pivot returns in its column names.

HyukjinKwon:
Seems fine otherwise. cc @ueshin

garawalid:
@HyukjinKwon thanks for the review. I'll address your comments after merging #637.

databricks/koalas/frame.py
```diff
@@ -3521,28 +3520,44 @@ def pivot_table(self, values=None, index=None, columns=None,
         C
         small  5.5  2.333333
         large  5.5  2.000000
+
+        The next example aggregates on multiple values.
+
+        >>> table = df.pivot_table(index=['C'], columns="A", values=['B', 'E'],
+        ...                         aggfunc={'B': 'mean', 'E': 'sum'})
```
ueshin: We should use `D` instead of `B` since `B` is a string column? it won't calculate anything.

BTW, pandas raises an error for the case:

```py
>>> pdf.pivot_table(index=['C'], columns="A", values=['B', 'E'], aggfunc={'B': 'mean', 'E': 'sum'})
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/frame.py", line 6067, in pivot_table
    observed=observed,
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/reshape/pivot.py", line 96, in pivot_table
    agged = grouped.agg(aggfunc)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/groupby/generic.py", line 1455, in aggregate
    return super().aggregate(arg, *args, **kwargs)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/groupby/generic.py", line 229, in aggregate
    result, how = self._aggregate(func, _level=_level, *args, **kwargs)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/base.py", line 506, in _aggregate
    result = _agg(arg, _agg_1dim)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/base.py", line 456, in _agg
    result[fname] = func(fname, agg_how)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/base.py", line 440, in _agg_1dim
    return colg.aggregate(how, _level=(_level or 0) + 1)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/groupby/generic.py", line 845, in aggregate
    return getattr(self, func_or_funcs)(*args, **kwargs)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 1205, in mean
    "mean", alt=lambda x, axis: Series(x).mean(**kwargs), **kwargs
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6_pd0.25/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 888, in _cython_agg_general
    raise DataError("No numeric types to aggregate")
pandas.core.base.DataError: No numeric types to aggregate
```

whereas:

```py
>>> pdf.pivot_table(index=['C'], columns="A", values=['D', 'E'], aggfunc={'D': 'mean', 'E': 'sum'})
         D             E
A      bar       foo bar foo
C
large  5.5  2.000000  15   9
small  5.5  2.333333  17  13
```

Should we follow the behavior and raise an error? cc @HyukjinKwon

HyukjinKwon: Yea, +1 to match.

garawalid: @ueshin Nice catch!  
I agree we should raise the same error!

garawalid:
@ueshin Sure! 
I added column name index and I skipped the docstring test because it fails.

databricks/koalas/frame.py
garawalid: @HyukjinKwon 
I added some comments, are they clear?

databricks/koalas/frame.py
garawalid: Thanks for the trick!

HyukjinKwon:
To me, it seems fine in general but @ueshin has worked on indexing stuff more than I do.. so I will leave it to him.

ueshin:
The reason why the doctests fail is you don't set the column index names.
Now that we always use `column_index` for even single index columns, and we can set the column index names for single index columns as well.
Could you try to set the column index names, or revert the changes for doctests not to fail and address it in a separate PR?

garawalid:
@ueshin 
Now `pivot_table` supports column index names (#636). Also, I updated the doctest of `pivot`.  
Would you mind reviewing the PR?

PR #639, opened 2019-08-13
Introduce default index with new three index types

HyukjinKwon:
This PR proposes a default index so that we can now forget about the case when index is missing in Koalas DataFrame  - when Koalas DataFrame is directly created from Spark DataFrame.

There are three types of default index that can be controlled by `DEFAULT_INDEX` environment variable.

- **one-by-one**: It implements an one-by-one sequence by Window function without
    specifying partition. Therefore, it ends up with whole partition in single node.
    This index type should be avoided when the data is large. This is default.
    **TL;DR:** `Window` with `row_number` without partitioning spec
 
- **distributed-one-by-one**: It implements an one-by-one sequence by group-by and
    group-map approach. It still generates a one-by-one sequential index globally.
    If the default index must be an one-by-one sequence in a large dataset, this
    index has to be used.
    **TL;DR:** `groupby(partition_id).count().collect()` and `groupby(partition_id).apply(f)`.

 - **distributed**: It implements a monotonically increasing sequence simply by using
    Spark's `monotonically_increasing_id` function. If the index does not have to be
    a one-by-one sequence, this index should be used. Performance-wise, this index
    almost does not have any penalty comparing to other index types.
    **TL;DR:** `moninically_increasing_id()`.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4596,11 +4596,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         >>> merged.sort_values(by=['lkey', 'value_x', 'rkey', 'value_y'])
           lkey  value_x rkey  value_y
         0  bar        2  bar        6
-        1  baz        3  baz        7
-        2  foo        1  foo        5
-        3  foo        1  foo        8
-        4  foo        5  foo        5
-        5  foo        5  foo        8
+        5  baz        3  baz        7
+        1  foo        1  foo        5
+        2  foo        1  foo        8
+        3  foo        5  foo        5
+        4  foo        5  foo        8
 
         >>> left_kdf = ks.DataFrame({'A': [1, 2]})
         >>> right_kdf = ks.DataFrame({'B': ['x', 'y']}, index=[1, 2])
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -436,7 +436,8 @@ class _InternalFrame(object):
         if default_index_type == "one-by-one":
             sequential_index = F.row_number().over(
                 Window.orderBy(F.monotonically_increasing_id().asc())) - 1
-            return sdf.select(sequential_index.alias("__index_level_0__"), *list(sdf))
+            scols = [scol_for(sdf, column) for column in sdf.columns]
+            return sdf.select(sequential_index.alias("__index_level_0__"), *scols)
         elif default_index_type == "distributed-one-by-one":
             # 1. Calculates counts per each partition ID. `counts` here is, for instance,
             #     {
@@ -474,8 +475,9 @@ class _InternalFrame(object):
             sdf = sdf.withColumn("__spark_partition_id", F.spark_partition_id())
             return sdf.groupBy("__spark_partition_id").apply(grouped_map_func)
         elif default_index_type == "distributed":
+            scols = [scol_for(sdf, column) for column in sdf.columns]
             return sdf.select(
-                F.monotonically_increasing_id().alias("__index_level_0__"), *list(sdf))
+                F.monotonically_increasing_id().alias("__index_level_0__"), *scols)
         else:
             raise ValueError("'DEFAULT_INDEX' environment variable should be one of 'one-by-one',"
                              " 'distributed-one-by-one' and 'distributed'")
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -368,12 +368,13 @@ class _InternalFrame(object):
             assert column_index is None
             assert column_index_names is None
 
-            index_map = [('__index_level_0__', None)]
+            if "__index_level_0__" not in sdf.schema.names:
+                # Create default index.
+                index_map = [('__index_level_0__', None)]
+                sdf = _InternalFrame.attach_default_index(sdf)
 
-            # Create default index.
-            sdf = _InternalFrame.attach_default_index(sdf)
-
-        assert all(isinstance(index_field, str)
+        assert index_map is None \
+            or all(isinstance(index_field, str)
                    and (index_name is None or isinstance(index_name, str))
                    for index_field, index_name in index_map)
         assert scol is None or isinstance(scol, spark.Column)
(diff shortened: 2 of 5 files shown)


HyukjinKwon:
will fix the tests soon.

databricks/koalas/internal.py
```diff
@@ -359,6 +363,16 @@ def __init__(self, sdf: spark.DataFrame,
         :param column_index_names: Names for each of the index levels.
         """
         assert isinstance(sdf, spark.DataFrame)
+        if index_map is None:
+            # Here is when Koalas DataFrame is created directly from Spark DataFrame.
+            assert column_index is None
+            assert column_index_names is None
+
+            if "__index_level_0__" not in sdf.schema.names:
```
HyukjinKwon: I will make a separate PR to completely disallow no-index Koalas DataFrame. Seems like there are multiple places to fix.

One side question is that how we will handle the roundtrip in Koalas and Spark DataFrame. If Koalas has an index, `to_spark()` loses index information and we have nowhere to store.

I think, until we figure out a way to store index information properly, we should strip index when we convert to Spark DataFrame.

Currently it's kind of funny:

```
>>> ks.DataFrame(ks.DataFrame({'a': [1,2,3]}).to_spark())
   __index_level_0__  a
0                  0  1
1                  1  2
2                  2  3
```

So basically I would like to propose below until we figure out a clever way to roundtrip.

```
>>> ks.DataFrame({'a': [1,2,3]}).to_spark().show()
+---+
|  a|
+---+
|  1|
|  2|
|  3|
+---+
```

HyukjinKwon: cc @ueshin and @rxin.

databricks/koalas/frame.py
```diff
@@ -4596,11 +4596,11 @@ def merge(self, right: 'DataFrame', how: str = 'inner',
...
-        1  baz        3  baz        7
-        2  foo        1  foo        5
-        3  foo        1  foo        8
-        4  foo        5  foo        5
-        5  foo        5  foo        8
+        5  baz        3  baz        7
+        1  foo        1  foo        5
+        2  foo        1  foo        8
+        3  foo        5  foo        5
+        4  foo        5  foo        8
```
HyukjinKwon: This is actually matched with pandas's result since the output is sorted.

databricks/koalas/namespace.py
```diff
@@ -273,7 +273,7 @@ def read_delta(path: str, version: Optional[str] = None, timestamp: Optional[str
     Examples
     --------
     >>> ks.range(1).to_delta('%s/read_delta/foo' % path)
-    >>> ks.read_delta('%s/read_delta/foo' % path)
+    >>> ks.read_delta('%s/read_delta/foo' % path)  # doctest: +SKIP
```
HyukjinKwon: Those tests are related with https://github.com/databricks/koalas/pull/639/files#r313350742.

I will make a separate PR to fix it.

databricks/koalas/internal.py
```diff
@@ -394,6 +408,80 @@ def __init__(self, sdf: spark.DataFrame,
...
+
+        """
+        default_index_type = os.environ.get("DEFAULT_INDEX", "one-by-one")
+        if default_index_type == "one-by-one":
+            sequential_index = F.row_number().over(
+                Window.orderBy(F.monotonically_increasing_id().asc())) - 1
+            scols = [scol_for(sdf, column) for column in sdf.columns]
+            return sdf.select(sequential_index.alias("__index_level_0__"), *scols)
+        elif default_index_type == "distributed-one-by-one":
+            # 1. Calculates counts per each partition ID. `counts` here is, for instance,
```
HyukjinKwon: Here, actually it mimics `zipWithIndex` in RDD API.

databricks/koalas/internal.py
```diff
@@ -394,6 +408,80 @@ def __init__(self, sdf: spark.DataFrame,
...
+            almost does not have any penalty comparing to other index types.
+
+        """
+        default_index_type = os.environ.get("DEFAULT_INDEX", "one-by-one")
+        if default_index_type == "one-by-one":
+            sequential_index = F.row_number().over(
+                Window.orderBy(F.monotonically_increasing_id().asc())) - 1
+            scols = [scol_for(sdf, column) for column in sdf.columns]
+            return sdf.select(sequential_index.alias("__index_level_0__"), *scols)
+        elif default_index_type == "distributed-one-by-one":
```
HyukjinKwon: Here, actually it mimics `zipWithIndex` in RDD API.

databricks/koalas/internal.py
```diff
@@ -394,6 +408,80 @@ def __init__(self, sdf: spark.DataFrame,
...
+        environment variable.
+
+        - one-by-one: It implements an one-by-one sequence by Window function without
+            specifying partition. Therefore, it ends up with whole partition in single node.
+            This index type should be avoided when the data is large. This is default.
+
+        - distributed-one-by-one: It implements an one-by-one sequence by group-by and
+            group-map approach. It still generates a one-by-one sequential index globally.
+            If the default index must be an one-by-one sequence in a large dataset, this
+            index has to be used.
```
ueshin: Maybe we should notice that, if we add more data to the data source after creating this index, then it doesn't guarantee the sequential index?

databricks/koalas/internal.py
```diff
@@ -394,6 +408,80 @@ def __init__(self, sdf: spark.DataFrame,
...
+
+        - distributed: It implements a monotonically increasing sequence simply by using
+            Spark's `monotonically_increasing_id` function. If the index does not have to be
+            a one-by-one sequence, this index should be used. Performance-wise, this index
+            almost does not have any penalty comparing to other index types.
```
ueshin: Also notice that we can't use this type of index for combining two dataframes?

databricks/koalas/internal.py
HyukjinKwon: But to do that, we should keep index mapping information somewhere. After converting Koalas DataFrame into Spark DataFrame, it's being lost.

ueshin: yeah, loosing the mapping should be okay, I just meant that if the index is named, then the name should be the column name for the Spark DataFrame.

HyukjinKwon: Oh, oh right. Yes.

HyukjinKwon:
I am merging this too. Let me know if you guys have more comments.

PR #640, opened 2019-08-13
add `fillna` `bfill` `pad` `ffill` function in groupby

RainFung:


RainFung:
wait to be code reviewed.

HyukjinKwon:
Builds were failed.

RainFung:
The function is different between pandas `0.25` and `0.24`. I pass the latest version but failed in older version. How do i solve the diffience?

databricks/koalas/tests/test_groupby.py
```diff
@@ -182,6 +182,41 @@ def test_diff(self):
...
+    def test_bfill(self):
+        pdf = pd.DataFrame({'A': [1, 1, 2, 2],
+                            'B': [2, 4, None, 3],
+                            'C': [None, None, None, 1],
+                            'D': [0, 1, 5, 4]}, columns=['A', 'B', 'C', 'D'])
+        kdf = koalas.DataFrame(pdf)
+        self.assert_eq(kdf.groupby("A").bfill(),
+                       pdf.groupby("A").bfill())
+        self.assert_eq(repr(kdf.groupby("A")['B'].bfill()),
+                       repr(pdf.groupby("A")['B'].bfill()))
```
HyukjinKwon:
```
E   AssertionError: False is not true : DataFrames are not equal: 
E   
E   Left:
E        B    C  D
E   0  2.0  NaN  0
E   1  4.0  NaN  1
E   2  3.0  1.0  5
E   3  3.0  1.0  4
E   B    float64
E   C    float64
E   D      int64
E   dtype: object
E   
E   Right:
E      A    B    C  D
E   0  1  2.0  NaN  0
E   1  1  4.0  NaN  1
E   2  2  3.0  1.0  5
E   3  2  3.0  1.0  4
E   A      int64
E   B    float64
E   C    float64
E   D      int64
E   dtype: object
```

Looks pandas DataFrame in lower versions has additional column. You can drop this column from `pdf` and add some comments so that the tests can pass in lower versions too.

For instance, 

```python
if 'a' in pdf.columns:
    # In low pandas versions, it returns columns 'a' whereas higher version does not. Koalas follows
    # latest pandas versions so it drops column 'a' to test with low pandas versions as well.
    pdf = pdf.drop(columns=["a"])
```

HyukjinKwon: Or,

```python
if LooseVersion("0.25") > LooseVersion(pd.__version__):
    # In pandas versions lower than 0.25, it returns columns 'a' whereas 0.25 version does not.
    # Koalas follows latest pandas versions so it drops column 'a' to test with low pandas
    # versions as well.
    pdf = pdf.drop(columns=["a"])

...

self.assert_eq(...)
```

ueshin: I'd prefer to use `LooseVersion` in this case.

HyukjinKwon:
Thanks for working on this, @RainFung

Issue #343, opened 2019-05-15
pandas.DataFrame.tail


tdliu:
Head is implemented but tail is not. Tail is not just useful for exploring data but also a very useful way of slicing data. 

Example: users have data ordered by date. They want to take the last 30 days. They can just call tail(30)

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.tail.html

rxin:
Is the most common case people run tail after sorting? It's actually somewhat poor performance to implement a general purpose tail.

ustcscgyer:
It is not as straightforward as it seems. In Pandas, all the data is already in memory, `tail` is not too much different to `head`. In SparkDataFrame, data is calculated at the time of consuming, getting tail means you have to evaluate the whole data set.

thunterdb:
spark tends to think of tail as what happened in a linked list (dropping just a few elements), but it is effectively used like .head or .limit on the end of the data.

I have started a small design doc on how we could support these use cases and at the same time ensure that the output is exactly the same as pandas despite the distributed nature of spark. I will share it this weekend or next week.

rodrigue89:
If you run a large data set in pandas, it may be a good idea to run both head and tail methods to confirm the integrity of your data while uploading or after processing your data.

brookewenig:
Any update on this? For the demo, I'd actually really like tail to be implemented.

HyukjinKwon:
^ WDYT @rxin, @thunterdb and @ueshin?

PR #641, opened 2019-08-13
[WIP] add tail function to frame

akhilputhiry:
Address https://github.com/databricks/koalas/issues/343

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -6489,6 +6489,65 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         # Koalas DataFrame.
         is_dataframe = None
 
+    def tail(self, n=5):
+        """
+        Return the last `n` rows.
+        This function returns last `n` rows from the object based on
+        position. It is useful for quickly verifying data, for example,
+        after sorting or appending rows.
+
+        Parameters
+        ----------
+        n : int, default 5
+            Number of rows to select.
+
+        Returns
+        -------
+        type of caller
+            The last `n` rows of the caller object.
+
+        Examples
+        --------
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -6489,6 +6489,65 @@ def __class_getitem__(cls, params):
...
+        2  zebra
+        """
+        # pick one column to perform order by
+        first_column = self.columns[0]
+
+        # calculate columns to be dropped
+        columns_to_drop = [col.name for col in self._sdf.schema if col.name not in self.columns]
+
+        # calculate last n records
+        data = self._sdf.orderBy(first_column).drop(*columns_to_drop).collect()[-n:]
```
HyukjinKwon: Actually this isn't quite performant. See https://github.com/databricks/koalas/issues/343#issuecomment-495416050 

If the DataFrame is big, it will collects all the data into the Spark's driver side (client) and it will easily hit an out-of-memory issue.

One possible way is that we collect tails from each partition, collect, and calculate the tail again.

One way might be to use `mapInPandas` which will be introduced in Spark 3.0 - https://github.com/apache/spark/blob/a5a5da78cfbb1b439c139fd2ffd5744fc60ebafe/python/pyspark/sql/functions.py#L3207-L3232

If we don't have a better idea now, let's leave this case closed for now.

akhilputhiry: Yaa. Let me try that.

WeichenXu123: If the rule is find the last n rows when sort by first column, then we can turn it into the operation "get first n rows", we only need to reverse the sorter compare function. What do you think ? @HyukjinKwon @akhilputhiry

akhilputhiry: I am not getting time to work on this. Closing this for now.

Issue #409, opened 2019-05-30
DataFrame.GroupBy.apply()


patryk-oleniuk:
I'm continuously missing this behavior when converting my pandas scripts to koalas:

```python
df = pd.DataFrame({"timestamp":[0.0, 0.5, 1.0, 0.0, 0.5],
              "car_id": ['A','A','A','B','B'], 
              "battery_charge": [100, 90, 80,100,90]
              })

print(df)

def calc_battery_usage(x):
    start_bat = x.sort_values('timestamp').iloc[0]['battery_charge']
    stop_bat = x.sort_values('timestamp').iloc[-1]['battery_charge']
    return start_bat - stop_bat

print(df.groupby('car_id').apply(calc_battery_usage))
```
That prints out:
```python
   timestamp car_id  battery_charge
0        0.0      A             100
1        0.5      A              90
2        1.0      A              80
3        0.0      B             100
4        0.5      B              90

car_id
A    20
B    10
```

On koalas I just get:
```
PandasNotImplementedError: The method `pd.groupby.GroupBy.apply()` is not implemented yet.
```

I can take over and start implementing it myself, but would probably need some general guidelines on how to do it (probably using pysparks pandas_udf).

rxin:
Is this similar to https://spark.apache.org/docs/latest/sql-pyspark-pandas-with-arrow.html#grouped-map and https://spark.apache.org/docs/latest/sql-pyspark-pandas-with-arrow.html#grouped-aggregate ? Perhaps we can just map to that.

patryk-oleniuk:
Yes, this is similar. In my current scripts these are the methods I'm using to "manually" make the apply on groupby's.

icexelloss:
Hi @patryk-oleniuk! I think it'd great to have groupby apply in koalas. 

Spark's groupby apply and Pandas's groupby apply are pretty similar. However, two main difference that I can think of are:

- Spark's groupby apply need full return type specification on the udf definition
- Spark's groupby apply doesn't keep index, while Spark's groupby apply use the grouping key as index.

As a start, I'd probably try this something like:
```
@pandas_wrap(full_return_schema):
def calc_battery_usage(x):
    start_bat = x.sort_values('timestamp').iloc[0]['battery_charge']
    stop_bat = x.sort_values('timestamp').iloc[-1]['battery_charge']
    return start_bat - stop_bat

kdf.groupby('car_id').apply(calc_battery_usage)
```
[...]

PR #633, opened 2019-08-09
Add an option to enable operations on different DataFrames

HyukjinKwon:
This PR proposes to add an environment variable, `OPS_ON_DIFF_FRAMES`, to enable operations on different DataFrames.

To use this feature, set `OPS_ON_DIFF_FRAMES` environment variable to `true` and run Koalas codes.

The changes here are a bit big to match the behaviours with pandas, and to generalize them. However, how it works is pretty straightforward. Basically, it joins in index columns and do the operations. See the rough Spark examples below:

```python
>>> df1.show()
```

```
+-----------------+---+
|__index_level_0__|  a|
+-----------------+---+
|                0|  1|
|                1|  2|
|                2|  3|
+-----------------+---+
```

```python
>>> df2.show()
```

```
+-----------------+---+
|__index_level_0__|  a|
+-----------------+---+
|                0|  1|
|                1|  2|
|                2|  3|
+-----------------+---+
```

```python
>>> df1.join(df2, on="__index_level_0__", how="full").show()
```
```
+-----------------+---+---+
|__index_level_0__|  a|  a|
+-----------------+---+---+
|                0|  1|  1|
|                1|  2|  2|
|                2|  3|  3|
+-----------------+---+---+
```

Annoying part here is that how to resolve duplicated column names `a`. In the current implementation, it's kind of tricky to use DataFrame's alias.

In this PR, I had to workaround by aliasing it. For instance:

```
+-----------------+--------+--------+
|__index_level_0__|__this_a|__that_a|
+-----------------+--------+--------+
|                0|       1|       1|
|                1|       2|       2|
|                2|       3|       3|
+-----------------+--------+--------+
```

and then, perform the operations, e.g.:

```
+-----------------+---------------------+
|__index_level_0__|(__this_a + __that_a)|
+-----------------+---------------------+
|                0|                    2|
|                1|                    4|
|                2|                    6|
+-----------------+---------------------+
```

and alias back:

```
+-----------------+---+
|__index_level_0__|  a|
+-----------------+---+
|                0|  2|
|                1|  4|
|                2|  6|
+-----------------+---+
```

Resolves #624

PR #646, opened 2019-08-14
Allow to omit type hint in GroupBy.transform, filter, apply

### LINKED ISSUE #628: Koalas groupby.apply does not keep the index whereas pandas does.
```python
pdf = pd.DataFrame({"timestamp":[0.0, 0.5, 1.0, 0.0, 0.5],
             "car_id": ['A','A','A','B','B']
              })

def do_something(x):
    return pd.DataFrame({"column": [0.0]})

print(pdf.groupby('car_id').apply(do_something))
```

```
          column
car_id
A      0     0.0
B      0     0.0
```

```python
kdf = ks.DataFrame({"timestamp":[0.0, 0.5, 1.0, 0.0, 0.5],
              "car_id": ['A','A','A','B','B']
              })

def do_something(x) -> ks.DataFrame[float]:
    return pd.DataFrame({"column": [0.0]})

print(kdf.groupby('car_id').apply(do_something))
```

```
    c0
0  0.0
1  0.0
```

HyukjinKwon:
This PR proposes to allow to omit type hint.

1. If type hint is not given, Koalas collects data as pandas DataFrame.
  1.1. If it returns less than 1000, it returns `ks.DataFrame(pdf)`.
  1.2. if it returns more than 1000, it just infers schema from `pdf`.
2. If type hint is found, uses that as return type directly.

- When the schema is inferred without type hint, index is cleanly kept.

- When the schema is given via type hint, index is lost - this is because pandas sometimes keeps the index and sometimes not. Koalas does not know without executing the given `func` once.

    Please see the example below:

    ```python
    pdf = pd.DataFrame({"timestamp":[0.0, 0.5, 1.0, 0.0, 0.5], "car_id": ['A','A','A','B','B']})
    print(pdf.groupby('car_id').apply(lambda _: pd.DataFrame({"column": [0.0]})))
    ```

    ```
              column
    car_id
    A      0     0.0
    B      0     0.0
    ```

    ```python
    pdf = pd.DataFrame({"timestamp":[0.0, 0.5, 1.0, 0.0, 0.5], "car_id": ['A','A','A','B','B']})
    print(pdf.groupby('car_id').apply(lambda x: x))
    ```

    ```
       timestamp car_id
    0        0.0      A
    1        0.5      A
    2        1.0      A
    3        0.0      B
    4        0.5      B
    ```

    Therefore, index information is lost in this case

Resolves #628
This PR also resolves 2. and 3. at https://github.com/databricks/koalas/issues/409#issuecomment-518762148

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -33,7 +33,7 @@ from pyspark.sql.types import FloatType, DoubleType, NumericType, StructField, S
 from pyspark.sql.functions import PandasUDFType, pandas_udf
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
-from databricks.koalas.typedef import _infer_return_type
+from databricks.koalas.typedef import _infer_return_type, as_spark_type
 from databricks.koalas.frame import DataFrame
 from databricks.koalas.internal import _InternalFrame
 from databricks.koalas.missing.groupby import _MissingPandasLikeDataFrameGroupBy, \
@@ -700,10 +700,18 @@ class GroupBy(object):
         be much faster than using `apply` for their specific purposes, so try to
         use them before reaching for `apply`.
 
-        .. note:: unlike pandas, it is required for ``func`` to specify return type hint.
+        .. note:: this API executes the function once to infer the type which is
+             potentially expensive, for instance, when the dataset is created after
+             aggregations or sorting.
 
-        .. note:: the output column names are `c0, c1, c2 ... cn`. These names
-            are positionally mapped to the returned DataFrame in ``func``. See examples below.
+             To avoid this, specify return type in ``func``, for instance, as below:
+
+             >>> def pandas_div_sum(x) -> ks.DataFrame[float, float]:
+             ...    return x[['B', 'C']] / x[['B', 'C']].sum()
+
+             If the return type is specified, the output column names become
+             `c0, c1, c2 ... cn`. These names are positionally mapped to the returned
+             DataFrame in ``func``. See examples below.
 
         .. note:: the dataframe within ``func`` is actually a pandas dataframe. Therefore,
             any pandas APIs within this function is allowed.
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
@patryk-oleniuk, with this PR and https://github.com/databricks/koalas/pull/633 PR, you will be able to proceed further. Hope this unblocks you.

databricks/koalas/groupby.py
```diff
@@ -817,46 +842,26 @@ def filter(self, func):
...
 
-            # Here, we restore the index column back in Spark DataFrame
-            # so that Koalas can understand it as an index.
+        kdf = self._apply(pandas_filter, data_schema, retain_index=True)
+        return DataFrame(self._kdf._internal.copy(sdf=kdf._sdf))
 
-            # TODO: deduplicate this logic with _InternalFrame.from_pandas
-            columns = pdf.columns
-            data_columns = [str(col) for col in columns]
+    def _apply(self, func, return_schema, retain_index):
```
HyukjinKwon: Mainly only deduplication.

databricks/koalas/groupby.py
```diff
@@ -881,17 +886,62 @@ def rename_output(pdf):
                     pdf.index.name = index_names[0]
 
             pdf = func(pdf)
-            # For now, just positionally map the column names to given schema's.
+
+            if retain_index:
```
HyukjinKwon: Mainly only deduplication here.

HyukjinKwon:
tests passed. codecov is due to missing coverage hits in Python worker side

databricks/koalas/groupby.py
```diff
@@ -817,46 +842,26 @@ def filter(self, func):
...
-            data_columns = [str(col) for col in columns]
+    def _apply(self, func, return_schema, retain_index):
+        should_infer_schema = return_schema is None
+        input_groupnames = [s.name for s in self._groupkeys]
 
-            index = pdf.index
+        if should_infer_schema:
+            # Here we execute with the first 1000 to get the return type.
+            # If the records were less than 1000, it uses pandas API directly for a shortcut.
+            limit = 1000
```
HyukjinKwon: I think I should have a configuration for such limit in a separate PR. I'll do it after this and your PR are merged @ueshin

ueshin: I merged my PR for the basic configuration.

PR #647, opened 2019-08-14
`get_dummies` should use internal functions.

ueshin:


ueshin:
I'd merge this for now.

PR #649, opened 2019-08-15
Groupby.idxmax and Groupby.idxmin

RainFung:
Now only support one-level index


RainFung:
wait to be code reviewed.

databricks/koalas/groupby.py
```diff
@@ -1008,6 +1008,120 @@ def rank(self, method='average', ascending=True):
...
+        databricks.koalas.Series.groupby
+        databricks.koalas.DataFrame.groupby
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'a': [1, 1, 2, 2, 3],
+        ...                    'b': [1, 2, 3, 4, 5],
+        ...                    'c': [5, 4, 3, 2, 1]}, columns=['a', 'b', 'c'])
+
+        >>> df.groupby(['a']).idxmin().sort_index() # doctest: +NORMALIZE_WHITESPACE
```
HyukjinKwon: Can we add some tests and implementation for series groupby too? For instance `df.b.groupby(df.a).idxmin()`

databricks/koalas/groupby.py
```diff
@@ -1010,6 +1010,135 @@ def rank(self, method='average', ascending=True):
...
+        databricks.koalas.Series.groupby
+        databricks.koalas.DataFrame.groupby
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'a': [1, 1, 2, 2, 3],
+        ...                    'b': [1, 2, 3, 4, 5],
+        ...                    'c': [5, 4, 3, 2, 1]}, columns=['a', 'b', 'c'])
+
+        >>> df.groupby(['a'])['b'].idxmin().sort_index() # doctest: +NORMALIZE_WHITESPACE
```
HyukjinKwon: pandas seems returning series:

```
a
1    0
2    2
3    4
Name: b, dtype: int64
```

probably we could do it via:

```python
class SeriesGroupBy(GroupBy):

    ...

    def idxmin(self, skipna=True):
        return _col(super(SeriesGroupBy, self). idxmin(func, skipna))
```

HyukjinKwon: @RainFung seems this comment isn't addressed.

RainFung:
I do nothing with  `groupby.rank` why i failed in `groupby.rank`. Can you help me check it?

databricks/koalas/groupby.py
```diff
@@ -1721,3 +1849,20 @@ def transform(self, func):
 
     def filter(self, func):
         raise NotImplementedError()
+
+    def rank(self, method='average', ascending=True):
+        kdf = super(SeriesGroupBy, self).rank(method, ascending).to_dataframe()
+        return _col(DataFrame(kdf._internal.copy(
+            sdf=kdf._sdf.select(kdf._internal.data_scols), index_map=[])))  # index is lost.
```
HyukjinKwon: Seems you do something with rank here. After https://github.com/databricks/koalas/commit/f2a718dcfdda965ac83128d8d979d3146e3ae76a, we should always set the index.

If index is lost, you can imply to:

```python
_col(DataFrame(kdf._internal.copy(sdf=kdf._sdf.select(kdf._internal.data_scols)))
```

PR #655, opened 2019-08-19
Exclude Index columns for exposed Spark DataFrame and disallow Koalas DataFrame with no index

HyukjinKwon:
This PR is a followup of  and proposes two things:
  - Exclude Index columns for exposed Spark DataFrame
  - Disallow Koalas DataFrame with no index

So, for instance, `to_spark()`  now shows:

```diff
-       __index_level_0__  x
-    0                  0  0
-    1                  1  1
+       x
+    0  0
+    1  1
```

and `index_map` is not expected to be empty in Koalas DataFrame, always. It sets the default index explicitly per https://github.com/databricks/koalas/pull/639

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1598,9 +1598,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         Index
         """
         from databricks.koalas.indexes import Index, MultiIndex
-        if len(self._internal.index_map) == 0:
-            return None
-        elif len(self._internal.index_map) == 1:
+        if len(self._internal.index_map) == 1:
             return Index(self)
         else:
             return MultiIndex(self)
@@ -1860,9 +1858,6 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         lion           mammal   80.5     run
         monkey         mammal    NaN    jump
         """
-        if len(self._internal.index_map) == 0:
-            raise NotImplementedError('Can\'t reset index because there is no index.')
-
         multi_index = len(self._internal.index_map) > 1
 
         def rename(index):
@@ -1877,7 +1872,15 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         if level is None:
             new_index_map = [(column, name if name is not None else rename(i))
                              for i, (column, name) in enumerate(self._internal.index_map)]
-            index_map = []
+            new_data_columns = [
+                self._internal.scol_for(column).alias(name) for column, name in new_index_map]
+            sdf = self._sdf.select(new_data_columns + self._internal.data_columns)
+
+            # Now, new internal Spark columns are named as same as index name.
+            new_index_map = [(name, name) for column, name in new_index_map]
+
+            index_map = [('__index_level_0__', None)]
+            sdf = _InternalFrame.attach_default_index(sdf)
         else:
             if isinstance(level, (int, str)):
                 level = [level]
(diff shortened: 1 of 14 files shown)


databricks/koalas/frame.py
```diff
@@ -1877,7 +1872,10 @@ def rename(index):
         if level is None:
             new_index_map = [(column, name if name is not None else rename(i))
                              for i, (column, name) in enumerate(self._internal.index_map)]
-            index_map = []
+            # TODO: this will end up with multiple columns that point the same index column.
+            #     For instance, if we do df.reset_index(), the new column `index` points the
+            #     same '__index_level_0__' internally. We should have new column.
```
HyukjinKwon: Seems like I should fix this one too to make the tests pass.

databricks/koalas/groupby.py
```diff
@@ -843,24 +866,11 @@ def filter(self, func):
...
-        if should_infer_schema:
-            # Here we execute with the first 1000 to get the return type.
-            # If the records were less than 1000, it uses pandas API directly for a shortcut.
-            limit = 1000
-            pdf = self._kdf.head(limit + 1).to_pandas()
-            pdf = pdf.groupby(input_groupnames).apply(func)
-            kdf = DataFrame(pdf)
-            return_schema = kdf._sdf.schema
-            if len(pdf) <= limit:
-                return kdf
```
HyukjinKwon: Mainly just a refactoring to disallow `_InternalDataFrame(sdf=sdf, index_map=index_map)`. Now `_InternalDataFrame` with Spark DataFrame

databricks/koalas/frame.py
```diff
@@ -1915,10 +1910,27 @@ def rename(index):
                      index_name if index_name is not None else rename(index_name)))
                 index_map.remove(info)
 
+        new_data_columns = [
```
HyukjinKwon: These codes are needed because previously it removed the index when `reset_index`. Now it sets the default index.

HyukjinKwon:
Thanks Takuya!!

thunterdb:
Why disallow the KDFs with no index?

HyukjinKwon:
Biggest reason was: https://github.com/databricks/koalas/pull/633 introduced operations on different Koalas DataFrame which is based upon index.

So, if no index is set, different Koalas DataFrames cannot perform an operation (because this feature is based upon a join). Namely, previously the cases below

```python
ks.range(10) + ks.range(10)
```

```python
# This was an actual usecase reported
spark_df = spark.read.redshift("...")
ks.DataFrame(spark_df) + ks.DataFrame(spark_df)
```

were disallowed because it has no index ^. Now, it has a default index - https://github.com/databricks/koalas/pull/639.

Secondly, it makes much easier implementing other APIs. Currently, many APIs such as cumulative sum, max, etc. don't work when index does not exist because the implementation is dependent on index column.

Lastly, since we set the default index, we can allow those cases now, which makes more sense and closer to pandas'

https://github.com/databricks/koalas/pull/655/files#diff-ebfd51978d1aa348373b958a1b27aa47R936
https://github.com/databricks/koalas/pull/655/files#diff-ebfd51978d1aa348373b958a1b27aa47R543
https://github.com/databricks/koalas/pull/655/files#diff-66320dc549688da8eecb81b165764a86R37-R40

HyukjinKwon:
cc @rxin, BTW, this is the PR we disabled no-index case.

PR #657, opened 2019-08-19
Fix assign for multi-index columns support.

ueshin:


databricks/koalas/frame.py
```diff
@@ -2770,10 +2770,13 @@ def assign(self, **kwargs):
                 sdf = sdf.withColumn(name, F.lit(c))
 
         data_columns = set(self._internal.data_columns)
+        adding_columns = [name for name, _ in pairs if name not in data_columns]
+        level = self._internal.column_index_level
+        adding_column_index = [tuple([col, *[''] * (level - 1)]) for col in adding_columns]
```
HyukjinKwon: no big deal but I would do `*([''] * (level - 1))` ..

ueshin:
Thanks! merging.

Issue #665, opened 2019-08-21
Plot APIs improvement in Series


HyukjinKwon:
Koalas: https://koalas.readthedocs.io/en/latest/reference/series.html#plotting
pandas: https://pandas.pydata.org/pandas-docs/stable/reference/series.html#plotting

We should improve Koalas' plot more.

HyukjinKwon:
We will track it in #293

PR #669, opened 2019-08-21
Add plot.pie in Series

HyukjinKwon:
This PR add series.plot.pie in Series.

Can be tested as below:

```python
ks.range(10).to_pandas().id.plot.pie().figure.savefig("image.png")
ks.range(10).id.plot.pie().figure.savefig("image.png")
```

image

If there are more than 1000 rows, it shows as below:

```
ks.range(1001).id.plot.pie().figure.savefig("image.png")
```

image

Most of people will use, for instance `df.x.value_counts().plot.pie()`.  This case is similar as bar plot.

Partially addresses #665

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -15,18 +15,18 @@
 #
 
 from distutils.version import LooseVersion
+
 import matplotlib
 import numpy as np
 import pandas as pd
-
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
I am merging this for today's release but please feel free to review after it's merged.

PR #671, opened 2019-08-21
Add plot.line in Series

HyukjinKwon:
This PR add series.plot.pie in Series.

Can be tested as below:

```python
import databricks.koalas as ks
import pandas as pd

pdf = pd.DataFrame({
          'a': [1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 50],
      }, index=[0, 1, 3, 5, 6, 8, 9, 9, 9, 10, 10])

kdf = ks.DataFrame(pdf)
pdf['a'].plot.line(colormap='Paired').figure.savefig("image1.png")
kdf['a'].plot.line(colormap='Paired').figure.savefig("image2.png")
```

image1

In case of this plot, we sample and match the row numbers around 1000.

```python
import databricks.koalas as ks
import pandas as pd

pdf = pd.DataFrame({'a': [1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 50, 100] * 100})
kdf = ks.DataFrame(pdf)
kdf['a'].plot.line(colormap='Paired').figure.savefig("image4.png")
```

image4

Partially addresses #665

PR #672, opened 2019-08-21
Add plot.kde in Series

HyukjinKwon:
This depends on https://github.com/databricks/koalas/pull/671


databricks/koalas/plot.py
```diff
@@ -514,6 +512,27 @@ def _make_plot(self):
         super(KoalasBarhPlot, self)._make_plot()
 
 
+class KoalasKdePlot(KdePlot):
```
HyukjinKwon: This seems it has to implement the logic with Koalas or Spark DataFrame operations rather then just simply sampling.

HyukjinKwon:
I am going to work on this later.

PR #674, opened 2019-08-21
Avoid copying pandas_df for internal use.

ueshin:
Currently we use a copy of cached pandas DataFrame/Series when calling `to_pandas()` because the cached pandas DataFrame/Series might be modified by a user and it breaks immutability of `InternalFrame`.
However, for internal use, we can know whether the DataFrame/Series will be modified or not and we don't need to use the copy when the DataFrame/Series will not be modified.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -370,7 +370,7 @@ class DataFrame(_Frame, Generic[T]):
             # Here we execute with the first 1000 to get the return type.
             # If the records were less than 1000, it uses pandas API directly for a shortcut.
             limit = 1000
-            pdf = self.head(limit + 1).to_pandas()
+            pdf = self.head(limit + 1)._to_pandas()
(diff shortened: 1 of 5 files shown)


databricks/koalas/frame.py
```diff
@@ -6425,8 +6425,16 @@ def _pd_getitem(self, key):
             return DataFrame(self._internal.copy(sdf=self._sdf.filter(bcol)))
         raise NotImplementedError(key)
 
+    def _to_pandas(self):
```
HyukjinKwon: Actually, how about naming it explicitly like ... `_to_internal_pandas` or .. `_to_pandas_without_copy` ... There is kiiiiind off ... similar one I did `spark_internal_df` vs `spark_df` for instance FYI.

HyukjinKwon:
I will fix the test in the other PR

PR #675, opened 2019-08-22
Deduplicate and cleanup new plot APIs

HyukjinKwon:


databricks/koalas/plot.py
```diff
@@ -720,12 +700,6 @@ def line(self, x=None, y=None, **kwargs):
         See Also
         --------
         matplotlib.pyplot.plot : Plot y versus x as lines and/or markers.
-
-        Examples
-        --------
-
-        >>> s = ks.Series([1, 3, 2])
-        >>> p = s.plot.line()
```
HyukjinKwon: This example is broken for same reason. I am going to remove it for now - this is tested in `test_series_plot.py`

PR #677, opened 2019-08-22
Implement DataFrame.to_csv and Series.to_csv base on spark.write.csv

RainFung:
Implement DataFrame.to_csv and Series.to_csv base on spark.write.csv.There are still some question. 
1. spark write file to hdfs, maybe user usually want to write to local.
2. It's not very convenient to use csv, for example, just view through excel.


databricks/koalas/frame.py
```diff
@@ -878,6 +878,85 @@ def iteritems(self):
         cols = list(self.columns)
         return list((col_name, self[col_name]) for col_name in cols)
 
+    def to_csv(self, path_or_buf=None, sep=',', na_rep='', columns=None, header=True,
```
HyukjinKwon: @RainFung, this is implemented under `_Frame.py` in `general.py`. You can remove that implementation or move those implementation to there.

HyukjinKwon:
Let me take this over at #749

PR #678, opened 2019-08-22
implement DataFrame.idxmin and DataFrame.idxmax

RainFung:


databricks/koalas/frame.py
```diff
@@ -2278,6 +2278,94 @@ def round(self, decimals=0):
...
+        c    (1, 1)
+        dtype: object
+        """
+        idxmax_data = []
+        idxmax_index = []
+        for col in self._internal.data_columns:
+            idxmax_tuple = self[col].idxmax(skipna=skipna)
+            idxmax_index.append(col)
+            idxmax_data.append(idxmax_tuple)
+        return pd.Series(data=idxmax_data, index=idxmax_index)
```
HyukjinKwon: Can we use Koalas DataFrame (e.g., by defining `Series._idxmax` and use it?)

databricks/koalas/series.py
```diff
@@ -2803,18 +2783,29 @@ def idxmin(self, skipna=True):
...
         index_scols = self._kdf._internal.index_scols
         # asc_nulls_(list|first)is used via Py4J directly because
         # it's not supported in Spark 2.3.
-        if skipna:
-            sdf = sdf.orderBy(Column(scol._jc.asc_nulls_last()))
-        else:
-            sdf = sdf.orderBy(Column(scol._jc.asc_nulls_first()))
+        if func == 'idxmin':
+            if skipna:
+                sdf = sdf.orderBy(Column(scol._jc.asc_nulls_last()))
```
HyukjinKwon: Hmmmm .. seems like it will execute sort for each column, which will be super expensive .. do you think we can sort once and extract idxmin and idxmax?

RainFung: Do you mean it will sort each column in `frame.idxmin`. I don't have good idea to solve it, because it need get max index in every independent column.

HyukjinKwon: hmmmmm .... let's don't add this API for now .. it will be crazy expensive ...

HyukjinKwon: Let's leave this closed for now.

PR #679, opened 2019-08-22
Rename the default index 'one-by-one' to 'sequence'

HyukjinKwon:
There was a concern about the naming 'one-by-one' since it's confusing (see https://groups.google.com/forum/m/#!topic/koalas-dev/xzZ0VvXFsGM)
This PR proposes to rename the default index 'one-by-one' to 'sequence'.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -422,32 +422,67 @@ class _InternalFrame(object):
         There are three types of default index that can be controlled by `DEFAULT_INDEX`
         environment variable.
 
-        - one-by-one: It implements an one-by-one sequence by Window function without
+        - sequence: It implements a sequence that increases one by one, by Window function without
             specifying partition. Therefore, it ends up with whole partition in single node.
             This index type should be avoided when the data is large. This is default.
+            See example below:
 
-        - distributed-one-by-one: It implements an one-by-one sequence by group-by and
-            group-map approach. It still generates a one-by-one sequential index globally.
-            If the default index must be an one-by-one sequence in a large dataset, this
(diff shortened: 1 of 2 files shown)


databricks/koalas/internal.py
```diff
@@ -422,32 +422,32 @@ def attach_default_index(sdf):
         There are three types of default index that can be controlled by `DEFAULT_INDEX`
         environment variable.
 
-        - one-by-one: It implements an one-by-one sequence by Window function without
+        - sequence: It implements an sequence sequence by Window function without
```
rxin: what do you mean by squence sequence?  you might want to just give some examples explaining the differences.

HyukjinKwon: Sorry, I will update it soon.

databricks/koalas/internal.py
```diff
@@ -422,32 +422,82 @@ def attach_default_index(sdf):
...
+            >>> prev = os.environ.get("DEFAULT_INDEX", "sequence")
+            >>> os.environ["DEFAULT_INDEX"] = "distributed-sequence"
+
+            >>> ks.range(3).index
+            Int64Index([0, 1, 2], dtype='int64')
+
+            This is conceptually equivalent to the Spark example as below:
+
+            >>> spark_df = ks.range(3).to_spark()
+            >>> spark_df.rdd.zipWithIndex().map(lambda p: p[1]).collect()
```
HyukjinKwon: Actually, RDD API is not used in actual index implementation but pandas group-map UDF is being used, and it mimics `zipWithIndex` in RDD API.

HyukjinKwon:
Merged for now.

PR #691, opened 2019-08-25
Temporaly repr of MultiIndex & Add some Tests for Indexes.py

itholic:
### \# 1

When I checked repr for MultiIndex,

It was confusing because it was printed as 'Index' instead of 'MultiIndex'.

```python
>>> import pandas as pd
>>> import numpy as np
>>>
>>> arrays = [[1, 1, 2, 2], ['red', 'blue', 'red', 'blue']]
>>> idx = pd.MultiIndex.from_arrays(arrays, names=('Index', 'color'))
>>> pdf = pd.DataFrame(np.random.randn(4, 5), idx)
>>> kdf = ks.from_pandas(pdf)
>>>
>>> kidx = kdf.index
>>> kidx
Index([(1, 'red'), (1, 'blue'), (2, 'red'), (2, 'blue')], dtype='object', name='number')
```

So of course this is not the correct answer,

I suggest changing the MuntiIndex repr temporarily to avoid confusion.

스크린샷 2019-08-26 오전 12 14 45

then we can get repr result like this

```python
>>> kidx = kdf.index
>>> kidx
MultiIndex([(1, 'red'), (1, 'blue'), (2, 'red'), (2, 'blue')], dtype='object', name='number')
```

### \# 2

And I found that there was not enough error testing for the names property.

image

I added tests for those errors,

And I've added tests for some other code that isn't being tested yet.

It's very appreciate it if someone check it out when available. :)

Thanks.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -273,6 +273,10 @@ class MultiIndex(Index):
         """
         return MultiIndex(self._kdf, scol)
 
+    # TODO: This is a temporary measure to avoid confusion, so the logic should be changed later.
+    def __repr__(self):
+        return super().__repr__().replace("Index", "MultiIndex", 1)
+
     def any(self, *args, **kwargs):
         raise TypeError("cannot perform any with this index type: MultiIndex")
 
@@ -297,4 +301,4 @@ class MultiIndex(Index):
         raise AttributeError("'MultiIndex' object has no attribute '{}'".format(item))
 
     def rename(self, name, inplace=False):
-        raise NotImplementedError()
+        raise PandasNotImplementedError(class_name='pd.MultiIndex')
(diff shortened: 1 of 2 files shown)


databricks/koalas/indexes.py
```diff
@@ -273,6 +273,10 @@ def _with_new_scol(self, scol: spark.Column) -> 'MultiIndex':
         """
         return MultiIndex(self._kdf, scol)
 
+    # TODO: This is a temporary measure to avoid confusion, so the logic should be changed later.
```
HyukjinKwon: If this is a temp fix, why do we need it in this PR?

leehaejoon: @HyukjinKwon Thanks for your feedback! :) As you said, I'll create each PR separately. Thank you again for your advice!

itholic: @HyukjinKwon Thanks for your feedback! :) As you said, I'll create each PR separately. Thank you again for your advice!

itholic:
I'll separate these commits and generate a PR again.

PR #694, opened 2019-08-26
Add test for reverse operator at Series

itholic:
Added missing parts for the reverse operator at Series.
such as radd, rsub, and rdiv and so on.

And to make the result different from the reverse operator,
I made some changes to the data in the existing example.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -73,20 +73,41 @@ Series.{reverse}
 _add_example_SERIES = """
 Examples
 --------
->>> df = ks.DataFrame({'a': [1, 1, 1, np.nan],
-...                    'b': [1, np.nan, 1, np.nan]},
+>>> df = ks.DataFrame({'a': [2, 2, 4, np.nan],
+...                    'b': [2, np.nan, 2, np.nan]},
 ...                   index=['a', 'b', 'c', 'd'], columns=['a', 'b'])
 >>> df
      a    b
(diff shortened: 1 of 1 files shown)


databricks/koalas/series.py
```diff
@@ -217,6 +343,27 @@
 Name: a, dtype: float64
 """
 
+_rfloordiv_example_SERIES = """
```
HyukjinKwon: @itholic, can you maybe just add the new example into `_floordiv_example_SERIES`? This doc looks unnecessarily getting long. For instance,

```python
_floordiv_example_SERIES = """
Examples
--------
>>> df = ks.DataFrame({'a': [2, 2, 4, np.nan],
...                    'b': [2, np.nan, 2, np.nan]},
...                   index=['a', 'b', 'c', 'd'], columns=['a', 'b'])
>>> df
     a    b
a  2.0  2.0
b  2.0  NaN
c  4.0  2.0
d  NaN  NaN

>>> df.a.floordiv(df.b)
a    1.0
b    NaN
c    2.0
d    NaN
Name: a, dtype: float64

>>> df.a.rfloordiv(df.b)
a    0.0
b    NaN
c   -2.0
d    NaN
Name: a, dtype: float64
"""
```

itholic: @HyukjinKwon , The test structure was changed based on your comments. Can you confirm that I made the change when you available? Thanks for your comment.

PR #695, opened 2019-08-26
ENH: Add bar plot for DataFrame

charlesdong1991:
kdf:

Screen Shot 2019-08-26 at 2 36 11 PM

pdf:

Screen Shot 2019-08-26 at 2 36 21 PM

PR #699, opened 2019-08-27
Implement frame Dataframe.plot.bar

itholic:
Implement frame Dataframe.plot.bar

스크린샷 2019-08-27 오후 2 10 33


itholic:
already implement at #695

PR #704, opened 2019-08-27
BUG&TST: Fix area plot for Series and add test

charlesdong1991:
There seems a little bug in SamplePlot that failed area plot. It is due to after sampling, since `scol` is not assigned, it will return a full DataFrame instead of Series. And looks like area plot turns good as well after fixing it.

Screen Shot 2019-08-27 at 10 21 03 PM

Before the bug fixing, the area plot looks like:

Screen Shot 2019-08-27 at 10 40 23 PM

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -90,8 +90,9 @@ class SampledPlot:
             sampled = data._sdf.sample(fraction=float(self.fraction))
             return DataFrame(data._internal.copy(sdf=sampled)).to_pandas()
         elif isinstance(data, Series):
+            scol = data._kdf._internal.data_scols[0]
             sampled = data._kdf._sdf.sample(fraction=float(self.fraction))
-            return DataFrame(data._kdf._internal.copy(sdf=sampled)).to_pandas()
+            return DataFrame(data._kdf._internal.copy(sdf=sampled, scol=scol)).to_pandas()
         else:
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Thanks!!!!

databricks/koalas/plot.py
```diff
@@ -90,8 +90,9 @@ def get_sampled(self, data):
             sampled = data._sdf.sample(fraction=float(self.fraction))
             return DataFrame(data._internal.copy(sdf=sampled)).to_pandas()
         elif isinstance(data, Series):
+            scol = data._kdf._internal.data_scols[0]
```
HyukjinKwon: oops, it should be `data._scol` so that it respects the expression stored in Series. Let me fix it quick.

charlesdong1991: oops... i didn't see this option in `Series.py`... my bad, sorry for this, i see it is not fixed yet, i will just create a small PR to fix it quick and validate if plot still works. @HyukjinKwon

HyukjinKwon: ah, I just made a one line fix just now - I found this comment after opening the PR :D. It's fine!

charlesdong1991: 😅 oh, sorry, i waked up and saw the comment, and found it hadn't been fixed yet in master branch, i thought you might forget this issue, so i opened a small PR for this ... but thanks for the tip, appreciate it!

PR #705, opened 2019-08-28
Implement `kde` for Series

itholic:
Implement `kde` for Series.

I confirmed that it works the same as kde in pandas.

스크린샷 2019-08-28 오후 12 10 42

스크린샷 2019-08-28 오후 12 10 56

I would be grateful if someone could confirm when available.

Thanks.

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -856,7 +856,7 @@ class KoalasSeriesPlotMethods(PandasObject):
             `ind` number of equally spaced points are used.
         **kwargs : optional
             Additional keyword arguments are documented in
-            :meth:`pandas.%(this-datatype)s.plot`.
+            :ref:`plot <api.series.plot>`
 
         Returns
         -------
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -41,11 +41,11 @@ def _get_standard_kind(kind):
 
 if LooseVersion(pd.__version__) < LooseVersion('0.25'):
     from pandas.plotting._core import _all_kinds, BarPlot, BoxPlot, HistPlot, MPLPlot, PiePlot, \
-        AreaPlot, LinePlot, BarhPlot
+        AreaPlot, LinePlot, BarhPlot, KdePlot
 else:
     from pandas.plotting._core import PlotAccessor
     from pandas.plotting._matplotlib import BarPlot, BoxPlot, HistPlot, PiePlot, AreaPlot, \
-        LinePlot, BarhPlot
+        LinePlot, BarhPlot, KdePlot
     from pandas.plotting._matplotlib.core import MPLPlot
     _all_kinds = PlotAccessor._all_kinds
 
@@ -508,6 +508,17 @@ class KoalasBarhPlot(BarhPlot, TopNPlot):
         super(KoalasBarhPlot, self)._make_plot()
 
 
+class KoalasKdePlot(KdePlot, TopNPlot):
+    max_rows = 1000
+
+    def __init__(self, data, **kwargs):
+        super(KoalasKdePlot, self).__init__(self.get_top_n(data), **kwargs)
+
+    def _make_plot(self):
+        self.set_result_text(self._get_ax(0))
+        super(KoalasKdePlot, self)._make_plot()
+
+
 _klasses = [
     KoalasHistPlot,
     KoalasBarPlot,
@@ -516,6 +527,7 @@ _klasses = [
     KoalasAreaPlot,
     KoalasLinePlot,
     KoalasBarhPlot,
+    KoalasKdePlot,
 ]
 _plot_klass = {getattr(klass, '_kind'): klass for klass in _klasses}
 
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
cc @charlesdong1991 mind taking a look when you're available?

databricks/koalas/plot.py
```diff
@@ -508,6 +508,17 @@ def _make_plot(self):
         super(KoalasBarhPlot, self)._make_plot()
 
 
+class KoalasKdePlot(KdePlot, TopNPlot):
+    max_rows = 1000
+
+    def __init__(self, data, **kwargs):
+        super(KoalasKdePlot, self).__init__(self.get_top_n(data), **kwargs)
```
HyukjinKwon: From what I know, we should implement the calculation logic (see histogram for instance) via Spark DataFrame unlike other simple charts.

charlesdong1991:
yeah, could have a look later today @HyukjinKwon

HyukjinKwon:
Yes, I think so too rather than simply top k or sample. Thank you @charlesdong1991

itholic:
@charlesdong1991 @HyukjinKwon Thank you both for taking the time to review. Please let me know if there is anything else you need to implement or anything.

itholic:
@charlesdong1991 @HyukjinKwon Thank you both for taking the time to review. Please let me know if there is anything else you need to.

databricks/koalas/plot.py
charlesdong1991: oh, apparently it seems it's okay to directly use, i took a double check on Pandas, it's a common kind there unlike boxplot. i tried different cases a bit and they are all good

databricks/koalas/plot.py
```diff
@@ -826,8 +838,31 @@ def hist(self, bins=10, **kwds):
...
+            If None (default), 'scott' is used.
+            See :class:`scipy.stats.gaussian_kde` for more information.
+        ind : NumPy array or integer, optional
+            Evaluation points for the estimated PDF. If None (default),
+            1000 equally spaced points are used. If `ind` is a NumPy array, the
+            KDE is evaluated at the points passed. If `ind` is an integer,
+            `ind` number of equally spaced points are used.
+        **kwargs : optional
+            Additional keyword arguments are documented in
+            :meth:`pandas.%(this-datatype)s.plot`.
```
charlesdong1991: could you change to `koalas` one?

databricks/koalas/tests/test_series_plot.py
```diff
@@ -244,11 +244,10 @@ def test_box_summary(self):
...
-            with self.assertRaisesRegex(PandasNotImplementedError,
-                                        "method.*Series.*{}.*not implemented".format(name)):
-                getattr(ks.plot, name)()
+    def test_kde_plot(self):
+        pdf = self.pdf1
+        kdf = self.kdf1
+
+        ax1 = pdf['a'].plot("kde", colormap='Paired')
+        ax2 = kdf['a'].plot("kde", colormap='Paired')
+        self.compare_plots(ax1, ax2)
```
charlesdong1991: could you also add test for `ind` and `bw_method`, just using the pandas example is fine.
e.g. `kdf['a'].plot.kde(bw_method=0.3)`

databricks/koalas/plot.py
HyukjinKwon: I didn;t take a look closely - it was just my impression. I will double check tomorrow.

charlesdong1991:
my concern is only for large dataset, then gaussian_kde might be calculated differently in Koalas than the real result @itholic

databricks/koalas/plot.py
charlesdong1991: for big data, calculation for `gaussian_kde` is not correct. I commented it below already. i think a calculation for this needs to be implemented. @HyukjinKwon

databricks/koalas/tests/test_series_plot.py
itholic: @charlesdong1991 Thanks for your comments!! I've fixed doc comment and add more tests as you have advised :)

itholic:
@charlesdong1991 Okay i'll check some more cases when it run with large size of dataset. Thank you for taking the time to my code review!

charlesdong1991:
@itholic  you are very welcome... personally, i think those tests are already enough and this PR is okay to merge (as long as you resolve the CI failure ^^)... but let maintainer decide, he has more insights about Koalas than I have

i had the doubt for a while actually: I am just curious why setting `maxrows` to 1000? this is a fixed number from TopNPlot class, does/should users have options to change how many data they want to plot? @HyukjinKwon

databricks/koalas/tests/test_series_plot.py
```diff
@@ -265,11 +265,30 @@ def test_box_summary(self):
...
-        unsupported_functions = ['kde']
-        for name in unsupported_functions:
-            with self.assertRaisesRegex(PandasNotImplementedError,
-                                        "method.*Series.*{}.*not implemented".format(name)):
-                getattr(ks.plot, name)()
+    def test_kde_plot(self):
+        pdf = self.pdf1
+        kdf = self.kdf1
+
+        kax = pdf['a'].plot("kde", colormap='Paired')
```
charlesdong1991: btw, do you mind swapping the `ax` name here... since this seems like `pax` and below should be `kax`

databricks/koalas/tests/test_series_plot.py
```diff
@@ -67,8 +67,8 @@ def test_bar_plot(self):
         pdf = self.pdf1
         kdf = self.kdf1
 
-        ax1 = pdf['a'].plot("bar", colormap='Paired')
-        ax2 = kdf['a'].plot("bar", colormap='Paired')
+        ax1 = pdf['a'].plot(kind='bar', colormap='Paired')
+        ax2 = kdf['a'].plot(kind='bar', colormap='Paired')
```
charlesdong1991: i think you could leave it as is... the failure should be caused by that missing function, if you remove it, i think CI will pass

databricks/koalas/tests/test_series_plot.py
itholic: @charlesdong1991 Oops you're right. I made a mistake. Thanks!

itholic:
@HyukjinKwon @charlesdong1991 There is something wrong with my local build so I will regenerate this PR.

PR #706, opened 2019-08-28
CLN: Cleaner Series col name assignment in SampledPlot

charlesdong1991:
area plot still works and tests all pass locally, @HyukjinKwon thanks for pointing it out!!

Screen Shot 2019-08-28 at 7 50 24 AM

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -90,7 +90,7 @@ class SampledPlot:
             sampled = data._sdf.sample(fraction=float(self.fraction))
             return DataFrame(data._internal.copy(sdf=sampled)).to_pandas()
         elif isinstance(data, Series):
-            scol = data._kdf._internal.data_scols[0]
+            scol = data._scol
             sampled = data._kdf._sdf.sample(fraction=float(self.fraction))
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
ooops, hahah LGTM

databricks/koalas/plot.py
```diff
@@ -90,7 +90,7 @@ def get_sampled(self, data):
             sampled = data._sdf.sample(fraction=float(self.fraction))
             return DataFrame(data._internal.copy(sdf=sampled)).to_pandas()
         elif isinstance(data, Series):
-            scol = data._kdf._internal.data_scols[0]
+            scol = data._scol
```
HyukjinKwon: BTW, the issue about it was, for instance,

```python
kdf = ks.range(10)
kdf.id = kdf.id - 1
```

In this case, `data._scol` will become `Column<id> - 1` but `_kdf._internal.data_scols[0]` will become just column `Column<id>`.

charlesdong1991: ahh, get it, good tip!! thanks, man! @HyukjinKwon

PR #709, opened 2019-08-28
Add missing tests for groupby

itholic:
Added missing tests or tests that weren't running properly for groupby.

I would be grateful if someone could check when possible.

Thanks.


HyukjinKwon:
@itholic, you can show the missing coverage next time with the screenshot like you did before.

itholic:
@HyukjinKwon, Oh I was worried that maybe it would be too much information. But I will upload the screen together in the future  as you have advised me. Thank you for your advice!

PR #710, opened 2019-08-28
Implement kde for Series

itholic:
# This PR is a recovery of #705 

Implement `kde` for Series.

I confirmed that it works the same as kde in pandas.

스크린샷 2019-08-28 오후 12 10 42

스크린샷 2019-08-28 오후 12 10 56

I would be grateful if someone could confirm when available.

Thanks.

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -41,11 +41,11 @@ def _get_standard_kind(kind):
 
 if LooseVersion(pd.__version__) < LooseVersion('0.25'):
     from pandas.plotting._core import _all_kinds, BarPlot, BoxPlot, HistPlot, MPLPlot, PiePlot, \
-        AreaPlot, LinePlot, BarhPlot
+        AreaPlot, LinePlot, BarhPlot, KdePlot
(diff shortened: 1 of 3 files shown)


itholic:
@HyukjinKwon @charlesdong1991 I made a PR #705  again My local build is tangled... @.@

charlesdong1991:
as i said in the closed PR, directly using KdePlot seems okay to me although since it's using scipy's `gaussian_kde` in the back that i thought an implementation was needed... let @HyukjinKwon decide and comment on this

one thing i also mentioned in the that closed PR, i am okay with your current tests as long as CI passes, but could you please remove those unrelated changes (mainly about adding `kind` for other plots), i don't think those need to be changed, especially in this PR. @itholic

itholic:
@charlesdong1991 Thanks for comment. I've reflected your opinion related with `kind` for other plot. (and also mine for code consistency)

HyukjinKwon:
Seems like we need some manual calculation with Spark DataFrame rather than just using sample or top n. Let me take this over at #767

PR #719, opened 2019-08-29
ENH: Add scatter plot for Frame

charlesdong1991:
Screen Shot 2019-08-30 at 8 47 22 AM
Screen Shot 2019-08-30 at 8 47 40 AM
Screen Shot 2019-08-30 at 8 47 50 AM

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -1119,21 +1119,8 @@ class KoalasFramePlotMethods(PandasObject):
             The column name or column position to be used as vertical
             coordinates for each point.
         s : scalar or array_like, optional
-            The size of each point. Possible values are:
-            - A single scalar so all points have the same size.
-            - A sequence of scalars, which will be used for each point's size
-              recursively. For instance, when passing [2, 14] all points size
-              will be either 2 or 14, alternatively.
         c : str, int or array_like, optional
-            The color of each point. Possible values are:
-            - A single color string referred to by name, RGB or RGBA code,
-              for instance 'red' or '#a98d19'.
-            - A sequence of color strings referred to by name, RGB or RGBA
-              code, which will be used for each point's color recursively. For
-              instance ['green', 'yellow'] all points will be filled in green or
-              yellow, alternatively.
-            - A column name or position whose values will be used to color the
-              marker points according to a colormap.
+
         **kwds: Optional
             Keyword arguments to pass on to :meth:`databricks.koalas.DataFrame.plot`.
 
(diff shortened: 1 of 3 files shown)


charlesdong1991:
ideally, should use docstring with is identical to what pandas have, however, i encountered quite annoying linting error issue with docstring, so for now, i keep `s` and `c` docstring very simple. Will change/complete them in future PR where I update docstring for all functions.

databricks/koalas/plot.py
```diff
@@ -509,6 +509,17 @@ def _make_plot(self):
         super(KoalasBarhPlot, self)._make_plot()
 
 
+class KoalasScatterPlot(ScatterPlot, TopNPlot):
+    max_rows = 1000
```
HyukjinKwon: I think we can remove this.

charlesdong1991: ahh, yeah, removed. and added pics in description, i thought i did it yesterday. my bad

HyukjinKwon:
@charlesdong1991, before we merge, can you check if Koalas plot adds "showing top 1,000 elements only" in the output chart when the rows are larger than 1000?

charlesdong1991:
@HyukjinKwon good call! cannot really read it because font size is a bit too small to me 😅 but we probably can tell from density of points that `koalas` plot less.

Screen Shot 2019-08-30 at 9 40 53 AM

HyukjinKwon:
Thanks for testing it out, @charlesdong1991.

Ankurneural:
How to use all the rows for the plot? 
Setting, ks.set_option('compute.max_rows', 50000)
ks.set_option("display.max_rows", 50000) seems not to be working

HyukjinKwon:
`plotting.max_rows` should work for that. There's a documentation here - https://koalas.readthedocs.io/en/latest/user_guide/options.html#available-options

Ankurneural:
> `plotting.max_rows` should work for that. There's a documentation here - https://koalas.readthedocs.io/en/latest/user_guide/options.html#available-options
Thanks! I somehow missed it. It is working as anticipated.

Ankurneural:
@HyukjinKwon I am trying to make subplots using Koalas Dataframe but it comes blank.
d1 is a koalas dataframe.
Sample Code:
fig, (ax1, ax2)  = plt.subplots(1, 2)
ax1 = d1.plot.scatter(x="col1", y="col2")
ax2 = d1.plot.scatter(x="col1", y="col2") 
display(fig)

Any suggestions or help on this?

HyukjinKwon:
@Ankurneural can you create a new ticket with the full codes and output?

Issue #718, opened 2019-08-29
Add configuration value "compute.max_rows"


itholic:
elated with #715 , I'm working on "compute.max_rows"

PR #721, opened 2019-08-30
Add configuration value: compute.max_rows

itholic:
related with #718  i added compute.max_rows

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -34,6 +34,11 @@ _registered_options = {
     # For example, this value determines whether the repr() for a dataframe prints out fully or
     # just a truncated repr.
     "display.max_rows": 1000,  # TODO: None should support unlimited.
+
+    # 'compute.max_rows sets the limit of the current DataFrame. Set `None` to unlimit
+    # the input length. When the limit is set, it is executed by the shortcut by collecting
+    # the data into driver side, and then using pandas API. If the limit is unset,
+    # the operation is executed by PySpark. Default is 1000.
     "compute.max_rows": 1000,  # TODO: None should support unlimited.
 }  # type: Dict[str, Any]
 
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -34,6 +34,7 @@ _registered_options = {
     # For example, this value determines whether the repr() for a dataframe prints out fully or
     # just a truncated repr.
     "display.max_rows": 1000,  # TODO: None should support unlimited.
+    "compute.max_rows": 1000,  # TODO: None should support unlimited.
 }  # type: Dict[str, Any]
 
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1460,10 +1460,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             pdf = self.head(max_compute_count + 1)._to_internal_pandas()
             if len(pdf) > max_compute_count:
                 raise ValueError(
-                    "Current DataFrame has more then the given limit 1000 rows. "
-                    "Please set '{0}' by using 'databricks.koalas.config.set_option' "
-                    "to retrieve to retrieve more than 1000 rows. Note that, before changing the "
-                    "'{0}', this operation is considerably expensive.".format(max_compute_count))
+                    "Current DataFrame has more then the given limit {0} rows. "
+                    "Please set 'compute.max_rows' by using 'databricks.koalas.config.set_option' "
+                    "to retrieve to retrieve more than {0} rows. Note that, before changing the "
+                    "'compute.max_rows', this operation is considerably expensive."
+                    .format(max_compute_count))
             return DataFrame(pdf.transpose())
 
         # Explode the data to be pairs.
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1461,9 +1461,9 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             if len(pdf) > max_compute_count:
                 raise ValueError(
                     "Current DataFrame has more then the given limit 1000 rows. "
-                    "Please set '{}' by using 'databricks.koalas.config.set_option' "
+                    "Please set '{0}' by using 'databricks.koalas.config.set_option' "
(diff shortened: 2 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -1354,7 +1354,7 @@ def to_latex(self, buf=None, columns=None, col_space=None, header=True, index=Tr
 
     # TODO: enable doctests once we drop Spark 2.3.x (due to type coercion logic
     #  when creating arrays)
-    def transpose(self, limit: Optional[int] = 1000):
+    def transpose(self, limit: Optional[int] = get_option("compute.max_rows")):
```
HyukjinKwon: I think we can entirely remove this `limit` parameter for now to be consistent with other APIs.

databricks/koalas/config.py
```diff
@@ -34,6 +34,7 @@
     # For example, this value determines whether the repr() for a dataframe prints out fully or
     # just a truncated repr.
     "display.max_rows": 1000,  # TODO: None should support unlimited.
+    "compute.max_rows": 1000,  # TODO: None should support unlimited.
```
HyukjinKwon: It should be better to add a comment above this configuration:

```
# 'compute.max_rows sets the limit of the current DataFrame. Set `None` to unlimit
# the input length. When the limit is set, it is executed by the shortcut by collecting
# the data into driver side, and then using pandas API. If the limit is unset,
# the operation is executed by PySpark. Default is 1000.
```

databricks/koalas/frame.py
HyukjinKwon: and don't forget to remove the documentation in its `Parameters`.

HyukjinKwon: Related changes can be easily found via:

1. View File
Screen Shot 2019-08-30 at 11 00 18 AM

2. Blame
Screen Shot 2019-08-30 at 11 00 24 AM

3. Go to the change and click the commit
Screen Shot 2019-08-30 at 11 00 46 AM

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1200,13 +1201,15 @@ def test_transpose(self):
             columns=['score', 'kids', 'age'])
         kdf2 = ks.from_pandas(pdf2)
 
+        set_option("compute.max_rows", None)
         self.assertEqual(
```
HyukjinKwon: Let's do:

```python
try:
    self.assertEqual(
        ...
finally:
    reset_option("compute.max_rows")
```

to avoid this test failure affects other tests related to `compute.max_rows` option.

databricks/koalas/frame.py
```diff
@@ -1461,14 +1455,15 @@ def transpose(self, limit: Optional[int] = 1000):
...
+        if max_compute_count:
+            pdf = self.head(max_compute_count + 1)._to_internal_pandas()
+            if len(pdf) > max_compute_count:
                 raise ValueError(
-                    "Current DataFrame has more then the given limit %s rows. Please use "
-                    "df.transpose(limit=<maximum number of rows>) to retrieve more than %s rows. "
-                    "Note that, before changing the given 'limit', this operation is considerably "
-                    "expensive." % (limit, limit))
+                    "Current DataFrame has more then the given limit {0} rows. "
+                    "Please use 'databricks.koalas.config.set_option' to retrieve more than "
```
HyukjinKwon:
```
Please set 'compute.max_rows' by using 'databricks.koalas.config.set_option' to retrieve ...
```

databricks/koalas/frame.py
```diff
@@ -1365,24 +1365,18 @@ def transpose(self, limit: Optional[int] = 1000):
...
 
+                >>> from databricks.koalas.config import get_option, set_option
+                >>> set_option('compute.max_rows', 1000)
                 >>> ks.DataFrame({'a': range(1001)}).transpose()  # doctest: +NORMALIZE_WHITESPACE
                 Traceback (most recent call last):
                   ...
                 ValueError: Current DataFrame has more then the given limit 1000 rows.
-                Please use df.transpose(limit=<maximum number of rows>) to retrieve more than
-                1000 rows. Note that, before changing the given 'limit', this operation is
+                Please use 'databricks.koalas.config.set_option' to retrieve more than
```
HyukjinKwon: ditto

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1222,9 +1225,11 @@ def test_transpose(self):
                                                              ('rg2', 'z')]))
         kdf3 = ks.from_pandas(pdf3)
 
+        set_option("compute.max_rows", None)
```
HyukjinKwon: ditto

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1222,9 +1227,13 @@ def test_transpose(self):
                                                              ('rg2', 'z')]))
         kdf3 = ks.from_pandas(pdf3)
 
-        self.assertEqual(
-            repr(pdf3.transpose().sort_index()),
-            repr(kdf3.transpose(limit=None).sort_index()))
+        try:
+            set_option("compute.max_rows", None)
```
HyukjinKwon: set_option("compute.max_rows", None) should be out of this try block to be consistent with other codes you're fixing.

databricks/koalas/frame.py
```diff
@@ -1461,14 +1455,16 @@ def transpose(self, limit: Optional[int] = 1000):
         1    float64
         dtype: object
         """
-        if limit is not None:
-            pdf = self.head(limit + 1)._to_internal_pandas()
-            if len(pdf) > limit:
+        max_compute_count = get_option("compute.max_rows")
+        if max_compute_count:
```
HyukjinKwon: It intentionally used `is not None` to make it clear `None` means unlimited case. Otherwise, empty list `[]` or other false-equivalent values could mean unlimited case as well.

It's rather preference matter but let's keep it consistent. Let's do ` if max_compute_count is not None:`

itholic:
@HyukjinKwon Thanks, I've resolved all you commented.

HyukjinKwon:
@itholic, can you resolve conflicts and rebase this? I will merge.

itholic:
@HyukjinKwon , Resolved conflicts. Thanks !

Issue #722, opened 2019-08-30
Add configuration value: "compute.default_index_type"


itholic:
elated with #715 , I'm working on "compute.default_index_type"

PR #723, opened 2019-08-30
Add configuration value: compute.default_index_type

itholic:
related with #722  i added compute.default_index_type

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -35,26 +35,8 @@ _registered_options = {
     # just a truncated repr.
     "display.max_rows": 1000,  # TODO: None should support unlimited.
 
-    # This sets the default index type
-    # There are three types of default index that can be configured by `compute.default_index_type`
-    # environment variable.
-    # - sequence: It implements a sequence that increases one by one, by Window function without
-    #     specifying partition. Therefore, it ends up with whole partition in single node.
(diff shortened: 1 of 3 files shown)


databricks/koalas/config.py
```diff
@@ -34,6 +34,27 @@
     # For example, this value determines whether the repr() for a dataframe prints out fully or
     # just a truncated repr.
     "display.max_rows": 1000,  # TODO: None should support unlimited.
+
+    # This sets the default index type
+    # There are three types of default index that can be configured by `compute.default_index_type`
+    # environment variable.
+    # - sequence: It implements a sequence that increases one by one, by Window function without
```
HyukjinKwon: Let's just say

```
This sets the default index type: sequence, distributed and distributed-sequence.
```

I will create a documentation page to describe this index type later separately in anther PR.

HyukjinKwon: you can fix https://github.com/databricks/koalas/blob/34088a4badc916285cc4872ea8364170880cc576/databricks/koalas/tests/test_default_index.py#L30 test as well

itholic: @HyukjinKwon Thanks! i've fixed both of them.

HyukjinKwon:
Thanks, merged.

Issue #724, opened 2019-08-30
Add configuration value: "compute.ops_on_diff_frames"


itholic:
elated with #715 , I'm working on "compute.ops_on_diff_frames"

PR #725, opened 2019-08-30
Add configuration value: compute.ops_on_diff_frames

itholic:
related with #724  i added compute.ops_on_diff_frames

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -39,7 +39,7 @@ _registered_options = {
     # For example, 'combine_frames' function internally performs a join operation which can be
     # expensive in general.
     # So, if `compute.ops_on_diff_frames` variable is not True, that method throws an exception.
-    "compute.ops_on_diff_frames": True,
+    "compute.ops_on_diff_frames": False,
 }  # type: Dict[str, Any]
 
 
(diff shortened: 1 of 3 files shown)


databricks/koalas/config.py
```diff
@@ -34,6 +34,12 @@
     # For example, this value determines whether the repr() for a dataframe prints out fully or
     # just a truncated repr.
     "display.max_rows": 1000,  # TODO: None should support unlimited.
+
+    # This determines whether or not to operate between two different dataframs.
+    # For example, 'combine_frames' function internally performs a join operation which can be
+    # expensive in general.
+    # So, if `compute.ops_on_diff_frames` variable is not True, that method throws an exception.
+    "compute.ops_on_diff_frames": True,
```
HyukjinKwon: It defaults `false`.

itholic: @HyukjinKwon Thanks. i've fix it.

databricks/koalas/utils.py
```diff
@@ -39,11 +38,12 @@ def combine_frames(this, *args, how="full"):
     the columns names from both DataFrames
 
     It internally performs a join operation which can be expensive in general.
-    So, if `OPS_ON_DIFF_FRAMES` environment variable is not set,
+    So, if `compute.ops_on_diff_frames` variable is not True,
```
HyukjinKwon: `variable` -> `option`.

HyukjinKwon: `is not True` -> `is False`

PR #728, opened 2019-08-30
Add plotting.max_rows

charlesdong1991:


databricks/koalas/tests/test_frame_plot.py
```diff
@@ -7,6 +7,7 @@
 import numpy as np
 
 from databricks import koalas
+from databricks.koalas.config import set_option
```
HyukjinKwon: @charlesdong1991, how about we just set? Although not all tests are dependent on this configuration but I think it's fine to set globally for the test cases below since this configuration itself is specific to plot and the test cases here are dedicated for plot.

```python
    @classmethod
    def setUpClass(cls):
        super(DataFramePlotTest, cls).setUpClass()
        set_option('plotting.max_rows', 1000)

    @classmethod
    def tearDownClass(cls):
        super(DataFramePlotTest, cls).tearDownClass()
        reset_option('plotting.max_rows')
```

databricks/koalas/tests/test_frame_plot.py
```diff
@@ -201,3 +202,13 @@ def test_missing(self):
             with self.assertRaisesRegex(PandasNotImplementedError,
                                         "method.*DataFrame.*{}.*not implemented".format(name)):
                 getattr(ks.plot, name)()
+
+    def test_topn_max_rows(self):
```
HyukjinKwon: +1 to have this test.

databricks/koalas/tests/test_frame_plot.py
```diff
@@ -201,3 +202,13 @@ def test_missing(self):
...
+        pdf = pd.DataFrame(np.random.rand(2500, 4), columns=['a', 'b', 'c', 'd'])
+        kdf = koalas.from_pandas(pdf)
+
+        set_option("plotting.max_rows", 2000)
```
HyukjinKwon: The other tests look failed because this configuration wasn't reset back.

databricks/koalas/tests/test_frame_plot.py
charlesdong1991: nice idea! thanks! @HyukjinKwon

databricks/koalas/plot.py
```diff
@@ -51,7 +52,7 @@ def _get_standard_kind(kind):
 
 
 class TopNPlot:
-    max_rows = 1000
+    max_rows = get_option("plotting.max_rows")
```
HyukjinKwon: Oh! this shouldn't be here. It will statically assign the value into the class value `TopNPlot. max_rows `. `get_option("plotting.max_rows")` should be called within each method, `get_top_n` and `set_result_text`.

databricks/koalas/tests/test_frame_plot.py
```diff
@@ -15,6 +16,17 @@
 
 
 class DataFramePlotTest(ReusedSQLTestCase, TestUtils):
+
+    @classmethod
```
HyukjinKwon: Can we add this same thing into `SeriesPlotTest` too?

databricks/koalas/plot.py
charlesdong1991: yes!! i also found this caused failure everytime... i am refactoring ^^

thanks!

databricks/koalas/tests/test_frame_plot.py
charlesdong1991: sure!

databricks/koalas/tests/test_frame_plot.py
```diff
@@ -201,3 +214,12 @@ def test_missing(self):
             with self.assertRaisesRegex(PandasNotImplementedError,
                                         "method.*DataFrame.*{}.*not implemented".format(name)):
                 getattr(ks.plot, name)()
+
+    def test_topn_max_rows(self):
+        self.setUpClass()
```
HyukjinKwon: Seems we can remove this since `setUpClass` will be called when this test class is started.

databricks/koalas/tests/test_series_plot.py
```diff
@@ -76,12 +87,13 @@ def test_bar_plot(self):
         self.compare_plots(ax1, ax2)
 
     def test_bar_plot_limited(self):
+        self.setUpClass()
```
HyukjinKwon: This one too. Unittest will automatically call `setUpClass` when it initialises the set of tests, and `tearDownClass` when the tests are finished.

databricks/koalas/tests/test_frame_plot.py
charlesdong1991: maybe this should stay? since only for this test, `max_rows` will be set to 2000? @HyukjinKwon

databricks/koalas/tests/test_series_plot.py
charlesdong1991: i will remove this.

databricks/koalas/tests/test_frame_plot.py
charlesdong1991: i see, will remove

HyukjinKwon:
Thanks for doing this, @charlesdong1991 ! merged.

PR #729, opened 2019-08-31
Add missing tests for datetime & Remove unsupported property

itholic:
Add missing test for `weekday`
스크린샷 2019-08-31 오후 3 01 12

Remove unsupported property `millisecond` for Series Datetime Properties.

스크린샷 2019-08-31 오후 2 49 55

### DIFF
--- a/databricks/koalas/datetimes.py
+++ b/databricks/koalas/datetimes.py
@@ -106,14 +106,6 @@ class DatetimeMethods(object):
         return _wrap_accessor_spark(self, F.second, LongType()).alias(self.name)
 
     @property
-    def millisecond(self) -> 'ks.Series':
-        """
-        The milliseconds of the datetime.
-        """
-        return _wrap_accessor_pandas(
-            self, lambda x: x.dt.millisecond, LongType()).alias(self.name)
-
-    @property
     def microsecond(self) -> 'ks.Series':
         """
         The microseconds of the datetime.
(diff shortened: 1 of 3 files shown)


databricks/koalas/datetimes.py
```diff
@@ -105,14 +105,6 @@ def second(self) -> 'ks.Series':
         """
         return _wrap_accessor_spark(self, F.second, LongType()).alias(self.name)
 
-    @property
-    def millisecond(self) -> 'ks.Series':
-        """
-        The milliseconds of the datetime.
-        """
-        return _wrap_accessor_pandas(
```
HyukjinKwon: Why did we remove this? Does this not work?

itholic: I thought these were the reuses of the dt functions from the Pandas Series.
However, in the screenshots I added to PR, the Pandas dt Property does not support milliseconds.
So I thought this was probably added by mistake.
And It doesn't actually work as you asked, (as in the following screenshot.)

스크린샷 2019-09-01 오후 2 06 51

itholic: @HyukjinKwon, Thanks to Review.
I thought these were the reuses of the Pandas Series dt Property.
However, in the screenshots I added to PR, the Pandas Series dt Property does not support milliseconds.
So I thought this was probably added by mistake.
And It doesn't actually work as you asked, (as in the following screenshot.)

스크린샷 2019-09-01 오후 2 06 51

PR #732, opened 2019-09-02
Add schema inference support at DataFrame.transform

HyukjinKwon:
This PR adds the support for schema inference support at DataFrame.transform.

Now, `DataFrame.transform` can be used as below:

```python

ks.range(10).transform(lambda x: x ** 2)
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -43,7 +43,7 @@ from pyspark.sql.types import (BooleanType, ByteType, DecimalType, DoubleType, F
                                IntegerType, LongType, NumericType, ShortType, StructType)
 from pyspark.sql.utils import AnalysisException
 from pyspark.sql.window import Window
-from pyspark.sql.functions import pandas_udf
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -1575,20 +1584,48 @@ def transform(self, func):
...
+        should_infer_schema = return_sig is None
 
-        wrapped = ks.pandas_wraps(func)
-        applied = []
-        for column in self._internal.data_columns:
-            applied.append(wrapped(self[column]).rename(column))
+        if should_infer_schema:
+            # Here we execute with the first 1000 to get the return type.
+            # If the records were less than 1000, it uses pandas API directly for a shortcut.
+            limit = 1000
```
HyukjinKwon: We should add this as a configuration as well later.

HyukjinKwon:
cc @RainFung FYI

HyukjinKwon:
Merged.

PR #733, opened 2019-09-02
Add missing tests for sql

itholic:
As you can see in the following screenshot,

There were some missing sql test using Series, so I added it to Doc test.

스크린샷 2019-09-02 오전 11 33 44

could someone check this change maybe if available?

Thanks.

### DIFF
--- a/databricks/koalas/sql.py
+++ b/databricks/koalas/sql.py
@@ -122,6 +122,14 @@ def sql(query: str, globals=None, locals=None, **kwargs) -> DataFrame:
     0  1  3
     1  2  4
     2  2  5
+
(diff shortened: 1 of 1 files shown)


databricks/koalas/sql.py
```diff
@@ -122,6 +122,14 @@ def sql(query: str, globals=None, locals=None, **kwargs) -> DataFrame:
     0  1  3
     1  2  4
     2  2  5
+
+    And also possible to query using Series.
```
HyukjinKwon: `Also, it is possible ...`

itholic: Thanks. fixed it!

PR #734, opened 2019-09-02
Add some missing exception tests for base.py

itholic:
Found some missing exception tests in base.py

스크린샷 2019-09-02 오후 2 40 33
스크린샷 2019-09-02 오후 2 41 22
스크린샷 2019-09-02 오후 2 42 41
스크린샷 2019-09-02 오후 2 42 57

i added them some for doctest,

one for pytest 

(but this test is currently being skipped for some reason 

스크린샷 2019-09-02 오후 3 14 26

so maybe does not directly affect coverage.)

could someone check this when if available.

Thanks.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -385,6 +385,9 @@ class IndexOpsMixin(object):
 
         >>> ser.rename("a").to_frame().set_index("a").index.astype('int64')
         Int64Index([1, 2], dtype='int64', name='a')
+
+        >>> ser.astype('int63')
+        ValueError: Type int63 not understood
         """
         from databricks.koalas.typedef import as_spark_type
         spark_type = as_spark_type(dtype)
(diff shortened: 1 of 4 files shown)


databricks/koalas/base.py
```diff
@@ -385,6 +385,9 @@ def astype(self, dtype):
 
         >>> ser.rename("a").to_frame().set_index("a").index.astype('int64')
         Int64Index([1, 2], dtype='int64', name='a')
+
+        >>> ser.astype('int63')
+        ValueError: Type int63 not understood
```
HyukjinKwon: For such test, we should add it under `koalas/tests/...` rather than doctest because it becomes the part of documentation.

itholic: Thanks for comment. i'll fix and commit again

itholic: i've fixed them!

PR #735, opened 2019-09-02
Fix rpow and rfloordiv to use proper operators in Series

itholic:
While checking the series.py file,

there was an incorrectly implemented operator.

(I think maybe someone forgot to change the code in the process of copy & paste by mistake)

스크린샷 2019-09-02 오후 3 18 56

could someone check this changes if available.

Thanks :)

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -204,11 +204,11 @@ d     NaN
 Name: a, dtype: float64
 
 >>> df.a.rpow(df.b)
-a    4.0
-b    NaN
-c   16.0
-d    NaN
-Name: a, dtype: float64
+a     4.0
+b     NaN
+c    16.0
+d     NaN
+Name: b, dtype: float64
 """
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -465,6 +465,8 @@ def rsub(self, other):
     def mod(self, other):
         return (self % other).rename(self.name)
 
+    mode = mod
```
HyukjinKwon: @itholic, is `mode` same as `mod`? Seems not:

```python
>>> pdf.a.mode(1)
0    1
1    2
2    3
dtype: int64
>>> pdf.a.mod(1)
0    0
1    0
2    0
Name: a, dtype: int64
```

HyukjinKwon: You can remove it here and implement this later in a separate PR.

itholic: @HyukjinKwon Oops, sorry it was mistaken. i've roll back this change

PR #737, opened 2019-09-02
ENH: Add plotting.sample_ratio in config

charlesdong1991:


databricks/koalas/plot.py
```diff
@@ -82,19 +82,32 @@ def set_result_text(self, ax):
...
+        sample_ratio = get_option("plotting.sample_ratio")
+
+        # if sample_ratio is default None, make sure the records are roughly 1000.
+        if sample_ratio is None:
+            fraction = 1 / (len(data) / 1000)
+        else:
+            fraction = sample_ratio
+
+        # check if fraction is larger than 1, and ceil it to 1 if so
+        if fraction > 1:
```
HyukjinKwon: We might need a way to check the value set in its options at https://github.com/databricks/koalas/pull/738 .. hm .. let me think about this a bit more.

charlesdong1991: hmm, how about checking the value here? so if fraction is given a number larger than 1, the raise an error/warning message? feel like checking the value as well in options sounds a bit overkill

HyukjinKwon: Actually there are some other cases where we need to check values like default index type .. I tried to integrate convert / check logics at #739 to make it less overkill. Please take a look when you are available.

HyukjinKwon:
Let me just merge and make a followup.

Thanks for working on this @charlesdong1991

PR #738, opened 2019-09-02
BUG: Allow default config option to be None

charlesdong1991:
In current setup, when using `set_option`, it will first check if the given value has the same type of default value. In this way, if the default is set to `None` which will result in `NoneType` in the type checking and thus fail to use `set_option`.

And because of this, the PR #737 will fail, if this PR is merged, the #737 will be changed to adopt this.


HyukjinKwon:
Let's track it here https://github.com/databricks/koalas/pull/739

PR #739, opened 2019-09-03
Add 'Option` class to support type check and value check in options

HyukjinKwon:
This PR takes over https://github.com/databricks/koalas/pull/738, and this fix was inspired by the PR. So the commit credits to, and is authored by @charlesdong1991.

This PR adds `Option` class to allow various cases such as type checking, value checking, etc.

See the usage below as a example

```python
Option(
    key='compute.default_index_type',
    doc=(
        "This sets the default index type: sequence, distributed and distributed-sequence."),
    default='sequence',
    types=str,
    check_func=(
        lambda v: v in ('sequence', 'distributed', 'distributed-sequence'),
        "Index type should be one of 'sequence', 'distributed', 'distributed-sequence'."))
```

While I am here, I also fixed `plotting.sample_ratio` respects `plotting.max_rows` by default if this is not set.

Also, let `DataFrame.transform` respect `compute.max_rows` as well with minor cleanups.

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -15,10 +15,10 @@
 #
 
 """
-Infrastructure of configuration for Koalas.
+Infrastructure of options for Koalas.
 """
 import json
-from typing import Dict, Union, Any
+from typing import Union, Any, Tuple, Callable, List
 
 from pyspark._globals import _NoValue, _NoValueType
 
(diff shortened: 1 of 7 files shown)


HyukjinKwon:
cc @charlesdong1991 and @itholic too

databricks/koalas/config.py
```diff
@@ -126,14 +138,18 @@ def reset_option(key: str) -> None:
...
 
-    if value is None:
-        return  # None is allowed for all types.
-    if value is not _NoValue and not isinstance(value, type(_registered_options[key])):
-        raise TypeError("The configuration value for '%s' was %s; however, %s is expected." % (
-            key, type(value), type(_registered_options[key])))
+
+def _convert_value(key: str, value: Any) -> Any:
+    convert_func = _registered_options[key][1]
+    if value is not _NoValue:
```
charlesdong1991: maybe still need to give an option if `value is None`?

charlesdong1991: right now, it seems to me for instance, if `config.test.something: (100, int)`, and then someone sets this option to `None`, and this `None` is allowed value for this option, i guess there will be an error raised?

HyukjinKwon: Yup, now it will throw an error. However, `None` can be also allowed explicitly such as https://github.com/databricks/koalas/pull/739/files#diff-4e0dcdf52811413fc3041eb7a9ef6c10R31

Currently, `None` is not allowed for the options we support currently.

charlesdong1991: ok, so for options that allow `None`, you mean something like `plotting.max_rows: (1000, (int, None))` will work then?

databricks/koalas/config.py
```diff
@@ -28,31 +28,42 @@
 __all__ = ['get_option', 'set_option', 'reset_option']
 
 
-# dict to store registered options and their default values (key -> default).
+def _make_check_func(func, msg):
+    def check(v):
+        if func(v):
+            return v
```
HyukjinKwon: Hmm, but this way won't allow custom confusion... Let me think a bit more about it

charlesdong1991:
is something like: `option-name: (default_value, allowed_type, restricted_values)` more readable and easier for maintenance? so it practically looks like: `display.max_columns: (1000, (None, int), None)`, if there are some restrictions on values like `default_index_type` has, so `default_index_type : ('sequence', (str), ('str1', 'str2', 'str3'))`? and then adjust the check function a bit?

databricks/koalas/config.py
HyukjinKwon: Ah yup. I was thinking that's better.

HyukjinKwon:
That sounds fine except that `restricted_values` might has to be a function in that case as well. For instance, `value > 0`.  As a reference, Spark is doing in a builder pattern like way to cover those possibilities.

https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/internal/SQLConf.scala#L301-L303

I don't want to make it that complicated here for now .. though.

charlesdong1991:
@HyukjinKwon ahh, i see... 
yeah, i prefer to keep things quite simple here as well...

databricks/koalas/config.py
```diff
@@ -28,33 +28,87 @@
...
-    # For example, 'combine_frames' function internally performs a join operation which can be
-    # expensive in general.
-    # So, if `compute.ops_on_diff_frames` variable is not True, that method throws an exception.
-    "compute.ops_on_diff_frames": False,
-
-    # This sets the default index type: sequence, distributed and distributed-sequence.
-    "compute.default_index_type": "sequence",
-}  # type: Dict[str, Any]
-
+class Option:
```
HyukjinKwon: I am going to add docstring later if we're good with the current approach.

HyukjinKwon:
How do you like the current way @charlesdong1991 and @ueshin?

databricks/koalas/config.py
HyukjinKwon: I am going to add docstring later tomorrow.

HyukjinKwon:
Oops, I am going to squash the commits into one so that it credits to @charlesdong1991

charlesdong1991:
looks really good at first glance!!! @HyukjinKwon i will take a deeper look once i have time, but looks very nice!!

pls don't need to `force-push` to give credit to me, appreciate a lot, but it's all your work, should be yours!! ^^

databricks/koalas/config.py
```diff
@@ -28,33 +28,87 @@
...
+    # TODO: None should support unlimited.
+    Option(
+        key='display.max_rows',
+        doc=(
+            "This sets the maximum number of rows koalas should output when printing out "
+            "various output. For example, this value determines whether the repr() for a "
+            "dataframe prints out fully or just a truncated repr."),
+        default=1000,
+        types=int,
+        check_func=(lambda v: v >= 0, "'display.max_rows' should be larger than 0.")),
```
charlesdong1991: maybe `lambda v: v>0`? since in the error message it says `should be larger than 0`.
Or alternatively, rephrase the error message a bit?

the same below

databricks/koalas/config.py
```diff
@@ -28,33 +28,87 @@
...
+            "the first 1000 data points will be used for plotting, and this number will be "
+            "seen on the right up corner."),
+        default=1000,
+        types=(int, type(None)),
+        check_func=(
+            lambda v: v is None or v >= 0, "'plotting.max_rows' should be larger than 0.")),
+
+    # TODO: None should support unlimited.
+    Option(
+        key='compute.max_row',
```
charlesdong1991: shouldn't this be called `compute.max_rows`? ^^

databricks/koalas/config.py
charlesdong1991: you could also do it in separate PR ^^

databricks/koalas/config.py
HyukjinKwon: Oops

HyukjinKwon:
I am merging this since the release date is tomorrow.

PR #742, opened 2019-09-04
Support None to unlimit 'display.max_rows'

HyukjinKwon:
This PR adds the support to unlimit the number of rows at 'display.max_rows'.

```python
>>> ks.set_option("display.max_rows", None)
>>> ks.range(1001)
        id
0        0
1        1
...
1018  1018
1019  1019

[1020 rows x 1 columns]
```

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -106,21 +106,23 @@ class Option:
 
 # Available options.
 _options = [
-    # TODO: None should support unlimited.
     Option(
         key='display.max_rows',
         doc=(
             "This sets the maximum number of rows koalas should output when printing out "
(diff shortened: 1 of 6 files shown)


databricks/koalas/config.py
```diff
@@ -130,7 +132,6 @@ def validate(self, v: Any) -> None:
             lambda v: v is None or v >= 0,
             "'compute.max_rows' should be greater than or equal to 0.")),
 
-    # TODO: None should support unlimited.
```
HyukjinKwon: `shortcut_limit` cannot support `None` as at least it needs empty dataframe to infer its type when its needed.

PR #702, opened 2019-08-27
Implemented `nbytes` of Series

itholic:
This is my first open source implementation so I'm not sure if it's the correct way,

but I've made it work the same as nbytes in pandas.

```python
>>> import databricks.koalas as ks
>>> import pandas as pd
>>> 
>>> pser = pd.Series(list('abc'))
>>> kser = ks.Series(list('abc'))
>>> 
>>> pser.nbytes
24
>>> kser.nbytes
24
```

I would be grateful if you could confirm when you available.

PR #744, opened 2019-09-04
Explicitly don't support nbytes in Series

itholic:
As discussed in issue #702 ,
not to explicitly use nbytes from the Series.

### DIFF
--- a/databricks/koalas/missing/series.py
+++ b/databricks/koalas/missing/series.py
@@ -156,6 +156,10 @@ class _MissingPandasLikeSeries(object):
     real = unsupported_property(
         'real',
         reason="If you want to collect your data as an NumPy array, use 'to_numpy()' instead.")
+    nbytes = unsupported_property(
+        'nbytes',
+        reason="Since nbytes trigger whole computation to get the size, which is pretty "
+               "critical when you handle large dataset, we don't support this for now")
 
     # Functions we won't support.
     memory_usage = common.memory_usage(unsupported_function)
--- a/databricks/koalas/missing/series.py
+++ b/databricks/koalas/missing/series.py
@@ -32,7 +32,6 @@ class _MissingPandasLikeSeries(object):
     # Properties
     axes = unsupported_property('axes')
     iat = unsupported_property('iat')
-    nbytes = unsupported_property('nbytes')
 
     # Deprecated properties
     blocks = unsupported_property('blocks', deprecated=True)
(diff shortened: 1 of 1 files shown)


itholic:
@HyukjinKwon, I've made a PR about nbytes of series here. could you check this when available?

databricks/koalas/missing/series.py
```diff
@@ -156,6 +156,10 @@ class _MissingPandasLikeSeries(object):
     real = unsupported_property(
         'real',
         reason="If you want to collect your data as an NumPy array, use 'to_numpy()' instead.")
+    nbytes = unsupported_property(
+        'nbytes',
+        reason="Since nbytes trigger whole computation to get the size, which is pretty "
+               "critical when you handle large dataset, we don't support this for now")
```
HyukjinKwon: @itholic, I think we can mention like check Spark UI to monitor memory usage or shuffle IO. Also, please avoid abbreviation like `don't` in public error message or documentation.

databricks/koalas/missing/series.py
```diff
@@ -156,6 +156,10 @@ class _MissingPandasLikeSeries(object):
     real = unsupported_property(
         'real',
         reason="If you want to collect your data as an NumPy array, use 'to_numpy()' instead.")
+    nbytes = unsupported_property(
+        'nbytes',
+        reason="Since nbytes trigger whole computation to get the size, which is pretty "
```
HyukjinKwon: You can type it like:

```
'nbytes' requires to compute whole dataset. You can calculate manually it, with its 'itemsize', 
by explicitly executing its count. Use Spark's web UI to monitor disk and memory usage of 
your application in general.
```

databricks/koalas/missing/series.py
itholic: oh i understand it. Thanks for comment !

databricks/koalas/missing/series.py
itholic: Thank you for help. i've fixed comment and commit it. :)

databricks/koalas/missing/series.py
```diff
@@ -156,6 +156,11 @@ class _MissingPandasLikeSeries(object):
     real = unsupported_property(
         'real',
         reason="If you want to collect your data as an NumPy array, use 'to_numpy()' instead.")
+    nbytes = unsupported_property(
```
HyukjinKwon: @itholic, can you remove `nbytes = unsupported_property('nbytes')` at `_MissingPandasLikeSeries`?

itholic: Sure, i've fixed it. Thanks !

PR #747, opened 2019-09-05
Update the lower version of pandas.

ueshin:
As a result of #746, update the lower version of pandas to 0.23.2.
Closes #746.


ueshin:
merging.

PR #753, opened 2019-09-06
Add read_json and let to_json use spark.write.json

HyukjinKwon:
This PR proposes to use `spark.write.json` API to enable distributed computation when `path` is specified. If `path` is not specified, it just calls pandas' `to_json` as was.

Also, this PR adds `read_json` API.


HyukjinKwon:
Let me merge this. although it's a breaking change, we should use spark's API whenever possible.

PR #754, opened 2019-09-06
CLN: Simplify TopN plot

charlesdong1991:
feel like `to_frame` is a bit redundant, remove it should also work.

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -58,9 +58,7 @@ class TopNPlot:
         max_rows = get_option("plotting.max_rows")
         # Simply use the first 1k elements and make it into a pandas dataframe
         # For categorical variables, it is likely called from df.x.value_counts().plot.xxx().
-        if isinstance(data, Series):
-            data = data.head(max_rows + 1).to_pandas().to_frame()
-        elif isinstance(data, DataFrame):
+        if isinstance(data, (Series, DataFrame)):
             data = data.head(max_rows + 1).to_pandas()
         else:
             ValueError("Only DataFrame and Series are supported for plotting.")


charlesdong1991:
may I ask a question about Spark? @HyukjinKwon 

These two days, I used docker and then the hostname of docker was set to `127.0.0.1`, which seems to be collided with spark. and got warning of `Your hostname, 127.0.0.1 resolves to a loopback address: 127.0.0.1; using 10.95.77.7 instead (on interface en0)`. And even I remove the hostname for docker in `/etc/hosts`, when i run tests, somehow, all tests are deselected, and I could not test my change at all. You are an expert of Spark, could you give some hints of how to fix this issue? Thanks a lot!

HyukjinKwon:
> Your hostname, 127.0.0.1 resolves to a loopback address: 127.0.0.1; using 10.95.77.7 instead (on interface en0)

Yes, I think it's related with `/etc/hosts` as you did. Is it possible to show the contents at `/etc/hosts`?

Maybe you should:

```
127.0.0.1      your_hostname
```

or `SPARK_LOCAL_IP="127.0.0.1" to load-spark-env.sh`

charlesdong1991:
thanks for your quick reply, here is how my `/etc/hosts` looks like, i commented out docker part, but seems it is still not working:

Screen Shot 2019-09-06 at 10 01 27 AM

charlesdong1991:
where does this `load-spark-env.sh` usually locate?

HyukjinKwon:
Are you able to ping out `10.95.77.7` by `ping 10.95.77.7`? If it's a testing purpose, you could just set `SPARK_LOCAL_IP="127.0.0.1"` to` load-spark-env.sh`. This is located under `$SPARK_HOME`.

If you installed PySpark via `pip`, you can find `SPARK_HOME` via:

```python
>>> import pyspark
>>> pyspark.__file__
'/.../.../.../.../spark/python/pyspark/__init__.py'
```

`'/.../.../.../.../spark` is the SPARK_HOME

charlesdong1991:
i think i installed in my virtualenv `koalas-dev`

Screen Shot 2019-09-06 at 10 11 37 AM

but i could not find `load-spark-env.sh`, need your help again, sorry for this. 😓

HyukjinKwon:
oh, it's alright. I usually use PySpark from what I manually build :D.

Can you try:

```bash
mkdir conf
cd conf
export SPARK_CONF_DIR=`pwd`
echo "spark.driver.host     localhost" > spark-defaults.conf
```

? `spark.driver.host` sets what `SPARK_LOCAL_IP` sets.

HyukjinKwon:
BTW, if possible, you can use the Apache channel to download (https://spark.apache.org/downloads.html).

```bash
curl -O http://apache.mirror.cdnetworks.com/spark/spark-2.4.4/spark-2.4.4-bin-hadoop2.7.tgz
tar xzvf spark-2.4.4-bin-hadoop2.7.tgz
cd spark-2.4.4-bin-hadoop2.7
export SPARK_HOME=`pwd`
export PYTHONPATH=$(ZIPS=("$SPARK_HOME"/python/lib/*.zip); IFS=:; echo "${ZIPS[*]}"):$PYTHONPATH
```

will make pyspark works identically.

In my case I set both `export SPARK_HOME ...` and `export PYTHONPATH ...` at my `~/.bashrc`.

charlesdong1991:
> oh, it's alright. I usually use PySpark from what I manually build :D.
ahh, haha, yeah, i get it.. ^^

where shall I create this `conf` folder? `/../../........./pyspark/`?

HyukjinKwon:
It's fine. At any location.

charlesdong1991:
> BTW, if possible, you can use the Apache channel to download (https://spark.apache.org/downloads.html).
> 
> ```shell
> curl -O http://apache.mirror.cdnetworks.com/spark/spark-2.4.4/spark-2.4.4-bin-hadoop2.7.tgz
> tar xzvf spark-2.4.4-bin-hadoop2.7.tgz
> cd spark-2.4.4-bin-hadoop2.7
> export SPARK_HOME=`pwd`
> export PYTHONPATH=$(ZIPS=("$SPARK_HOME"/python/lib/*.zip); IFS=:; echo "${ZIPS[*]}"):$PYTHONPATH
> ```
> 
> will make pyspark works identically.
> 
> In my case I set both `export SPARK_HOME ...` and `export PYTHONPATH ...` at my `~/.bashrc`.

ok! i will try doing it!

charlesdong1991:
emm, could you please check this? still have the same issue 😓 :

Screen Shot 2019-09-06 at 10 51 48 AM

HyukjinKwon:
oh! actually looks your Spark seems fine .. maybe the tests are being deselected for another reason ... do you use `./dev/pyspark` to run the tests?

charlesdong1991:
nah, i am always using `./dev/pytest -k tests/test_config.py` for instance.

Two days ago, right after i installed docker for my work purpose, and then this command always deselected all tests somehow... shall I switch to `./dev/pyspark`? i just tried, doesnt seem correct to me.

HyukjinKwon:
Hmm .. when you specify `-k ...` it's suppose to deselect some tests. If you run the tests with `./dev/pytest` it should select all.  Maybe you can try either:

1.  `--doctest-modules databricks` option to see if it selects doctests
2. `--trace` to debug why some tests are being deselected.

charlesdong1991:
ok, I will try to figure out what's going on there later... it's gonna take
me a while ... we could focus on this PR first... ^^

charlesdong1991:
ahh, it is figured out, it turns out that i need to remove `-k` in order to test a single test file, e.g. `./dev/pytest databricks/koalas/tests/test_series_plot.py`, this works for me now. Not sure if docstring needs to be adjusted a bit for different scenarios (different pr though)

thanks for your great help and explanation!  @HyukjinKwon

databricks/koalas/plot.py
```diff
@@ -58,9 +58,7 @@ def get_top_n(self, data):
         max_rows = get_option("plotting.max_rows")
         # Simply use the first 1k elements and make it into a pandas dataframe
         # For categorical variables, it is likely called from df.x.value_counts().plot.xxx().
-        if isinstance(data, Series):
-            data = data.head(max_rows + 1).to_pandas().to_frame()
-        elif isinstance(data, DataFrame):
+        if isinstance(data, (Series, DataFrame)):
             data = data.head(max_rows + 1).to_pandas()
```
HyukjinKwon: BTW how come it is same? to_pandas previously always returned DataFrame but now seems it can return series too.

charlesdong1991: ehh, i think the outcome of this `to_pandas` is either series or frame, both cases they can directly be used for plotting, so no need to convert to frame using `to_frame`. I tried a bit, and seems okay to me.

Issue #755, opened 2019-09-06
Add precommit in workflow to standardize code style


charlesdong1991:
This is not a function but more like coding/contributing related.

I know now quite many open source projects add `black` for instance in their pre-commit pipeline to kind of standardize the code style. I tried to `black` several files and found some look more 'pretty'. Not sure if it is something Koalas wants to have as well.

If you like it, I would be happy to work on this.

ueshin:
Basically I'd prefer to have a source code formatter to standardize the code style.

I'm wondering whether the pre-commit hook can format only the updated codes since it's possibly difficult to review so many code changes in a PR for enabling the formatter. If it can, we will eventually have fully formatted codes after a while. If we won't have so many changes in the PR, then it'd be no problem.

Also, do you know other formatters? We might want to compare formatters.

charlesdong1991:
Thanks for your comment! @ueshin 

> I'm wondering whether the pre-commit hook can format only the updated codes since it's possibly difficult to review so many code changes in a PR for enabling the formatter. If it can, we will eventually have fully formatted codes after a while. If we won't have so many changes in the PR, then it'd be no problem.

Indeed, most of files I believe will need to be reformatted, and it's difficult to review at once. What I plan to do is to have a WIP PR to implement the pre-commit hook, and then create PRs for each or several files using a selected formatter to avoid huge amount of changes at the same time, once all files or most of files are reformatted, then we merge the pre-cmmit hook PR and then afterwards, all new changes have to follow the new formatting( actually users can just do `black filename.py` before they push the PR to avoid failure on code format. And since this will be basically only reformatting, so reviewing will become pretty easy, i think as long as the tests can be passed, and then should be good. In the long-term, I think having a source code formatter is nice for a project, especially an open source project. 

> Also, do you know other formatters? We might want to compare formatters if any.

Sorry about this, I only have experience in `black` and it's quite popular and pretty good. I think google also has a `yapf`, but never used it.

ueshin:
Sounds great. I agree that the formatter is nice for a project.
Sounds like `black` is good enough for us. `yapf` says it's still in "alpha" stage.
WDYT? @HyukjinKwon

Let me try the `black` in my local for a while in the meantime.

ueshin:
Oh, almost all files changed. We definitely shouldn't do in one PR. haha

charlesdong1991:
[...]

PR #756, opened 2019-09-06
ENH: Add config for pre-commit

charlesdong1991:
this is just a draft which kind of shows what i wanna say for #755 

i set it to WIP, so feel free to comment and i will change. @ueshin @HyukjinKwon


charlesdong1991:
ahh, just checked `.travis.yml`, and find out it's building py35, but `black` only runs on py36 or newer, so pip install will fail..

ueshin:
Maybe we don't need to add it as a dependency, but in that case, definitely we have to add some instruction to install the library for contributors.

charlesdong1991:
yeah, agree... have to remove it, i will add some instructions in `CONTRIBUTING.md`.

pyproject.toml
```diff
@@ -0,0 +1,26 @@
...
+  | \.mypy_cache
+  | \.tox
+  | \.venv
+  | _build
+  | buck-out
+  | build
+  | dist
+  | venv
+)/
+'''
```
ueshin: nit: add a line break at the end of a file.

charlesdong1991: thanks, added!
i also notice line length used in koalas is 100, not 88 when reading through `contributing.md`, also changed in this file.

HyukjinKwon:
To me, I'm okie. Seems worth trying.

PR #759, opened 2019-09-08
Add missing tests for DataFrame

itholic:
Found some missing doctest & unittests for DataFrame as shown below.

- test for operator truediv & truediv

스크린샷 2019-09-08 오후 5 42 57

- test for nunique

스크린샷 2019-09-08 오후 5 42 28

- test for dropna

스크린샷 2019-09-08 오후 5 42 07

스크린샷 2019-09-08 오후 5 42 00

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -116,7 +116,7 @@ circle          1      361
 triangle        4      181
 rectangle       5      361
 
-Divide & True Divide by constant with reverse version.
+Divide and true divide by constant with reverse version.
 
 >>> df / 10
            angles  degrees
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -116,7 +116,7 @@
 triangle        4      181
 rectangle       5      361
 
-Divide by constant with reverse version.
+Divide & True Divide by constant with reverse version.
```
HyukjinKwon: `& True Divide` -> `and true divide`

leehaejoon: Thanks! i've fixed it

itholic: Thanks! i've fixed it

PR #760, opened 2019-09-08
STYLE: Reformat testing dir

charlesdong1991:
This will reformat files in `databricks/koalas/testing/` dir, max length has been set to 100.

### DIFF
--- a/databricks/koalas/testing/utils.py
+++ b/databricks/koalas/testing/utils.py
@@ -32,13 +32,12 @@ from databricks.koalas.series import Series
 
 
 class PySparkTestCase(unittest.TestCase):
-
     def setUp(self):
         self._old_sys_path = list(sys.path)
         if SparkContext._active_spark_context is not None:
             SparkContext._active_spark_context.stop()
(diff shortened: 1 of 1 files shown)


databricks/koalas/testing/utils.py
```diff
@@ -32,21 +32,19 @@
 
 
 class PySparkTestCase(unittest.TestCase):
-
```
ueshin: I'd prefer to have a blank line between the class declaration and its first member. Is it possible?

charlesdong1991: emm... I am afraid not

charlesdong1991: @ueshin  but there is a walkaround, if we add docstring, could be one sentence only:
```python
class PySparkTestCase(TestCase):
    """pyspark customize test case"""
    def __init__(args):
         pass
```
then, Black will automatically add a blank line in between:
```python
class PySparkTestCase(TestCase):
    """pyspark customize test case"""

    def __init__(args):
         pass
```

PR #761, opened 2019-09-08
STYLE: Reformat missing dir

charlesdong1991:
this pr will reformat `missing` dir, and max length has set to 100.


databricks/koalas/missing/frame.py
```diff
@@ -18,119 +18,121 @@
...
-    ftypes = unsupported_property('ftypes', deprecated=True)
-    is_copy = unsupported_property('is_copy', deprecated=True)
-    ix = unsupported_property('ix', deprecated=True)
+    blocks = unsupported_property("blocks", deprecated=True)
+    ftypes = unsupported_property("ftypes", deprecated=True)
+    is_copy = unsupported_property("is_copy", deprecated=True)
+    ix = unsupported_property("ix", deprecated=True)
 
     # Functions
-    agg = unsupported_function('agg')
```
HyukjinKwon: Hm, '`' vs `"`. both are legitimate per PEP 8. Does this show an error?

databricks/koalas/missing/__init__.py
```diff
@@ -18,28 +18,30 @@
 
 
 def _unsupported_function(class_name, method_name, deprecated=False, reason=""):
-
     def unsupported_function(*args, **kwargs):
-        raise PandasNotImplementedError(class_name=class_name, method_name=method_name,
-                                        reason=reason)
```
HyukjinKwon: Hmm .. this is legitimate per PEP 8 if I didn't miss anything.

databricks/koalas/missing/common.py
```diff
@@ -16,32 +16,36 @@
 
 
 memory_usage = lambda f: f(
-    'memory_usage',
+    "memory_usage",
     reason="Unlike pandas, most DataFrames are not materialized in memory in Spark "
-           "(and Koalas), and as a result memory_usage() does not do what you intend it "
-           "to do. Use Spark's web UI to monitor disk and memory usage of your application.")
+    "(and Koalas), and as a result memory_usage() does not do what you intend it "
```
HyukjinKwon: and, this actually looks a bit odd. Because alined indentation within parentheses is legitimate per PEP 8 as well (i use this style quite often too)

HyukjinKwon:
@charlesdong1991, is it possible to make the changes smaller by ignoring some rules? Seems it's too aggressive.

databricks/koalas/missing/frame.py
charlesdong1991: yeah, they standardize quotes to `"` in Black. You are right, both are legitimate in PEP8

databricks/koalas/missing/__init__.py
charlesdong1991: yeah, i think almost all lines in the current codebase are legitimate. This is just how Black reformat codes to standardize code style i think

charlesdong1991:
yeah, i agree with you some look a bit odd and aggressive. There are several ways to write codes aligned with PEP8, and I think Black just choose one of them so that in a way all codes follow the same style. I will take a deeper look later to see if there are less aggressive way to figure this out. @HyukjinKwon 

with my past experience, i think sometimes code reformatted by Black is not aligned with `flake8`, so some people suggest to ignore some rules in `flake8`. I will let you know later.

HyukjinKwon:
Hmmmm .. is it possible to disable the formatting itself less aggressively?

HyukjinKwon:
or  .. is it possible to only reformat the github diff optionally? In case of Apache Spark, it has a script called `dev/scalafmt` which can optionally reformat the files but only for the diff IIRC. I wonder if Black has similar feature.

ueshin:
I'd prefer to only reformat the github diff, but seems like black does not support partial format (https://github.com/psf/black/issues/320).
There is an extension to support it, but I'm not sure how it works (https://github.com/wbolster/black-macchiato).

charlesdong1991:
> Hmmmm .. is it possible to disable the formatting itself less aggressively?

i think only forcing the string quotes can be easily ignored in Black...

charlesdong1991:
sorry, i was away for a holiday, and back now, I am not sure how you want to proceed on this? or do you want to proceed? maybe give some directions you prefer? @HyukjinKwon @ueshin 

otherwise, i would like to pick up other issues then :)

HyukjinKwon:
Can we only apply the formatting for github diff? I think @ueshin has a point at https://github.com/databricks/koalas/pull/761#issuecomment-529639332

charlesdong1991:
i am afraid partial formatting is not supported by black because it goes against black's philosophy which is to have standard/unified code format in the project even it's aligned with PEP8.

Regarding the alternative, i also saw that repo quite long ago, and I don't know exactly how that works, but when i read it, in README it says: `Note that this tool is a stopgap measure, and you should avoid using it if you can`, so I was doubting if koalas should adopt this.

HyukjinKwon:
Hm .. let's drop this for now then. I think it's too aggressive and forces us the style even if you don't like it ..

PR #764, opened 2019-09-08
STYLE: Reformat main function files

charlesdong1991:
this will reformat all other files under `databricks/*` dir


charlesdong1991:
emm, seems it's a bug in `black`... the errors look like the same... just did some search, and i think need to ignore some error codes in flake8,,, I need think a bit more...

PR #766, opened 2019-09-10
raise error when hist on non-numeric cols

HyukjinKwon:
This PR takes over #536. It raises an exception for non-numeric columns in `plot.hist()`.


HyukjinKwon:
Merged to master.

PR #767, opened 2019-09-10
Implement Series.plot.kde

HyukjinKwon:
This PR takes over https://github.com/databricks/koalas/pull/710

KDE, unlike other plots like line or area, it needs to calculates values via Spark so that we can compute it in a distributed manner.

This PR uses MLlib's KernelDensity API to calculate KDE. Since Spark only support scalar bandwidth, unlike SciPy that pandas' uses, Koalas will currently only support fixed scalar bandwidth only.

Implementation is different so the values are slightly different but seems good enough:

```python
import pandas as pd
pd.Series([1, 2, 2.5, 3, 3.5, 4, 5]).plot.kde(bw_method=0.3).figure.savefig("image.png")
```

image

```python
ks.Series([1, 2, 2.5, 3, 3.5, 4, 5]).plot.kde(bw_method=0.3).figure.savefig("image.png")
```

image

```python
import pandas as pd
pd.Series([1, 2, 2.5, 3, 3.5, 4, 5]).plot.kde(bw_method=3.0).figure.savefig("image.png")
```

image

```python
ks.Series([1, 2, 2.5, 3, 3.5, 4, 5]).plot.kde(bw_method=3.0).figure.savefig("image.png")
```

image

### DIFF
--- a/databricks/koalas/tests/test_series_plot.py
+++ b/databricks/koalas/tests/test_series_plot.py
@@ -78,12 +78,12 @@ class SeriesPlotTest(ReusedSQLTestCase, TestUtils):
         pdf = self.pdf1
         kdf = self.kdf1
 
-        ax1 = pdf['a'].plot('bar', colormap='Paired')
-        ax2 = kdf['a'].plot('bar', colormap='Paired')
+        ax1 = pdf['a'].plot("bar", colormap='Paired')
+        ax2 = kdf['a'].plot("bar", colormap='Paired')
         self.compare_plots(ax1, ax2)
 
-        ax1 = pdf['a'].plot('bar', colormap='Paired')
-        ax2 = kdf['a'].plot('bar', colormap='Paired')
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
Let me merge this to move forward.

databricks/koalas/tests/test_series_plot.py
```diff
@@ -276,14 +275,21 @@ def test_box_summary(self):
...
 
-    def test_missing(self):
-        ks = self.kdf1['a']
+    def test_kde_plot(self):
+        pdf = self.pdf1
+        kdf = self.kdf1
+
+        pax = pdf['a'].plot('kde', bw_method=0.3)
+        kax = kdf['a'].plot('kde', bw_method=0.3)
+        self.compare_plots(pax, kax)
```
ueshin: @HyukjinKwon Why does this still work when the values are slightly different?

PR #774, opened 2019-09-12
Add axis parameter to dataframe.diff

itholic:
Since pandas dataframe support axis parameter in dataframe.diff.

스크린샷 2019-09-12 오후 1 47 32

I think we need to support it either. (And it was already commented as TODO)

In addition, now the nunique function only support the axis parameter as 0 (without 'index'), also fix it.

스크린샷 2019-09-12 오후 1 53 41

And currently we only support axis 0, or axis 1 for some functions, so i plan to implement it soon.


ueshin:
Thanks! merging.

PR #769, opened 2019-09-11
Add index_col for spark IO reads.

itholic:
Resolves #765 , I've added index_col for spark IO reads

If we know the index column already, We can prevent the creation of a default index by explicitly typing an index column as function arguments.

For example, now we can use 'read_table' like below.

```python
>>> ks.read_table('test_table1', index_col='i32')
     i64     f  bhello
i32
0      0   0.0  people
2      1  11.0   hello
0      2  12.0  people
1      0  10.0   hello
0      0  15.0   hello
2      0   5.0   hello
0      3   3.0      yo
1      4   4.0  people
0      1   6.0  people
1      2   7.0      yo
0      3  18.0      yo
1      4  19.0   hello
1      1   1.0   hello
2      2   2.0      yo
1      1  16.0   hello
2      2  17.0      yo
2      3   8.0      yo
0      4   9.0   hello
1      3  13.0      yo
2      4  14.0   hello
>>> ks.read_table('test_table1', index_col=['i32', 'i64'])
            f  bhello
i32 i64
0   0     0.0  people
2   1    11.0   hello
0   2    12.0  people
1   0    10.0   hello
0   0    15.0   hello
2   0     5.0   hello
0   3     3.0      yo
1   4     4.0  people
0   1     6.0  people
1   2     7.0      yo
0   3    18.0      yo
1   4    19.0   hello
    1     1.0   hello
2   2     2.0      yo
1   1    16.0   hello
2   2    17.0      yo
    3     8.0      yo
0   4     9.0   hello
1   3    13.0      yo
2   4    14.0   hello
```

Currently only added to 'read_table' functions.

And If you think this way is okay, I'm going to create a PR with all the other functions.

So could you take a look at this maybe if you available??

Thanks :)

PR #775, opened 2019-09-13
Add index_col for spark IO reads

### LINKED ISSUE #765: Add index_col for spark IO reads.
We should provide `index_col` argument for spark IO reads to avoid creating default indexes in the case index columns are known.

- `read_csv`
- `read_delta`
- `read_table`
- `read_spark_io`
- `read_parquet`

itholic:
Resolves #765 ,

I applied the same logic(as worked on #769) to all of the functions mentioned in above issue.

So when we work with spark IO read, and also know about index column name,

now we can use these functions with index_col like below and avoid creation of default index:

```python
>>> ks.read_parquet(path, index_col=['i32', 'i64'])
           f  bhello
i32 i64
0   1    6.0  people
1   2    7.0      yo
```

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -26,6 +26,7 @@ import itertools
 import numpy as np
 import pandas as pd
 
+from pyspark import sql as spark
 from pyspark.sql import functions as F
 from pyspark.sql.types import ByteType, ShortType, IntegerType, LongType, FloatType, \
     DoubleType, BooleanType, TimestampType, DecimalType, StringType, DateType, StructType
@@ -241,7 +242,7 @@ def read_csv(path, header='infer', names=None, index_col=None,
     else:
         sdf = default_session().createDataFrame([], schema=StructType())
 
-    index_map = _get_index_map(sdf, index_col)  # type: Optional[List[IndexMap]]
+    index_map = _get_index_map(sdf, index_col)
 
     return DataFrame(_InternalFrame(sdf=sdf, index_map=index_map))
 
@@ -325,7 +326,7 @@ def read_table(name: str, index_col: Optional[Union[str, List[str]]] = None) -> 
     0   0
     """
     sdf = default_session().read.table(name)
-    index_map = _get_index_map(sdf, index_col)  # type: Optional[List[IndexMap]]
+    index_map = _get_index_map(sdf, index_col)
 
     return DataFrame(_InternalFrame(sdf=sdf, index_map=index_map))
 
@@ -373,7 +374,7 @@ def read_spark_io(path: Optional[str] = None, format: Optional[str] = None,
     0   0
     """
     sdf = default_session().read.load(path=path, format=format, schema=schema, options=options)
-    index_map = _get_index_map(sdf, index_col)  # type: Optional[List[IndexMap]]
+    index_map = _get_index_map(sdf, index_col)
 
     return DataFrame(_InternalFrame(sdf=sdf, index_map=index_map))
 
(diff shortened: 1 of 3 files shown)


databricks/koalas/namespace.py
```diff
@@ -124,7 +124,7 @@ def range(start: int,
 
 
 def read_csv(path, header='infer', names=None, usecols=None,
-             mangle_dupe_cols=True, parse_dates=False, comment=None):
+             mangle_dupe_cols=True, parse_dates=False, comment=None, index_col=None):
```
ueshin: We should put the `index_col` argument betwen `names` and `usecols` to follow pandas'?

databricks/koalas/namespace.py
```diff
@@ -238,11 +240,23 @@ def read_csv(path, header='infer', names=None, usecols=None,
...
+    index_map = None  # type: Optional[List[IndexMap]]
+
+    if index_col is not None:
+        if isinstance(index_col, str):
+            index_col = [index_col]
+        sdf_columns = set(sdf.columns)
+        for col in index_col:
+            if col not in sdf_columns:
+                raise KeyError(col)
+        index_map = [(col, col) for col in index_col]
```
ueshin: We should extract this to a function to share the code?

itholic: oh i think it's a better way either. I've extract it as a `_get_inde_map` function :)

databricks/koalas/namespace.py
itholic: thanks for comment. i agree with you. i've fixed it! 😃

databricks/koalas/namespace.py
```diff
@@ -1635,6 +1626,23 @@ def _to_datetime2(arg_year, arg_month, arg_day,
         format=format,
         infer_datetime_format=infer_datetime_format)
 
+    index_map = None  # type: Optional[List[IndexMap]]
```
ueshin: What's this for?

databricks/koalas/namespace.py
```diff
@@ -1635,6 +1626,23 @@ def _to_datetime2(arg_year, arg_month, arg_day,
...
+
+
+def _get_index_map(sdf, index_col=None):
```
ueshin: Could you add type hints?

ueshin: Then I think we don't need the type hints at each returned place above.

databricks/koalas/namespace.py
```diff
@@ -815,16 +824,7 @@ def read_sql_table(table_name, con, schema=None, index_col=None, columns=None, *
...
-        if isinstance(index_col, str):
-            index_col = [index_col]
-        sdf_columns = set(sdf.columns)
-        for col in index_col:
-            if col not in sdf_columns:
-                raise KeyError(col)
-        index_map = [(col, col) for col in index_col]
-    else:
-        index_map = None
+    index_map = _get_index_map(sdf, index_col)  # type: Optional[List[IndexMap]]
```
ueshin: Thanks for applying here as well!

databricks/koalas/namespace.py
```diff
@@ -875,16 +875,7 @@ def read_sql_query(sql, con, index_col=None, **options):
...
```
ueshin: ditto.

databricks/koalas/namespace.py
itholic: Oops maybe i added it by mistake. i'll fix it. 😭

databricks/koalas/namespace.py
itholic: Thanks for comment! I've fixed them. and remove unnecessary type hints above

PR #781, opened 2019-09-16
Refactor and move KoalasBoxPlot to be consistent in plot implementations.

HyukjinKwon:
This PR make current plot implement consistent, and remove `KoalasBoxPlotSummary` for now. Seems like the generalization can be done differently later. Currently seems a bit overkill to me.

Mainly this PR moves codes just to be consistent.


HyukjinKwon:
Merged.

PR #785, opened 2019-09-17
Add namespace support in options

HyukjinKwon:
This PR adds the support for options to be set like attributes.

```python
>>> ks.options.display.max_rows
1000
>>> ks.options.display.max_rows = 10
>>> ks.options.display.max_rows
10
```

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -318,12 +318,10 @@ class DictWrapper:
             k for k in d.keys() if all(x in k.split(".") for x in canonical_key.split("."))]
         if len(candidates) == 1 and candidates[0] == canonical_key:
             return set_option(canonical_key, val)
-        elif len(candidates) == 0:
+        else:
             raise OptionError(
                 "No such option: '{}'. Available options are [{}]".format(
                     key, ", ".join(list(_options_dict.keys()))))
-        else:
-            return DictWrapper(d, canonical_key)
 
     def __getattr__(self, key):
         prefix = object.__getattribute__(self, "prefix")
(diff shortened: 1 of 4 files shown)


databricks/koalas/config.py
```diff
@@ -298,3 +298,53 @@ def _check_option(key: str) -> None:
...
+        d = object.__getattribute__(self, "d")
+        if prefix:
+            prefix += "."
+        canonical_key = prefix + key
+
+        candidates = [
+            k for k in d.keys() if all(x in k.split(".") for x in canonical_key.split("."))]
+        if len(candidates) == 1 and candidates[0] == canonical_key:
+            return set_option(canonical_key, val)
+        elif len(candidates) == 0:
```
ueshin: We don't need to check here but just raise an error? Otherwise, we can do without errors, e.g.,:

```py
ks.options.compute = 1
```

Maybe we should raise an error like this case.

databricks/koalas/config.py
```diff
@@ -298,3 +298,53 @@ def _check_option(key: str) -> None:
...
+            return get_option(canonical_key)
+        elif len(candidates) == 0:
+            raise OptionError(
+                "No such option: '{}'. Available options are [{}]".format(
+                    key, ", ".join(list(_options_dict.keys()))))
+        else:
+            return DictWrapper(d, canonical_key)
+
+    def __dir__(self):
+        return list(self.d.keys())
```
ueshin: We might want to filter with prefix and drop the prefix? E.g.,:

```py
>>> dir(ks.options.compute)
['default_index_type', 'max_rows', 'ops_on_diff_frames', 'shortcut_limit']
```

instead of:

```py
>>> dir(ks.options.compute)
['compute.default_index_type', 'compute.max_rows', 'compute.ops_on_diff_frames', 'compute.shortcut_limit', 'display.max_rows', 'plotting.max_rows', 'plotting.sample_ratio']
```

WDYT?

ueshin: Btw, pandas' options only show current child categories:

```py
>>> dir(pd.options)
['compute', 'display', 'html', 'io', 'mode', 'plotting']
>>> dir(pd.options.compute)
['use_bottleneck', 'use_numexpr']
```

I don't think we need to follow this now, but we might want if we have more and more options in the future.

HyukjinKwon: Yea sure. Let me think about a better one and fix it today.

databricks/koalas/config.py
HyukjinKwon: Yea, I was trying to find a simpler way .. let me check today.

HyukjinKwon:
Thanks for detailed review @ueshin always!

HyukjinKwon:
Today is the release date. I am merging it in

PR #786, opened 2019-09-17
Show consistent repr output for index, series and frame

HyukjinKwon:
Currently, Koalas `display.max_rows` has a weird relationship with pandas' `display.max_rows` because Koalas directly uses pandas' `repr`.

```python
import pandas as pd

ks.set_option("display.max_rows", 5)
pd.set_option("display.max_rows", 5)

ks.range(6)
```

```
   id
0   0
1   1
2   2
3   3
4   4
```

```python
ks.set_option("display.max_rows", 5)
pd.set_option("display.max_rows", 4)

ks.range(6)
```

```
    id
0    0
1    1
..  ..
3    3
4    4

[Showing only the first 5 rows x 1 columns]
```

**The reason** is because pandas appends footer like:

```python
>>> ks.range(6).to_pandas()
    id
0    0
1    1
..  ..
4    4
5    5

[6 rows x 1 columns]
```

only when the value exceeds `display.max_rows`. 

Koalas replaces this footer `[6 rows x 1 columns]` to `[Showing only the first 6 rows x 1 columns]`.

If pandas does now show the footer:

```python
>>> ks.range(3).to_pandas()
   id
0   0
1   1
2   2
```

Koalas cannot replace it.

**The solution** is that to avoid use `repr` but use `to_string` and ignore pandas' `display.max_rows` but always print out fully. Koalas will control it.

### DIFF
--- a/databricks/koalas/tests/test_repr.py
+++ b/databricks/koalas/tests/test_repr.py
@@ -19,7 +19,7 @@ from databricks.koalas.testing.utils import ReusedSQLTestCase
 
 
 class ReprTests(ReusedSQLTestCase):
-    max_display_count = 123
+    max_display_count = 23
 
     @classmethod
     def setUpClass(cls):
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -6641,12 +6641,13 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
     def __repr__(self):
         max_display_count = get_option("display.max_rows")
         if max_display_count is None:
-            return repr(self._to_internal_pandas())
+            return self._to_internal_pandas().to_string()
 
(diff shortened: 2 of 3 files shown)


databricks/koalas/tests/test_repr.py
```diff
@@ -19,7 +19,7 @@
 
 
 class ReprTests(ReusedSQLTestCase):
-    max_display_count = 123
+    max_display_count = 23
```
HyukjinKwon: pandas' `display.max_rows` value is 60 by default. So, lowering then that will test the changes correctly.

HyukjinKwon:
hm, NVM. let me just implement to match pandas' to make API management less complicated in Koalas for now.

HyukjinKwon:
Let me merge this one - I am pretty sure on this.

PR #790, opened 2019-09-17
Explicitly don't support is_[TYPE] APIs in Index

HyukjinKwon:
```python
>>> import pandas as pd
>>> index = pd.Index([1,2,3])
```

```python
>>> index.dtype
dtype('int64')
>>> index.is_numeric()
True
>>> index.is_integer()
True
>>> index.is_floating()
False
```

This just can be checked via `dtype`.


HyukjinKwon:
hm, NVM. let me just implement to match pandas' to make API management less complicated in Koalas for now.

PR #791, opened 2019-09-17
Implement DataFrame.pop

### LINKED ISSUE #788: Implement DataFrame.pop
Select one Series and drop it in the DataFrame in-place (see https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.pop.html?highlight=pop#pandas.DataFrame.pop)

Related to https://github.com/databricks/koalas/issues/362

itholic:
Resolves #788 

like pandas.DataFrame.pop (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.pop.html?highlight=pop#pandas.DataFrame.pop) now we can pop column from koalas DataFrame like below:

```python
>>> import numpy as np
>>>
>>> df = ks.DataFrame([('falcon', 'bird', 389.0),
...                   ('parrot', 'bird', 24.0),
...                   ('lion', 'mammal', 80.5),
...                    ('monkey','mammal', np.nan)],
...                   columns=('name', 'class', 'max_speed'))
>>>
>>> df.pop('class')
0      bird
1      bird
2    mammal
3    mammal
Name: class, dtype: object
>>>
>>> df
     name  max_speed
0  falcon      389.0
1  parrot       24.0
2    lion       80.5
3  monkey        NaN
```
(use pandas.DataFrame.pop example)

Unlike pandas, we should remove column information via internal frame.

and when we remove column from internal frame, maybe we should remove related metadata also (like _data_columns, _column_index ...).


databricks/koalas/internal.py
```diff
@@ -416,6 +416,13 @@ def __init__(self, sdf: spark.DataFrame,
         else:
             self._column_index_names = column_index_names
 
+    def __delitem__(self, key):
+        self._sdf = self._sdf.drop(key)
+        self._data_columns.remove(key)
+        for i, col in enumerate(self._column_index):
+            if col[0] == key:
+                self._column_index.pop(i)
```
ueshin: Please keep `_InternalFrame` as immutable. It has some cache internally, so if we break the immutability, then the cache doesn't make sense anymore.

itholic: @ueshin , Thanks for review! okay i understand it.
Anyway in pandas, they use a method named `_reset_cacher` (maybe reset the cache literally) when they delete column from DataFrame with `pop` method like below.

스크린샷 2019-09-18 오전 8 55 08

maybe are we have a similar thing for reset the internal cache? or for manually set & unset cache?

Thanks for review again, Have a good day :)

harupy: What is a proper way to do this (remove a specific key and corresponding data)? Replace the current `_InternalFrame` with a new one like below?

```
internal = DataFrame._internal.copy(new_sdf, **new_metadata)
DataFrame._internal = internal
```

HyukjinKwon: Yes. Currently the proper way is:

```python
internal = DataFrame._internal.copy(new_sdf, **new_metadata)
DataFrame(internal)
```

or

```python
internal = _InternalFrame(sdf=...)
DataFrame(internal)
```

HyukjinKwon: @itholic, let's stick to the way we're doing for now. Cache stuff is orthogonal with the current API.

itholic: @harupy @HyukjinKwon 
Thank you guys! they're really helpful for me. i'm going to fix them right away.

databricks/koalas/frame.py
```diff
@@ -6773,6 +6799,27 @@ def __class_getitem__(cls, params):
         # Koalas DataFrame.
         is_dataframe = None
 
+    def _get_dropped_internal_frame(self, columns):
```
HyukjinKwon: Can you fix drop to remove fields from Spark DataFrame first and reuse `drop` inside of `pop` API?

itholic: @HyukjinKwon got it, i've created related PR at #794

itholic: @HyukjinKwon Fixed it :)

ueshin:
Thanks! merging.

itholic:
@ueshin Thanks :)

PR #793, opened 2019-09-17
Reduce usages of `data_columns`.

ueshin:
With this change, the following DataFrame APIs support multi-index columns:

- `applymap`
- `shift`
- `diff`
- `fillna`
- `rank`


ueshin:
Thanks! merging.

PR #794, opened 2019-09-18
Fix DataFrame.drop() to remove fields from Spark DataFrame also.

itholic:
When we drop columns from dataframe with `DataFrame.drop()`,

We can get a dataframe which columns are dropped properly like below.

```python
>>> df
     name   class  max_speed
0  falcon    bird      389.0
1  parrot    bird       24.0
2    lion  mammal       80.5
3  monkey  mammal        NaN
>>>
>>> df = df.drop('name')
>>> df
    class  max_speed
0    bird      389.0
1    bird       24.0
2  mammal       80.5
3  mammal        NaN
```

But when we try to get an internal spark dataframe after then,

it shows us original one which is not delete columns like below.

```
>>> df._sdf.show()
+-----------------+------+------+---------+
|__index_level_0__|  name| class|max_speed|
+-----------------+------+------+---------+
|                0|falcon|  bird|    389.0|
|                1|parrot|  bird|     24.0|
|                2|  lion|mammal|     80.5|
|                3|monkey|mammal|     null|
+-----------------+------+------+---------+
```

(Although I dropped a column 'name' above example, it still shown in internal spark dataframe)

so i think maybe we need to drop them, too.

like:

```
>>> df._sdf.show()
+-----------------+------+---------+
|__index_level_0__| class|max_speed|
+-----------------+------+---------+
|                0|  bird|    389.0|
|                1|  bird|     24.0|
|                2|mammal|     80.5|
|                3|mammal|     null|
+-----------------+------+---------+
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4371,10 +4371,12 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                               for column, idx
                               in zip(self._internal.data_columns, self._internal.column_index)
                               if idx not in drop_column_index))
-            internal = self._internal.copy(data_columns=list(cols), column_index=list(idx))
-            # To drop internal Spark DataFrame's columns also
-            for col in columns:
-                internal._sdf = internal._sdf.drop(col[0])
+            # make column string list to drop internal spark dataframe fields
+            columns = [col[0] for col in columns]
+            internal = self._internal.copy(
+                sdf=self._sdf.drop(*columns),
+                data_columns=list(cols),
+                column_index=list(idx))
             return DataFrame(internal)
         else:
             raise ValueError("Need to specify at least one of 'labels' or 'columns'")
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4348,14 +4348,14 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         >>> columns = [('a', 'x'), ('a', 'y'), ('b', 'z'), ('b', 'w')]
         >>> pdf.columns = pd.MultiIndex.from_tuples(columns)
         >>> kdf = ks.DataFrame(pdf)
-        >>> kdf
-           a     b   
+        >>> kdf  # doctest: +NORMALIZE_WHITESPACE
+           a     b
            x  y  z  w
         0  1  3  5  7
         1  2  4  6  8
 
-        >>> kdf.drop('a')
-           b   
+        >>> kdf.drop('a')  # doctest: +NORMALIZE_WHITESPACE
+           b
            z  w
         0  5  7
         1  6  8
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4371,10 +4371,9 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                               for column, idx
                               in zip(self._internal.data_columns, self._internal.column_index)
                               if idx not in drop_column_index))
-            # make column string list to drop internal spark dataframe fields
-            columns = [col[0] for col in columns]
             internal = self._internal.copy(
-                sdf=self._sdf.drop(*columns),
+                sdf=self._sdf.select(
+                    self._internal.index_scols + [self._internal.scol_for(col) for col in cols]),
                 data_columns=list(cols),
                 column_index=list(idx))
             return DataFrame(internal)
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4348,13 +4348,15 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         >>> columns = [('a', 'x'), ('a', 'y'), ('b', 'z'), ('b', 'w')]
         >>> pdf.columns = pd.MultiIndex.from_tuples(columns)
         >>> kdf = ks.DataFrame(pdf)
-        >>> kdf # doctest: +NORMALIZE_WHITESPACE
+        >>> kdf
+        ... # doctest: +NORMALIZE_WHITESPACE
            a     b
            x  y  z  w
         0  1  3  5  7
         1  2  4  6  8
 
-        >>> kdf.drop('a') # doctest: +NORMALIZE_WHITESPACE
+        >>> kdf.drop('a')
+        ... # doctest: +NORMALIZE_WHITESPACE
            b
            z  w
         0  5  7
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4344,6 +4344,22 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         0  1  7
         1  2  8
 
+        >>> pdf = pd.DataFrame({'x': [1, 2], 'y': [3, 4], 'z': [5, 6], 'w': [7, 8]})
+        >>> columns = [('a', 'x'), ('a', 'y'), ('b', 'z'), ('b', 'w')]
+        >>> pdf.columns = pd.MultiIndex.from_tuples(columns)
+        >>> kdf = ks.DataFrame(pdf)
+        >>> kdf
+           a     b   
+           x  y  z  w
+        0  1  3  5  7
+        1  2  4  6  8
+
+        >>> kdf.drop('a')
+           b   
+           z  w
+        0  5  7
+        1  6  8
+
         Notes
         -----
         Currently only axis = 1 is supported in this function,
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4344,15 +4344,17 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         0  1  7
         1  2  8
 
-        >>> df = ks.DataFrame({('a', 'x'): [1, 2], ('a', 'y'): [3, 4],
-        ...                    ('b', 'z'): [5, 6], ('b', 'w'): [7, 8]},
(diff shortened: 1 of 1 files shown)


databricks/koalas/frame.py
```diff
@@ -4372,6 +4372,9 @@ def drop(self, labels=None, axis=1,
                               in zip(self._internal.data_columns, self._internal.column_index)
                               if idx not in drop_column_index))
             internal = self._internal.copy(data_columns=list(cols), column_index=list(idx))
```
HyukjinKwon: I think you can:

```python
    sdf=self._sdf.drop(*columns),
    data_columns=list(cols),
    column_index=list(idx))
```

You're still modifying immutable `internal` instance.

databricks/koalas/frame.py
```diff
@@ -4372,6 +4372,9 @@ def drop(self, labels=None, axis=1,
...
+            # To drop internal Spark DataFrame's columns also
+            for col in columns:
+                internal._sdf = internal._sdf.drop(col[0])
```
HyukjinKwon: Each call of `drop` creates Spark DataFrame which accumulates a bit of overhead.

databricks/koalas/frame.py
itholic: Wonderful! Thanks for your advice and i just fixed them not to modifying immutable internal instance anymore.

databricks/koalas/frame.py
itholic: ditto !

databricks/koalas/frame.py
```diff
@@ -4371,7 +4371,12 @@ def drop(self, labels=None, axis=1,
                               for column, idx
                               in zip(self._internal.data_columns, self._internal.column_index)
                               if idx not in drop_column_index))
-            internal = self._internal.copy(data_columns=list(cols), column_index=list(idx))
+            # make column string list to drop internal spark dataframe fields
+            columns = [col[0] for col in columns]
+            internal = self._internal.copy(
+                sdf=self._sdf.drop(*columns),
```
HyukjinKwon: Actually, this seems not covering multi-index case.

I think you can just select existing data_columns:

```python
self._sdf.select(self._internal.index_scols + [self._internal.scol_for(col) for col in cols])
```

HyukjinKwon:
Let me leave it to @ueshin

databricks/koalas/frame.py
itholic: Thanks for comment! 😃 

I just tested it for multi-index case like below,

```
>>> import pandas as pd
>>> import numpy as np
>>>
>>> arrays = [[1, 1, 2, 2], ['red', 'blue', 'red', 'blue']]
>>> idx = pd.MultiIndex.from_arrays(arrays, names=('Index', 'color'))
>>> pdf = pd.DataFrame(np.random.randn(4, 5), idx)
>>> kdf = ks.from_pandas(pdf)
>>>
>>> kdf
                    0         1         2         3         4
Index color
1     red    0.304819  1.373173 -0.095708 -0.165494 -0.922387
      blue  -0.538924  0.623598  0.705721 -0.006320  1.173270
2     red    1.397902 -1.870591 -0.294745 -0.100288  0.802501
      blue   1.922724 -0.314832  1.279700  0.414461 -0.010711
>>>
>>> kdf = kdf.drop(['0', '1', '2'])
>>> kdf
                    3         4
Index color
1     red   -0.165494 -0.922387
      blue  -0.006320  1.173270
2     red   -0.100288  0.802501
      blue   0.414461 -0.010711
>>>
>>> kdf._sdf.show()
+-----+-----+--------------------+--------------------+
|Index|color|                   3|                   4|
+-----+-----+--------------------+--------------------+
|    1|  red| -0.1654939967509364|  -0.922387275283718|
|    1| blue|-0.00632027805463...|  1.1732703308776347|
|    2|  red|-0.10028785194871741|  0.8025013695853971|
|    2| blue| 0.41446113311041816|-0.01071114630144753|
+-----+-----+--------------------+--------------------+
```

I think it also works for multi-index case,

But maybe is there a something wrong that i had tested about multi-index?

harupy: I think @HyukjinKwon meant multi-index **columns**.

itholic: @harupy , 
You're the best! it is really helpful for me. Thanks ! 😃

harupy: For example, `kdf.drop('e')` should remove `('e', 'a')` and `('e', 'b')`, but the current implementation does `self._sdf.drop('e')`.

itholic: @harupy ,Really appreciate for your helping. Now maybe i really finished. could you check this out if when you available??

itholic: @harupy Could i use your multi-index example as an additional example for `DataFrame.drop()` maybe if you don't mind??

harupy: @itholic 
Yes of courser you can 😃😃😃

itholic: @harupy Thanks. i just added it. 😸

databricks/koalas/frame.py
```diff
@@ -4344,24 +4344,6 @@ def drop(self, labels=None, axis=1,
         0  1  7
         1  2  8
 
-        >>> pdf = pd.DataFrame({'x': [1, 2], 'y': [3, 4], 'z': [5, 6], 'w': [7, 8]})
```
ueshin: Let's have this doctests.

Need to add `columns=[...]` for py3.5.

```py
>>> df = ks.DataFrame({'x': [1, 2], 'y': [3, 4], 'z': [5, 6], 'w': [7, 8]}, columns=['x', 'y', 'z', 'w'])
```

or we can do:

```py
>>> df = ks.DataFrame({('a', 'x'): [1, 2], ('a', 'y'): [3, 4], ('b', 'z'): [5, 6], ('b', 'w'): [7, 8]},
...                   columns=[('a', 'x'), ('a', 'y'), ('b', 'z'), ('b', 'w')])
```

databricks/koalas/frame.py
```diff
@@ -4371,7 +4371,11 @@ def drop(self, labels=None, axis=1,
                               for column, idx
                               in zip(self._internal.data_columns, self._internal.column_index)
                               if idx not in drop_column_index))
-            internal = self._internal.copy(data_columns=list(cols), column_index=list(idx))
+            internal = self._internal.copy(
+                sdf=self._sdf.select(
+                    self._internal.index_scols + [self._internal.scol_for(col) for col in cols]),
```
ueshin: We should use `idx` rather than `cols`.
Maybe we should rename `idx` as `idxes` or something, then:

```py
[self._internal.scol_for(idx) for idx in idxes]
```

databricks/koalas/frame.py
itholic: @ueshin  Really appreciate your comment Takuya!! I was really suffered all last night from it had never passed lint-test 😭

databricks/koalas/frame.py
itholic: @ueshin it makes sense. i fixed them. Thanks!! :)

databricks/koalas/frame.py
harupy: @itholic 
In python < 3.6, `pandas.DataFrame` sorts the key order when it takes  `dict` using the function below.

```
def dict_keys_to_ordered_list(mapping):
    # when pandas drops support for Python < 3.6, this function
    # can be replaced by a simple list(mapping.keys())
    if PY36 or isinstance(mapping, OrderedDict):
        keys = list(mapping.keys())
    else:
        keys = try_sort(mapping)
    return 
```

https://github.com/pandas-dev/pandas/blob/0a082d40c4a750655a7f1ce127c58ca26cd5905e/pandas/core/common.py#L218-L225

itholic: @harupy Love this comment! now my question all solved. Thanks for sharing your knowledge.

databricks/koalas/frame.py
```diff
@@ -4344,6 +4344,22 @@ def drop(self, labels=None, axis=1,
         0  1  7
         1  2  8
 
+        >>> pdf = pd.DataFrame({'x': [1, 2], 'y': [3, 4], 'z': [5, 6], 'w': [7, 8]},
+        ...                    columns=['x', 'y', 'z', 'w'])
+        >>> columns = [('a', 'x'), ('a', 'y'), ('b', 'z'), ('b', 'w')]
+        >>> pdf.columns = pd.MultiIndex.from_tuples(columns)
+        >>> kdf = ks.DataFrame(pdf)
+        >>> kdf  # doctest: +NORMALIZE_WHITESPACE
```
ueshin: Shall we skip pandas part?

```py
>>> df = ks.DataFrame({'x': [1, 2], 'y': [3, 4], 'z': [5, 6], 'w': [7, 8]},
...                   columns=['x', 'y', 'z', 'w'])
>>> columns = [('a', 'x'), ('a', 'y'), ('b', 'z'), ('b', 'w')]
>>> df.columns = pd.MultiIndex.from_tuples(columns)
>>> df  # doctest: +NORMALIZE_WHITESPACE
```

itholic: @ueshin Sure, It looks better! Thanks for review :)

HyukjinKwon:
Merged.

itholic:
@HyukjinKwon Thanks! i'm going to move on to #791 again.

PR #798, opened 2019-09-19
Handle TimestampType separately when convert to pandas' dtype.

### LINKED ISSUE #772: Filtering rows throws error when result is empty
In Pandas, if I filter for rows and there are none that satisfy the condition, it returns an empty DataFrame.

With Koalas, it throws an error. I think it should return an empty dataframe, like Pandas.

Note: data is a koalas dataframe in the screenshot below
Screen Shot 2019-09-11 at 4 11 21 PM

ueshin:
When the initialization of pandas in pyarrow is not done yet, it can't convert `pa.TimestampType` to pandas' dtype. In that case, the following example raises an error:

```py
from datetime import datetime

kdf = ks.DataFrame({'t': [datetime(2019, 1, 1, 0, 0, 0), datetime(2019, 1, 2, 0, 0, 0), datetime(2019, 1, 3, 0, 0, 0)]})
kdf[kdf['t'] != kdf['t']]
```

```py
>>> kdf[kdf['t'] != kdf['t']]
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/frame.py", line 6646, in __repr__
    pdf = self.head(max_display_count + 1)._to_internal_pandas()
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/frame.py", line 6639, in _to_internal_pandas
    return self._internal.pandas_df
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/utils.py", line 338, in _lazy_property
    setattr(self, attr_name, fn(self))
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/internal.py", line 638, in pandas_df
    for field in sdf.schema})
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/internal.py", line 638, in <dictcomp>
    for field in sdf.schema})
  File "pyarrow/types.pxi", line 404, in pyarrow.lib.TimestampType.to_pandas_dtype
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pyarrow/pandas_compat.py", line 625, in make_datetimetz
    return _pandas_api.datetimetz_type('ns', tz=tz)
TypeError: 'NoneType' object is not callable
```

We know the dtype should be `np.dtype('datetime64[ns]')`, so we don't need to rely on pyarrow's implementation.

```py
>>> kdf[kdf['t'] != kdf['t']]
Empty DataFrame
Columns: [t]
Index: []
```

Resolves #772.

### DIFF
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 
+from datetime import date, datetime
 import inspect
 
 import numpy as np
@@ -1642,3 +1643,11 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
 
         with self.assertRaisesRegex(AssertionError, "the first argument should be a callable"):
             kdf.transform(1)
+
+    def test_empty_timestamp(self):
+        pdf = pd.DataFrame({'t': [datetime(2019, 1, 1, 0, 0, 0),
+                                  datetime(2019, 1, 2, 0, 0, 0),
+                                  datetime(2019, 1, 3, 0, 0, 0)]})
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(kdf[kdf['t'] != kdf['t']], pdf[pdf['t'] != pdf['t']])
+        self.assert_eq(kdf[kdf['t'] != kdf['t']].dtypes, pdf[pdf['t'] != pdf['t']].dtypes)
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -26,12 +26,11 @@ import pandas as pd
 from pandas.api.types import is_list_like
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Window
-from pyspark.sql.types import DoubleType, FloatType, LongType, StringType, TimestampType, \
-    to_arrow_type
+from pyspark.sql.types import DoubleType, FloatType, LongType, StringType, TimestampType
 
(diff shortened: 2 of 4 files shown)


ueshin:
FYI: Seems like here is a bug:

https://github.com/apache/arrow/blob/master/python/pyarrow/pandas-shim.pxi#L150-L151

Maybe we need to add `self._check_import()`.

ueshin:
also cc @harupy

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1642,3 +1643,11 @@ def test_transform(self):
...
         with self.assertRaisesRegex(AssertionError, "the first argument should be a callable"):
             kdf.transform(1)
+
+    def test_empty_timestamp(self):
+        pdf = pd.DataFrame({'t': [datetime(2019, 1, 1, 0, 0, 0),
+                                  datetime(2019, 1, 2, 0, 0, 0),
+                                  datetime(2019, 1, 3, 0, 0, 0)]})
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(kdf[kdf['t'] != kdf['t']], pdf[pdf['t'] != pdf['t']])
+        self.assert_eq(kdf[kdf['t'] != kdf['t']].dtypes, pdf[pdf['t'] != pdf['t']].dtypes)
```
ueshin: Actually this test doesn't reproduce the error in the description since pyarrow is properly initialized while running other tests.

PR #800, opened 2019-09-19
Fix transform to support multi-index columns.

ueshin:


databricks/koalas/frame.py
```diff
@@ -39,11 +39,11 @@
 from pandas.core.dtypes.inference import is_sequence
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Column
+from pyspark.sql.functions import pandas_udf
```
HyukjinKwon: looks like we don't need this.

ueshin: Actually it is used at line 446.
https://github.com/databricks/koalas/blob/a90966b76a655074a12521b0d814f65f68ca9b94/databricks/koalas/frame.py#L446-L448

ueshin: We can use `F.pandas_udf` there, but let me merge this for now.
Please feel free to submit the PR if needed.

ueshin:
Thanks! merging.

PR #806, opened 2019-09-20
Add missing method rename for koalas dataframe

WeichenXu123:
Add missing method rename for koalas dataframe.

Some limitation:
* Do not support in-place operation.
* Require the mapper function include return type hint. such as:
```
def f1(x) -> int:
    return x*10
```
* When rename index labels, it is possible to raise SparkException instead of KeyError (Discussion: Could we get the nested exception "KeyError" and re-throw it ?)


WeichenXu123:
One thing we need to discuss is:
when we want to rename labels in index, we'd better to let user specify the output label type. Because in koalas, the underlying spark dataframe transforming the index labels via pandas udf, which require the udf return type.
Currently, I force the return type to be the same with the old index label type.

WeichenXu123:
@HyukjinKwon @ueshin Ready for review. I leave some discussions in PR description. When these discussion resolved, I will add some edge tests for them.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -368,6 +368,51 @@ def test_rename_columns(self):
...
+        self.assert_eq(result_kdf.columns,
+                       pd.MultiIndex.from_tuples([('x', 'A'), ('x', 'B'), ('y', 'C'), ('y', 'D')]))
+
+        result_kdf = kdf2.rename(columns=str.lower, level=1)
+        self.assert_eq(result_kdf.columns,
+                       pd.MultiIndex.from_tuples([('X', 'a'), ('X', 'b'), ('Y', 'c'), ('Y', 'd')]))
+
+        kdf3 = ks.DataFrame([[1, 2], [3, 4], [5, 6], [7, 8]], index=idx, columns=list('ab'))
+
+        self.spark.conf.set('spark.sql.execution.arrow.enabled', False)
```
HyukjinKwon: @WeichenXu123, I think we don't have to do this since we don't support Spark 3 yet. The test was being failed against current master of Spark because the current master, with arrow optimization, converts a struct into a dictionary whereas without arrow optimization it becomes a row that inherits a tuple.

So the multiindex contained a dictionary instead of tuple that expresses index structure. cc @ueshin

HyukjinKwon:
I discussed about this offline and reviewed roughly @ueshin. Let me leave it to you or I review a day after tomorrow (after my vacation).

databricks/koalas/tests/test_dataframe.py
```diff
@@ -368,6 +368,64 @@ def test_rename_columns(self):
...
+
+        result_kdf = kdf2.rename(columns=str_lower, level=1)
+        self.assert_eq(result_kdf.columns,
+                       pd.MultiIndex.from_tuples([('X', 'a'), ('X', 'b'), ('Y', 'c'), ('Y', 'd')]))
+
+        kdf3 = ks.DataFrame([[1, 2], [3, 4], [5, 6], [7, 8]], index=idx, columns=list('ab'))
+
+        result_kdf = kdf3.rename(index=str_lower)
+        self.assert_eq(result_kdf.index,
+                       pd.Index([{'__index_level_0__': 'x', '__index_level_1__': 'a'},
```
HyukjinKwon: Oh wait, I thought it becomes a dictionary only against Spark 3 which is not released yet. Does this return a dictionary (due to Arrow optimization) in Spark 2.4 and 2.3 too.

databricks/koalas/frame.py
```diff
@@ -6810,6 +6810,213 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+            and `columns`.
+        index : dict-like or function
+            Alternative to specifying axis ("mapper, axis=0" is equivalent to "index=mapper").
+        columns : dict-like or function
+            Alternative to specifying axis ("mapper, axis=1" is equivalent to "columns=mapper").
+        axis : int or str, default 'index'
+            Axis to target with mapper. Can be either the axis name ('index', 'columns') or
+            number (0, 1).
+        copy : bool, default True
+            Also copy underlying data.
```
ueshin: Maybe we don't need to support this argument.

databricks/koalas/frame.py
```diff
@@ -6810,6 +6810,213 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+            else:
+                raise ValueError("rename should take either `index` or `columns` mapper.")
+
+        if inplace is None:
+            do_copy = copy
+        else:
+            do_copy = not inplace
+
+        if not do_copy:
+            raise RuntimeError("Koalas dataframe rename method do not support in-place operation.")
```
ueshin: Why do we not support in-place?

databricks/koalas/frame.py
```diff
@@ -6810,6 +6810,213 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+            if level:
+                if level < 0 or level >= num_indices:
+                    raise ValueError("level should be an integer between [0, num_indices)")
+
+            def gen_new_index_column(level):
+                index_col_name = index_columns[level]
+
+                index_mapper_udf = pandas_udf(lambda s: s.map(mapper_fn),
+                                              returnType=spark_return_type)
+                return index_mapper_udf(col(index_col_name))
```
ueshin: We should use `scol_for()` instead of `col()` to resolve Spark Column safely.

databricks/koalas/frame.py
```diff
@@ -6810,6 +6810,213 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+                for i in range(num_indices):
+                    sdf = sdf.withColumn(index_columns[i], gen_new_index_column(i))
+            else:
+                sdf = sdf.withColumn(index_columns[level], gen_new_index_column(level))
+
+            internal = self._internal.copy(sdf=sdf)
+        else:
+            # rename column name.
+            # we only need to modify the `kdf._internal._column_index
+            # replace each column name by mapper dict or mapper function.
```
ueshin: The comment is right in the future, but unfortunately so far we don't fully support it.
We still need to transform the underlying Spark DataFrame to have the same column name for single-index columns, or string of the tuple for multi-index columns.

databricks/koalas/tests/test_dataframe.py
ueshin: Actually returning `pd.Index` from `MultiIndex.to_pandas()` is a bug. It should be `pd.MultiIndex`.

HyukjinKwon: Yea but another problem was that it contained dictionaries (not tuples).. 

Given:

```python
>>> import pandas as pd
>>>
>>> pd.Index([('x', 'a'), ('x', 'b'), ('y', 'c'), ('y', 'd')], dtype='object')
MultiIndex(levels=[['x', 'y'], ['a', 'b', 'c', 'd']],
           codes=[[0, 0, 1, 1], [0, 1, 2, 3]])
```

I thought it might not be a big deal from a rough look. Will take a closer look tonight or tomorrow as planned.

ueshin: I think we need to override `MultiIndex.to_pandas()` to return `pd.MultiIndex`.

WeichenXu123: We could fix this in separate PR.

HyukjinKwon: Yea, I think it should be a separate PR.

HyukjinKwon: Can we fix that issue first then?

databricks/koalas/frame.py
WeichenXu123: My understanding is, the underlying spark df is immutable, so in any case the modification should not be called "in-place"

WeichenXu123:
Updated code to support renaming `index` and `columns` at the same time. @ueshin 
(although the `level` arguments will be shared by them, it looks a little weird)

databricks/koalas/tests/test_dataframe.py
ueshin: Let me fix it.

databricks/koalas/frame.py
ueshin: Usually we just replace its `internal` with the new one, e.g., https://github.com/databricks/koalas/blob/daa0ccd5673c300e53b1d27863406cf360adbb54/databricks/koalas/frame.py#L2037-L2040

databricks/koalas/frame.py
```diff
@@ -6810,6 +6810,223 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+
+            if internal.column_index_level == 1:
+                new_data_columns = [col[0] for col in new_column_index]
+            else:
+                new_data_columns = [str(col) for col in new_column_index]
+            new_data_scols = [scol_for(internal.sdf, old_col_name).alias(new_col_name)
+                              for old_col_name, new_col_name
+                              in zip(internal.data_columns, new_data_columns)]
+            sdf = internal.sdf.select(*(internal.index_scols + new_data_scols))
+            # Q: Should we set `column_index_names` for new internal dataframe ?
```
ueshin: A: If the column index level is different between new and old, then we should modify it, but in this case, the level is not changed, so we don't need to handle it.

databricks/koalas/tests/test_dataframe.py
ueshin: Submitted a PR #832.

WeichenXu123:
Why the codecov failed ? the travis passed.

HyukjinKwon:
It fails due to decreasing test coverage (see https://codecov.io/gh/databricks/koalas/compare/9186870f42749382451b14bf15c8723e5d68105d...e1839febd9cdfc896cca8b7f55e24fad8eb910ba/diff)

databricks/koalas/frame.py
```diff
@@ -1893,6 +1893,9 @@ def index(self):
 
         Currently not supported when the DataFrame has no index.
 
+        ..Note:: In spark 2.3, this method only works when 'spark.sql.execution.arrow.enabled'
+                 disabled.
```
ueshin: We can remove this if my comment above is right.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -379,6 +380,67 @@ def test_rename_columns(self):
...
+        result_kdf = kdf3.rename(index=str_lower, level=0)
+        self.assert_eq(result_kdf.index,
+                       pd.MultiIndex.from_tuples([('x', 'A'), ('x', 'B'), ('y', 'C'), ('y', 'D')]))
+
+        result_kdf = kdf3.rename(index=str_lower, level=1)
+        self.assert_eq(result_kdf.index,
+                       pd.MultiIndex.from_tuples([('X', 'a'), ('X', 'b'), ('Y', 'c'), ('Y', 'd')]))
+
+        if LooseVersion(pyspark.__version__) <= LooseVersion("2.3"):
+            self.spark.conf.set('spark.sql.execution.arrow.enabled', True)
```
ueshin: ditto.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -20,6 +20,7 @@
 
 import numpy as np
 import pandas as pd
+import pyspark
```
ueshin: ditto.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -379,6 +380,67 @@ def test_rename_columns(self):
...
+        self.assert_eq(result_kdf.columns,
+                       pd.MultiIndex.from_tuples([('X', 'a'), ('X', 'b'), ('Y', 'c'), ('Y', 'd')]))
+
+        kdf3 = ks.DataFrame([[1, 2], [3, 4], [5, 6], [7, 8]], index=idx, columns=list('ab'))
+
+        # for spark 2.3, disable arrow optimization. Because koalas multi-index do not support
+        # arrow optimization in spark 2.3.
+
+        if LooseVersion(pyspark.__version__) <= LooseVersion("2.3"):
+            self.spark.conf.set('spark.sql.execution.arrow.enabled', False)
```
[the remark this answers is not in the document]
HyukjinKwon: Actually I asked to add it because toPandas's fallback seems added from Spark 2.4 as of https://github.com/apache/spark/commit/d6632d185e147fcbe6724545488ad80dce20277e

If we don't, it fails in Spark 2.3 as below:
```
                    return pd.DataFrame.from_records([], columns=self.columns)
            except Exception as e:
                msg = (
                    "Note: toPandas attempted Arrow optimization because "
                    "'spark.sql.execution.arrow.enabled' is set to true. Please set it to false "
                    "to disable this.")
>               raise RuntimeError("%s\n%s" % (_exception_message(e), msg))
E               RuntimeError: Unsupported type in conversion to Arrow: StructType(List(StructField(__index_level_0__,StringType,true),StructField(__index_level_1__,StringType,true)))
E               Note: toPandas attempted Arrow optimization because 'spark.sql.execution.arrow.enabled' is set to true. Please set it to false to disable this.
```

databricks/koalas/frame.py
```diff
@@ -1893,6 +1893,9 @@ def index(self):
 
         Currently not supported when the DataFrame has no index.
 
+        ..Note:: In spark 2.3, this method only works when 'spark.sql.execution.arrow.enabled'
```
HyukjinKwon: `..Note::` -> `.. note::` - it affects documentation rendering.  spark -> Spark or PySpark

HyukjinKwon:
Thanks @WeichenXu123

Issue #804, opened 2019-09-20
pivot_table function with renaming columns


thoo:
Here is the example from the documentation. 

```python
df = ks.DataFrame({"A": ["foo", "foo", "foo", "foo", "foo",
                          "bar", "bar", "bar", "bar"],
                    "C": ["small", "large", "large", "small",
                          "small", "large", "small", "small",
                          "large"],
                    "E": [2, 4, 5, 5, 6, 6, 8, 9, 9]},
                   columns=['A',  'C',  'E'])
df.pivot_table(values='E', index=['A'],
                        columns='C', aggfunc='sum')
```

The above code works as expected.  Then I rename the columns like 

```python
df.columns =['New_A','New_C','New_E']
print(df.columns)
```
This works fine with `applymap` and other functions like `to_datetime` but as soon as I use `pivot_table` , I got an error. 

```python
df.pivot_table(values='New_E', index=['New_A'],
                        columns='New_C', aggfunc='sum')
```
The above code will give the following error:

```python

KeyError: 'No StructField named New_E'

```

Is it not allowed to change the column name right now??  Thanks.

harupy:
Seems like `df.columns = new_columns` doesn't change the internal dataframe columns.

```
>>> df._sdf.show()
+-----------------+---+-----+---+
|__index_level_0__|  A|    C|  E|
+-----------------+---+-----+---+
|                0|foo|small|  2|
|                1|foo|large|  4|
|                2|foo|large|  5|
|                3|foo|small|  5|
|                4|foo|small|  6|
|                5|bar|large|  6|
|                6|bar|small|  8|
|                7|bar|small|  9|
|                8|bar|large|  9|
+-----------------+---+-----+---+

```

ueshin:
@harupy Actually the root cause is the `pivot_table` doesn't handle metadata (`_InternalFrame`) properly. This is the same as `merge` case at #782. So far fixing `columns` is easier way, though, but it doesn't fix the case multi-index columns are used.

HyukjinKwon:
BTW, thanks for reporting this @thoo. the issue looks very easy to understand and reproduce.

ueshin:
Shall we fix the `columns` setter to rename the Spark columns?
@itholic @harupy Do you want to fix it?

I'll handle `merge` and `pivot_table` soon.

harupy:
I'll work on this.

ueshin:
@harupy Great, thanks!

itholic:
[...]

Issue #782, opened 2019-09-16
When using drop on koalas dataframe and then merge, dropped columns persist


kermany:
It seems that after dropping a col from a kdf, if we merge it with another one and create a new dataframe, the dropped column persists. 
This is in databricks environment
e.g:
```
ks.__version__
'0.17.0'
df1=ks.DataFrame({'c1':[1,2,3],'c2':['a','b','c'],'c3':[0,1,2]})
df2=ks.DataFrame({'c1':[1,2,3],'c4':['x','y','z']})
df1=df1.drop('c2')
df3 = df1.merge(df2,on=['c1'])
print(df1.columns,df3.columns)
Index(['c1', 'c3'], dtype='object') Index(['c1', 'c2', 'c3', 'c4'], dtype='object')
```

itholic:
Hi kermany :)

Maybe if you can, Could you give some detail environment your code are running at?
(maybe OS, python version, etc?)

because In my case, i failed to reproduce situation what you faced, 

and there're no problem in jupyter notebook, python interpreter, python script either like below:

<img width="847" alt="스크린샷 2019-09-17 오후 3 20 26" src="https://user-images.githubusercontent.com/44108233/65016252-c2364300-d95e-11e9-88ac-d6781d4266d0.png">

```python
>>> df1=ks.DataFrame({'c1':[1,2,3],'c2':['a','b','c'],'c3':[0,1,2]})
>>> df2=ks.DataFrame({'c1':[1,2,3],'c4':['x','y','z']})
>>> df1=df1.drop('c2')
>>> df3 = df1.merge(df2,on=['c1'])
>>> print(df1.columns,df3.columns)
Index(['c1', 'c3'], dtype='object') Index(['c1', 'c3', 'c4'], dtype='object')
```

```python
# issue782.py
import databricks.koalas as ks

df1=ks.DataFrame({'c1':[1,2,3],'c2':['a','b','c'],'c3':[0,1,2]})
df2=ks.DataFrame({'c1':[1,2,3],'c4':['x','y','z']})
df1=df1.drop('c2')
df3 = df1.merge(df2,on=['c1'])
print(df1.columns,df3.columns)
```

```shell
$ python issue782.py
Index(['c1', 'c3'], dtype='object') Index(['c1', 'c3', 'c4'], dtype='object')
```

Thanks for reporting, Have a nice day :)

HyukjinKwon:
hey @kermany thanks for reporting and @itholic thanks for trying it out.

[...]

PR #810, opened 2019-09-20
Add a workaround to add_prefix/add_suffix for APIs which don't fully support multi-index columns.

ueshin:
Current `add_prefix` and `add_suffix` might cause the same issue as #782 and #804.
This explicitly adds the updated Spark DataFrame for APIs which don't support multi-index columns properly yet as a workaround.


ueshin:
also cc @harupy

PR #812, opened 2019-09-21
add squeeze argument to read_csv

### LINKED ISSUE #808: Pandas read_csv(squeeze='True') to enable koalas Series and not Dataframe
Would be extremely useful to directly read in a koalas.Series from read_csv and not have to convert Dataframe to Series (which also does not exist in Koalas). 
Thank you!

thoo:
Fix #808 

Add `squeeze` argument to `DataFrame.read_csv` so that it is more aligned with `pandas.DataFrame.read_csv` and the behavior will be the same as in `pandas`. Also add pytests for this feature.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -125,7 +125,8 @@ def range(start: int,
 
 
 def read_csv(path, sep=',', header='infer', names=None, index_col=None,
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Thanks for the contribution, @thoo

HyukjinKwon:
Thanks, @thoo

PR #813, opened 2019-09-21
Fix columns setter to rename Spark DataFrame columns

### LINKED ISSUE #782: When using drop on koalas dataframe and then merge, dropped columns persist

### LINKED ISSUE #804: pivot_table function with renaming columns

harupy:
The current columns setter doesn't rename the Spark columns and might cause issues such as #782 and #804.
This PR fixes that.

Resolves #782
Resolves #804

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4214,7 +4214,14 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                     "Length mismatch: Expected axis has %d elements, new values have %d elements"
                     % (len(old_names), len(column_index)))
             column_index_names = columns.names
-            self._internal = self._internal.copy(column_index=column_index,
+            data_columns = [str(idx) if len(idx) > 1 else idx[0] for idx in column_index]
+            sdf = self._sdf.select(
+                self._internal.index_scols +
+                [self._internal.scol_for(idx).alias(name)
+                 for idx, name in zip(self._internal.column_index, data_columns)])
+            self._internal = self._internal.copy(sdf=sdf,
+                                                 data_columns=data_columns,
+                                                 column_index=column_index,
                                                  column_index_names=column_index_names)
         else:
             old_names = self._internal.column_index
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Also let's fix PR title to be more descriptive.

harupy:
@HyukjinKwon 
Thanks for the comment! We have a test function named [`test_rename_columns`](https://github.com/databricks/koalas/blob/1bf41b8b35f20a88b72a9dc29d24467e57352fc2/databricks/koalas/tests/test_dataframe.py#L318-L361). Should I update this one or make a new one?

HyukjinKwon:
Either way should be fine but I would just add a new test

HyukjinKwon:
Thanks, @harupy

PR #816, opened 2019-09-22
Implement Series.aggregate and agg

itholic:
Like pandas Series.aggregate (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.aggregate.html)

I implemented aggregate function for series.

Example:

```python
>>> s = ks.Series([1, 2, 3, 4])
>>> s
0    1
1    2
2    3
3    4
Name: 0, dtype: int64

>>> s.agg('min')
1

>>> s.agg(['min', 'max'])
min    1
max    4
Name: 0, dtype: int64
```

(above example is using pandas one)

스크린샷 2019-09-23 오전 12 01 32

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -21,7 +21,7 @@ import re
 import inspect
 from collections import Iterable
 from functools import partial, wraps
-from typing import Any, Generic, List, Optional, Tuple, TypeVar, Union
+from typing import Any, Generic, List, Optional, Tuple, TypeVar, Union, Dict
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


itholic:
oops.. maybe python3.5 sequence order issue raised again. i'm going to fix them ASAP.

databricks/koalas/series.py
```diff
@@ -2294,6 +2294,56 @@ def apply(self, func, args=(), **kwds):
...
+        """
+        if isinstance(func, list):
+            if all((isinstance(f, str) for f in func)):
+                rows = OrderedDict((f, eval("self.{}()".format(f), dict(self=self))) for f in func)
+                return Series(rows)
+            else:
+                raise ValueError("If the given function is a list, it "
+                                 "should only contains function names as strings.")
+        elif isinstance(func, str):
+            return eval("self.{}()".format(func))
```
ueshin: We should avoid `eval()` as far as possible. `getattr(self, func)()` instead?

databricks/koalas/series.py
```diff
@@ -2294,6 +2294,56 @@ def apply(self, func, args=(), **kwds):
...
+
+        >>> s.agg(['min', 'max'])
+        min    1
+        max    4
+        Name: 0, dtype: int64
+        """
+        if isinstance(func, list):
+            if all((isinstance(f, str) for f in func)):
+                rows = OrderedDict((f, eval("self.{}()".format(f), dict(self=self))) for f in func)
+                return Series(rows)
```
ueshin: This runs Spark jobs many times. In this case, I think we can reuse `DataFrame`'s aggregate.

databricks/koalas/series.py
itholic: @ueshin Thanks for review ueshin :) I totally agree. fixed it !

databricks/koalas/series.py
```diff
@@ -19,7 +19,7 @@
 """
 import re
 import inspect
-from collections import Iterable
+from collections import Iterable, OrderedDict
```
ueshin: nit: no need this anymore.

databricks/koalas/series.py
itholic: @ueshin Thanks again!! fixed it :)
Anyway I have a question, (It may be a very basic question though 😿 )
is it right the every function call at eval in OrderedDict comprehension call spark job each time?

databricks/koalas/series.py
itholic: @ueshin right. i just removed it :)

databricks/koalas/series.py
ueshin: yes, each aggregate function for Series call triggers `sdf.head(2)` in https://github.com/databricks/koalas/blob/cbcb502e12df87cd948ed6ce3c23ece8cb342ba9/databricks/koalas/series.py#L3139-L3149 to return the result value.

itholic: @ueshin Oh now i totally got it. Thanks ueshin!! 😃

ueshin:
Thanks! merging.

PR #820, opened 2019-09-23
Raise a more helpful error for duplicated columns in Join

### LINKED ISSUE #819: Raise error for duplicate columns in join
```python
df = ks.DataFrame([['a', 1,'dog'], ['b', 2,'cat']],
                    columns=['letter', 'number','animal'])
s1 = ks.Series(['a', 'e'],index=[0,1],name='letter')
df.join(s1, how='outer')
```
The above code will raise  the following error.
```python
>>> AnalysisException: "Reference 'letter' is ambiguous, could be: letter, letter.;"
```
In `pandas` it will raise the following error which is more helpful. 
```python
>>> ValueError: columns overlap but no suffix specified: ['letter']
``

### PR DESCRIPTION
**Pull request by thoo**
Fix #819 

Raise an error `>>> ValueError: columns overlap but no suffix specified: ['letter']` if the duplicated column(s) is found but no suffix is specified. 

Also added the pytests.

### DIFF
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -1001,7 +1001,7 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
                              'A': ['A0', 'A1', 'A2', 'A3']}, columns=['key', 'A'])
         kdf2 = ks.DataFrame({'key': ['K0', 'K1', 'K2'],
                              'B': ['B0', 'B1', 'B2']}, columns=['key', 'B'])
-        ks1 = ks.Series(['A1','A5'], index=[1,2], name='A')
+        ks1 = ks.Series(['A1', 'A5'], index=[1, 2], name='A')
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: databricks/koalas/tests/test_dataframe.py [hunk 1]
```diff
@@ -1009,6 +1010,11 @@ def test_join(self):
 
         self.assert_eq(join_pdf, join_kdf)
 
+        # join with duplicated columns in Series and DataFrame
+        with self.assertRaisesRegex(ValueError,
+                                    "columns overlap but no suffix specified"):
+            kdf1.join(ks1, how='outer')
+            kdf1.join(kdf2, how='outer')
```
ueshin: Actually this only tests the first line.

thoo: Sorry. I will fix it.

HyukjinKwon: Ah yes. Sorry for missing this one  will take a closer look next time @ueshin

PR #824, opened 2019-09-23
Add upper bound for libprotobuf caused by newer pyarrow.

ueshin:
Currently master brach build for `py3.7` is broken.
The conda on Travis-CI suddenly starts to use `libprotobuf==3.9.2` instead of `3.8.0` as before, then it causes downgrade of pyarrow to `0.11.1`. Seems like `libprotobuf==3.9.2` and `pyarrow==0.13.0` are not compatible and tries to remove or downgrade `pyarrow`.

```
$ conda install libprotobuf=3.9.2 pyarrow=0.13.0
Collecting package metadata: done
Solving environment: failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - libprotobuf=3.9.2
  - pyarrow=0.13.0 -> arrow-cpp[version='>=0.13.0,<0.14.0a0,>=0.13.0,<1.0a0'] -> libprotobuf[version='>=3.6.0,<3.6.1.0a0']
Use "conda search <package> --info" to see the dependencies for each package.
```

According to the dependency of `pyarrow==0.13.0`, it also conflicts with `libprotobuf==3.8.0`, but it does not try to downgrade or remove `pyarrow`, but `libprotobuf==3.9.2` does (this is my local, not pyarrow 0.14 instead of 0.13, but the behavior is the same):

```
$ conda install libprotobuf=3.9.2

...

The following packages will be REMOVED:

  arrow-cpp-0.14.0-py36h43d7656_0
  grpc-cpp-1.22.0-h5c95ec2_0
  parquet-cpp-1.5.1-2
  pyarrow-0.14.0-py36he1943e6_0
  zstd-1.4.0-ha9f0a20_0
```

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,6 +7,9 @@ numpy>=1.14
 # MLflow
 mlflow>=1.0
 
+# Dependency Issues
+protobuf<3.9.0
+
 # Documentation build
 sphinx==2.0.1
 nbsphinx


ueshin:
I'd merge this for now to unblock other PRs.

PR #828, opened 2019-09-24
Allow str and list in aggfunc in DataFrameGroupby.agg

charlesdong1991:
right now, when I look at Groupby, it does not accept str or list, but in pandas, it's allowed. So before implementing named aggregation, i think this is a better thing to deal first.
e.g. in pandas we could have:

Screen Shot 2019-09-24 at 10 58 52 PM

now koalas can also accept this:

Screen Shot 2019-09-24 at 11 00 13 PM

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -115,7 +115,7 @@ class GroupBy(object):
         """
         if not isinstance(func_or_funcs, (str, list)):
             if not isinstance(func_or_funcs, dict) or \
-                    not all(isinstance(key, str) and
+                    not all(isinstance(key, (str, tuple)) and
                             (isinstance(value, str) or isinstance(value, list) and
                                 all(isinstance(v, str) for v in value))
                             for key, value in func_or_funcs.items()):
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -137,9 +137,7 @@ class GroupBy(object):
                                  "to aggregate functions (string or list of strings).")
 
         else:
-            group_keyname = [key.name for key in self._groupkeys]
-            agg_cols = [key for key in self._kdf.columns if key not in group_keyname]
-
+            agg_cols = [col.name for col in self._agg_columns]
             func_or_funcs = OrderedDict([(col, func_or_funcs) for col in agg_cols])
 
         kdf = DataFrame(GroupBy._spark_groupby(self._kdf, func_or_funcs, self._groupkeys))
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -57,7 +57,7 @@ class GroupBy(object):
 
         Parameters
         ----------
-        func : dict
+        func : dict, str or list
              a dict mapping from column name (string) to
              aggregate functions (string or list of strings).
 
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -115,11 +115,11 @@ class GroupBy(object):
         if not isinstance(func_or_funcs, (str, list)):
             if not isinstance(func_or_funcs, dict) or \
                     not all(isinstance(key, str) and
-                            (isinstance(value, str) or
-                             isinstance(value, list) and all(isinstance(v, str) for v in value))
+                            (isinstance(value, str) or isinstance(value, list) and
+                                all(isinstance(v, str) for v in value))
                             for key, value in func_or_funcs.items()):
-                raise ValueError("aggs must be a dict mapping from column name (string) to aggregate "
-                                 "functions (string or list of strings).")
+                raise ValueError("aggs must be a dict mapping from column name (string) "
+                                 "to aggregate functions (string or list of strings).")
 
         else:
             func_dict = OrderedDict()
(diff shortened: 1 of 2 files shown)


charlesdong1991:
I am not sure if this will pass or not, since group key order is changed as you could see from the picture i attach. Any suggestions on a fix is welcome.

charlesdong1991:
emm, even with current setup, `animals.groupby('kind').agg({'height': ['min', 'max']})` will have different order in index. Is it mandatory to have the exactly same output with pandas, the result looks correct except the order of rows changes. Probably should reorder the rows then before outputting the result. @ueshin @HyukjinKwon any thoughts?

ueshin:
@charlesdong1991 We don't guarantee the row order without a special reason, e.g., `Series.value_counts` has a `sort` argument.

charlesdong1991:
thanks for your comment, i slightly changed the test a bit to fix this order issue that failed tests, and added some docstrings in agg. Feel free to take a look. @ueshin

databricks/koalas/groupby.py
```diff
@@ -112,14 +112,39 @@ def aggregate(self, func_or_funcs, *args, **kwargs):
...
+            if not isinstance(func_or_funcs, dict) or \
+                    not all(isinstance(key, (str, tuple)) and
+                            (isinstance(value, str) or isinstance(value, list) and
+                                all(isinstance(v, str) for v in value))
+                            for key, value in func_or_funcs.items()):
+                raise ValueError("aggs must be a dict mapping from column name (string or tuple) "
+                                 "to aggregate functions (string or list of strings).")
+
+        else:
+            func_dict = OrderedDict()
```
harupy: You could shorten this part to:
```
group_keyname = [key.name for key in self._groupkeys]
agg_cols = [key for key in self._kdf.columns if key not in group_keyname]
func_or_funcs = OrderedDict([(col, func_or_funcs) for col in agg_cols])
```

databricks/koalas/groupby.py
```diff
@@ -112,14 +112,39 @@ def aggregate(self, func_or_funcs, *args, **kwargs):
...
+            group_keyname = [key.name for key in self._groupkeys]
+            agg_cols = [key for key in self._kdf.columns if key not in group_keyname]
+
+            for col in agg_cols:
+                func_dict[col] = func_or_funcs
+
+            func_or_funcs = func_dict
```
harupy: You could shorten this part to:
```
group_keyname = [key.name for key in self._groupkeys]
agg_cols = [key for key in self._kdf.columns if key not in group_keyname]
func_or_funcs = OrderedDict([(col, func_or_funcs) for col in agg_cols])
```

charlesdong1991: yeah, indeed! thx!

charlesdong1991:
any follow-up review will be appreciated a lot ^^ @ueshin @HyukjinKwon

databricks/koalas/groupby.py
```diff
@@ -112,14 +112,35 @@ def aggregate(self, func_or_funcs, *args, **kwargs):
...
+                    not all(isinstance(key, (str, tuple)) and
+                            (isinstance(value, str) or isinstance(value, list) and
+                                all(isinstance(v, str) for v in value))
+                            for key, value in func_or_funcs.items()):
+                raise ValueError("aggs must be a dict mapping from column name (string or tuple) "
+                                 "to aggregate functions (string or list of strings).")
+
+        else:
+            group_keyname = [key.name for key in self._groupkeys]
+            agg_cols = [key for key in self._kdf.columns if key not in group_keyname]
```
ueshin: I guess `self._agg_columns` should work?

charlesdong1991: Nice one! thanks! @ueshin

ueshin:
@charlesdong1991 oh, one more thing I'd like to ask.
Could you add tests for multi-index columns as well, just in case?

charlesdong1991:
@ueshin sure! Added and it passes tests locally, let's see if it's okay on CI.

ueshin:
@charlesdong1991 Thanks!
LGTM again, pending tests.

databricks/koalas/tests/test_groupby.py
```diff
@@ -151,6 +151,38 @@ def test_aggregate(self):
...
+                            'C': [0.362, 0.227, 1.267, -0.562]})
+        kdf = koalas.from_pandas(pdf)
+
+        columns = pd.MultiIndex.from_tuples([('X', 'A'), ('X', 'B'), ('Y', 'C')])
+        pdf.columns = columns
+        kdf.columns = columns
+
+        for aggfunc in agg_funcs:
+            sorted_agg_kdf = kdf.groupby('kind').agg(aggfunc)
+            sorted_agg_pdf = pdf.groupby('kind').agg(aggfunc)
```
ueshin: We might need `.sort_index()`?

charlesdong1991: emm, seems it has the same index as pandas when it's numeric. but could add one to ensure

ueshin:
The latest failure seems not related to this PR.
Let me retrigger the tests.

ueshin:
Thanks! merging.

PR #831, opened 2019-09-25
Put the logo on the docs page.

ueshin:
Screen Shot 2019-09-25 at 5 53 47 PM

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,5 +1,8 @@
 .. Koalas' documentation master file
 
+.. image:: ../../Koalas-logo-nolabel.png
+    :width: 160px
+
 Koalas: pandas API on Apache Spark
 ============================================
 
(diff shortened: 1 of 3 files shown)


ueshin:
I'm not sure this is the best place, though. cc @gioa @HyukjinKwon @gatorsmile

docs/source/index.rst
```diff
@@ -1,5 +1,8 @@
 .. Koalas' documentation master file
 
+.. image:: ../../Koalas-logo-nolabel.png
+    :width: 160px
```
HyukjinKwon: How about placing the logo like pyspark? (https://spark.apache.org/docs/latest/api/python/index.html)
I think it's also fine to just don't put the logo - seems pandas doesn't do that as well

ueshin: I got an advice from our designer.

HyukjinKwon: +1 for the current one

ueshin:
Thanks @gioa for the great advice! merging.

PR #832, opened 2019-09-26
Fix `MultiIndex.to_pandas()` and `__repr__()`.

### LINKED ISSUE #805: Issue with multiple indexes.
```python
df = ks.DataFrame({"A": ["foo", "foo", "foo", "foo", "foo",
                          "bar", "bar", "bar", "bar"],
                    "B": ["one", "one", "one", "two", "two",
                          "one", "one", "two", "two"],
                    "C": ["small", "large", "large", "small",
                          "small", "large", "small", "small",
                          "large"],
                    "D": [1, 2, 2, 3, 3, 4, 5, 6, 7],
                    "E": [2, 4, 5, 5, 6, 6, 8, 9, 9]},
                   columns=['A', 'B', 'C', 'D', 'E'])

table = df.pivot_table(values='D', index=['A', 'B'],
                       columns='C', aggfunc='sum', fill_value=0)
```
If I run ,
```python
>>> table.index
Index([('foo', 'one'), ('foo', 'two'), ('bar', 'two'), ('bar', 'one')], 
dtype='object', name='A')

>>>table.index.to_pandas().tolist()
[Row(A='foo', B='one'),
 Row(A='foo', B='two'),
 Row(A='bar', B='two'),
 Row(A='bar', B='one')]


>>>table.to_pandas().index
MultiIndex([('foo', 'one'),
            ('foo', 'two'),
            ('bar', 'two'),
            ('bar', 'one')],
           names=['A', 'B'])
```

I think `table.index` should have both `names: ['A','B']`.  Is `MultiIndex` currently not supported? I looked into the source and `MultiIndex` was there.

ueshin:
Resolves #805.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -61,7 +61,6 @@ class Index(IndexOpsMixin):
     """
 
     def __init__(self, kdf: DataFrame, scol: Optional[spark.Column] = None) -> None:
-        assert len(kdf._internal._index_map) == 1
         if scol is None:
             scol = kdf._internal.index_scols[0]
         internal = kdf._internal.copy(scol=scol,
@@ -413,6 +412,9 @@ class MultiIndex(Index):
                     ('d', 'h')],
                    )
         """
+        # TODO: We might need to handle internal state change.
+        # So far, we don't have any functions to change the internal state of MultiIndex except for
+        # series-like operations. In that case, it creates new Index object instead of MultiIndex.
         return self._kdf[[]]._to_internal_pandas().index
 
     toPandas = to_pandas
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -329,20 +329,10 @@ class Index(IndexOpsMixin):
 
     def __repr__(self):
         max_display_count = get_option("display.max_rows")
-        sdf = self._kdf._sdf.select(self._scol)
-
         if max_display_count is None:
-            return repr(DataFrame(self._kdf._internal.copy(
-                sdf=sdf,
-                index_map=[(sdf.schema[0].name, self._kdf._internal.index_names[0])],
-                data_columns=[], column_index=[], column_index_names=None)).index.to_pandas())
+            return repr(self.to_pandas())
 
-        sdf = sdf.limit(max_display_count + 1)
-        internal = self._kdf._internal.copy(
-            sdf=sdf,
-            index_map=[(sdf.schema[0].name, self._kdf._internal.index_names[0])],
-            data_columns=[], column_index=[], column_index_names=None)
-        pindex = DataFrame(internal).index.to_pandas()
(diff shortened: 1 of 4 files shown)


databricks/koalas/indexes.py
```diff
@@ -378,6 +395,30 @@ def name(self) -> str:
     def name(self, name: str) -> None:
         raise PandasNotImplementedError(class_name='pd.MultiIndex', property_name='name')
 
+    def to_pandas(self) -> pd.MultiIndex:
+        """
+        Return a pandas MultiIndex.
+
+        .. note:: This method should only be used if the resulting Pandas object is expected
+                  to be small, as all the data is loaded into the driver's memory. If the input
+                  is large, set max_rows parameter.
```
HyukjinKwon: Seems `max_rows` parameter is not existent :).

ueshin: oh haha

databricks/koalas/indexes.py
```diff
@@ -378,6 +395,30 @@ def name(self) -> str:
...
+        ...                   columns=['dogs', 'cats'],
+        ...                   index=[list('abcd'), list('efgh')])
+        >>> df['dogs'].index.to_pandas()  # doctest: +SKIP
+        MultiIndex([('a', 'e'),
+                    ('b', 'f'),
+                    ('c', 'g'),
+                    ('d', 'h')],
+                   )
+        """
+        return self._kdf[[]]._to_internal_pandas().index
```
HyukjinKwon: I maybe missed something but why we cannot do this in `Index` and share the implementation?

ueshin: `Index`'s `scol` might be changed. We need to apply the changes.

HyukjinKwon: oh, NVM. got it. We should select `self._scol` in case it has some column expressions there. Don't we have to do the same thing here too?

HyukjinKwon: oh, okay, but we don't currently support multi index operations like:

```python
>>> ks.range(10).index + 1
Int64Index([1, 2, 3, 4, 5, 6, 7, 8, 9, 10], dtype='int64')
```

Okay, LGTM but might be good to have some comments here like `TODO:`.

ueshin: So far, we don't have any functions to change the internal state of `MultiIndex` except for series-like operations. In that case, it creates new `Index` object instead of `MultiIndex`.

HyukjinKwon: Yes, I checked this assert `assert len(kdf._internal._index_map) == 1`.

At the current moment, the anchor Koalas DataFrame should select Spark column to apply series-like opertations. If we only select from anchor Koalas DataFrame, the Spark column will be lost. Since we don't currently support multi-index, I suspect it's fine for now.

PR #835, opened 2019-09-27
Move `SeriesGroupBy` specific functions and update docs for them.

ueshin:


ueshin:
We will support `SeriesGroupBy.filter()`, right? So I kept it there as well as `aggregate` and `agg` for now. I'm not sure we will be able to reuse the implementation, though.

HyukjinKwon:
Thanks, LGTM!

PR #838, opened 2019-09-28
Make plot tests actually check the difference between pandas' and Koalas'

HyukjinKwon:
We should close figure or clear axis; otherwise, it draws another plot in the same window. This ends up with comparing, for instance:

a1
b1

However we should compare each plot:

a
b

So, previously the comparison was always correct since both plots are overlapped. `plot_to_base64` closes the plot so we should call `plot_to_base64` for each axis before calling another plot.

### DIFF
--- a/databricks/koalas/tests/test_frame_plot.py
+++ b/databricks/koalas/tests/test_frame_plot.py
@@ -51,58 +51,61 @@ class DataFramePlotTest(ReusedSQLTestCase, TestUtils):
         plt.close(ax.figure)
         return b64_data
 
-    def compare_plots(self, ax1, ax2):
-        self.assert_eq(self.plot_to_base64(ax1), self.plot_to_base64(ax2))
-
     def test_line_plot(self):
-
-        def _test_line_plot(pdf, kdf):
+        def check_line_plot(pdf, kdf):
             ax1 = pdf.plot(kind="line", colormap='Paired')
+            bin1 = self.plot_to_base64(ax1)
             ax2 = kdf.plot(kind="line", colormap='Paired')
-            self.compare_plots(ax1, ax2)
+            bin2 = self.plot_to_base64(ax2)
+            self.assertEqual(bin1, bin2)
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_frame_plot.py
```diff
@@ -278,81 +311,115 @@ def test_pie_plot_error_message(self):
...
-    #     self.compare_plots(ax1, ax2)
-
-    #     ax1 = pdf.plot.kde(bw_method=3, ind=[1, 2, 3])
-    #     ax2 = kdf.plot.kde(bw_method=3, ind=[1, 2, 3])
-    #     self.compare_plots(ax1, ax2)
+        pdf1.columns = columns
+        kdf1.columns = columns
+        check_hist_plot(pdf1, kdf1)
+
+    def test_kde_plot(self):
```
HyukjinKwon: While I am here, I made a test case that compares pandas' and Koalas KDEs.

databricks/koalas/tests/test_frame_plot.py
```diff
@@ -278,81 +311,115 @@ def test_pie_plot_error_message(self):
...
-        _test_scatter_plot(pdf, kdf, x=('x', 'a'), y=('x', 'b'), c=('y', 'c'))
+        pdf1.columns = columns
+        kdf1.columns = columns
+        check_scatter_plot(pdf1, kdf1, x=('x', 'a'), y=('x', 'b'), c=('y', 'c'))
 
     def test_hist_plot(self):
-
-        def _test_hist_plot(pdf, kdf):
+        def check_hist_plot(pdf, kdf):
             _, ax1 = plt.subplots(1, 1)
```
ueshin: qq: some tests have lines like this, but what's those for? Do we need those?

HyukjinKwon: I think it was just to test with subplot (to allocate multiple plots (?)). We should close it to properly compare in this case too anyway.. should be fine.

HyukjinKwon:
Thanks 😍

PR #842, opened 2019-09-28
Generate "Release Notes" in official documentation

HyukjinKwon:
This PR proposes to generate "Release Notes" from https://github.com/databricks/koalas/releases
Basically, it gets the release MD text from the URL, converts it into rst file via `pandoc` (with `pypandoc`).

Reference: https://pandas.pydata.org/pandas-docs/stable/whatsnew/index.html#

Seems like working fine as below:

Screen Shot 2019-09-29 at 2 06 20 AM

Screen Shot 2019-09-29 at 2 06 26 AM

### DIFF
--- a/dev/gendoc.py
+++ b/dev/gendoc.py
@@ -64,7 +64,7 @@ def get_json(url):
     except HTTPError as e:
         if "X-RateLimit-Remaining" in e.headers and e.headers["X-RateLimit-Remaining"] == '0':
             print("Exceeded the GitHub API rate limit; see the instructions in " +
-                  "dev/merge_spark_pr.py to configure an OAuth token for making authenticated " +
+                  "dev/gendoc.py to configure an OAuth token for making authenticated " +
                   "GitHub requests.", sys.stderr)
         else:
             print("Unable to fetch URL, exiting: %s" % url, sys.stderr)
@@ -88,7 +88,7 @@ def list_releases_to_document(cur_version):
 def gen_release_notes(path):
     """
     Generate reStructuredText files for "Release Notes". It generates 'index.rst' file and
-    each rst file for each version's release note.under 'whatsnew' directory.
+    each rst file for each version's release note under 'whatsnew' directory.
     The contents are from Github release notes.
     """
     whatsnew_dir = "%s/whatsnew" % path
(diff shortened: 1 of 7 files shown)


docs/Makefile
```diff
@@ -8,6 +8,8 @@ SOURCEDIR     = source
 BUILDDIR      = build
 
 export PYTHONPATH := $(realpath ..):$(PYTHONPATH)
+# Add a sitecustomize.py to generate release notes
+export PYTHONPATH := $(realpath .)/prehook:$(PYTHONPATH)
```
HyukjinKwon: One downside is, though, it will always generates release notes even if we do, for example, `make clean`. But seems like it doesn't take a while (like a seconds).

dev/gendoc.py
```diff
@@ -0,0 +1,195 @@
...
+                pandoc_download._make_executable(dst)
+            src = os.path.join(tempfolder, "usr", "share", "doc", "pandoc", "copyright")
+            dst = os.path.join(targetfolder, "copyright.pandoc")
+            print("* Copying copyright to %s ..." % (targetfolder))
+            shutil.copyfile(src, dst)
+        finally:
+            os.chdir(cur_wd)
+            shutil.rmtree(tempfolder)
+
+    pandoc_urls, _ = pandoc_download._get_pandoc_urls("latest")
```
HyukjinKwon: This logic was just ported from pypandoc to:

1. check if already downloaded
2. to patch `https://github.com/bebraw/pypandoc/pull/154`

dev/gendoc.py
```diff
@@ -0,0 +1,195 @@
...
+                release_file.write("\n")
+                release_file.write("=" * len(name))
+                release_file.write("\n")
+                release_file.write("\n")
+                release_file.write(release_doc)
+                release_file.write("\n")
+                release_file.write("\n")
+
+
+def download_pandoc_if_needed(path):
```
HyukjinKwon: This whole function was basically just ported from pypandoc with some minor changes, in order to:
1. check if already downloaded
2. to patch `https://github.com/bebraw/pypandoc/pull/154`

HyukjinKwon:
Thanks! For 10 minutes koalas as a doc, I will do it in a separate PR @ueshin!

PR #844, opened 2019-09-29
Fix on wrong error raise in DataFrame.fillna

charlesdong1991:
https://github.com/databricks/koalas/blob/7ee2a8dff3d61f2fe66f85385bc5af20d2f154f5/databricks/koalas/frame.py#L3534

seems this line won't be touched, and the reason it still passed tests is because it triggered the `Series.fillna()` error raise mechanism in the `else` below.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3531,8 +3531,6 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 axis = 0
             if not (axis == 0 or axis == "index"):
                 raise NotImplementedError("fillna currently only works for axis=0 or axis='index'")
-            if (value is None) and (method is None):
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Thanks. Looks good but can we add a simple test @charlesdong1991?

charlesdong1991:
i think there is a test already there:

https://github.com/databricks/koalas/blob/7ee2a8dff3d61f2fe66f85385bc5af20d2f154f5/databricks/koalas/tests/test_dataframe.py#L513

but this triggered the error raising in `Series.fillna`. Do you still want a test? I could add one quickly @HyukjinKwon

HyukjinKwon:
Ah, gotya. LGTM

Issue #834, opened 2019-09-27
Understanding Groupbyapply


Devarsh-UTD:
Hello there, firstly thank you for such an amazing package that bridges the gap between Pandas and PySpark. 
I started using koalas approximately 1 week back and everything was intuitive till the time i stumbled upon `koalas.Groupby.Apply`. 

Code:
```
if __name__ == '__main__':

        ks_df = ks.DataFrame(features_data)
        ks_df_info_abt_train = ks_df.groupby(['div_nbr', 'store_nbr']).apply(_koalas_train)
        
        def _koalas_train(frame):
                  out_frame = frame.copy()
                  out_frame = frame['trans_type_value'].sum()
                  return out_frame
```
Here features_data is a  pd.Dataframe.

Output from Koalas.Groupby.Apply:

![Screen Shot 2019-09-27 at 2 27 26 PM](https://user-images.githubusercontent.com/33138515/65793013-f2d26580-e132-11e9-9a39-9e0fecb5bd3f.png)

Output from Pandas.Groupby.Apply:
![Screen Shot 2019-09-27 at 2 30 13 PM](https://user-images.githubusercontent.com/33138515/65793255-65dbdc00-e133-11e9-8cde-d47aed06ce3b.png)

As you can see, the output from pandas Groupby apply is as expected, but the output from Koalas Groupby apply is not right. Could you guid me towards the right direction by pointing out any logical mistake that i might have made or anything else. 
Thank you once again.

Koalas version - 0.18.0 
Pandas version - 0.23.4
PySpark -  2.4.3

harupy:
The issue is the code I highlighted below in `Groupby.apply` ([link](https://github.com/databricks/koalas/blob/1c11cc71f1710325f8ee2911f4f78a57940205c4/databricks/koalas/groupby.py#L819-L828)).

```
if should_infer_schema:
    # Here we execute with the first 1000 to get the return type.
    # If the records were less than 1000, it uses pandas API directly for a shortcut.
    limit = get_option("compute.shortcut_limit")
    pdf = self._kdf.head(limit + 1)._to_internal_pandas()
    
    ##### This line is causing the issue #####
    pdf = pdf.groupby(input_groupnames).apply(func)

    kdf = DataFrame(pdf)
    return_schema = kdf._sdf.schema
    if len(pdf) <= limit:
        return kdf
```

[...]

PR #845, opened 2019-09-30
Infer type for transform and apply in Series, and applymap in DataFrame

HyukjinKwon:
This PR adds the type inference for:

- `transform` in Series
- `apply` in Series
- `applymap` in DataFrame

Note that this PR intentionally does not implement shortcut to avoid the operations on different DataFrames. Maybe we should add different configuration to indicate the sample size for type inference.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -828,11 +828,16 @@ class DataFrame(_Frame, Generic[T]):
         This method applies a function that accepts and returns a scalar
         to every element of a DataFrame.
 
-        .. note:: unlike pandas, it is required for `func` to specify return type hint.
-            See https://docs.python.org/3/library/typing.html. For instance, as below:
+        .. note:: this API executes the function once to infer the type which is
+             potentially expensive, for instance, when the dataset is created after
+             aggregations or sorting.
 
-            >>> def function() -> int:
-            ...     return 1
+             To avoid this, specify return type in ``func``, for instance, as below:
+
+             >>> def square(x) -> np.int32:
+             ...     return x ** 2
+
+             Koalas uses return type hint and does not try to infer the type.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -2282,15 +2292,44 @@ def apply(self, func, args=(), **kwds):
...
+        Name: 0, dtype: float64
+
         """
-        assert callable(func), "the first argument should be a callable function."
-        spec = inspect.getfullargspec(func)
-        return_sig = spec.annotations.get("return", None)
-        if return_sig is None:
-            raise ValueError("Given function must have return type hint; however, not found.")
+        assert callable(func) or hasattr(func, "__call__"), \
+            "the first argument should be a callable function."
```
HyukjinKwon: `np.log` failed `callable(...)` check ..

databricks/koalas/frame.py
```diff
@@ -865,10 +870,19 @@ def applymap(self, func):
...
+
+        >>> df.applymap(lambda x: x ** 2)
+                   0          1
+        0   1.000000   4.494400
+        1  11.262736  20.857489
         """
 
         applied = []
         for idx in self._internal.column_index:
+            # TODO: We can implement shortcut theoretically since it creates new DataFrame
```
HyukjinKwon: This also will collects for every column .. we should implement separate inference logic here. will do separately.

databricks/koalas/series.py
ueshin: FYI: seems working in my local:

```py
>>> import numpy as np
>>> np.__version__
'1.16.4'
>>> callable(np.log)
True
```

databricks/koalas/series.py
```diff
@@ -2282,15 +2292,44 @@ def apply(self, func, args=(), **kwds):
...
+            should_infer_schema = True
 
         apply_each = wraps(func)(lambda s, *a, **k: s.apply(func, args=a, **k))
-        wrapped = ks.pandas_wraps(return_col=return_sig)(apply_each)
+
+        if should_infer_schema:
+            # TODO: In this case, it avoids the shortcut for now (but only infers schema)
+            #  because it returns a series from a different DataFrame and it has a different
+            #  anchor. We should fix this to allow the shortcut or only allow to infer
+            #  schema.
```
HyukjinKwon: For infer only config, yes, I left it as a todo here for now.

databricks/koalas/series.py
HyukjinKwon: Mine was older and the error message was misreading in my case ... let me fix it. seems it returns `True` in my case too anyway.

HyukjinKwon:
@harupy, you can fix it like this for #834. I will add a separate configuration later once you fix it.

HyukjinKwon:
Merged for today's release.

PR #847, opened 2019-09-30
Add some examples for plot implementations in their docstrings

itholic:
Basically we finished the rough implementation of all plots (except 2 in DataFrame).

So it should be better to have some examples for them.


databricks/koalas/plot.py
```diff
@@ -1328,6 +1473,20 @@ def area(self, x=None, y=None, stacked=True, **kwds):
...
         matplotlib.axes.Axes or numpy.ndarray
             Area plot, or array of area plots if subplots is True.
+
+        Examples
+        --------
+
+        .. plot::
+            :context: close-figs
+
+        >>> df = ks.DataFrame({
```
HyukjinKwon: Indentation seems different from others.

databricks/koalas/plot.py
```diff
@@ -992,6 +992,16 @@ def line(self, x=None, y=None, **kwargs):
...
+
+        Examples
+        --------
+        Basic plot.
+
+        .. plot::
+            :context: close-figs
+
+            >>> s = ks.Series([1, 3, 2])
+            >>> ax = s.plot.line()  # doctest: +SKIP
```
HyukjinKwon: Why should we skip?

databricks/koalas/plot.py
```diff
@@ -1148,6 +1215,10 @@ def area(self, **kwds):
 
         Examples
         --------
+
+        .. plot::
+            :context: close-figs
+
         >>> df = ks.DataFrame({
```
HyukjinKwon: indentation seems inconsistent too

HyukjinKwon:
This is nice. Thanks for working on this @itholic

databricks/koalas/plot.py
leehaejoon: @HyukjinKwon This doctest continuously failed on my local test bed like below:


스크린샷 2019-09-30 오후 4 40 07

But actually it works well on without exception:

스크린샷 2019-09-30 오후 4 39 37

 so i wanna discussion with others members.

could you test this on your local if when available or have some insight of above failure?

itholic: @HyukjinKwon This doctest continuously failed on my local test bed like below:


스크린샷 2019-09-30 오후 4 40 07

But it actually works well without exception:

스크린샷 2019-09-30 오후 4 39 37

 so i wanna discussion with others members.

could you test this on your local if when available or have some insight of above failure?

databricks/koalas/plot.py
itholic: Thanks for comment!! fixed both of them.

databricks/koalas/plot.py
itholic: I just undo this skip and push again,

itholic:
@HyukjinKwon my pleasure.

databricks/koalas/plot.py
itholic: Failed again. 😢 Do you have any guesswork ??

itholic: Well i just try this doctest again with exactly same test in `test_series_plot`. But still failed.

스크린샷 2019-09-30 오후 5 57 49

above passed, but below not.

스크린샷 2019-09-30 오후 5 58 45

charlesdong1991: what if you explicitly specify `index` in Series, like `ks.Series([1, 3, 2], index=[0,1,2])`, pretty wired it directs to `timeseries` plot instead of a normal plot.

itholic: @charlesdong1991 Thanks for comment !! I tried that way you recommend but they still throws me exception that `UNEXPECTED EXCEPTION: TypeError('index type not supported')` 😢

charlesdong1991: and what about defining `dtype` in Series? i feel it looks like a bug to me, because I am pretty sure pandas will check the `type` of Series, and if it infers from data and it's datetime-like, then will use `timeseries` plot, otherwise, use normal plot.

harupy: Testing only `KoalasSeriesPlotMethods.line` worked.

`./dev/pytest -k KoalasSeriesPlotMethods.line --doctest-modules databricks --verbose`

itholic: @charlesdong1991 oh i think maybe there are some bugs in series type check either (but why this bug only reproduced in doctest?!).

i just ran test below, and it passed. 

```python
            >>> s = ks.Series(range(5),
            ...     index=pd.date_range('2018-01-01', periods=5, freq='H'))
            >>> ax = s.plot.line()
```

(and also defining `dtype` is not work neither 😢 )

itholic: @harupy Thanks for comment!! i just try like same with you, it really is. wow it's make me very confused. @.@

databricks/koalas/plot.py
```diff
@@ -1243,6 +1315,38 @@ def line(self, x=None, y=None, **kwargs):
...
+            >>> df = ks.DataFrame({
+            ...    'pig': [20, 18, 489, 675, 1776],
+            ...    'horse': [4, 25, 281, 600, 1900]
+            ...    }, index=[1990, 1997, 2003, 2009, 2014])
+            >>> lines = df.plot.line()
+
+        .. plot::
+           :context: close-figs
+
+           An example with subplots, so an array of axes is returned.
```
harupy: different indentation from others

itholic: @harupy Eagle eye! thanks for find my mistake out. i fixed it. Thanks 😃

databricks/koalas/plot.py
harupy: Changed the doctest as follows and it passed.
```
.. plot::
    :context: close-figs

    >>> import matplotlib.pyplot as plt
    >>> plt.close('all')
    >>> s = ks.Series([1, 3, 2])
    >>> ax = s.plot.line()
```

itholic: @harupy oh?? then these failures before we got maybe conflicted with other existing plots??

charlesdong1991: emm, yes, indeed it could pass tests, but for users who read the documentation, I dont know if this will bring them some confusion, so I have mixed feeling about this trick.

itholic: Anyway, it works! Thanks ! 😸

harupy: `close-figs` doesn't seem to close figures.
```
995 
996         Examples
997         --------
998         Basic plot.
999 
1000         .. plot::
1001             :context: close-figs
1002 
1003             >>> import matplotlib.pyplot as plt
1004             >>> plt.get_fignums()
Expected nothing
Got:
    [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21]

/Users/harutaka/Desktop/projects/koalas/databricks/koalas/plot.py:1004: DocTestFailure
```

itholic: @charlesdong1991 yeah i totally agree with it. I think i should be find better solution for solve this issue to keep consistency of docs.
@harupy you're right i think maybe close-figs doesn't work properly either.

harupy: Does `close-figs` has any effect when running tests?

databricks/koalas/plot.py
```diff
@@ -992,6 +992,18 @@ def line(self, x=None, y=None, **kwargs):
...
+
+        Examples
+        --------
+        Basic plot.
+
+        .. plot::
+            :context: close-figs
+
+            >>> import matplotlib.pyplot as plt
+            >>> plt.close('all')
```
ueshin: Do we still need this?
Seems like others don't have this, otherwise we need to add for the others?

ueshin:
Btw, seems like `dev/lint-python` will fail after docs are built since some python files are generated in `./docs/build/html/reference/api`.

```
starting pycodestyle test...
pycodestyle checks failed:
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-bar-1.py:1:25: E231 missing whitespace after ':'
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-bar-1.py:1:48: E231 missing whitespace after ':'
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-hist-1.py:3:12: E251 unexpected spaces around keyword / parameter equals
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-hist-1.py:3:14: E251 unexpected spaces around keyword / parameter equals
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-pie-1.py:1:40: E203 whitespace before ','
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-line-1.py:5:4: E121 continuation line under-indented for hanging indent
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-line-1.py:7:4: E123 closing bracket does not match indentation of opening bracket's line
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-barh-1.py:1:25: E231 missing whitespace after ':'
./docs/build/html/reference/api/databricks-koalas-frame-DataFrame-plot-barh-1.py:1:48: E231 missing whitespace after ':'
```

Can we ignore those files when running `pycodestyle`?

databricks/koalas/plot.py
itholic: @ueshin at above discussions, maybe there is an issue that `:context: close-figs` doesn't work properly. so now it wouldn't work if we remove the line `plt.close('all')` to explicitly close all plots.  i'm going to figure out how can we solve this problem.

databricks/koalas/plot.py
itholic: @harupy Oops.. sorry i missed your reply
 Anyway, yes it seems have any effect, because they are showing same result when i remove all `close-figs` options.

databricks/koalas/plot.py
harupy: @itholic 
I think `:context: close-figs` is just for Sphinx, and does nothing when pytest runs tests.

itholic:
@ueshin Thanks for comment!
I think maybe we already have examples for `Series.plot.barh` and `Series.plot.pie` , maybe is there a problem with that examples or wanna additional ones? 😃

itholic:
@ueshin And I'm going to make them ignore in pycodestyle! Thanks for review :)

databricks/koalas/plot.py
harupy: So when running tests, all the plots are drawn onto the same `ax` and get messed up.

itholic: @harupy Thanks for comment!! It really makes sense. Then i think maybe we need to skip these all examples on pytest, and added them separate unittest (actually most of them already tested in unittest suite though)

ueshin:
@itholic I mean, when I generated the doc, I can't see the plots for `Series.plot.barh` and `Series.plot.pie`. I thought we need 

```
        .. plot::
            :context: close-figs
```

tags for them as well, then we can see the plots?

itholic:
@ueshin oh i got it :) yes we can see the plots in generated docs by adding
```
 .. plot::
            :context: close-figs
```

i'm going to add them to `Series.plot.barh` & `Series.plot.pie` either for visualize plots in docs 😄

databricks/koalas/plot.py
harupy: I added `plt.savefig` to see what's happening and ran `./dev/pytest -k plot.py --doctest-modules databricks`.
```
# doctest of KoalasSeriesPlotMethods.line

>>> import matplotlib.pyplot as plt
>>> plt.savefig('foo.png')
>>> s = ks.Series([1, 3, 2])
>>> ax = s.plot.line()
``` 

And I got the figure below.

foo

itholic: @harupy Thanks for researching! 😄  now i think it is clear that `close-fig` doesn't work in pytest. (maybe this syntax only for sphinx). isn't it?

harupy: Adding the code below solves this issue.

```python
@pytest.fixture(autouse=True)
def close_figs():
    yield
    plt.close("all")
```

This is actually what `seaborn` does.(https://github.com/mwaskom/seaborn/blob/master/seaborn/conftest.py#L6)

itholic: @harupy Genius, It's works! Thanks for researching really so much !! 😺

harupy: but I'm not sure if this is right because `plt.close("all")` is called on tests which are not related to plots.

itholic: ah? sorry i can't get it perfectly 😢 maybe could you let me know where they are using for tests not related to plots?

harupy: `plt.close` is called on test functions where we actually don't need to call `plt.close`.

itholic: got it :)
but maybe i think they are only called when we explicitly write the `:context: close-figs` in doctest, ain't they?

harupy: no

databricks/koalas/plot.py
```diff
@@ -1183,7 +1259,11 @@ def pie(self, **kwds):
...
         --------
-        >>> df = ks.DataFrame({'mass': [0.330, 4.87 , 5.97],
+
+        .. plot::
+            :context: close-figs
+
+        >>> df = ks.DataFrame({'mass': [0.330, 4.87, 5.97],
         ...                    'radius': [2439.7, 6051.8, 6378.1]},
         ...                   index=['Mercury', 'Venus', 'Earth'])
         >>> plot = df.mass.plot.pie(figsize=(5, 5))
```
ueshin: nit: indent should be one more level?

itholic: Thanks for comment!! fixed it :)

databricks/koalas/plot.py
```diff
@@ -1450,5 +1732,36 @@ def scatter(self, x, y, s=None, c=None, **kwds):
...
+            ...                       c='species',
+            ...                       colormap='viridis')
         """
         return self(kind="scatter", x=x, y=y, s=s, c=c, **kwds)
+
+
+@pytest.fixture(autouse=True)
+def close_figs():
+    yield
+    plt.close("all")
```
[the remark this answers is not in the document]
itholic: Thanks again, i moved them to `conftest.py` :)

ueshin:
oh, I forgot about https://github.com/databricks/koalas/pull/847#issuecomment-536759721.
Seems like `pycodestyle` still fails after creating docs.

itholic:
@ueshin  oops, sorry i'm going to fix them soon

itholic:
@ueshin Suddenly, i have some question. Is it right way we ignore these tests rather than fix lint fails on doctest in created docs?

ueshin:
We should definitely keep lint consistency even in documentations.
But the problem here is the generated `.py` files under the `./docs/build` won't follow the rule.
Could you try `dev/lint-python` after generating docs and see what happens?
Now `flake8` shows a lot of errors.
But maybe we should fix them in a separate PR. WDYT? @HyukjinKwon

itholic:
@ueshin ah, i just tried, understood why we should ignore them. thanks for explanation! 👍

databricks/koalas/plot.py
```diff
@@ -1183,12 +1257,16 @@ def pie(self, **kwds):
...
-        >>> plot = df.mass.plot.pie(subplots=True, figsize=(6, 3))
+        .. plot::
+            :context: close-figs
+
+            >>> df = ks.DataFrame({'mass': [0.330, 4.87, 5.97],
+            ...                    'radius': [2439.7, 6051.8, 6378.1]},
+            ...                   index=['Mercury', 'Venus', 'Earth'])
+            >>> plot = df.mass.plot.pie(figsize=(5, 5))
+
+            >>> plot = df.mass.plot.pie(subplots=True, figsize=(6, 3))
```
ueshin: Do we need `.. plot::` here as well? Seems like the documents are properly generated, though.

ueshin:
Hmm, our official doc doesn't show all the plots. https://koalas.readthedocs.io/en/latest/reference/series.html#plotting
It shows `bar`, `barh`, or `pie`, but doesn't the others.
Even for `bar`, `barh`, or `pie`, the last plots seem missing.
We might miss something for Read the Docs service.

ueshin:
```
Unsupported class file major version 55
```

databricks/koalas/plot.py
harupy: @ueshin is correct. pandas put `.. plot::` on each plot.

https://github.com/pandas-dev/pandas/blob/v0.25.1/pandas/plotting/_core.py#L1310-L1363

HyukjinKwon:
@itholic can you take a look and make a fix?

itholic:
@ueshin @HyukjinKwon , okay i'm going to take a look at it and summit another PR for fix them.

databricks/koalas/plot.py
itholic: @ueshin @harupy Thanks for comment! i'll add them for all proper location. :)

ueshin: This was resolved at #865.

PR #862, opened 2019-10-02
Remove shortcut from GroupBy.apply

### LINKED ISSUE #834: Understanding Groupbyapply

harupy:
The current `GroupBy.apply` uses a shortcut (proposed in #646) when schema inference is triggered and the number of records is less than or equal to `compute.shortcut_limit` (1000 by default), but this might cause issues such as #834.

This PR proposes to remove the shortcut and make `GroupBy.apply` only infer the schema.

Closes #834

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -799,7 +799,7 @@ class GroupBy(object):
 
         >>> def plus_min(x):
         ...    return x + x.min()
-        >>> g.apply(plus_min)  # doctest: +NORMALIZE_WHITESPACE
+        >>> g.apply(plus_min).sort_index()  # doctest: +NORMALIZE_WHITESPACE
             A  B   C
         0  aa  2   8
         1  aa  3  10
@@ -838,17 +838,16 @@ class GroupBy(object):
 
         if should_infer_schema:
             # Here we execute with the first 1000 to get the return type.
-            # If the records were less than 1000, it uses pandas API directly for a shortcut.
             limit = get_option("compute.shortcut_limit")
-            pdf = self._kdf.head(limit + 1)._to_internal_pandas()
+            pdf = self._kdf.head(limit)._to_internal_pandas()
             pdf = pdf.groupby(input_groupnames).apply(func)
             kdf = DataFrame(pdf)
             return_schema = kdf._sdf.schema
-            if len(pdf) <= limit:
-                return kdf
 
         sdf = self._spark_group_map_apply(
-            func, return_schema, retain_index=should_infer_schema)
+            lambda pdf: pdf.groupby(input_groupnames).apply(func),
+            return_schema,
+            retain_index=should_infer_schema)
 
         if should_infer_schema:
             # If schema is inferred, we can restore indexes too.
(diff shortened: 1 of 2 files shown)


harupy:
Is it okay to remove shortcut tests in `test_apply` and `test_apply_with_new_dataframe` ([code](https://github.com/databricks/koalas/blob/1462291490c4a3e084ece28edaef92721bce14fd/databricks/koalas/tests/test_groupby.py#L544-#622))?

ueshin:
@harupy Thanks for working on this!
I think we can remove the shortcut tests, but I'd like to ask you to add a test for the case the result was explicitly wrong instead. I guess the example in your comment (https://github.com/databricks/koalas/issues/834#issuecomment-536829367) should be good.

harupy:
@ueshin Thanks! I'll add it.

harupy:
```
______________ [doctest] databricks.koalas.groupby.GroupBy.apply _______________
793            c0  c1  c2
794         0  bb   6  10
795         1  aa   3  10
796         2  aa   4  12
797 
798         You can omit the type hint and let Koalas infer its type.
799 
800         >>> def plus_min(x):
801         ...    return x + x.min()
802         >>> g.apply(plus_min)  # doctest: +NORMALIZE_WHITESPACE
    @@ -1,4 +1,4 @@
         A  B   C
    +2  bb  6  10
     0  aa  2   8
     1  aa  3  10
    -2  bb  6  10
```
The doctest failed. We might need `sort_index()`.

ueshin:
Yup, that's fine. Please add `sort_index()`.

databricks/koalas/groupby.py
```diff
@@ -838,17 +838,16 @@ def apply(self, func):
...
             kdf = DataFrame(pdf)
             return_schema = kdf._sdf.schema
-            if len(pdf) <= limit:
-                return kdf
 
         sdf = self._spark_group_map_apply(
-            func, return_schema, retain_index=should_infer_schema)
+            lambda pdf: pdf.groupby(input_groupnames).apply(func),
+            return_schema,
+            retain_index=should_infer_schema)
```
HyukjinKwon: This fix LGTM

HyukjinKwon:
Merged. Thanks!

harupy:
@HyukjinKwon Thanks for supporting!

PR #889, opened 2019-10-04
Fix plots on readthedocs.

ueshin:


ueshin:
Thanks! merging.

harupy:
What was causing this issue?

ueshin:
@harupy JDK in the build server was 8 by default, so PySpark didn't work.

HyukjinKwon:
The cause was that the build in Read the Docs (https://readthedocs.org/projects/koalas/builds/) failed, where we host our Koalas documentation.

When it generates documentations, now it has to run plot codes as of https://github.com/databricks/koalas/pull/847 .

Seems the environment in Read the Docs has JDK 11 by default; however, Spark versions up to 2.4 don't support JDK 11 yet.

So, we had to force to install JDK 8 at the build environment in Read the Docs.

harupy:
@HyukjinKwon @ueshin Thanks! I understand why the doc build was failing.

Why some plots didn't show up in the doc (as @ueshin pointed out [here](https://github.com/databricks/koalas/pull/847#issuecomment-537683743))?

HyukjinKwon:
Ah right. that's because Spark 2.4 works with JDK 11 partialy (unofficially). In some APIs, it doesn't work.

HyukjinKwon:
cc @itholic fyi

harupy:
@HyukjinKwon Thanks!

itholic:
@HyukjinKwon Thanks for cc me, and applaud to everyone have got into fix this issue. :)

PR #899, opened 2019-10-06
Implement Series.mode

### LINKED ISSUE #888: Series.mode
Implement `Series.mode`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.mode.html

itholic:
Resolves #888 

```python
>>> s = ks.Series([0, 0, 1, 1, 1, 2, 2, 2, 3, 3, 3,
...                np.nan, np.nan, np.nan])
>>> s
0     0.0
1     0.0
2     1.0
3     1.0
4     1.0
5     2.0
6     2.0
7     2.0
8     3.0
9     3.0
10    3.0
11    NaN
12    NaN
13    NaN
Name: 0, dtype: float64

>>> s.mode()
0    1.0
1    2.0
2    3.0
Name: 0, dtype: float64

>>> s.mode(False)
0    NaN
1    1.0
2    2.0
3    3.0
Name: 0, dtype: float64
```


databricks/koalas/series.py
```diff
@@ -3184,6 +3184,91 @@ def copy(self) -> 'Series':
...
+        2    2.0
+        3    3.0
+        Name: 0, dtype: float64
+        """
+        ser_count = self.value_counts(dropna=dropna)
+        sdf_count = ser_count._internal.sdf
+        most_value = sdf_count.head(1)[0][1]
+        sdf_most_value = (sdf_count.where("count == {}".format(most_value))
+                                   .sort('__index_level_0__')
+                                   .rdd.zipWithIndex().toDF())
```
ueshin: What's this doing?

itholic: Thanks for commenting and sorry for confusing.

let me explain these with some examples below.

first, we have series `ser`.
```python
>>> ser
lama    speed      45.0
        weight    250.0
        length      1.0
cow     speed      30.0
        weight    250.0
        length      1.0
falcon  speed     250.0
        weight      1.0
        length      0.3
Name: 0, dtype: float64
```

`sdf_count` is literally spark data frame that has count of each values

`ser`'s `sdf_count` looks like

```
>>> sdf_count.show()
+-----------------+-----+
|__index_level_0__|count|
+-----------------+-----+
|              1.0|    3|
|            250.0|    3|
|             30.0|    1|
|             45.0|    1|
|              0.3|    1|
+-----------------+-----+
```
and `most_value` is the mode value of `ser`.

in this case, it's 3

```
>>> most_value
3
```

And i used zipWithIndex() to make a column that have ascending number from 0.

because if there are plural mode values,

pandas shows them in ascending order with ascending number from 0 like below.

```python
>>> s.to_pandas().mode()
0      1.0
1    250.0
dtype: float64
```

and `sdf_most_value` looks like:

```
>>> sdf_most_value.show()
+----------+---+
|        _1| _2|
+----------+---+
|  [1.0, 3]|  0|
|[250.0, 3]|  1|
+----------+---+
```

so finally, i make `sdf` using data from `sdf_most_value`

```
>>> sdf.show()
+---+-----+
| _2|    0|
+---+-----+
|  0|  1.0|
|  1|250.0|
+---+-----+
```

actually while i'm doing this, I thought this code might be a bit harder to understand for someone.

However, there are no other idea came up from my brain, and sorry to make confuse you with complicated code. 😢

ueshin: Koalas automatically handle adding index if emtpy. We just need to set up the `sdf_most_value` without index.

itholic: @ueshin Thanks!! i'm going to fix them. :)

databricks/koalas/series.py
```diff
@@ -3227,6 +3227,86 @@ def copy(self) -> 'Series':
...
+        >>> s.mode(False)
+        0    NaN
+        1    1.0
+        2    2.0
+        3    3.0
+        Name: 0, dtype: float64
+        """
+        ser_count = self.value_counts(dropna=dropna)
+        sdf_count = ser_count._internal.sdf
+        most_value = sdf_count.head(1)[0][1]
```
ueshin: `ser_count.max()` should work?

databricks/koalas/series.py
```diff
@@ -3227,6 +3227,86 @@ def copy(self) -> 'Series':
...
+        sdf_most_value = (sdf_count.where("count == {}".format(most_value))
+                                   .sort('__index_level_0__'))
+        sdf = sdf_most_value.select(F.col('__index_level_0__').alias(self.name))
```
ueshin: `.alias(self.name)` doesn't work for tuple named Series.

databricks/koalas/series.py
```diff
@@ -3227,6 +3227,86 @@ def copy(self) -> 'Series':
...
+        1    1.0
+        2    2.0
+        3    3.0
+        Name: 0, dtype: float64
+        """
+        ser_count = self.value_counts(dropna=dropna)
+        sdf_count = ser_count._internal.sdf
+        most_value = sdf_count.head(1)[0][1]
+        sdf_most_value = (sdf_count.where("count == {}".format(most_value))
+                                   .sort('__index_level_0__'))
```
ueshin: Do we need `sort` here?

databricks/koalas/series.py
```diff
@@ -3227,6 +3227,86 @@ def copy(self) -> 'Series':
...
+        With 'dropna' set to 'False', we can also see NaN in the result
+
+        >>> s.mode(False)
+        0    NaN
+        1    1.0
+        2    2.0
+        3    3.0
+        Name: 0, dtype: float64
+        """
+        ser_count = self.value_counts(dropna=dropna)
```
ueshin: We should use `sort=False`?

databricks/koalas/series.py
itholic: i think not necessary, thanks! :)

databricks/koalas/series.py
itholic: yeah it works. it should be better thanks 😄

databricks/koalas/series.py
```diff
@@ -3223,6 +3223,89 @@ def copy(self) -> 'Series':
...
+        2    3.0
+        3    2.0
+        Name: 0, dtype: float64
+        """
+        ser_count = self.value_counts(dropna=dropna, sort=False)
+        sdf_count = ser_count._internal.sdf
+        most_value = ser_count.max()
+        sdf_most_value = sdf_count.where("count == {}".format(most_value))
+        sdf = sdf_most_value.select(
+            F.col(SPARK_INDEX_NAME_FORMAT(0)).alias('__temp__'))
```
HyukjinKwon: Why do we need this `__temp__` column?

HyukjinKwon: If this is to remove existing index, you can use `reset_index`.

HyukjinKwon: Ah, got it. So `__temp__` becomes the actual data. Can you use `'0'` because that's the default name of a Koalas series (`ks.Series([1,2,3]).name`)?

databricks/koalas/series.py
```diff
@@ -3223,6 +3223,89 @@ def copy(self) -> 'Series':
...
+        1    1.0
+        2    3.0
+        3    2.0
+        Name: 0, dtype: float64
+        """
+        ser_count = self.value_counts(dropna=dropna, sort=False)
+        sdf_count = ser_count._internal.sdf
+        most_value = ser_count.max()
+        sdf_most_value = sdf_count.where("count == {}".format(most_value))
+        sdf = sdf_most_value.select(
```
HyukjinKwon: I would also explain that this index becomes a column with a comment there.

databricks/koalas/series.py
itholic: thanks! i just fixed it.

databricks/koalas/series.py
ueshin: ah, we might need to sort the result values since the doc says it returns values in sorted order. [here](https://github.com/databricks/koalas/pull/899/files#diff-f38e425003958ac0d1bd6ffa418f2613R3240), also https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.mode.html Otherwise we should remove "in sorted order" from the doc.
See #1047
cc @HyukjinKwon

itholic: In my opinion, maybe it would be better sort the order like pandas for now, since they wrote down explicitly "sorted order" only for `Series.mode()`.
btw, For those people who prefer to see sorted results, maybe we can make an option value to make possible to choose whether we sort the result or not?

HyukjinKwon: let's fix it to sort first and see if people actually need it and report an issue.
Thanks @ueshin. I missed https://github.com/databricks/koalas/pull/899#discussion_r346011895

PR #902, opened 2019-10-07
Support PyArrow 0.15.0 in Travis CI

HyukjinKwon:


HyukjinKwon:
Woah, Spark 2.4 don't support PyArrow 0.15.0.

databricks/koalas/internal.py
```diff
@@ -638,6 +642,15 @@ def spark_df(self) -> spark.DataFrame:
...
         """ Return as pandas DataFrame. """
         sdf = self.spark_internal_df
+
+        if LooseVersion(pyarrow.__version__) >= LooseVersion("0.15") and \
+                LooseVersion(pyspark.__version__) < LooseVersion("3.0"):
+            # There seems a bug in PyArrow 0.15 with PySpark 2.4 and below.
+            # TODO(Hyukjin): I have no idea why only inserting limit works.
+            #   Other APIs like noop map, explicitly shuffle and sample with
+            #   fraction 1 still don't work.
+            sdf = sdf.limit(2147483647)
```
HyukjinKwon: I have no idea at all why only `limit` works. I tried some other APIs I could think like map with empty function, sort, sample with fraction 1.0 but none of them work.

HyukjinKwon:
cc @ueshin ... at least the tests passed ..

databricks/koalas/internal.py
```diff
@@ -638,7 +642,26 @@ def spark_df(self) -> spark.DataFrame:
...
         sdf = self.spark_internal_df
-        pdf = sdf.toPandas()
+
+        if LooseVersion(pyarrow.__version__) >= LooseVersion("0.15") and \
+                LooseVersion(pyspark.__version__) < LooseVersion("3.0"):
+            # There seems a bug in PyArrow 0.15 with PySpark 2.4 and below.
+            # TODO(Hyukjin): I have no idea why only inserting limit works.
+            #   Other APIs like noop map, explicitly shuffle and sample with
+            #   fraction 1 still don't work.
+            max_count = 2147483646  # JVM integer's max value - 1.
```
HyukjinKwon: BTW, if we guess each row is, say, at least 8 bytes or say 16 bytes, then the output is at least 16 GB or 32 GB. I don't think this is not quite common.

databricks/koalas/internal.py
```diff
@@ -638,7 +642,26 @@ def spark_df(self) -> spark.DataFrame:
...
+            sdf = sdf.limit(max_count + 1)
```
HyukjinKwon: I hope we can find a better alternative instead of doing a `limit` ...

HyukjinKwon:
15.1 seems going to be release soon .. let me try that one later.

HyukjinKwon:
Nice! @ueshin, tests all passed!

HyukjinKwon:
I made a followup to fix setup.py too at https://github.com/databricks/koalas/commit/a0fc9af0517b9df536031502acef19737775ba5e

Issue #823, opened 2019-09-23
Implement Named Aggregation


charlesdong1991:
Pandas has put forward a named aggregation for groupby aggregation since 0.25.0, I am working on an implementation on this.

PR #904, opened 2019-10-07
Implement nested renaming for groupby agg

charlesdong1991:
This allows koalas to have nested renaming/selection:
```python
kdf = ks.DataFrame({'A': [1, 1, 2, 2],
                   'B': [1, 2, 3, 4],
                   'C': [0.362, 0.227, 1.267, -0.562]},columns=['A', 'B', 'C'])
```

Screen Shot 2019-10-07 at 11 15 22 PM

WIP: still need to add some tests and docstrings

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -1969,6 +1969,8 @@ def _is_multi_agg_with_relabel(**kwargs):
     >>> _is_multi_agg_with_relabel()
     False
     """
+    if not kwargs:
+        return False
     return all(
         isinstance(v, tuple) and len(v) == 2
         for v in kwargs.values()
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -1972,7 +1972,7 @@ def _is_multi_agg_with_relabel(**kwargs):
     return all(
         isinstance(v, tuple) and len(v) == 2
         for v in kwargs.values()
-    ) and kwargs
+    )
 
 
 def _normalize_keyword_aggregation(kwargs):
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -2030,7 +2030,7 @@ def _normalize_keyword_aggregation(kwargs):
     order = []
     columns, pairs = list(zip(*kwargs.items()))
 
-    for _, (column, aggfunc) in zip(columns, pairs):
+    for column, aggfunc in pairs:
         if column in aggspec:
             aggspec[column].append(aggfunc)
         else:
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -18,6 +18,7 @@
 A wrapper for GroupedData to behave similar to pandas GroupBy.
 """
 
+import sys
 import inspect
 from collections import Callable, OrderedDict
 from functools import partial
@@ -53,7 +54,7 @@ class GroupBy(object):
 
     # TODO: Series support is not implemented yet.
     # TODO: not all arguments are implemented comparing to Pandas' for now.
-    def aggregate(self, func_or_funcs, *args, **kwargs):
+    def aggregate(self, func_or_funcs=None, *args, **kwargs):
         """Aggregate using one or more operations over the specified axis.
 
         Parameters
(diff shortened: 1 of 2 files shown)


ueshin:
@charlesdong1991 Could you put the result of the example in the description?

charlesdong1991:
sorry, i actually didn't know how to add the result dataframe there, so didn't put it in description, but it is the same one in the test. I added a screenshot for this instead. @ueshin

ueshin:
@charlesdong1991 thanks, the screenshot is fine.

harupy:
@charlesdong1991 
Is this related to #823 ?

charlesdong1991:
ahh, yeah, kinda first step to implement named aggregation. @harupy

databricks/koalas/groupby.py
```diff
@@ -1931,3 +1945,78 @@ def value_counts(self, sort=None, ascending=None, dropna=True):
...
+    PY36 = sys.version_info >= (3, 6)
+    if not PY36:
+        kwargs = OrderedDict(sorted(kwargs.items()))
+
+    # TODO(Py35): When we drop python 3.5, change this to defaultdict(list)
+    aggspec = OrderedDict()
+    order = []
+    columns, pairs = list(zip(*kwargs.items()))
+
+    for name, (column, aggfunc) in zip(columns, pairs):
```
ueshin: Seems like `name` is not used?

charlesdong1991: ahh, yeah

ueshin: yeah, seems not needed to zip columns.

charlesdong1991: thanks, this is my bad, i shouldn't be lazy at this. @harupy

databricks/koalas/groupby.py
```diff
@@ -127,7 +128,34 @@ def aggregate(self, func_or_funcs, *args, **kwargs):
         A
         1    1    2  0.227  0.362
         2    3    4 -0.562  1.267
+
+        To control the output names with different aggregations per column, koalas
```
ueshin: `koalas` -> `Koalas`

databricks/koalas/tests/test_groupby.py
```diff
@@ -183,6 +183,27 @@ def test_aggregate_func_str_list(self):
...
+    def test_aggregate_relabel(self):
+        # this is to test named aggregation in groupby
+        pdf = pd.DataFrame({"group": ['a', 'a', 'b', 'b'],
+                            "A": [0, 1, 2, 3],
+                            "B": [5, 6, 7, 8]})
+        kdf = koalas.from_pandas(pdf)
+
+        # this is only applied in version after 0.25.0
+        if pd.__version__ < "0.25.0":
+            return
```
ueshin: Could you use `@unittest.skipIf( ... )` instead?

databricks/koalas/tests/test_groupby.py
```diff
@@ -805,3 +826,10 @@ def test_missing(self):
                                         "property.*GroupBy.*{}.*is deprecated"
                                         .format(name)):
                 getattr(kdf.a.groupby('a'), name)
+
+    @staticmethod
+    def test_is_multi_agg_with_relabel():
+        from databricks.koalas.groupby import _is_multi_agg_with_relabel
```
ueshin: Shall we move this to the file header?

databricks/koalas/tests/test_groupby.py
charlesdong1991: nice!

ueshin:
Hmm, I found a problem with multi-index columns:

```py
>>> kdf = ks.DataFrame({"group": ['a', 'a', 'b', 'b'], "A": [0, 1, 2, 3], "B": [5, 6, 7, 8]})
>>> kdf.columns = pd.MultiIndex.from_tuples([('x', 'group'), ('y', 'A'), ('y', 'B')])
>>> kdf
      x  y
  group  A  B
0     a  0  5
1     a  1  6
2     b  2  7
3     b  3  8
>>> kdf.groupby(('x', 'group')).agg(a_max=(('y', 'A'), "max"))
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/groupby.py", line 179, in aggregate
    kdf = kdf[order]
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/frame.py", line 7272, in __getitem__
    return self._pd_getitem(key)
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/frame.py", line 7208, in _pd_getitem
    return self.loc[:, key]
  File "/Users/ueshin/workspace/databricks-koalas/master/databricks/koalas/indexing.py", line 447, in __getitem__
    raise ValueError('All the key level should be the same as column index level.')
ValueError: All the key level should be the same as column index level.
```

@charlesdong1991 Is it possible to support this case?

charlesdong1991:
i don't think pandas supports such case. I haven't checked, will check it once i am back home. Or we could have it in a separate PR if you agree. @ueshin

ueshin:
I'm fine with having it in a separate PR. I'd merge this now.

ueshin:
Thanks! merging.

PR #901, opened 2019-10-07
Add an API compatibility guide for pandas and PySpark users

HyukjinKwon:
Screen Shot 2019-10-10 at 1 26 53 PM
Screen Shot 2019-10-10 at 1 26 58 PM

PR #906, opened 2019-10-08
Add 'index_col' parameter to DataFrame.to_spark

HyukjinKwon:
This PR adds `index_col` parameter to `DataFrame.to_spark`. This PR relates to https://github.com/databricks/koalas/pull/901.

After this PR, we can do a roundtrip in Koalas and Spark DataFrame easily.

```python
>>> kdf = ks.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6], 'c': [7, 8, 9]})
>>> sdf = kdf.to_spark(index_col="index").filter("a == 2")  # PySpark API
>>> kdf = sdf.to_koalas(index_col="index")
>>> kdf
       a  b  c
index
1      2  5  8
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2741,7 +2741,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                                              index_map=[(index_column,
                                                          self._internal.index_names[0])])))
 
-    def to_koalas(self, index_col: Optional[Union[str, List]] = None):
+    def to_koalas(self, index_col: Optional[Union[str, List[str]]] = None):
         """
         Converts the existing DataFrame into a Koalas DataFrame.
 
@@ -3063,7 +3063,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         self.to_spark().write.save(
             path=path, format=format, mode=mode, partitionBy=partition_cols, options=options)
 
-    def to_spark(self, index_col: Optional[Union[str, List]] = None):
+    def to_spark(self, index_col: Optional[Union[str, List[str]]] = None):
         """
         Return the current DataFrame as a Spark DataFrame.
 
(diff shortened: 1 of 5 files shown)


databricks/koalas/internal.py
```diff
@@ -599,34 +605,30 @@ def column_index_names(self) -> Optional[List[str]]:
         """ Return names of the index levels. """
         return self._column_index_names
 
+    def _data_column_names(self):
+        index_columns = set(self.index_columns)
+        data_columns = []
+        for i, (column, idx) in enumerate(zip(self._data_columns, self.column_index)):
+            assert column not in index_columns
```
HyukjinKwon: @ueshin, mind if I ask to double check if this is correct?

databricks/koalas/frame.py
```diff
@@ -2742,7 +2741,7 @@ def duplicated(self, subset=None, keep='first'):
                                              index_map=[(index_column,
                                                          self._internal.index_names[0])])))
 
-    def to_koalas(self, index_col=None):
+    def to_koalas(self, index_col: Optional[Union[str, List]] = None):
```
ueshin: `Optional[Union[str, List[str]]`?

databricks/koalas/frame.py
```diff
@@ -3064,18 +3063,90 @@ def to_spark_io(self, path: Optional[str] = None, format: Optional[str] = None,
         self.to_spark().write.save(
             path=path, format=format, mode=mode, partitionBy=partition_cols, options=options)
 
-    def to_spark(self):
+    def to_spark(self, index_col: Optional[Union[str, List]] = None):
```

databricks/koalas/frame.py
```diff
@@ -3064,18 +3063,90 @@ def to_spark_io(self, path: Optional[str] = None, format: Optional[str] = None,
...
+        index_1 index_2
+        0       1        4  7
+        1       2        5  8
+        2       3        6  9
+        """
+        if index_col is None:
+            return self._internal.spark_df
+        else:
+            if isinstance(index_col, str):
+                index_col = [index_col]
```
ueshin: We should check the length of `index_col` to match the number of index?

ueshin: Also we should check the names are not duplicated with data column names.

databricks/koalas/internal.py
```diff
@@ -398,11 +403,12 @@ def __init__(self, sdf: spark.DataFrame,
         assert isinstance(sdf, spark.DataFrame)
         if index_map is None:
             # Here is when Koalas DataFrame is created directly from Spark DataFrame.
-            assert "__index_level_0__" not in sdf.schema.names, \
-                "Default index column should not appear in columns of the Spark DataFrame"
+            assert any(SPARK_INDEX_NAME_PATTERN.match(name) for name in sdf.schema.names), \
```
ueshin: `not any(SPARK_INDEX_NAME_PATTERN.match(name) for name in sdf.schema.names)` or `all(not SPARK_INDEX_NAME_PATTERN.match(name) for name in sdf.schema.names)`?

databricks/koalas/internal.py
ueshin: Actually we do have such a case. We can't raise an error.

HyukjinKwon: Hmmmm .. okie let me revert it back for now.

databricks/koalas/internal.py
```diff
@@ -36,6 +36,11 @@
 from databricks.koalas.utils import column_index_level, default_session, lazy_property, scol_for
 
 
+# A function to turn given numbers to Spark columns that represent Koalas index.
+SPARK_INDEX_NAME_FORMAT = "__index_level_{}__".format
+# A pattern to check if the name of a Spark column is a Koalas index name or not.
+SPARK_INDEX_NAME_PATTERN = re.compile(r"__index_level_[0-9]+__")
```
HyukjinKwon: These changes are not related actually ..

ueshin:
Thanks! merging.

HyukjinKwon:
Thanks!!

PR #911, opened 2019-10-09
Add NamedAgg

### LINKED ISSUE #823: Implement Named Aggregation

charlesdong1991:
This is equivalent to nested renaming, so like pandas, we have `koalas.NamedAgg` now for this:
```python
        >>> aggregated = df.groupby('A').agg(b_max=ks.NamedAgg(column='B', aggfunc='max'))
        >>> aggregated
             b_max
        A
        1        2
        2        4
```

Resolves #823

### DIFF
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -45,7 +45,7 @@ from databricks.koalas.groupby import NamedAgg
 __all__ = ['read_csv', 'read_parquet', 'to_datetime', 'from_pandas',
            'get_dummies', 'DataFrame', 'Series', 'Index', 'MultiIndex', 'pandas_wraps',
            'sql', 'range', 'concat', 'melt', 'get_option', 'set_option', 'reset_option',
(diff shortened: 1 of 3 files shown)


databricks/koalas/__init__.py
```diff
@@ -40,6 +40,7 @@ def assert_pyspark_version():
 from databricks.koalas.series import Series
 from databricks.koalas.typedef import pandas_wraps
 from databricks.koalas.config import get_option, set_option, reset_option, options
+from databricks.koalas.groupby import NamedAgg
```
HyukjinKwon: Can we add it to `__all__`?

charlesdong1991: thanks! added!

HyukjinKwon:
Thanks @charlesdong1991!

Issue #880, opened 2019-10-03
Index.unique


ueshin:
Implement `Index.unique`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.unique.html

charlesdong1991:
i will take this one

PR #912, opened 2019-10-09
Add Index unique

charlesdong1991:
#880 

```python
ks.DataFrame({'a': ['a', 'b', 'c']}, index=[1, 1, 3]).index.unique()
Int64Index([1, 3], dtype='int64')
```


charlesdong1991:
emm, any idea how to fix `missing function` failure? @ueshin the error message doesn't seem very intuitive.

ueshin:
@charlesdong1991 Thanks for working on this!
You also need to remove the line from `_MissingPandasLikeMultiIndex`.

charlesdong1991:
ahh, thanks for your quick reply! @ueshin

databricks/koalas/tests/test_indexes.py
```diff
@@ -140,6 +140,21 @@ def test_multi_index_names(self):
         with self.assertRaises(PandasNotImplementedError):
             kidx.name = 'renamed'
 
+    def test_index_unique(self):
```
HyukjinKwon: Sorry if I missed but did we have a test case for multi-index?

charlesdong1991: no, i think their implementation is slightly different, and i thought the scope of this PR is for Index, not MultiIndex (I was supposed to do in a separate PR).

But do you prefer to have this `unique` function added for both Index and MultiIndex in this PR? @HyukjinKwon

HyukjinKwon: Ah, it's fine to target to implement it only in `Index`. Can we explicitly throw a notimplemented exception manually?

charlesdong1991: i added a `PandasNotImplementedError`, pls let me know if this is what you meant, thanks.

charlesdong1991:
any follow up review will be appreciated! :)

databricks/koalas/missing/indexes.py
```diff
@@ -104,7 +104,6 @@ class _MissingPandasLikeIndex(object):
     to_numpy = unsupported_function('to_numpy')
     transpose = unsupported_function('transpose')
     union = unsupported_function('union')
-    unique = unsupported_function('unique')
     value_counts = unsupported_function('value_counts')
     view = unsupported_function('view')
     where = unsupported_function('where')
```
HyukjinKwon: @charlesdong1991, sorry can you remove `unique` at `_MissingPandasLikeMultiIndex` as well? 

Techinically `MultiIndex` has the method called `unique`. It's a bit complicated because `MultiIndex` inherits `Index`, so I suggested to throw an exception :-)..

charlesdong1991: ok, removed! @HyukjinKwon 

i will get some free time recently, and see if I could add some properties and functions for `Index` and `MultiIndex` and then implement `unique` for `MultiIndex` if itholic hasn't implemented them all beforehand xD 😝

HyukjinKwon: That would be awesome!!

HyukjinKwon:
LGTM. I will merge once the conflicts are resolved.

charlesdong1991:
oops, should have merged to master before pushing commits.

PR #918, opened 2019-10-11
Set an upper bound for pyarrow version.

ueshin:
I'd set the upper bound for pyarrow version for now since we need to set an environment variable to work with pyarrow 0.15.

https://github.com/apache/spark/pull/26045:
> Arrow 0.15.0 introduced a change in format which requires an environment variable to maintain compatibility.


ueshin:
Thanks! merging.

PR #920, opened 2019-10-11
Fix melt for multi-index columns support.

ueshin:


databricks/koalas/frame.py
```diff
@@ -6684,7 +6684,7 @@ def _reindex_columns(self, columns):
 
         return self._internal.copy(sdf=sdf, data_columns=columns, column_index=idx)
 
-    def melt(self, id_vars=None, value_vars=None, var_name='variable',
+    def melt(self, id_vars=None, value_vars=None, var_name=None,
```
HyukjinKwon: Seems the default value of `melt`'s `var_name` at namespace.py should be changed as well.

ueshin: ah, good catch!

databricks/koalas/frame.py
```diff
@@ -6769,29 +6771,52 @@ def melt(self, id_vars=None, value_vars=None, var_name='variable',
         """
         if id_vars is None:
             id_vars = []
-        if not isinstance(id_vars, (list, tuple, np.ndarray)):
-            id_vars = list(id_vars)
+        elif isinstance(id_vars, str):
+            id_vars = [(id_vars,)]
+        elif isinstance(id_vars, tuple):
```
HyukjinKwon: Seems the tuple alone is not allowed when multi-index:

```python
>>> pdf.melt(id_vars=('X', 'A'))
  File "/usr/local/lib/python3.7/site-packages/pandas/core/frame.py", line 6500, in melt
    col_level=col_level,
  File "/usr/local/lib/python3.7/site-packages/pandas/core/reshape/melt.py", line 42, in melt
    "id_vars must be a list of tuples when columns" " are a MultiIndex"
ValueError: id_vars must be a list of tuples when columns are a MultiIndex
```

vs

```python
>>> kdf.melt(id_vars=('X', 'A'))
   ('X', 'A') variable_0 variable_1  value
0           1          X          B      2
1           1          Y          C      7
2           3          X          B      4
3           3          Y          C      8
4           5          X          B      6
5           5          Y          C      9
```

Maybe we should check if via `len(self._internal.index_map) == 1:` and disallow a tuple alone.

and .. I think it should be considered as multiple columns. See

```
>>> kdf.melt(id_vars=('A', 'B'))
  File "/.../koalas/databricks/koalas/frame.py", line 6818, in melt
    for name in var_name[:self._internal.column_index_level]] +
  File "/.../koalas/databricks/koalas/frame.py", line 6816, in <listcomp>
    for idx in id_vars] +
  File "/.../koalas/databricks/koalas/internal.py", line 534, in scol_for
    return scol_for(self._sdf, self.column_name_for(column_name_or_index))
  File "/.../koalas/databricks/koalas/internal.py", line 523, in column_name_for
    raise KeyError(column_name_or_index)
KeyError: ('A', 'B')
```

vs

```python
>>> pdf.melt(id_vars=('A', 'B'))
   A  B variable  value
0  1  2        C      7
1  3  4        C      8
2  5  6        C      9
```

ueshin: Actually, I thought it's a weird behavior but let's follow pandas for now.

HyukjinKwon:
Thanks, merged.

PR #921, opened 2019-10-12
Implement Series.xs

itholic:
Like pandas Series.xs (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.xs.html#pandas.Series.xs)

implemented function `xs` for series.

```python
>>> s
a  lama    speed      45.0
           weight    200.0
           length      1.2
b  cow     speed      30.0
           weight    250.0
           length      1.5
c  falcon  speed     320.0
           weight      1.0
           length      0.3
Name: 0, dtype: float64


>>> s.xs('a')
lama  speed      45.0
      weight    200.0
      length      1.2
Name: 0, dtype: float64



>>> s.xs(('a', 'lama'))
speed      45.0
weight    200.0
length      1.2
Name: 0, dtype: float64



>>> s.xs('lama', level=1)
a  speed      45.0
   weight    200.0
   length      1.2
Name: 0, dtype: float64
```


databricks/koalas/series.py
```diff
@@ -3409,6 +3409,104 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+                [self._internal.scol_for(self._internal.column_index[0])])
+        rows = [self._internal.scols[lvl] == index
+                for lvl, index in enumerate(key, level)]
+        sdf = self._internal.sdf \
+            .select(cols) \
+            .where(reduce(lambda x, y: x & y, rows))
+
+        if len(self._index_map) == len(key):
+            # if sdf has one column and one data, return data only without frame
+            if sdf.select(self.name).limit(2).count() == 1:
```
HyukjinKwon: I think you can collect first and use it for both places. For instance,

```python
pdf = sdf.limit(2).to_pandas()
length = len(pdf)
If length == 1:
    return pdf[self.name].iloc[0]
```

itholic: @HyukjinKwon Thanks for reviewing!! well i think spark dataframe has no attribute to_pandas(), hasn't it?

itholic: @HyukjinKwon Thanks for reviewing!!! 😄

databricks/koalas/series.py
```diff
@@ -3409,6 +3409,106 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+            # if sdf has one column and one data, return data only without frame
+            pdf = sdf.limit(2).toPandas()
+            length = len(pdf)
+            if length == 1:
+                return pdf[self.name].iloc[0]
+
+        index_scols = [col for col in sdf.columns if col not in self._internal.data_columns]
+        internal = self._internal.copy(
+            sdf=sdf,
+            index_map=[(index_scol, None) for index_scol in index_scols])
```
ueshin: We should manage `index_map` properly. This will lose the index names.

itholic: Thanks for commenting!! fixed it.

databricks/koalas/series.py
```diff
@@ -3436,6 +3436,107 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+
+        Get values at specified index and level
+
+        >>> s.xs('lama', level=1)
+        a  speed      45.0
+           weight    200.0
+           length      1.2
+        Name: 0, dtype: float64
+        """
+        if not isinstance(key, (str, tuple)):
```
HyukjinKwon: This one too. I don't think we should restrict the value to be a string.

databricks/koalas/series.py
```diff
@@ -3436,6 +3436,107 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+
+        if len(self._internal._index_map) == len(key):
+            # if sdf has one column and one data, return data only without frame
+            pdf = sdf.limit(2).toPandas()
+            length = len(pdf)
+            if length == 1:
+                return pdf[self.name].iloc[0]
+
+        index_scols = [col for col in sdf.columns if col not in self._internal.data_columns]
+        index_map_dict = {imap[0]: imap[1] for imap in self._internal.index_map}
```
HyukjinKwon: `dict(self._internal.index_map)`

databricks/koalas/series.py
```diff
@@ -3436,6 +3436,107 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+            .where(reduce(lambda x, y: x & y, rows))
+
+        if len(self._internal._index_map) == len(key):
+            # if sdf has one column and one data, return data only without frame
+            pdf = sdf.limit(2).toPandas()
+            length = len(pdf)
+            if length == 1:
+                return pdf[self.name].iloc[0]
+
+        index_scols = [col for col in sdf.columns if col not in self._internal.data_columns]
```
HyukjinKwon: I wouldn't call it `scols` because it's not Spark columns but column names. you could just call it something like `index_columns`

HyukjinKwon:
Seems fine but let me leave it to @ueshin

databricks/koalas/series.py
itholic: ah, this naming might be confusing. thanks!!

databricks/koalas/series.py
itholic: wow i didn't know it can be worked like this. thanks! 👍

databricks/koalas/series.py
itholic: yeah, i fixed it also. thanks!!

itholic:
@ueshin Thanks 😄

PR #922, opened 2019-10-13
Implement Series.where

itholic:
Like pandas Series.where (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.where.html)

implemented function `where` for series.

```python
>>> s1 = ks.Series([0, 1, 2, 3, 4])
>>> s2 = ks.Series([100, 200, 300, 400, 500])
>>> s1.where(s1 > 0)
0    NaN
1    1.0
2    2.0
3    3.0
4    4.0
Name: 0, dtype: float64


>>> s1.where(s1 > 1, 10)
0    10
1    10
2     2
3     3
4     4
Name: 0, dtype: int64

>>> s1.where(s1 > 1, s1 + 50)
0    50
1    51
2     2
3     3
4     4
Name: 0, dtype: int64


>>> s1.where(s1 > 1, s2)
0    100
1    200
2      2
3      3
4      4
Name: 0, dtype: int64
```


databricks/koalas/series.py
```diff
@@ -3409,6 +3409,80 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+
+        >>> s1.where(s1 > 1, s2)
+        0    100
+        1    200
+        2      2
+        3      3
+        4      4
+        Name: 0, dtype: int64
+        """
+        current = F.when(cond._internal.data_scols[0], self._scol)
```
ueshin: We should check the input? e.g., what if cond is `DataFrame`?

databricks/koalas/series.py
```diff
@@ -3409,6 +3409,80 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+            if self_sdf == other_sdf:
+                current = current.otherwise(other._scol)
+                result = self._with_new_scol(current)
+            else:
+                temp_col = self.name + "_"
+                temp_idx = self._index_map[0][0]
+                other_temp = other_sdf.withColumn(temp_col, other_sdf[other.name])
+                new_sdf = self_sdf.join(other_temp, temp_idx).sort(temp_idx)
+                current = current.otherwise(other_temp[temp_col])
+                result = _col(DataFrame(_InternalFrame(sdf=new_sdf.select(current))))
```
ueshin: I think we can manage index.

databricks/koalas/series.py
itholic: yeah we should check! Thanks for reviewing 😄

databricks/koalas/series.py
```diff
@@ -3409,6 +3409,80 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+            self_sdf = self._internal.sdf
+            other_sdf = other._internal.sdf
+            if self_sdf == other_sdf:
+                current = current.otherwise(other._scol)
+                result = self._with_new_scol(current)
+            else:
+                temp_col = self.name + "_"
+                temp_idx = self._index_map[0][0]
+                other_temp = other_sdf.withColumn(temp_col, other_sdf[other.name])
+                new_sdf = self_sdf.join(other_temp, temp_idx).sort(temp_idx)
```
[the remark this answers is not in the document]
HyukjinKwon: Yea, I think you can do, for instance,

```python
self_sdf[temp_col] = ...
current.otherwise(other_temp[temp_col])
```

Keep it same as https://github.com/databricks/koalas/commit/8367c93343da181fe1a3b0702efe75b97f522460#diff-35732ddb08a2af7a8c8a0c3c4ad54b52R514 

Adding temp columns isn't a good idea. We might have to use `utils.align_diff_series` there as well.

databricks/koalas/series.py
```diff
@@ -3409,6 +3409,80 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
 
         return self._with_new_scol(current)
 
+    def where(self, cond, other=np.nan):
```
HyukjinKwon: @itholic seems like pandas shares the same implementation internally. After this PR is merged, can you move this into `_Frame` class and implement `DataFrame.where` as well?

HyukjinKwon:
Seems fine to me otherwise.

databricks/koalas/series.py
itholic: okay, i'm going to work right after this PR is merged

databricks/koalas/series.py
itholic: Thanks for reviewing!! i'm going to manage it.

databricks/koalas/series.py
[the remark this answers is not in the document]
itholic: Thanks for commenting!! i'm trying to use `align_diff_series`, and have got a little confused since there is no usecase examples of `align_diff_series` 😭

databricks/koalas/series.py
```diff
@@ -3605,6 +3605,86 @@ def replace(self, to_replace=None, value=None, regex=False) -> 'Series':
...
+        # |__index_level_0__|  0|__tmp_cond_col__|__tmp_other_col__|
+        # +-----------------+---+----------------+-----------------+
+        # |                0|  0|           false|              100|
+        # |                1|  1|           false|              200|
+        # |                3|  3|            true|              400|
+        # |                2|  2|            true|              300|
+        # |                4|  4|            true|              500|
+        # +-----------------+---+----------------+-----------------+
+        data_col_name = self._internal.column_name_for(self._internal.column_index[0])
+        index_column = self._internal.index_columns[0]
```
HyukjinKwon: @itholic, I think this doesn't support multi-level index cases. Can you fix this please?

HyukjinKwon: `index_columns` can be multiple and we cannot just use the first one only.

databricks/koalas/tests/test_series.py
```diff
@@ -31,10 +31,21 @@
...
 
 class SeriesTest(ReusedSQLTestCase, SQLTestUtils):
 
+    @classmethod
+    def setUpClass(cls):
+        super(SeriesTest, cls).setUpClass()
+        set_option("compute.ops_on_diff_frames", True)
+
+    @classmethod
+    def tearDownClass(cls):
```
HyukjinKwon: @itholic disable this. `compute.ops_on_diff_frames` is disabled by default because it costs a lot. We should move the test cases into `OpsOnDiffFramesEnabledTest`

databricks/koalas/tests/test_series.py
```diff
@@ -742,6 +753,23 @@ def test_duplicates(self):
         self.assert_eq(pser.drop_duplicates().sort_values(),
                        kser.drop_duplicates().sort_values())
 
+    def test_where(self):
+        pser1 = pd.Series([0, 1, 2, 3, 4], name=0)
```
HyukjinKwon: Can you add a test when `compute.ops_on_diff_frames` is off? I think we can still use a scalar values for `other` such as `int`.

PR #928, opened 2019-10-15
Implement Series.truncate

itholic:
Like pandas Series.truncate (https://pandas.pydata.org/pandas-docs/version/0.21/generated/pandas.Series.truncate.html#pandas.Series.truncate)

implemented function `truncate` for Series.

```python
>>> s = ks.Series([10, 20, 30, 40, 50, 60, 70],
...               index=[1, 2, 3, 4, 5, 6, 7])
>>> s
1    10
2    20
3    30
4    40
5    50
6    60
7    70
Name: 0, dtype: int64

>>> s.truncate(2, 5)
2    20
3    30
4    40
5    50
Name: 0, dtype: int64

>>> s = ks.Series([10, 20, 30, 40, 50, 60, 70],
...               index=['a', 'b', 'c', 'd', 'e', 'f', 'g'])
>>> s
a    10
b    20
c    30
d    40
e    50
f    60
g    70
Name: 0, dtype: int64

>>> s.truncate('b', 'e')
b    20
c    30
d    40
e    50
Name: 0, dtype: int64
```


databricks/koalas/series.py
```diff
@@ -3223,6 +3223,90 @@ def copy(self) -> 'Series':
...
+
+        >>> s.truncate('b', 'e')
+        b    20
+        c    30
+        d    40
+        e    50
+        Name: 0, dtype: int64
+        """
+        indexes = self.index.to_pandas()
+        indexes_increasing = indexes.is_monotonic_increasing
```
HyukjinKwon: `is_monotonic_increasing` will need to sort anyway. We don't you just sort?

databricks/koalas/series.py
```diff
@@ -3223,6 +3223,90 @@ def copy(self) -> 'Series':
...
+        Name: 0, dtype: int64
+
+        >>> s.truncate('b', 'e')
+        b    20
+        c    30
+        d    40
+        e    50
+        Name: 0, dtype: int64
+        """
+        indexes = self.index.to_pandas()
```
HyukjinKwon: @itholic, this will collect everything in index into driver's memory. We should avoid this.

HyukjinKwon: Koalas index implements `is_monotonic_increasing` and `is_monotonic_decreasing`. The problem is that this is an expensive operation as warned in their documentations as it leads all data into single node.

Can you add a note that this API will be expensive (see `is_monotonic_increasing` as an example). We even might have to explicitly don't implement. cc @ueshin

itholic: @HyukjinKwon Thanks for commenting!

yeah you're right. i used `is_monotonic_increasing` of pandas' because there was a bug for koalas'. (so i fixed in #930 )

now we can use koalas', but there still have an issue about cost as you said.

databricks/koalas/series.py
```diff
@@ -3224,6 +3224,100 @@ def copy(self) -> 'Series':
...
+        >>> s.truncate('b', 'e')
+        b    20
+        c    30
+        d    40
+        e    50
+        Name: 0, dtype: int64
+        """
+        indexes = self.index
+        indexes_increasing = indexes.is_monotonic_increasing
+        if not any([indexes_increasing, indexes.is_monotonic_decreasing]):
```
HyukjinKwon: @itholic can you just use this condition?

```python
not indexes.is_monotonic_increasing or not indexes.is_monotonic_decreasing
```

It will allow a short circuit and we can avoid to call both `is_monotonic_increasing` and `is_monotonic_decreasing`.

databricks/koalas/series.py
```diff
@@ -3224,6 +3224,100 @@ def copy(self) -> 'Series':
...
+            raise ValueError("truncate requires a sorted index")
+        if (before is None) and (after is None):
+            return self.copy() if copy else self
+
+        if before is None:
+            sdf = indexes._internal.sdf
+            idx_col_name = self._internal.index_columns[0]
+            before = sdf.first()[idx_col_name] if indexes_increasing \
+                else sdf.orderBy(self._internal.scol_for(idx_col_name).desc()).first()[idx_col_name]
```
HyukjinKwon: Why do we need to calculate the last or first value? Can't we just do `self.to_frame()[before:]` or `self.to_frame()[:after]`? It looks too expensive

databricks/koalas/series.py
itholic: @HyukjinKwon thanks for commenting!

but maybe this will raise `ValueError` even when `indexes.is_monotonic_increasing` is `False` and `indexes.is_monotonic_decreasing` is `True` ?? 

in this case, we should pass them without raises Error.

databricks/koalas/series.py
itholic: @HyukjinKwon Thanks for commenting!

oops i didn't realize that we can use `None` value when slicing.

i'm going to fix it right away.

databricks/koalas/series.py
HyukjinKwon: Oh, then maybe `not (indexes.is_monotonic_increasing or indexes.is_monotonic_decreasing)`

HyukjinKwon: I see. can we simply like this: `not indexes.is_monotonic_increasing and not indexes.is_monotonic_decreasing`?

itholic: okay i agree with that is the more readable. thanks for reviewing!!

PR #930, opened 2019-10-15
Make is_monotonic~ work properly for index

### LINKED ISSUE #931: is_monotonic_decreasing & is_monotonic_increasing for index not work properly
When we use `is_monotonic_decreasing` or `is_monotonic_increasing` for index,

pandas works like below:

```python
>>> s = pd.Series([7, 6, 5, 4, 3, 2, 1], index=[7, 6, 5, 4, 3, 2, 1])
>>> s.index.is_monotonic_decreasing
True
```
Since the index order is literally motononic decreasing, they return `True`

But our case,

```python
>>> s = ks.Series([7, 6, 5, 4, 3, 2, 1], index=[7, 6, 5, 4, 3, 2, 1])
>>> s.index.is_monotonic_decreasing
False
```

as seen above, it returns `False`.

because our existing logic always order by index column before calculate the each rows increasing or decreasing.

So they always calculate result based on data column, not index column. (index column is always sorted since order by)

So i think maybe it is better to fix this logic to follow behavior of pandas one for index.

itholic:
Resolve #931 (The contents below is same as the contents of this issue)

When we use `is_monotonic_decreasing` or `is_monotonic_increasing` for index,

pandas works like below:

```python
>>> s = pd.Series([7, 6, 5, 4, 3, 2, 1], index=[7, 6, 5, 4, 3, 2, 1])
>>> s.index.is_monotonic_decreasing
True
```
Since the index order is literally motononic decreasing, they return `True`

But our case,

```python
>>> s = ks.Series([7, 6, 5, 4, 3, 2, 1], index=[7, 6, 5, 4, 3, 2, 1])
>>> s.index.is_monotonic_decreasing
False
```

as seen above, it returns `False`.

because our existing logic always order by index column before calculate the each rows increasing or decreasing.

So they always calculate result based on data column, not index column. (index column is always sorted since order by)

So i think maybe it is better to fix this logic to follow behavior of pandas one for index.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -311,16 +311,15 @@ class IndexOpsMixin(object):
         """
         from databricks.koalas.frame import DataFrame
         from databricks.koalas.series import _col
-        if self.hasnans:
-            return False
         col = self._scol
         idx_tmp = SPARK_INDEX_NAME_FORMAT('temp')
         sdf = self._internal.sdf.withColumn(idx_tmp, monotonically_increasing_id())
         window = Window.orderBy(sdf[idx_tmp]).rowsBetween(-1, -1)
-        internal = _InternalFrame(
-            sdf=sdf.select(sdf[idx_tmp], col >= F.lag(col, 1).over(window)),
-            index_map=[(idx_tmp, None)])
-        return _col(DataFrame(internal)).all()
+        sdf = sdf.select(
+                  sdf[idx_tmp],
+                  F.when((col >= F.lag(col, 1).over(window)) & col.isNotNull(), True)
+                   .otherwise(False).alias('__temp__'))
+        return sdf.where(F.col('__temp__') == 'false').count() < 2
 
     is_monotonic_increasing = is_monotonic
 
@@ -371,16 +370,15 @@ class IndexOpsMixin(object):
         """
         from databricks.koalas.frame import DataFrame
         from databricks.koalas.series import _col
-        if self.hasnans:
-            return False
         col = self._scol
         idx_tmp = SPARK_INDEX_NAME_FORMAT('temp')
         sdf = self._internal.sdf.withColumn(idx_tmp, monotonically_increasing_id())
         window = Window.orderBy(sdf[idx_tmp]).rowsBetween(-1, -1)
-        internal = _InternalFrame(
-            sdf=sdf.select(sdf[idx_tmp], col <= F.lag(col, 1).over(window)),
-            index_map=[(idx_tmp, None)])
-        return _col(DataFrame(internal)).all()
+        sdf = sdf.select(
+                  sdf[idx_tmp],
+                  F.when((col <= F.lag(col, 1).over(window)) & col.isNotNull(), True)
+                   .otherwise(False).alias('__temp__'))
+        return sdf.where(F.col('__temp__') == 'false').count() < 2
 
     def astype(self, dtype):
         """
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -32,6 +32,7 @@ from databricks import koalas as ks  # For running doctests and reference resolu
 from databricks.koalas.internal import _InternalFrame
 from databricks.koalas.typedef import pandas_wraps, spark_type_to_pandas_dtype
 from databricks.koalas.utils import align_diff_series, scol_for
+from databricks.koalas.internal import SPARK_INDEX_NAME_FORMAT
 from pyspark.sql.functions import monotonically_increasing_id
 
 
@@ -313,9 +314,12 @@ class IndexOpsMixin(object):
         if self.hasnans:
             return False
         col = self._scol
-        sdf = self._internal.sdf.withColumn('id', monotonically_increasing_id())
-        window = Window.orderBy(sdf['id']).rowsBetween(-1, -1)
-        internal = _InternalFrame(sdf.select(col >= F.lag(col, 1).over(window)))
+        idx_tmp = SPARK_INDEX_NAME_FORMAT('temp')
+        sdf = self._internal.sdf.withColumn(idx_tmp, monotonically_increasing_id())
+        window = Window.orderBy(sdf[idx_tmp]).rowsBetween(-1, -1)
+        internal = _InternalFrame(
+            sdf=sdf.select(sdf[idx_tmp], col >= F.lag(col, 1).over(window)),
+            index_map=[(idx_tmp, None)])
         return _col(DataFrame(internal)).all()
 
     is_monotonic_increasing = is_monotonic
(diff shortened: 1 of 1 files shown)


ueshin:
@itholic Good catch!
btw, the proposed fix seems still something missing:

```py
>>> pser
0.097069    0.097069
0.674804    0.674804
0.417065    0.417065
dtype: float64
>>> pser.is_monotonic
False

>>> ks.from_pandas(pser).is_monotonic
True
>>> ks.from_pandas(pser).index.is_monotonic
False
```

maybe we need to apply the fix for Index to Series' case as well.

itholic:
@ueshin , ah i missed it. just fixed it, Thanks for commenting :)

itholic:
oops, I've never seen such that build failure. 😢 
(maybe internal dependency issue..?)

ueshin:
Let me retrigger the build and let's see.

itholic:
@ueshin oops my comment removed by mistake; yeah let's see and maybe hyukjin is something doing for speed up build process. 😃

databricks/koalas/base.py
```diff
@@ -299,10 +300,23 @@ def is_monotonic(self):
...
+        True
         """
+        from databricks.koalas.frame import DataFrame
+        from databricks.koalas.series import _col
+        if self.hasnans:
+            return False
         col = self._scol
-        window = Window.orderBy(self._kdf._internal.index_scols).rowsBetween(-1, -1)
-        return self._with_new_scol((col >= F.lag(col, 1).over(window)) & col.isNotNull()).all()
+        sdf = self._internal.sdf.withColumn('id', monotonically_increasing_id())
```
ueshin: Shall we use more complex name like `__id__` or something? `id` will conflict with user columns.

databricks/koalas/base.py
```diff
@@ -299,10 +300,23 @@ def is_monotonic(self):
...
+        window = Window.orderBy(sdf['id']).rowsBetween(-1, -1)
+        internal = _InternalFrame(sdf.select(col >= F.lag(col, 1).over(window)))
```
ueshin: This will invoke attaching default index which might cause performance penalty.
Shall we include the dummy id column and use it as index for the new DataFrame?

ueshin: Can we remove ` & col.isNotNull()`, btw?

itholic: Thanks for reviewing!

i think it's not necessary because the handling of null value is already processed above as

```python
if self.hasnans:
    return False
```

databricks/koalas/base.py
itholic: thanks for commenting! resolved! 😄 
(fix to use `__index_level_temp__`)

databricks/koalas/base.py
ueshin: I see. hm, it runs a Spark job once, maybe we should avoid it?

itholic: @ueshin yeah, i also think it should be better.

maybe we can avoid this by using spark dataframe only. i fixed it just before

databricks/koalas/base.py
```diff
@@ -299,10 +301,25 @@ def is_monotonic(self):
...
+        from databricks.koalas.frame import DataFrame
+        from databricks.koalas.series import _col
         col = self._scol
-        window = Window.orderBy(self._kdf._internal.index_scols).rowsBetween(-1, -1)
-        return self._with_new_scol((col >= F.lag(col, 1).over(window)) & col.isNotNull()).all()
+        idx_tmp = SPARK_INDEX_NAME_FORMAT('temp')
+        sdf = self._internal.sdf.withColumn(idx_tmp, monotonically_increasing_id())
+        window = Window.orderBy(sdf[idx_tmp]).rowsBetween(-1, -1)
+        sdf = sdf.select(
+            sdf[idx_tmp],
```
HyukjinKwon: This column seems not needed.

databricks/koalas/base.py
```diff
@@ -299,10 +301,25 @@ def is_monotonic(self):
...
+            F.when((col >= F.lag(col, 1).over(window)) & col.isNotNull(), True)
+             .otherwise(False).alias('__temp__'))
+        return sdf.where(F.col('__temp__') == 'false').count() < 2
```
HyukjinKwon: ?? why does this compare to a string `false`?

databricks/koalas/base.py
```diff
@@ -299,10 +301,25 @@ def is_monotonic(self):
...
+
+        >>> ser = ks.Series([5, 4, 3, 2, 1], index=[1, 2, 3, 4, 5])
+        >>> ser.is_monotonic
+        False
+
+        >>> ser.index.is_monotonic
+        True
         """
+        from databricks.koalas.frame import DataFrame
+        from databricks.koalas.series import _col
```
HyukjinKwon: This imports seem not needed.

databricks/koalas/base.py
```diff
@@ -299,10 +301,25 @@ def is_monotonic(self):
...
         col = self._scol
-        window = Window.orderBy(self._kdf._internal.index_scols).rowsBetween(-1, -1)
```
HyukjinKwon: I think the only fix needed is replace `self._kdf._internal.index_scols` to `monotonically_increasing_id()`. Other changes look not needed.

itholic: @HyukjinKwon Thanks for reviewing..! ugh... it was such a simple problem if i took more carefully 😿

databricks/koalas/base.py
itholic: you're right, i fixed whole logic. Thanks!!

databricks/koalas/base.py
itholic: when i `show()` of the spark data frame, it looks string of 'false' in there, but  actually it is boolean internally, my mistake.

databricks/koalas/base.py
```diff
@@ -32,6 +32,8 @@
 from databricks.koalas.internal import _InternalFrame
 from databricks.koalas.typedef import pandas_wraps, spark_type_to_pandas_dtype
 from databricks.koalas.utils import align_diff_series, scol_for
+from databricks.koalas.internal import SPARK_INDEX_NAME_FORMAT
```
HyukjinKwon: This import seems not used.

databricks/koalas/base.py
```diff
@@ -32,6 +32,8 @@
...
+from pyspark.sql.functions import monotonically_increasing_id
```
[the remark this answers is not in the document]
itholic: ah! i see. thanks for commenting!

databricks/koalas/base.py
itholic: i just removed. thanks for commenting! :D

PR #932, opened 2019-10-15
Use one channel in Conda to speed up the build

HyukjinKwon:


HyukjinKwon:
the failure seems not related with this change.

HyukjinKwon:
readthedocs build seems passing at https://readthedocs.org/projects/koalas/builds/9805966/

HyukjinKwon:
I'm merging this since the related codes are executed fine and the Travis is already being failed with the same error already. this change shouldn't affect Travis.

PR #934, opened 2019-10-17
Fix Series.__getitem__ to work properly

### LINKED ISSUE #933: getitem for Series doesn't work properly
When we get item from Series, pandas works like below:

```python
>>> pser = pd.Series([10, 20, 15, 30, 45], ['A', 'A', 'B', 'C', 'D'])
>>> pser['A']
A    10
A    20
dtype: int64
>>>
>>> pser['B']
15
```

but in koalas, it raises `pyspark.sql.utils.AnalysisException: "Can't extract value from 0#6L: need struct type but got bigint;"`

so i think we should fix it to work properly like same as pandas.

itholic:
Resolves #933 

```python
>>> kser = ks.Series([10, 20, 15, 30, 45], ['A', 'A', 'B', 'C', 'D'])
>>> kser['A']
A    10
A    20
Name: 0, dtype: int64
>>>
>>> kser['B']
15
```

For MultiIndex

```python
...                       [0, 1, 2, 0, 1, 2, 0, 1, 2]])
>>> kser = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3], index=midx)
>>> pser = kser.to_pandas()

>>> kser['lama']
speed      45.0
weight    200.0
length      1.2
Name: 0, dtype: float64

>>> pser['lama']
speed      45.0
weight    200.0
length      1.2
Name: 0, dtype: float64
```


itholic:
while i'm here, i'm going to fix to cover below case.

```python
...                       [0, 1, 2, 0, 1, 2, 0, 1, 2]])
>>> kser = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3], index=midx)
>>> pser = kser.to_pandas()

>>> pser[('lama', 'speed')]
45.0

>>> kser[('lama', 'speed')]
...
```

ueshin:
Seems like the examples for MultiIndex are different between Koalas and pandas?

itholic:
@ueshin oops, i think i was just my mistake

they actually work same.

```python
...                       [0, 1, 2, 0, 1, 2, 0, 1, 2]])
>>> kser = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3], index=midx)
>>> pser = kser.to_pandas()

>>> kser['lama']
speed      45.0
weight    200.0
length      1.2
Name: 0, dtype: float64

>>> pser['lama']
speed      45.0
weight    200.0
length      1.2
Name: 0, dtype: float64
```

databricks/koalas/series.py
```diff
@@ -3529,7 +3529,42 @@ def __len__(self):
         return len(self.to_dataframe())
 
     def __getitem__(self, key):
-        return self._with_new_scol(self._scol.__getitem__(key))
+        if not isinstance(key, (str, tuple)):
```
HyukjinKwon: Why should we restrict the type to only strings? From rough reading, it looks going to work. e.g.:

```python
>>> pd.Series([10, 20, 15, 30, 45], [1, 1, 3, 4, 5])[1]
1    10
1    20
```

databricks/koalas/series.py
```diff
@@ -3529,7 +3529,42 @@ def __len__(self):
...
+            .where(reduce(lambda x, y: x & y, rows))
+
+        if len(self._internal._index_map) == len(key):
+            # if sdf has one column and one data, return data only without frame
+            pdf = sdf.limit(2).toPandas()
+            length = len(pdf)
+            if length == 1:
+                return pdf[self.name].iloc[0]
+
+            key_string = ('(' + ', '.join(key) + ')') if len(key) > 1 else key[0]
```
HyukjinKwon: @itholic, seems you can just use direct string implementation of `key`.

itholic: @HyukjinKwon Thanks for reviewing! you mean like this?

```python
>>> key = ('a', 'b')
>>> str(key)
"('a', 'b')"
```

if right, the result maybe like

```python
('cow', 'weight')    200.0
('cow', 'weight')      1.2
('cow', 'weight')    250.0
Name: 0, dtype: float64
```

not 

```python
(cow, weight)    200.0
(cow, weight)      1.2
(cow, weight)    250.0
Name: 0, dtype: float64
```

is it okay?

databricks/koalas/series.py
itholic: ah. it should not restrict. i just fixed them. Thanks!! 😄

databricks/koalas/series.py
HyukjinKwon: ah, okay. let's just leave it as is. It matches with pandas' right?

itholic: @HyukjinKwon right :)

itholic: @ueshin sure, then first i'll make util function for make string like this in the separate PR, and after then, fix the existing strings including this one.

databricks/koalas/series.py
```diff
@@ -3560,7 +3560,37 @@ def __len__(self):
...
```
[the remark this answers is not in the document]
itholic: ah it looks way better. thanks!! 👍

databricks/koalas/utils.py
```diff
@@ -358,3 +358,16 @@ def column_index_level(column_index: List[Tuple[str, ...]]) -> int:
         levels = set(0 if idx is None else len(idx) for idx in column_index)
         assert len(levels) == 1, levels
         return list(levels)[0]
+
+
+def tuple_like_strings(items):
```
HyukjinKwon: I don't like this naming lol but I think we can change in the separate PR.

databricks/koalas/series.py
itholic: i just made util function `tuple-like strings` and applied.

databricks/koalas/utils.py
itholic: yeah it's hard to name. any recommended??

HyukjinKwon: let's do it later in a separate PR when we handle https://github.com/databricks/koalas/pull/934#discussion_r336932433 comment.

itholic: okay i will. Thanks for commenting!! 😄

PR #936, opened 2019-10-17
Implement Series.first_valid_index

itholic:
Like pandas Series.first_valid_index (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.first_valid_index.html#pandas.Series.first_valid_index) 

implemented function `first_valid_index` for Series.

```python
>>> s = ks.Series([None, None, 3, 4, 5], index=[100, 200, 300, 400, 500])
>>> s
100    NaN
200    NaN
300    3.0
400    4.0
500    5.0
Name: 0, dtype: float64

>>> s.first_valid_index()
300
>>> s.to_pandas().first_valid_index()
300
```

And act like below when Series has MultiIndex

```python
...                       [0, 1, 2, 0, 1, 2, 0, 1, 2]])
>>> s = ks.Series([None, None, None, None, 250, 1.5, 320, 1, 0.3], index=midx)
>>> s
lama    speed       NaN
        weight      NaN
        length      NaN
cow     speed       NaN
        weight    250.0
        length      1.5
falcon  speed     320.0
        weight      1.0
        length      0.3
Name: 0, dtype: float64

>>> s.first_valid_index()
('cow', 'weight')
>>> s.to_pandas().first_valid_index()
('cow', 'weight')
```


databricks/koalas/series.py
```diff
@@ -3223,6 +3223,65 @@ def copy(self) -> 'Series':
...
+        Name: 0, dtype: float64
+
+        >>> s.first_valid_index()
+        ('cow', 'weight')
+        """
+        sdf = self._internal.sdf
+        data_scol = self._internal.data_scols[0]
+
+        first_valid_value = sdf.select(
+            F.first(data_scol, True).alias(str(self.name))).first()[str(self.name)]
```
ueshin: This can't guarantee the first value since `F.first()` is non-deterministic.

itholic: @ueshin Thanks for reviewing, but could you explain a little bit more detail?

below is the point why i hard to understand `F.first()` is non-deterministic

if `koalas.Series` looks like below.

```python
>>> s = ks.Series([None, 1, 2, 3, 4], index=[4, 5, 2, 1, 8])
>>> s
4    NaN
5    1.0
2    2.0
1    3.0
8    4.0
Name: 0, dtype: float64
```

then, the internal spark dataframe looks like below.

```
>>> s._internal.sdf.show()
+-----------------+----+
|__index_level_0__|   0|
+-----------------+----+
|                4|null|
|                5| 1.0|
|                2| 2.0|
|                1| 3.0|
|                8| 4.0|
+-----------------+----+
```

for this `sdf`, isn't the `F.first()` returns always same result? (in other words, isn't the `F.first()` is deterministic for same `sdf`?)

```
>>> sdf.select(F.first('0', True)).show()
+--------------+
|first(0, true)|
+--------------+
|           1.0|
+--------------+
```

HyukjinKwon: I think he meant the dataframe order itself is not guaranteed and therefore the first value can be non-deterministic. If this was correct, not sure what's the right solution for this case. Indexes can have its own natural order and it's lost.

Maybe we should explicitly note that Koalas DataFrame currently does not keep its natural order and it might cause unexpected results (and maybe ti should explicitly sort).

databricks/koalas/series.py
```diff
@@ -3223,6 +3223,65 @@ def copy(self) -> 'Series':
...
+        falcon  speed     320.0
+                weight      1.0
+                length      0.3
+        Name: 0, dtype: float64
+
+        >>> s.first_valid_index()
+        ('cow', 'weight')
+        """
+        sdf = self._internal.sdf
+        data_scol = self._internal.data_scols[0]
```
HyukjinKwon: maybe `_internal.scol`

databricks/koalas/series.py
HyukjinKwon: The `first` function doc says:

```
    .. note:: The function is non-deterministic because its results depends on order of rows which
        may be non-deterministic after a shuffle.
```

databricks/koalas/series.py
```diff
@@ -3223,6 +3223,65 @@ def copy(self) -> 'Series':
...
+
+        first_valid_value = sdf.select(
+            F.first(data_scol, True).alias(str(self.name))).first()[str(self.name)]
+        sdf = sdf.select(self._internal.index_scols) \
+                 .where(data_scol == first_valid_value) \
```
HyukjinKwon: I think you can simply filter via `isNotNull`

databricks/koalas/series.py
itholic: oh i got it. i just fixed to not use `F.first()`

databricks/koalas/series.py
itholic: thanks!! fixed it!

PR #937, opened 2019-10-18
Impelment DataFrame.keys

itholic:
An alias for columns.

```python
...                   index=['cobra', 'viper', 'sidewinder'],
>>> df
            max_speed  shield
cobra               1       2
viper               4       5
sidewinder          7       8

>>> df.keys()
Index(['max_speed', 'shield'], dtype='object')

>>> df.to_pandas().keys()
Index(['max_speed', 'shield'], dtype='object')
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7413,6 +7413,20 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         -------
         Index
             Columns of the DataFrame.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame([[1, 2], [4, 5], [7, 8]],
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -7405,6 +7405,17 @@ def gen_new_column_index_entry(column_index_entry):
...
 
+    def keys(self):
+        """
+        Return alias for columns.
+
+        Returns
+        -------
+        Index
+            Columns of the DataFrame.
+        """
```
ueshin: Shall we have an example? The example in the PR description should be fine.

itholic:
@ueshin @HyukjinKwon Thanks for commenting!! added it :)

PR #938, opened 2019-10-18
Implement Index.dropna & MultiIndex.dropna

itholic:
There is functionality `pandas.Index.dropna` (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.dropna.html#pandas.Index.dropna)

support it also for Koalas same as pandas.

```python
...                   index=['cobra', 'viper', None],
>>> df
       max_speed  shield
cobra          1       2
viper          4       5
NaN            7       8

>>> df.index.dropna()
Index(['cobra', 'viper'], dtype='object')

>>> df.to_pandas().index.dropna()
Index(['cobra', 'viper'], dtype='object')
```


itholic:
While i'm here, i'm going to make this also support to MultiIndex also.

itholic:
While i'm here, i'm going to make this possible support to MultiIndex also.

itholic:
While i'm here, i'm going to make this possible to support MultiIndex also.

databricks/koalas/indexes.py
```diff
@@ -422,6 +446,17 @@ def to_pandas(self) -> pd.MultiIndex:
 
     toPandas = to_pandas
 
+    def dropna(self):
+        """
+        Return MultiIndex without NA/NaN values
+        """
```
HyukjinKwon: Can you add an example?

databricks/koalas/indexes.py
```diff
@@ -422,6 +446,17 @@ def to_pandas(self) -> pd.MultiIndex:
...
+        kdf = self._kdf
+        sdf = self._kdf._internal.sdf
+        for index_scol in self._kdf._internal.index_scols:
+            sdf = sdf.where(index_scol.isNotNull())
```
HyukjinKwon: I think you can make the condition with `and` and `reduce` as you did in other places.

HyukjinKwon: In fact, can we share the same implementation?

databricks/koalas/indexes.py
itholic: sure, i'm gonna add it. thanks :)

databricks/koalas/indexes.py
itholic: Thanks for reviewing! i just changed. i'll commit after resolve conflicts :)

itholic: okay i'll commit them soon. Thanks for reviewing!!

databricks/koalas/indexes.py
```diff
@@ -317,6 +317,62 @@ def is_object(self):
...
+        MultiIndex([(   'cow', 'weight'),
+                    (   'cow', 'weight'),
+                    (   'cow', 'weight'),
+                    (   'cow', 'length'),
+                    ('falcon', 'weight'),
+                    ('falcon', 'weight'),
+                    ('falcon', 'length')],
+                   )
+        """
+        kdf = self._kdf.copy()
```
HyukjinKwon: @itholic can we reuse DataFrame.dropna impl?

itholic: Not pretty sure, i think they are slightly different when they filter the null value

In case DataFrame.dropna, they assume `column_index` as index,

```python
idxes = self._internal.column_index
reduce(lambda x, y: x + y,
       [F.when(self[idx].notna()._scol, 1).otherwise(0)
       for idx in idxes]
```

but in case Index.dropna, we assume `index_scols` as index.
```python
index_scol.isNotNull() for index_scol in kdf._internal.index_scols
```

HyukjinKwon: I see. can we use Spark's `DataFrame.dropna` API?

HyukjinKwon: ping @itholic can we use Spark's DataFrame.dropna API?

itholic: @HyukjinKwon yeah i think we can simply use that one. thanks !! 👍

databricks/koalas/indexes.py
```diff
@@ -317,6 +317,60 @@ def is_object(self):
...
```
HyukjinKwon: I think we can do as below:

```python
sdf = kdf._internal.sdf.select(index_scols).dropna()
internal = kdf._internal.copy(sdf=sdf, column_index=[], ...)
kdf = DataFrame(internal)
return Index(kdf) if type(self) == Index else MultiIndex(kdf)
```

itholic: thanks for the comment!
i fixed with using `_InternalFrame` rather than using `copy` since i think we can't copy internal because `sdf` has only index columns while `kdf` has also data columns

PR #939, opened 2019-10-18
Implement Index.copy & MultiIndex.copy

itholic:
like pandas (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.copy.html#pandas.Index.copy)

implement function `copy` for Index

```python
...                   index=['cobra', 'viper', 'sidewinder'],
>>> df
            max_speed  shield
cobra               1       2
viper               4       5
sidewinder          7       8
>>> df.index
Index(['cobra', 'viper', 'sidewinder'], dtype='object')

>>> df.index.copy()
Index(['cobra', 'viper', 'sidewinder'], dtype='object')

>>> df.index.copy(name='snake')
Index(['cobra', 'viper', 'sidewinder'], dtype='object', name='snake')
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -460,6 +460,62 @@ class MultiIndex(Index):
 
     toPandas = to_pandas
 
+    # TODO: add 'name' parameter after pd.MultiIndex.name is implemented
+    def copy(self):
+        """
+        Make a copy of this object. name sets those attributes on the new object.
+
+        Examples
+        --------
+        >>> midx = pd.MultiIndex([['a', 'b', 'c'],
+        ...                       ['lama', 'cow', 'falcon'],
+        ...                       ['speed', 'weight', 'length']],
+        ...                      [[0, 0, 0, 0, 0, 0, 1, 1, 1],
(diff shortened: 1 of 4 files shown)


itholic:
While i'm here, i'm going to make this also support to MultiIndex also.

itholic:
While i'm here, i'm going to make this possible support to MultiIndex also.

itholic:
While i'm here, i'm going to make this possible to support MultiIndex also.

databricks/koalas/indexes.py
```diff
@@ -422,6 +460,15 @@ def to_pandas(self) -> pd.MultiIndex:
 
     toPandas = to_pandas
 
+    # TODO: add 'name' parameter after pd.MultiIndex.name is implemented
+    def copy(self):
+        """
+        Make a copy of this object.
+        """
```
HyukjinKwon: Maybe we should add an example but I think it's fine for now since it's incomplete for now due to missing name anyway.

itholic: thanks for reviewing. i'm planning to implement `name`, after then i'll add the example here also.

PR #946, opened 2019-10-23
Add has_duplicates property for Index and MultiIndex

charlesdong1991:


databricks/koalas/indexes.py
```diff
@@ -125,6 +125,18 @@ def spark_type(self):
...
 
+    @property
+    def has_duplicates(self) -> bool:
+        """
+        If index has duplicates, return True, otherwise False.
+        """
+        idx_column = self._kdf._sdf.select(self._scol)
+        deduplicate_idx_column = idx_column.drop_duplicates()
+
+        if idx_column.count() == deduplicate_idx_column.count():
```
ueshin: This runs Spark jobs twice. Can we use `F.count` and `F.countDistinct` instead?

charlesdong1991: ahh! Nice! thanks!!!

databricks/koalas/indexes.py
```diff
@@ -125,6 +125,34 @@ def spark_type(self):
...
+
+        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=[list('aac'), list('eef')])
+        >>> kdf.index.has_duplicates
+        True
+        """
+        scol = self._kdf._sdf.select(self._scol)
+        col = scol.columns[0]
+
+        count = scol.agg(F.count(col).alias('count')).collect()
+        dedup_count = scol.agg(F.countDistinct(col).alias('count')).collect()
```
HyukjinKwon: @charlesdong1991 can we do it in one pass?

```
df.select(F.count("id") == F.countDistinct("id")).first()
```

charlesdong1991: Nice, this is much simpler! thanks! @HyukjinKwon

PR #949, opened 2019-10-24
Implement value_counts() for Index & MultiIndex

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.value_counts.html#pandas.Index.value_counts

```python
>>> s = ks.Series([0, 1, 2, 3, 4, 5], index=[3, 1, 2, 3, 4, np.nan])
>>> s.index
Float64Index([3.0, 1.0, 2.0, 3.0, 4.0, nan], dtype='float64')

>>> s.index.value_counts().sort_index()
1.0    1
2.0    1
3.0    2
4.0    1
Name: count, dtype: int64

>>> s.index.value_counts(sort=True)  # doctest: +SKIP
3.0    2
4.0    1
1.0    1
2.0    1
Name: count, dtype: int64

>>> s.index.value_counts(normalize=True).sort_index()
1.0    0.2
2.0    0.2
3.0    0.4
4.0    0.2
Name: count, dtype: float64

>>> s.index.value_counts(dropna=False).sort_index()
1.0    1
2.0    1
3.0    2
4.0    1
NaN    1
Name: count, dtype: int64
```

For MultiIndex

```python
...                       [1, 1, 1, 1, 1, 2, 1, 2, 2]])
>>> s = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3], index=midx)
>>> s.index
MultiIndex([(  'lama', 'weight'),
            (  'lama', 'weight'),
            (  'lama', 'weight'),
            (   'cow', 'weight'),
            (   'cow', 'weight'),
            (   'cow', 'length'),
            ('falcon', 'weight'),
            ('falcon', 'length'),
            ('falcon', 'length')],
           )

>>> s.index.value_counts().sort_index()
(cow, length)       1
(cow, weight)       2
(falcon, length)    2
(falcon, weight)    1
(lama, weight)      3
Name: count, dtype: int64

>>> s.index.value_counts(normalize=True).sort_index()
(cow, length)       0.111111
(cow, weight)       0.222222
(falcon, length)    0.222222
(falcon, weight)    0.111111
(lama, weight)      0.333333
Name: count, dtype: float64
```


databricks/koalas/indexes.py
```diff
@@ -399,6 +400,140 @@ def copy(self, name=None):
...
+        The resulting object will be in descending order so that the
+        first element is the most frequently-occurring element.
+        Excludes NA values by default.
+
+        Parameters
+        ----------
+        normalize : boolean, default False
+            If True then the object returned will contain the relative
+            frequencies of the unique values.
+        sort : boolean, default False
```
ueshin: `default True`?

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,140 @@ def copy(self, name=None):
...
+        if dropna:
+            sdf_count = sdf_count.dropna()
+        if sort:
+            sdf_count = sdf_count.sort('count', ascending=ascending)
+        if normalize:
+            sum_values = sdf_count.select(
+                F.sum(sdf_count['count'])).collect()[0][0]
+            sdf_count = sdf_count.select(
+                idx_name,
+                F.round(sdf_count['count'] / sum_values, 6).alias('count'))
```
ueshin: Do we need to round here?

itholic: @ueshin Thanks for the comment!

Ah, i thought if we don't round here the result shows more than 6 decimal point which pandas' behavior.

but i just checked it works well when round is removed.

maybe is koalas already limit the decimal point to 6 somewhere before display the result?

databricks/koalas/indexes.py
itholic: @ueshin Thanks for the comment!

I set the default value of `sort` as `False` intentionally for avoid performance decreasing.

but maybe is it the better way to set this `True` in terms of mimic pandas'?

itholic:
@HyukjinKwon yes i think we can, but need some fix to apply `MultiIndex` also

HyukjinKwon:
Sure sounds good.

itholic:
@HyukjinKwon 

oops, even pandas not support `isnull()` for `MultiIndex`.

```python
>>> self.to_pandas().isnull()
...
NotImplementedError: isna is not defined for MultiIndex
```

i think it is better separate the logic for now. what are you think?

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,139 @@ def copy(self, name=None):
...
+        sdf = self._kdf._sdf
+        sdf_count = sdf.groupBy(self._scol.alias(idx_name)).count()
+
+        if dropna:
+            sdf_count = sdf_count.dropna()
+        if sort:
+            sdf_count = sdf_count.sort('count', ascending=ascending)
+        if normalize:
+            sum_values = sdf_count.select(
+                F.sum(sdf_count['count'])).collect()[0][0]
```
HyukjinKwon: You can use `first()[0]` instead of `collect()[0][0]`

HyukjinKwon: Can we do this in a single pass without executing the job via `collect`?

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,139 @@ def copy(self, name=None):
...
+        2.0    1
+        3.0    2
+        4.0    1
+        Name: count, dtype: int64
+
+        **sort**
+
+        With `sort` set to `False`, the result wouldn't be sorted by number of count.
+
+        >>> s.index.value_counts(sort=True)  # doctest: +SKIP
```
HyukjinKwon: Why did we skip?

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,139 @@ def copy(self, name=None):
...
+        2.0    0.2
+        3.0    0.4
+        4.0    0.2
+        Name: count, dtype: float64
+
+        **dropna**
+
+        With `dropna` set to `False` we can also see NaN index values.
+
+        >>> s.index.value_counts(dropna=False).sort_index()  # doctest: +SKIP
```
HyukjinKwon: ditto. why did we skip?

databricks/koalas/indexes.py
itholic: ah, that looks way better for performance.
thanks for the comment!

databricks/koalas/indexes.py
itholic: because the `__repr__` of MultiIndex seems different depends on python version.

databricks/koalas/indexes.py
itholic: also ditto.

databricks/koalas/indexes.py
itholic: @HyukjinKwon Thanks for the comment! fixed it.

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,138 @@ def copy(self, name=None):
...
+        2.0    0.2
+        3.0    0.4
+        4.0    0.2
+        Name: count, dtype: float64
+
+        **dropna**
+
+        With `dropna` set to `False` we can also see NaN index values.
+
+        >>> s.index.value_counts(dropna=False).sort_index()
```
HyukjinKwon: Seems this test is failed:

```
____________ [doctest] databricks.koalas.indexes.Index.value_counts ____________
530         2.0    0.2
531         3.0    0.4
532         4.0    0.2
533         Name: count, dtype: float64
534 
535         **dropna**
536 
537         With `dropna` set to `False` we can also see NaN index values.
538 
539         >>> s.index.value_counts(dropna=False).sort_index()
    @@ -1,6 +1,6 @@
    -1.0    1
    -2.0    1
    -3.0    2
    -4.0    1
    -NaN    1
    + 1.0    1
    + 2.0    1
    + 3.0    2
    + 4.0    1
    +NaN     1
     Name: count, dtype: int64
```

You can skip this test for now I think.

itholic: ah, now i remember why i skipped those tests.
Thanks for the comment! 😄

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,138 @@ def copy(self, name=None):
...
+
+        if dropna:
+            sdf_count = sdf_count.dropna()
+        if sort:
+            sdf_count = sdf_count.sort('count', ascending=ascending)
+        if normalize:
+            sdf_count = sdf_count.select(
+                idx_name,
+                (sdf_count['count'] / sdf_count.select(F.sum(sdf_count['count']))
+                                               .first()[0]).alias('count'))
```
HyukjinKwon: @itholic can't we do this in a single pass without execution job here? e.g. `sdf_count['count'] / F.sum(sdf_count['count'])`

HyukjinKwon: ping @itholic

itholic: thanks for the comment! 

i tried, but they raises an exception like below:

```python
>>> sdf_count.select(sdf_count['count'] / F.sum(sdf_count['count']))
...
pyspark.sql.utils.AnalysisException: "grouping expressions sequence is empty, and '`count`' is not an aggregate function. Wrap '((CAST(`count` AS DOUBLE) / CAST(sum(`count`) AS DOUBLE)) AS `(count / sum(count))`)' in windowing function(s) or wrap '`count`' in first() (or first_value) if you don't care which value you get.;;\nAggregate [(cast(count#54L as double) / cast(sum(count#54L) as double)) AS (count / sum(count))#304]\n+- Aggregate [named_struct(__index_level_0__, __index_level_0__#0, __index_level_1__, __index_level_1__#1)], [named_struct(__index_level_0__, __index_level_0__#0, __index_level_1__, __index_level_1__#1) AS __index_level_0__#49, count(1) AS count#54L]\n   +- LogicalRDD [__index_level_0__#0, __index_level_1__#1, 0#2], false\n"
```

i think maybe if we want to code like that way, we should use crossJoin before like below:

```
sdf_count = sdf_count.crossJoin(
    sdf_count.select(
        F.sum(sdf_count['count']).alias('count_sum')))

sdf_count.select(F.col('count') / F.col('count_sum')).show()
+-------------------+
|(count / count_sum)|
+-------------------+
| 0.2222222222222222|
| 0.1111111111111111|
| 0.3333333333333333|
| 0.1111111111111111|
| 0.2222222222222222|
+-------------------+
```

how about this? i'm not pretty sure what solution is more efficient.

HyukjinKwon: Ah, actually there's `Series.value_counts` implementation. Can we move `Series.value_counts` into `IndexOpsMixin` and reuse?

itholic: @HyukjinKwon oh i see. i fixed them to reuse Series one 👍

databricks/koalas/base.py
```diff
@@ -760,3 +761,174 @@ def _shift(self, periods, fill_value, part_cols=()):
...
+            sdf = sdf.withColumn('count', F.col('count') / F.lit(sum))
+
+        # column_index & column_index_name are need for Series, but not for Index/MtutiIndex
+        internal = _InternalFrame(sdf=sdf,
+                                  index_map=[(index_name, None)],
+                                  column_index=self._internal.column_index,
+                                  column_scols=[scol_for(sdf, 'count')],
+                                  column_index_names=self._internal.column_index_names) \
+            if isinstance(self, Series) else \
+            _InternalFrame(sdf=sdf,
```
HyukjinKwon: Can avoid to make it all inlined? Seems difficult to read

databricks/koalas/indexes.py
```diff
@@ -36,8 +36,8 @@
 from databricks.koalas.frame import DataFrame
 from databricks.koalas.internal import _InternalFrame
 from databricks.koalas.missing.indexes import _MissingPandasLikeIndex, _MissingPandasLikeMultiIndex
-from databricks.koalas.series import Series
-from databricks.koalas.internal import _InternalFrame
+from databricks.koalas.internal import _InternalFrame, SPARK_INDEX_NAME_FORMAT
+from databricks.koalas.series import Series, _col
```
HyukjinKwon: Let's revert unneeded changes.

databricks/koalas/base.py
```diff
@@ -760,3 +761,176 @@ def _shift(self, periods, fill_value, part_cols=()):
...
+        if isinstance(self, Series):
+            internal = _InternalFrame(sdf=sdf,
+                                      index_map=[(index_name, None)],
+                                      column_index=self._internal.column_index,
+                                      column_scols=[scol_for(sdf, 'count')],
+                                      column_index_names=self._internal.column_index_names)
+        else:
+            internal = _InternalFrame(sdf=sdf,
+                                      index_map=[(index_name, None)],
+                                      column_scols=[scol_for(sdf, 'count')])
```
ueshin: This should still hande `column_index`. e.g.,

```py
>>> s = pd.Series([0, 1, 2, 3, 4, 5], index=[3, 1, 2, 3, 4, np.nan])
>>> s.index.name = 'idx'
>>> s.index.value_counts()
3.0    2
4.0    1
2.0    1
1.0    1
Name: idx, dtype: int64
```

itholic: thanks for the comment :)

databricks/koalas/base.py
```diff
@@ -763,3 +763,187 @@ def _shift(self, periods, fill_value, part_cols=()):
...
+        2    1
+        3    1
+        Name: koalas, dtype: int64
+        """
+        from databricks.koalas.series import Series, _col
+        if bins is not None:
+            raise NotImplementedError("value_counts currently does not support bins")
+
+        if dropna:
+            sdf_dropna = self._internal._sdf.dropna()
```
ueshin: We should've specified a proper `subset` here, otherwise returns wrong values:

```py
>>> kdf
   a    b
0  1  NaN
1  2  1.0
2  3  NaN
>>> kdf.a
0    1
1    2
2    3
Name: a, dtype: int64
>>> kdf.a.value_counts()
2    1
Name: a, dtype: int64
```

but it should return

```py
>>> kdf.a.value_counts()
3    1
1    1
2    1
Name: a, dtype: int64
```

ueshin: @itholic Do you want to work on this fix? Thanks.

itholic: @ueshin ah, thanks for the check this. let me try to fix 😄

PR #952, opened 2019-10-25
Explicitly not support codes and levels property for MultiIndex

charlesdong1991:


databricks/koalas/indexes.py
```diff
@@ -471,6 +472,41 @@ def any(self, *args, **kwargs):
...
+        i = 0
+        for col in cols:
+            df = sdf.select(col).distinct()
+            w = Window.orderBy(col)
+            df = df.withColumn("__code_{}".format(str(i)), F.row_number().over(w) - 1)
+            sdf = sdf.join(df, sdf[col] == df[col]).drop(df[col])
+            i += 1
+
+        sdf = sdf.orderBy("__order__").select(sdf.colRegex("`__code_.`"))
+        return DataFrame(_InternalFrame(sdf=sdf)).astype('int').to_numpy().T.tolist()
```
HyukjinKwon: Hmmmmm... this collects everything into driver side which can be against [the design principles](https://koalas.readthedocs.io/en/latest/development/design.html#return-koalas-data-structure-for-big-data-and-pandas-data-structure-for-small-data) ... I think we might have to rather explicitly don't support, and let users call `to_pandas()` directly .. ? cc @ueshin

ueshin: ah, in that case, we might also rather explicitly not support `levels`.

charlesdong1991: emm, I see. Then if so, what do you suggest to do next? @ueshin @HyukjinKwon

charlesdong1991: In my work experience, I think most of the time, those method/property related to `index` are used on columns, and pandas should be enough.

HyukjinKwon: oh, sorry. @charlesdong1991 and @ueshin. I was confused. it's mapped to levels
Sure, we should add this then. I thought it's mapped to the actual data.

HyukjinKwon: It requires to collect all data into driver so we shouldn't implement. @charlesdong1991 do you use this property and `levels` often? If not, let's explicitly don't support. You can refer https://github.com/databricks/koalas/pull/744 to explicitly don't support.

charlesdong1991: sorry for my late reply, didn't have time these days @HyukjinKwon 

tbh, i barely use this `levels` as well as `codes`often, but this can be different across different people and work. I will also drop supports for `levels` as well

databricks/koalas/missing/indexes.py
```diff
@@ -225,6 +223,16 @@ class _MissingPandasLikeMultiIndex(object):
...
+    codes = unsupported_property(
+        'codes',
+        reason="'codes' requires to collect all data into the driver which is against the "
+               "design principle of Koalas. Alternatively, you could call 'to_pandas()' and"
+               " use 'codes' property in pandas.")
+    levels = unsupported_property(
+        'levels',
+        reason="'levels' requires to collect all data into the driver which is against the "
+               "design principle of Koalas. Alternatively, you could call 'to_pandas()' and"
+               " use 'levels' property in pandas.")
```
HyukjinKwon: @ueshin let me merge this one for now .. but we can always flip this decision if I missed anything.

HyukjinKwon:
Thanks @charlesdong1991

PR #953, opened 2019-10-25
Implement symmetric_difference of Index & MultiIndex

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.symmetric_difference.html#pandas.Index.symmetric_difference

```python
>>> s1 = ks.Series([1, 2, 3, 4], index=[1, 2, 3, 4])
>>> s2 = ks.Series([1, 2, 3, 4], index=[2, 3, 4, 5])

>>> s1.index.symmetric_difference(s2.index)
Int64Index([5, 1], dtype='int64')

>>> s1.index.symmetric_difference(s2.index, result_name='koalas')
Int64Index([5, 1], dtype='int64', name='koalas')

>>> s1.index.symmetric_difference(s2.index, sort=True)
Int64Index([1, 5], dtype='int64')

>>> s1.index ^ s2.index
Int64Index([5, 1], dtype='int64')
```

For MultiIndex

```python
>>> midx1 = pd.MultiIndex([['lama', 'cow', 'falcon'],
...                        ['speed', 'weight', 'length']],
...                       [[0, 0, 0, 1, 1, 1, 2, 2, 2],
...                        [0, 0, 0, 0, 1, 2, 0, 1, 2]])
>>> midx2 = pd.MultiIndex([['koalas', 'cow', 'falcon'],
...                        ['speed', 'weight', 'length']],
...                       [[0, 0, 0, 1, 1, 1, 2, 2, 2],
...                        [0, 0, 0, 0, 1, 2, 0, 1, 2]])
>>> s1 = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3],
...                index=midx1)
>>> s2 = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3],
...              index=midx2)

>>> s1.index.symmetric_difference(s2.index)
MultiIndex([('koalas', 'speed'),
            (  'lama', 'speed')],
           )

>>> s1.index.symmetric_difference(s2.index, result_name=['a', 'b'])
MultiIndex([('koalas', 'speed'),
            (  'lama', 'speed')],
           names=['a', 'b'])

>>> s1.index.symmetric_difference(s2.index, sort=True)
MultiIndex([('koalas', 'speed'),
            (  'lama', 'speed')],
           )

>>> s1.index ^ s2.index
MultiIndex([('koalas', 'speed'),
            (  'lama', 'speed')],
           )
```


databricks/koalas/indexes.py
```diff
@@ -399,6 +400,70 @@ def copy(self, name=None):
...
+        sdf_other = other._kdf._sdf.select(other._internal.index_scols)
+
+        sdf_symdiff = sdf_self.union(sdf_other) \
+                              .subtract(sdf_self.intersect(sdf_other))
+
+        if sort:
+            sdf_symdiff = sdf_symdiff.sort(self._internal.index_scols)
+
+        internal = self._kdf._internal.copy(sdf=sdf_symdiff)
+        result = Index(DataFrame(internal))
```
ueshin: This should be `DataFrame(internal).index`? E.g., what if the index is multi-index?

databricks/koalas/indexes.py
```diff
@@ -517,6 +585,88 @@ def copy(self):
         result = MultiIndex(ks.DataFrame(internal))
         return result
 
+    def symmetric_difference(self, other, result_name=None, sort=None):
```
ueshin: The logic is exactly the same as of `Index`. I don't think we need this with the fix mentioned above.

databricks/koalas/indexes.py
```diff
@@ -399,6 +400,70 @@ def copy(self, name=None):
...
+
+        >>> s1.index.symmetric_difference(s2.index, sort=True)
+        Int64Index([1, 5], dtype='int64')
+
+        You can also use the ``^`` operator:
+
+        >>> s1.index ^ s2.index
+        Int64Index([5, 1], dtype='int64')
+        """
+        sdf_self = self._kdf._sdf.select(self._internal.index_scols)
```
ueshin: We need to check the index levels and data types here? E.g., what if `Index ^ MultiIndex`?
Could you add such tests?

databricks/koalas/indexes.py
itholic: @ueshin , thanks for the review! that sounds reasonable. i'll fix them 👍

databricks/koalas/indexes.py
itholic: Sure, i'll add

BTW, i found some funny behavior in pandas, not pretty sure, maybe a bug in pandas.

There is an index and a MultiIndex of pandas here.

```python
>>> idx = pd.Index([1, 2, 3, 4])
...                        ['speed', 'weight', 'length']],
...                       [[0, 0, 0, 1, 1, 1, 2, 2, 2],
...                        [0, 0, 0, 0, 1, 2, 0, 1, 2]])
```

then, below seems to work pretty well,

```python
>>> idx ^ midx
Index([                   1,                    2,                    3,
                          4,    ('lama', 'speed'),     ('cow', 'speed'),
          ('cow', 'weight'),    ('cow', 'length'),  ('falcon', 'speed'),
       ('falcon', 'weight'), ('falcon', 'length')],
      dtype='object')
```

but below not. (although just switch the position `idx` and `midx`)

```python
>>> midx ^ idx
...
TypeError: Expected tuple, got int
>>>
```

itholic: @ueshin @HyukjinKwon 
I think there is some issue when we're handling `Index ^ MultiIndex`

i'm going to brief this issue with simple example, 

Here, we have an `Index` & a `MultiIndex` like below.

```python
>>> idx
Int64Index([1, 2, 3, 4], dtype='int64')
>>> midx
MultiIndex([('koalas',  'speed'),
            ('koalas',  'speed'),
            ('koalas',  'speed'),
            (   'cow',  'speed'),
            (   'cow', 'weight'),
            (   'cow', 'length'),
            ('falcon',  'speed'),
            ('falcon', 'weight'),
            ('falcon', 'length')],
           )
```

in pandas, they will show 'symdiff' result like below:

```python
>>> idx ^ midx
Index([                   1,                    2,                    3,
                          4,    ('lama', 'speed'),     ('cow', 'speed'),
          ('cow', 'weight'),    ('cow', 'length'),  ('falcon', 'speed'),
       ('falcon', 'weight'), ('falcon', 'length')],
      dtype='object')
```

the problem is here,

because i think now we can't support Index like this, since Internal spark dataframe can only support single type for single column, not multiple type

So i though how about we would support result of `Index ^ MultiIndex` as a string type.

Below example shows an internal spark dataframe for `symdiff` which type is string , 1 ~ 4 are from `Index` which type is bigint,  and the others from `MultiIndex` which type is struct<string, string>.

```
>>> sdf_symdiff.show()
+-----------------+
|__index_level_0__|
+-----------------+
|                1|
|                2|
|                3|
|                4|
|  [koalas, speed]|
|  [koalas, speed]|
|  [koalas, speed]|
|     [cow, speed]|
|    [cow, weight]|
|    [cow, length]|
|  [falcon, speed]|
| [falcon, weight]|
| [falcon, length]|
+-----------------+
```

then the final result will be shown as:

```python 
Index(['1', '2', '3', '4', '[koalas, speed]', '[koalas, speed]',
       '[koalas, speed]', '[cow, speed]', '[cow, weight]', '[cow, length]',
       '[falcon, speed]', '[falcon, weight]', '[falcon, length]'],
      dtype='object')
```

of course this result is slightly different from pandas' , but i think it still makes sense.

and we can make '[]' to '()', If we don't care about the cost of `replace`.

Have any suggestion to solve this problem with a better way?

ueshin: Maybe we should just not support such cases for now.

databricks/koalas/indexes.py
```diff
@@ -517,6 +585,88 @@ def copy(self):
...
+        ...              index=midx2)
+
+        >>> s1.index.symmetric_difference(s2.index)  # doctest: +SKIP
+        MultiIndex([('koalas', 'speed'),
+                    (  'lama', 'speed')],
+                   )
+
+        You can set names of result Index.
+
+        >>> s1.index.symmetric_difference(s2.index, result_name=['a', 'b'])  # doctest: +SKIP
```
HyukjinKwon: Why did we skip this @itholic ?

HyukjinKwon:
Seems fine.

databricks/koalas/indexes.py
itholic: @HyukjinKwon because the way of print out for MultiIndex are difference depends on their python version 😲

databricks/koalas/indexes.py
itholic: @ueshin it seems better for now 👍

ueshin:
Shall we explicitly block the cases `MultiIndex ^ Index` or `Index ^ MultiIndex` for now?

HyukjinKwon:
The tests are being failed after this commit anyway. I will revert this commit. @itholic do you mind opening a PR again to address test failures and Takuya's comment above?

itholic:
@HyukjinKwon sure, i'm going to address in separated PR.

PR #957, opened 2019-10-26
Add datetime as the equivalent python type to TimestampType

### LINKED ISSUE #951: `datetime.datetime` is not recognized as the equivalent python type to `types.TimestampType()`
Actually, this isn't an issue. I was looking for a workaround for [this comment](https://github.com/databricks/koalas/pull/478#issuecomment-545888595), wrote the code below and got `ValueError`.

```python
from datetime import datetime, timedelta

def tomorrow(date) -> datetime:
    return date + timedelta(days=1)

test_df = ks.DataFrame([(datetime(2019, 10, 24))], columns=['date'])
test_df['date'] = test_df['date'].map(tomorrow)

>>> test_df
...
ValueError: Invalid returnType: returnType can not be None
```

The cause is that `datetime.datetime` is not recognized as the equivalent python type to `types.TimestampType()`.

```python
_base = {
    types.StringType(): [str, 'str', 'string'],
    types.BinaryType(): [bytes],
    types.ByteType(): [np.int8, 'int8', 'byte'],
    types.ShortType(): [np.int16, 'int16', 'short'],
    types.IntegerType(): [int, 'int', np.int, np.int32],
    types.LongType(): [np.int64, 'int64', 'long', 'bigint'],
    types.FloatType(): [float, 'float', np.float],
    types.DoubleType(): [np.float64, 'float64', 'double'],

    types.TimestampType(): [np.datetime64],  # <- should be included here.

    types.DateType(): [datetime.date],
    types.BooleanType(): [bool, 'boolean', 'bool', np.bool],
    types.ArrayType(types.StringType()): []
}
```
https://github.com/databricks/koalas/blob/master/databricks/koalas/typedef.py#L112

I think `datetime.datetime` should be included in `types.TimestampType()`.

@ueshin @HyukjinKwon What do you think?

harupy:
Resolves #951

### DIFF
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -19,6 +19,7 @@ from collections import defaultdict
 from distutils.version import LooseVersion
 import inspect
 from io import BytesIO
+from datetime import datetime, timedelta
 
 import matplotlib
 matplotlib.use('agg')
@@ -470,6 +471,15 @@ class SeriesTest(ReusedSQLTestCase, SQLTestUtils):
         self.assertEqual(
             repr(kser.map(d)),
             repr(pser.map(d).rename(0)))
+
+        def tomorrow(date) -> datetime:
+            return date + timedelta(days=1)
+
+        pser = pd.Series([datetime(2019, 10, 24)])
+        kser = ks.from_pandas(pser)
+        self.assertEqual(
+            repr(kser.map(tomorrow)),
+            repr(pser.map(tomorrow).rename(0)))
 
     def test_add_prefix(self):
         pser = pd.Series([1, 2, 3, 4], name='0')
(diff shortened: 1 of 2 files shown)


harupy:
Do I need to write a test?

itholic:
@harupy i'm not sure but maybe it should be better in terms of usecase example? :smile_cat:

harupy:
@itholic Thanks for the comment. I'll wait for them.

harupy:
@HyukjinKwon Ok, I'll add a test.

harupy:
I'm thinking of adding the following code to `test_map` in `test_series.py`.

```python
def tomorrow(date) -> datetime:
    return date + timedelta(days=1)

test_df = ks.DataFrame([datetime(2019, 10, 24)], columns=['date'])
test_df['date'] = test_df['date'].map(tomorrow)
```

HyukjinKwon:
Sounds good!

harupy:
@HyukjinKwon test passed

HyukjinKwon:
Thanks, @harupy

PR #968, opened 2019-10-28
Fix passing options as keyword arguments

### LINKED ISSUE #967: The options argument is not passed to Spark reader properly.
The issue was reported at dev-list (['read_delta' with timestamp option](https://groups.google.com/forum/#!topic/koalas-dev/6FJAPLbnTN8)).

```py
df = ks.read_delta('/delta/OI_test', timestamp="2019-10-25")
```

retrieves only the most recent version of the table.

This is because the `options` argument is not passed to Spark reader properly.

ueshin:
Resolves #967.


rxin:
This is embarrassing :( We should add a test case.

ueshin:
Yup, let me add some.

PR #970, opened 2019-10-29
Implement MultiIndex.from_tuples

### LINKED ISSUE #958: MultiIndex.from_tuples
```Python
df = ks.DataFrame(
    {"a" : [4 ,5, 6],
     "b" : [7, 8, 9],
     "c" : [10, 11, 12]},
index = ks.MultiIndex.from_tuples(
    [('d',1),('d',2),('e',2)],
       names=['n','v']))
```

Due to the missing MultiIndex.from_tuples, it returns the following error:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-19-ba2ef765a723> in <module>
      3 "b" : [7, 8, 9],
      4 "c" : [10, 11, 12]},
----> 5 index = ks.MultiIndex.from_tuples(
      6 [('d',1),('d',2),('e',2)],
      7 names=['n','v']))

AttributeError: type object 'MultiIndex' has no attribute 'from_tuples'
```

itholic:
Resolves #958 

```python
>>> tuples = [(1, 'red'), (1, 'blue'),
...           (2, 'red'), (2, 'blue')]
>>> ks.MultiIndex.from_tuples(tuples, names=('number', 'color'))
MultiIndex([(1,  'red'),
            (1, 'blue'),
            (2,  'red'),
            (2, 'blue')],
           names=['number', 'color'])
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -25,8 +25,6 @@ import pandas as pd
 from pandas.api.types import is_list_like, is_interval_dtype, is_bool_dtype, \
     is_categorical_dtype, is_integer_dtype, is_float_dtype, is_numeric_dtype, is_object_dtype
 
-import numpy as np
-
 from pyspark import sql as spark
 from pyspark.sql import functions as F
 
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -479,21 +479,23 @@ class MultiIndex(Index):
     def from_tuples(cls, tuples, sortorder=None, names=None):
         """
         Convert list of tuples to MultiIndex.
+
         Parameters
         ----------
         tuples : list / sequence of tuple-likes
             Each tuple is the index of one row/column.
         sortorder : int or None
-            Level of sortedness (must be lexicographically sorted by that
-            level).
+            Level of sortedness (must be lexicographically sorted by that level).
         names : list / sequence of str, optional
             Names for the levels in the index.
+
         Returns
         -------
         index : MultiIndex
 
         Examples
         --------
+
         >>> tuples = [(1, 'red'), (1, 'blue'),
         ...           (2, 'red'), (2, 'blue')]
         >>> ks.MultiIndex.from_tuples(tuples, names=('number', 'color'))  # doctest: +SKIP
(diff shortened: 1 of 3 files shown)


databricks/koalas/indexes.py
```diff
@@ -472,6 +475,54 @@ def any(self, *args, **kwargs):
     def all(self, *args, **kwargs):
         raise TypeError("cannot perform all with this index type: MultiIndex")
 
+    @classmethod
```
ueshin: `@staticmethod` is enough?

databricks/koalas/indexes.py
```diff
@@ -472,6 +475,54 @@ def any(self, *args, **kwargs):
...
+                msg = "Cannot infer number of levels from empty list"
+                raise TypeError(msg)
+
+        index = [[idx[i] for idx in tuples] for i in range(len(tuples[0]))]
+        result = DataFrame(index=index).index
+
+        if names is not None:
+            result.names = names
+
+        return result
```
ueshin: How about:

```py
return DataFrame(index=pd.MultiIndex.from_tuples(tuples=tuples, sortorder=sortorder, names=names)).index
```

itholic: ah, yes! seems way better. Thanks for the comment! :)

databricks/koalas/indexes.py
itholic: oh i just researched the more about their difference, and it's better to use `@staticmethod` as you said. thanks for the comment!!

databricks/koalas/indexes.py
```diff
@@ -472,6 +475,49 @@ def any(self, *args, **kwargs):
...
+        """
+        if not is_list_like(tuples):
+            raise TypeError("Input must be a list / sequence of tuple-likes.")
+        elif is_iterator(tuples):
+            tuples = list(tuples)
+
+        if len(tuples) == 0:
+            if names is None:
+                msg = "Cannot infer number of levels from empty list"
+                raise TypeError(msg)
```
ueshin: Do we need to check those by ourselves?
Actually `DataFrame` or `Series` don't check those and rely on pandas. cc @HyukjinKwon

itholic: @ueshin ah, yes you right. we don't need to check them ourselves since we reuse `pd.MultiIndex.from_tuples`. i just removed these unnecessary logics.

ueshin:
Thanks! merging.

itholic:
@ueshin Thanks :)

PR #973, opened 2019-10-29
Add levels property for MultiIndex

charlesdong1991:


databricks/koalas/indexes.py
```diff
@@ -571,6 +571,32 @@ def __getattr__(self, item: str) -> Any:
...
+        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=mi)
+        >>> kdf.index.levels
+        [['a', 'b', 'c'], ['e', 'e', 'f']]
+        """
+        idx_cols = self._kdf._internal.index_columns
+        sdf = self._kdf._sdf.select(idx_cols).dropDuplicates()
+
+        # use sorting is because pandas doesn't care the appearance order of level
+        # names, so e.g. if ['b', 'd', 'a'] will return as ['a', 'b', 'd']
+        return [sorted([row[col] for row in sdf.collect()]) for col in idx_cols]
```
ueshin: This invokes Spark job multiple times.
How about using `F.collect_set`? And then we don't need `dropDuplicates`.

```py
self._kdf._sdf.select([F.collect_set(scol) for scol in self._kdf._internal.index_scols]).collect()
```

databricks/koalas/indexes.py
```diff
@@ -571,6 +571,32 @@ def __getattr__(self, item: str) -> Any:
...
+        >>> mi.names = ['level_1', 'level_2']
+        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=mi)
+        >>> kdf.index.levels
+        [['a', 'b', 'c'], ['d', 'e', 'f']]
+
+        >>> mi = pd.MultiIndex.from_arrays((list('bac'), list('fee')))
+        >>> mi.names = ['level_1', 'level_2']
+        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=mi)
+        >>> kdf.index.levels
+        [['a', 'b', 'c'], ['e', 'e', 'f']]
```
ueshin: This result seems wrong.

```py
>>> kdf.to_pandas().index.levels
FrozenList([['a', 'b', 'c'], ['e', 'f']])
```

databricks/koalas/indexes.py
charlesdong1991: nice one!

databricks/koalas/indexes.py
charlesdong1991: oh, yeah, changed

databricks/koalas/indexes.py
```diff
@@ -571,6 +571,32 @@ def __getattr__(self, item: str) -> Any:
     def rename(self, name, inplace=False):
         raise NotImplementedError()
 
+    @property
+    def levels(self) -> list:
+        """
+        Names of index columns in list.
+
```
ueshin: We might need a note that this might cause OOM since the result could still be huge?

charlesdong1991: emm, but this is similar to `unique` or `dropduplicates` in a way. And practically, maybe it is used for column index most of the time, so I feel it's okay not to address OOM. Any specific reason to address OOM here? would love to hear some thoughts and learn things!! ^^ @ueshin

ueshin: This is different from `unique` or `drop_duplicates` since they are returning `Series` which is distributed across the cluster so it's safe, whereas this collects data into Driver and returns the lists.

ueshin: Basically we should add a note when:

- collect data into Driver
- use `Window` without partition cols.

charlesdong1991: ahh, thanks for detailed explanation! Appreciate a lot!! @ueshin

ueshin: Could you use `.. note:: ` (e.g., see `to_pandas`)?

databricks/koalas/indexes.py
```diff
@@ -571,6 +571,33 @@ def __getattr__(self, item: str) -> Any:
...
+        [['a', 'b', 'c'], ['d', 'e', 'f']]
+
+        >>> mi = pd.MultiIndex.from_arrays((list('bac'), list('fee')))
+        >>> mi.names = ['level_1', 'level_2']
+        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=mi)
+        >>> kdf.index.levels
+        [['a', 'b', 'c'], ['e', 'f']]
+        """
+        scols = self._kdf._internal.index_scols
+        sdf = self._kdf._sdf.select([F.collect_set(scol) for scol in scols]).collect()[0]
```
HyukjinKwon: nit: `row = self._kdf._sdf.select([F.collect_set(scol) for scol in scols]).first()`

databricks/koalas/indexes.py
charlesdong1991: sure!

databricks/koalas/indexes.py
charlesdong1991: looks nicer!! thanks! ^^

databricks/koalas/indexes.py
```diff
@@ -571,6 +571,35 @@ def __getattr__(self, item: str) -> Any:
     def rename(self, name, inplace=False):
         raise NotImplementedError()
 
+    @property
+    def levels(self) -> list:
+        """
+        Names of index columns in list.
+
+        .. note:: Be aware of the possibility of running into out
+        of memory issue if returned list is huge.
```
ueshin: nit: indent

databricks/koalas/indexes.py
```diff
@@ -571,6 +571,35 @@ def __getattr__(self, item: str) -> Any:
...
+
+        Examples:
+        --------
```
ueshin: nit: need one more `-`?

databricks/koalas/indexes.py
charlesdong1991: done

databricks/koalas/indexes.py
charlesdong1991: I removed `:` instead to align with other docstring. And add it in `doc/source`

ueshin:
LGTM.

ueshin:
Thanks! merging.

HyukjinKwon:
I had a short offline discussion with @ueshin .. let's take this out in the today's release.

I don't know how common `levels` and `codes` are. Maybe we should consider adding it back if this is used pretty much common ..?

HyukjinKwon:
I had a offline discussion with @ueshin. It requires to collect all data into driver side (similar discussion at https://github.com/databricks/koalas/pull/952). Let me take this out in today's release for now.

PR #974, opened 2019-10-29
Fix bug when  assign None to name as Index

### LINKED ISSUE #971: BUG: Index setter is not working when `None` is assigned to `name`
with Pandas:
```python
>>> idx = pd.Index([1, 2, 3], name='a')
>>> idx
Int64Index([1, 2, 3], dtype='int64', name='a')
```
if set name to `None` as below:
```python
idx.name = None
```
then idx will turn to 
```python
Int64Index([1, 2, 3], dtype='int64')
```
but in Koalas, this will raise an error, e.g. 
```python
kdf = ks.DataFrame({"a": [1,2,3], index=pd.Index([1, 2, 3], name='a')})
kidx = kdf.index
kidx.name = None
AssertionError
```

itholic:
Resolves #971 

```python
>>> kidx
Int64Index([1, 2, 3], dtype='int64', name='a')
>>> kidx.name = None
>>> kidx
Int64Index([1, 2, 3], dtype='int64')
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -172,10 +172,11 @@ class Index(IndexOpsMixin):
             raise ValueError('Names must be a list-like')
         internal = self._kdf._internal
         if len(internal.index_map) != len(names):
+
             raise ValueError('Length of new names must be {}, got {}'
                              .format(len(internal.index_map), len(names)))
-        if None not in names:
-            names = [name if isinstance(name, tuple) else (name,) for name in names]
+
+        names = [name if isinstance(name, (tuple, type(None))) else (name,) for name in names]
         self._kdf._internal = internal.copy(index_map=list(zip(internal.index_columns, names)))
 
     @property
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -172,7 +172,6 @@ class Index(IndexOpsMixin):
             raise ValueError('Names must be a list-like')
         internal = self._kdf._internal
         if len(internal.index_map) != len(names):
-
             raise ValueError('Length of new names must be {}, got {}'
                              .format(len(internal.index_map), len(names)))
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/indexes.py
```diff
@@ -174,7 +174,8 @@ def names(self, names: List[Union[str, Tuple[str, ...]]]) -> None:
         if len(internal.index_map) != len(names):
             raise ValueError('Length of new names must be {}, got {}'
                              .format(len(internal.index_map), len(names)))
-        names = [name if isinstance(name, tuple) else (name,) for name in names]
+        if None not in names:
+            names = [name if isinstance(name, tuple) else (name,) for name in names]
```
HyukjinKwon: @itholic, does it work if some of names are only `None`? E.g.)

```python
pd.MultiIndex.from_tuples([('a', 'b'), ('a', 'c')], names=['first', None]).names
```

itholic: @HyukjinKwon , thanks for the comment!

i missed that case 😢 , and just added it!

databricks/koalas/indexes.py
```diff
@@ -172,9 +172,11 @@ def names(self, names: List[Union[str, Tuple[str, ...]]]) -> None:
             raise ValueError('Names must be a list-like')
         internal = self._kdf._internal
         if len(internal.index_map) != len(names):
+
```
HyukjinKwon: I would remove this line. Otherwise looks fine cc @ueshin

itholic: oops... i should have double checked that.
Thanks for the comment!

databricks/koalas/indexes.py
```diff
@@ -174,7 +174,8 @@ def names(self, names: List[Union[str, Tuple[str, ...]]]) -> None:
         if len(internal.index_map) != len(names):
             raise ValueError('Length of new names must be {}, got {}'
                              .format(len(internal.index_map), len(names)))
-        names = [name if isinstance(name, tuple) else (name,) for name in names]
+
+        names = [name if isinstance(name, (tuple, type(None))) else (name,) for name in names]
```
HyukjinKwon: It should usually be `is None` but `isinstance(type(None))` is legitimate too. so I think it's fine.

PR #976, opened 2019-10-30
Add Test for config.show_options()

fwani:


databricks/koalas/tests/test_config.py
```diff
@@ -132,3 +132,14 @@ def test_dir_options(self):
...
+    def test_show_options(self):
+        from unittest import mock
+        from io import StringIO
+
+        with mock.patch("sys.stdout", new=StringIO()) as mock_stdout:
+            config.show_options()
+            self.assertEqual(
+                " ".join(mock_stdout.getvalue().split("\n")[1].split()),
+                "Option Default Description"
+            )
```
HyukjinKwon: Thanks for working on this @fwani. Actually, WDYT about adding it as a doctest? we won't have to patch stdout as well.

What we need to test if the first option is properly printed out, and ignore other options in the doctest

```
>>> show_options()
Option                          Default        Description
display.max_rows                1000           This sets the maximum number of rows koalas should
...
=====================================================
```

fwani: Thank you. I'll change that code.

PR #979, opened 2019-10-30
Add test of drop_duplicates with inplace parameter

fwani:
There is no test that tested with `inplace=True` parameter at drop_duplicates function.
So I update test case that is `drop_duplicates(inplace=True)`.

image

Now the `test_drop_duplicates()` check 4 stages. (`single-index columns with/without inplace`, `multi-index columns with/without inplace`)

### DIFF
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -1699,6 +1699,7 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
         pdf = pd.DataFrame({'a': [1, 2, 2, 2, 3], 'b': ['a', 'a', 'a', 'c', 'd']})
         kdf = ks.from_pandas(pdf)
 
+        # inplace is False
         self.assert_eq(pdf.drop_duplicates().sort_index(),
                        kdf.drop_duplicates().sort_index())
         self.assert_eq(pdf.drop_duplicates('a').sort_index(),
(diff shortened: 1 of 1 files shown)


databricks/koalas/tests/test_dataframe.py
```diff
@@ -1699,14 +1699,15 @@ def test_drop_duplicates(self):
...
+        # inplace is False
         self.assert_eq(pdf.drop_duplicates().sort_index(),
                        kdf.drop_duplicates().sort_index())
         self.assert_eq(pdf.drop_duplicates('a').sort_index(),
                        kdf.drop_duplicates('a').sort_index())
         self.assert_eq(pdf.drop_duplicates(['a', 'b']).sort_index(),
                        kdf.drop_duplicates(['a', 'b']).sort_index())
 
-        # multi-index columns
+        # multi-index columns, inplace is False
```
itholic: Not pretty sure, but IMO, maybe this line is not necessary since there are also another options not only `inplace`

HyukjinKwon: It's alright as long as it's easier to read. explicitly classifying the case is okie.

HyukjinKwon:
@fwani can you fill up PR description although the changes are pretty straightforward?

HyukjinKwon:
Looks good otherwise.

databricks/koalas/tests/test_dataframe.py
itholic: @HyukjinKwon thanks for the comment!

PR #984, opened 2019-10-31
Implement DataFrame.quantile

### LINKED ISSUE #959: DataFrame.quantile
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.quantile.html

```Python
df = ks.DataFrame(
  {"a" : [4, 5, 6],
   "b" : [7, 8, 9],
   "c" : [10, 11, 12]},
  index = [1, 2, 3]
)
df.quantile([0.25,0.75])
```

```
PandasNotImplementedError                 Traceback (most recent call last)
<ipython-input-36-ade0575f6b3c> in <module>
----> 1 df.quantile([0.25,0.75])

/srv/conda/envs/notebook/lib/python3.7/site-packages/databricks/koalas/missing/__init__.py in unsupported_function(*args, **kwargs)
     22     def unsupported_function(*args, **kwargs):
     23         raise PandasNotImplementedError(class_name=class_name, method_name=method_name,
---> 24                                         reason=reason)
     25 
     26     def deprecated_function(*args, **kwargs):

PandasNotImplementedError: The method `pd.DataFrame.quantile()` is not implemented yet.

```

itholic:
Resolves #959 

```python
>>> kdf = ks.DataFrame({'a': [1, 2, 3, 4, 5], 'b': [6, 7, 8, 9, 0]})
>>> kdf
   a  b
0  1  6
1  2  7
2  3  8
3  4  9
4  5  0

>>> kdf.quantile(.5)
a    3
b    7
Name: 0.5, dtype: int64

>>> kdf.quantile([.25, .5, .75])
      a  b
0.25  2  6
0.5   3  7
0.75  4  8
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7527,7 +7527,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         args = ", ".join(map(str, quantiles))
 
         percentile_cols = []
-        for i, column in enumerate(self.columns):
+        for i, column in enumerate(self._internal.data_columns):
             sdf = self._sdf
             percentile_cols.append(F.expr(
                 "approx_percentile(`%s`, array(%s), %s)" % (column, args, accuracy))
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7503,6 +7503,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         0.5   3  7
         0.75  4  8
         """
+        from collections import defaultdict
         result_as_series = False
         if axis not in [0, 'index']:
             raise ValueError('axis should be either 0 or "index" currently.')
@@ -7512,29 +7513,59 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             result_as_series = True
             key = str(q)
             q = (q,)
+
         quantiles = q
+        # First calculate the percentiles from all columns and map it to each `quantiles`
+        # by creating each entry as a struct. So, it becomes an array of structs as below:
+        #
+        # +-----------------------------------------+
+        # |                                   arrays|
+        # +-----------------------------------------+
+        # |[[0.25, 2, 6], [0.5, 3, 7], [0.75, 4, 8]]|
+        # +-----------------------------------------+
+        sdf = self._sdf
         args = ", ".join(map(str, quantiles))
-        internal_index_column = SPARK_INDEX_NAME_FORMAT(0)
-        sdf_list = []
+
+        percentile_cols = []
         for i, column in enumerate(self.columns):
             sdf = self._sdf
-            percentile_col = F.expr(
+            percentile_cols.append(F.expr(
                 "approx_percentile(`%s`, array(%s), %s)" % (column, args, accuracy))
-            sdf = sdf.select(percentile_col.alias("percentiles"))
-            value_column = column
-            cols = []
-            for i, quantile in enumerate(quantiles):
-                cols.append(F.struct(
-                    F.lit("%s" % quantile).alias(internal_index_column),
-                    F.expr("percentiles[%s]" % i).alias(value_column)))
-            sdf = sdf.select(F.array(*cols).alias("arrays"))
-            sdf = sdf.select(F.explode(F.col("arrays"))).selectExpr("col.*")
-            sdf_list.append(sdf)
+                .alias(column))
+        sdf = sdf.select(percentile_cols)
+        # Here, after select percntile cols, a sdf looks like below:
+        # +---------+---------+
+        # |        a|        b|
+        # +---------+---------+
+        # |[2, 3, 4]|[6, 7, 8]|
+        # +---------+---------+
 
-        sdf = reduce(lambda x, y: x.join(y, on=internal_index_column), sdf_list)
+        cols_dict = defaultdict(list)
+        for column in self._internal.data_columns:
+            for i in range(len(quantiles)):
+                cols_dict[column].append(F.expr("%s[%s]" % (column, i)).alias(column))
+
+        internal_index_column = SPARK_INDEX_NAME_FORMAT(0)
+        cols = []
+        for i, col in enumerate(zip(*cols_dict.values())):
+            cols.append(F.struct(
+                F.lit("%s" % quantiles[i]).alias(internal_index_column),
+                *col))
+        sdf = sdf.select(F.array(*cols).alias("arrays"))
+
+        # And then, explode it and manually set the index.
+        # +-----------------+---+---+
+        # |__index_level_0__|  a|  b|
+        # +-----------------+---+---+
+        # |             0.25|  2|  6|
+        # |              0.5|  3|  7|
+        # |             0.75|  4|  8|
+        # +-----------------+---+---+
+        sdf = sdf.select(F.explode(F.col("arrays"))).selectExpr("col.*")
+
         internal = self._internal.copy(
             sdf=sdf,
-            data_columns=self.columns,
+            data_columns=self._internal.data_columns,
             index_map=[(internal_index_column, None)],
             column_index=self._internal.column_index,
             column_index_names=None)
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7451,6 +7451,76 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         return self.columns
 
+    # TODO: fix parameter 'axis' and 'numeric_only' to work same as pandas'
+    def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
+        """
+        Return value at the given quantile.
+
+        .. note:: Unlike pandas', the quantile in Koalas is an approximated quantile based upon
+            approximate percentile computation because computing quantile across a large dataset
+            is extremely expensive.
+
+        Parameters
+        ----------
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,76 @@ def keys(self):
...
+        0.25  2  6
+        0.5   3  7
+        0.75  4  8
+        """
+        if axis not in [0, 'index']:
+            raise ValueError('axis should be either 0 or "index" currently.')
+        if numeric_only is not True:
+            raise ValueError("quantile currently doesn't supports numeric_only")
+        if isinstance(q, float):
+            q = (q,)
```
ueshin: Can we do this?

```py
>>> pdf.quantile(.5)
a    3.0
b    7.0
Name: 0.5, dtype: float64
>>> pdf.quantile([.5])
       a    b
0.5  3.0  7.0
```

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,76 @@ def keys(self):
...
+        quantile_columns = [self[col_name].quantile(q) for col_name in self.columns]
+        if len(q) == 1:
+            from databricks.koalas.series import _col
+            return _col(reduce(
+                lambda x, y: x.to_frame().merge(y.to_frame(), left_index=True, right_index=True),
+                quantile_columns).T)
+        else:
+            return reduce(
+                lambda x, y: x.to_frame().merge(y.to_frame(), left_index=True, right_index=True),
+                quantile_columns)
```
ueshin: Actually these are too expensive. I guess we would rather implement here and reuse it in `Series`.

itholic: @ueshin oops i get it. i misunderstood the comment in https://github.com/databricks/koalas/issues/959#issuecomment-546852684. i thought that means we can use it directly in DataFrame but it means just we can refer them and reuse the logic 😅 

Anyway, i'll fix them soon, Thanks for the comment!!

databricks/koalas/frame.py
itholic: Thanks for the comment!! yes sure we can 😄 i'll fix them soon

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,96 @@ def keys(self):
...
+            raise ValueError("quantile currently doesn't supports numeric_only")
+        if isinstance(q, float):
+            result_as_series = True
+            key = str(q)
+            q = (q,)
+        quantiles = q
+        args = ", ".join(map(str, quantiles))
+        internal_index_column = SPARK_INDEX_NAME_FORMAT(0)
+        sdf_list = []
+        for i, column in enumerate(self.columns):
```
ueshin: We should use `self._internal.data_columns`.
We can't use external APIs for accessing Spark DataFrame.
E.g., what if with multi-index columns?

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,96 @@ def keys(self):
...
+            cols = []
+            for i, quantile in enumerate(quantiles):
+                cols.append(F.struct(
+                    F.lit("%s" % quantile).alias(internal_index_column),
+                    F.expr("percentiles[%s]" % i).alias(value_column)))
+            sdf = sdf.select(F.array(*cols).alias("arrays"))
+            sdf = sdf.select(F.explode(F.col("arrays"))).selectExpr("col.*")
+            sdf_list.append(sdf)
+
+        sdf = reduce(lambda x, y: x.join(y, on=internal_index_column), sdf_list)
```
ueshin: I guess we can avoid the join chain? Otherwise this is still too expensive.

itholic: yeah it still looks so, i just fixed it to avoid expensive calculation.
Thanks for the comment! :)

databricks/koalas/frame.py
ueshin: This is not addressed yet.

itholic: oops i missed it. just addressed this. thanks for the comment!! :)

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+
+        quantiles = q
+        # First calculate the percentiles from all columns and map it to each `quantiles`
+        # by creating each entry as a struct. So, it becomes an array of structs as below:
+        #
+        # +-----------------------------------------+
+        # |                                   arrays|
+        # +-----------------------------------------+
+        # |[[0.25, 2, 6], [0.5, 3, 7], [0.75, 4, 8]]|
+        # +-----------------------------------------+
```
ueshin: Seems like these comments are not right?
I don't see any `arrays` column here?
Could you reorganize comments?

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+        sdf = self._sdf
+        args = ", ".join(map(str, quantiles))
+
+        percentile_cols = []
+        for i, column in enumerate(self._internal.data_columns):
```
ueshin: `i` is not used?

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+        # +---------+---------+
+
+        cols_dict = defaultdict(list)
+        for column in self._internal.data_columns:
+            for i in range(len(quantiles)):
+                cols_dict[column].append(F.expr("%s[%s]" % (column, i)).alias(column))
+
+        internal_index_column = SPARK_INDEX_NAME_FORMAT(0)
+        cols = []
+        for i, col in enumerate(zip(*cols_dict.values())):
```
ueshin: Is this deterministic even in Python 3.5?

databricks/koalas/frame.py
itholic: ah, that means we can get those DataFrame after finished below logic.

before ```# And then``` comment line

스크린샷 2019-11-05 오전 9 55 02

This comment mimics Series' one like:

스크린샷 2019-11-05 오전 9 57 48

databricks/koalas/frame.py
itholic: Thanks for the check this out. that's my mistake.

databricks/koalas/frame.py
itholic: i'll check it and come back after then. :)
thanks for the comment!

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+                .alias(column))
+        sdf = sdf.select(percentile_cols)
+        # Here, after select percntile cols, a sdf looks like below:
+        # +---------+---------+
+        # |        a|        b|
+        # +---------+---------+
+        # |[2, 3, 4]|[6, 7, 8]|
+        # +---------+---------+
+
+        cols_dict = defaultdict(list)
```
HyukjinKwon: I think you can use `OrderedDict` to address https://github.com/databricks/koalas/pull/984/files#r342333566

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+        for column in self._internal.data_columns:
+            for i in range(len(quantiles)):
+                cols_dict[column].append(F.expr("%s[%s]" % (column, i)).alias(column))
```
HyukjinKwon: I think you can use `cols_dict[column].getItem(...)` or `cols_dict[column][...]` instead of `F.expr`.

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+        b    7
+        Name: 0.5, dtype: int64
+
+        >>> kdf.quantile([.25, .5, .75])
+              a  b
+        0.25  2  6
+        0.5   3  7
+        0.75  4  8
+        """
+        from collections import defaultdict
```
HyukjinKwon: I would move the import into the top of this file.

databricks/koalas/frame.py
```diff
@@ -7451,6 +7451,127 @@ def keys(self):
...
+        # |[[0.25, 2, 6], [0.5, 3, 7], [0.75, 4, 8]]|
+        # +-----------------------------------------+
+        sdf = self._sdf
+        args = ", ".join(map(str, quantiles))
+
+        percentile_cols = []
+        for column in self._internal.data_columns:
+            sdf = self._sdf
+            percentile_cols.append(F.expr(
+                "approx_percentile(`%s`, array(%s), %s)" % (column, args, accuracy))
```
HyukjinKwon: Ah, https://github.com/databricks/koalas/pull/984/files#r342332798 this is for here.

databricks/koalas/frame.py
itholic: @HyukjinKwon thanks for the comment! :)
maybe you mean instead of `append` ??

databricks/koalas/frame.py
itholic: ah, yeah. i'll do it.

databricks/koalas/frame.py
itholic: thanks for the comment!! yes i get it :)

databricks/koalas/frame.py
itholic: it should be better to use `OrderedDict` rather than `defaultdict`

databricks/koalas/frame.py
itholic: i just fixed it :)

PR #985, opened 2019-10-31
Implement min, max, sum and mean of Expanding in Series and DataFrame

HyukjinKwon:
This PR implement min, max, sum and mean of Expanding in Series and DataFrame

```python
>>> ks.Series([2, 3, float("nan"), 10]).expanding().sum()
0     2.0
1     5.0
2     5.0
3    15.0
Name: 0, dtype: float64
>>> ks.DataFrame({'a': [1, float('nan'), 3], 'b': [1.0, 2.0, 3.0]}).expanding().sum()
     a    b
0  1.0  1.0
1  1.0  3.0
2  4.0  6.0
```

relates to #984 977


databricks/koalas/window.py
```diff
@@ -80,6 +80,34 @@ def __getattr__(self, item: str) -> Any:
                 return partial(property_or_func, self)
         raise AttributeError(item)
 
+    def _apply_as_series_or_frame(self, func):
```
HyukjinKwon: Should I maybe separate `Expanding` into `SeriesExpandng` and `FrameExpanding` .. ? not sure yet.

databricks/koalas/missing/window.py
```diff
@@ -71,14 +67,10 @@ class _MissingPandasLikeRolling(object):
...
     cov = unsupported_function_rolling("cov")
     kurt = unsupported_function_rolling("kurt")
-    max = unsupported_function_rolling("max")
-    mean = unsupported_function_rolling("mean")
     median = unsupported_function_rolling("median")
-    min = unsupported_function_rolling("min")
     quantile = unsupported_function_rolling("quantile")
     skew = unsupported_function_rolling("skew")
     std = unsupported_function_rolling("std")
-    sum = unsupported_function_rolling("sum")
```
itholic: maybe you only implemented that functions for `Expanding`??

databricks/koalas/missing/window.py
```diff
@@ -120,14 +108,10 @@ class _MissingPandasLikeRollingGroupby(object):
...
```
itholic: ditto

databricks/koalas/missing/window.py
HyukjinKwon: Yeah, it's for simplicity. `ExpandingGroupBy` extends `Expanding` so it inherits the implementation. As a workaround for now, I am manually throwing an exception at `ExpandingGroupBy`.

This missing methods are tested so this is just to make the tests pass for now.

itholic: @HyukjinKwon Thanks for the comment!

okay i got it, but i mentioned about these commented lines are indicates functions of `Rolling`, not `Expanding` 😄

HyukjinKwon: Oh, you're right. let me check.

HyukjinKwon:
Let me merge this to proceed forward .. I'm still touching the codes around here so please let me know if you guys have more comments.

PR #987, opened 2019-10-31
Fix Dataframe.melt function & Add doctest case for melt function

fwani:
There is no test that tested with `str, tuple type` of `value_vars` parameter at Dataframe.melt function. 
I update doctest case.
Screen Shot 2019-10-31 at 5 36 37 PM

And, I thought, There is a bug like the image below.
Screen Shot 2019-10-31 at 5 46 57 PM

I changed some codes.


itholic:
nice finding

databricks/koalas/frame.py
```diff
@@ -6801,8 +6825,6 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
             value_vars = []
         elif isinstance(value_vars, str):
             value_vars = [(value_vars,)]
-        elif isinstance(value_vars, tuple):
-            value_vars = [value_vars]
```
ueshin: The root cause is not here.
We should check the specified vars are included in `column_index`.
E.g.,

```py
>>> df.melt(value_vars='Z')
...
pyspark.sql.utils.AnalysisException: "Can't extract value from pairs#15: need struct type but got string;"
```

This is necessary for multi-index columns support, but we might want to do the same thing for `id_vars` to follow pandas' behavior. I don't like the behavior, though. :)

fwani: Thank you for the review. 👍 

I'll fix it in the other direction after comparing the melt function of pandas and koalas.

fwani:
@ueshin 
I don't know if I understand your opinion well,
but I added some lines to catch some exceptions in melt function.

databricks/koalas/frame.py
```diff
@@ -6797,14 +6818,29 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
 
         column_index = self._internal.column_index
 
+        missing = [True for id_var in id_vars if id_var not in column_index]
+        if True in missing:
+            raise KeyError("The following 'id_vars' are not present"
+                           " in the DataFrame: {}"
+                           .format(id_vars))
```
ueshin: Shall we use `name_like_string` for `id_vars`? E.g., `.format([nam_like_string(id) for id in id_vars])`

databricks/koalas/frame.py
```diff
@@ -6797,14 +6818,29 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
+            else:
+                raise ValueError('value_vars must be a list of tuples'
+                                 ' when columns are a MultiIndex')
         else:
             value_vars = [valv if isinstance(valv, tuple) else (valv,) for valv in value_vars]
+        missing = [True for value_var in value_vars if value_var not in column_index]
+        if True in missing:
+            raise KeyError("The following 'value_vars' are not present"
+                           " in the DataFrame: {}"
+                           .format(value_vars))
```

databricks/koalas/frame.py
```diff
@@ -6782,6 +6782,27 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
+        >>> df.columns = pd.MultiIndex.from_tuples([('X', 'A'), ('X', 'B'), ('Y', 'C')])
+        >>> df.melt(id_vars=('A',))  # doctest: +ELLIPSIS
+        Traceback (most recent call last):
+          ...
+        ValueError: id_vars must be a list of tuples when columns are a MultiIndex
+
+        >>> df.melt(value_vars=('A',))  # doctest: +ELLIPSIS
+        Traceback (most recent call last):
+          ...
+        ValueError: value_vars must be a list of tuples when columns are a MultiIndex
```
ueshin: Could you move these to `DataFrameTest.test_melt`?

databricks/koalas/frame.py
fwani: I understand.
but I have a question, 
If id_vars is ['A', 'Z'], should we raise all of id_vars ('A', 'Z') or only non-existent column ('Z')?

databricks/koalas/frame.py
fwani: Thanks!

databricks/koalas/frame.py
fwani: Ok, Thanks for reviewing.

databricks/koalas/frame.py
ueshin: Could you check pandas' behavior? and shall we follow it?

fwani: Ok.
I checked result of pandas' behavior.
```
>>> pdf.melt(value_vars=['A', 'Z'])
  ...
KeyError: "The following 'value_vars' are not present in the DataFrame: ['Z']"
>>> pdf.melt(value_vars=[('X', 'A'),('X', 'Z')])
  ...
KeyError: "The following 'value_vars' are not present in the DataFrame: ['Z']"
```
I changed the codes of the result of that exception.

databricks/koalas/frame.py
```diff
@@ -6783,28 +6789,56 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
+                    raise ValueError('id_vars must be a list of tuples'
+                                     ' when columns are a MultiIndex')
+            else:
+                id_vars = [idv if isinstance(idv, tuple) else (idv,)
+                           for idv in id_vars]
 
-        column_index = self._internal.column_index
+            raveled_column_index = np.ravel(column_index)
+            missing = [idv for idv in np.ravel(id_vars)
+                       if idv not in raveled_column_index]
```
ueshin: These seem not like a good strategy. E.g., for multi-index columns,

```py
>>> df
   X     Y
   A  B  C
0  a  1  2
1  b  3  4
2  c  5  6
>>> df.melt(value_vars=[('Y', 'A')])
...
pyspark.sql.utils.AnalysisException: "Can't extract value from pairs#68: need struct type but got string;"
```

still throws `AnalysisException`.

fwani: Wow! I didn’t catch that case.
Thank you. :)

fwani:
I checked lint, pytest, doctest, building document in my local environment.
But lint-python test is failed at just version 3.5 in the Travis CI. not 3.6, 3.7
How can I fix it...?

databricks/koalas/frame.py
```diff
@@ -6783,28 +6789,64 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
+            if len(non_existence_col) != 0:
+                raveled_column_index = np.ravel(column_index)
+                missing = [nec for nec in np.ravel(non_existence_col)
+                           if nec not in raveled_column_index]
+                if len(missing) != 0:
+                    raise KeyError("The following 'id_vars' are not present"
+                                   " in the DataFrame: {}".format(missing))
+                else:
+                    raise KeyError("None of [{}] are in the [columns]"
+                                   .format(pd.MultiIndex.from_tuples(non_existence_col)))
```
ueshin: The error message is strange?

```
KeyError: "None of [MultiIndex(levels=[['X'], ['Y']],\n           codes=[[0], [0]])] are in the [columns]"
```

How about:

```diff
-                    raise KeyError("None of [{}] are in the [columns]"
-                                   .format(pd.MultiIndex.from_tuples(non_existence_col)))
+                    raise KeyError("None of {} are in the {}"
+                                   .format(non_existence_col, column_index))
```

then

```
KeyError: "None of [('X', 'Y')] are in the [('X', 'A'), ('X', 'B'), ('Y', 'C')]"
```

databricks/koalas/frame.py
```diff
@@ -6783,28 +6789,64 @@ def melt(self, id_vars=None, value_vars=None, var_name=None,
...
+            if len(non_existence_col) != 0:
+                raveled_column_index = np.ravel(column_index)
+                missing = [nec for nec in np.ravel(non_existence_col)
+                           if nec not in raveled_column_index]
+                if len(missing) != 0:
+                    raise KeyError("The following 'value_vars' are not present"
+                                   " in the DataFrame: {}".format(missing))
+                else:
+                    raise KeyError("None of [{}] are in the [columns]"
+                                   .format(pd.MultiIndex.from_tuples(non_existence_col)))
```

databricks/koalas/frame.py
fwani: I agree with you. 
I'll change error message the case of id_vars, value_vars.

ueshin:
Thanks! merging.

Issue #977, opened 2019-10-30
Implement missing Expanding and Rolling APIs in Series, Frame, SeriesGroupBy, FrameGroupBy


HyukjinKwon:
See https://github.com/databricks/koalas/blob/master/databricks/koalas/missing/window.py

It mimics the class hierarchy from pandas internally. Ideally, we will be able to share implementation across APIs (like `cumsum`).

HyukjinKwon:
I am starting with Expanding in Series and Frame.

itholic:
I am starting with Rolling in Series and Frame.

HyukjinKwon:
Please go ahead!

PR #990, opened 2019-11-01
Implement rolling().count() in Series and Frame

itholic:
This PR implements rolling().count():

```python
>>> s = ks.Series([2, 3, float("nan"), 10])
>>> s.rolling(1).count()
0    1.0
1    1.0
2    0.0
3    1.0
Name: 0, dtype: float64

>>> s.to_frame().rolling(1).count()
     0
0  1.0
1  1.0
2  0.0
3  1.0
```

Relates to #977


HyukjinKwon:
Let's fix expanding ones first, and them get those in.

databricks/koalas/window.py
```diff
@@ -42,6 +60,91 @@ def __getattr__(self, item: str) -> Any:
...
+        3  1.0
+
+        >>> s.to_frame().rolling(3).count()
+             0
+        0  1.0
+        1  2.0
+        2  2.0
+        3  2.0
+        """
+        def count(scol, _):
```
HyukjinKwon: why `_` is needed?

itholic: Because in the Decorator `_apply_as_series_or_frame`, Rolling functions are called `func(ker._scol, ker._internal.index_scols)` like below.


스크린샷 2019-11-05 오후 12 51 18


But only for the case of `count`, we don't need `ker._internal.index_scols` parameter.

itholic: i fixed them!

databricks/koalas/window.py
```diff
@@ -30,8 +30,26 @@ class _RollingAndExpanding(object):
...
+        self._window_val = window
+        self._min_periods = min_periods
+        self.kdf_or_kser = kdf_or_kser
+        if not isinstance(kdf_or_kser, (DataFrame, Series, DataFrameGroupBy, SeriesGroupBy)):
+            raise TypeError(
+                "kdf_or_kser must be a series or dataframe; however, got: %s" % type(kdf_or_kser))
+        if isinstance(kdf_or_kser, (DataFrame, Series)):
+            self._index_scols = kdf_or_kser._internal.index_scols
+            self._window = Window.orderBy(self._index_scols).rowsBetween(
+                Window.currentRow - window, Window.currentRow)
```
HyukjinKwon: You can simply send `-window` instead of `Window.currentRow - window` IIRC.

itholic: thanks for the comment! what a convenient.

PR #995, opened 2019-11-01
Fix median to support multi-index columns.

ueshin:
This PR fixes `_Frame.median()` to:

- follow `_scol` for `Series`.
- support multi-index columns.
- avoid attaching default index for `DataFrame`.

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -1422,7 +1422,7 @@ class _Frame(object):
             .withColumn('__DUMMY__', F.monotonically_increasing_id())
         # This is expected to be small so it's fine to transpose.
         return DataFrame(kdf._internal.copy(sdf=sdf, index_map=[('__DUMMY__', None)])) \
-                   ._to_internal_pandas().transpose().iloc[:, 0]
+            ._to_internal_pandas().transpose().iloc[:, 0]
 
     def rolling(self, *args, **kwargs):
         return Rolling(self)
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -1404,22 +1404,25 @@ class _Frame(object):
         from databricks.koalas.frame import DataFrame
         from databricks.koalas.series import Series, _col
 
-        kdf_or_ks = self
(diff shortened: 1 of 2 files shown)


databricks/koalas/generic.py
```diff
@@ -1369,29 +1369,60 @@ def median(self, accuracy=10000):
...
         from databricks.koalas.frame import DataFrame
-        from databricks.koalas.series import Series
+        from databricks.koalas.series import Series, _col
 
         kdf_or_ks = self
         if isinstance(kdf_or_ks, Series):
-            ks = kdf_or_ks
-            return self._reduce_for_stat_function(
+            ks = _col(kdf_or_ks.to_frame())
+            return ks._reduce_for_stat_function(
```
itholic: i don't think it is important, but maybe we better use `kser` rather than `ks` when indicate `Series` since the `ks` is used when import koalas conventionally like `import databricks.koalas as ks` ?

Although maybe it is really not so important, I mentioned it because i thought you think about these conventions are important when you work like #919

ueshin: Good catch! Yeah we should use `kser` here.

databricks/koalas/generic.py
```diff
@@ -1369,29 +1369,63 @@ def median(self, accuracy=10000):
...
         # to get a column name from Spark column but we need it to pass it through `expr`.
-        kdf = kdf_or_ks
+        kdf = kdf_or_kser
         sdf = kdf._sdf
         median = lambda name: F.expr("approx_percentile(`%s`, 0.5, %s)" % (name, accuracy))
-        sdf = sdf.select([median(col).alias(col) for col in kdf.columns])
+        sdf = sdf.select([median(col).alias(col) for col in kdf._internal.data_columns])
+
+        # Attach a dummy column for index to avoid default index.
+        sdf = sdf.withColumn('__DUMMY__', F.monotonically_increasing_id())
```
HyukjinKwon: Hmmm ..  maybe we should just document `__...__` columns are reserved in Koalas .. of course in a separate PR..

PR #997, opened 2019-11-04
Implement Series.between

harupy:
Implement Series.between

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -668,6 +668,83 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         """
         return (self != other).rename(self.name)
 
+    def between(self, left, right, inclusive=True):
+        """
+        Return boolean Series equivalent to left <= series <= right.
+        This function returns a boolean vector containing `True` wherever the
+        corresponding Series element is between the boundary values `left` and
+        `right`. NA values are treated as `False`.
+
+        Parameters
+        ----------
+        left : scalar or list-like
+            Left boundary.
+        right : scalar or list-like
+            Right boundary.
+        inclusive : bool, default True
+            Include boundaries.
+
+        Returns
+        -------
+        Series
+            Series representing whether each element is between left and
+            right (inclusive).
+
+        See Also
+        --------
+        Series.gt : Greater than of series and other.
+        Series.lt : Less than of series and other.
+
+        Notes
+        -----
+        This function is equivalent to ``(left <= ser) & (ser <= right)``
+
+        Examples
+        --------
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -668,6 +668,80 @@ def ne(self, other):
...
+        This function returns a boolean vector containing `True` wherever the
+        corresponding Series element is between the boundary values `left` and
+        `right`. NA values are treated as `False`.
+
+        Parameters
+        ----------
+        left : scalar or list-like
+            Left boundary.
+        right : scalar or list-like
+            Right boundary.
```
ueshin: Could you add a test if we can support list-like?

ueshin: Btw, pandas only support `scalar` value? https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.between.html

harupy: https://github.com/pandas-dev/pandas/issues/28435

ueshin: cool!
Then shall we add tests for them?

harupy:
```
>>> a = ks.Series([1, 2, 3])
>>> b = ks.Series([4, 5, 6])
>>> a > b

  File "/Users/harutaka/Desktop/projects/koalas/databricks/koalas/base.py", line 66, in wrapper
    return align_diff_series(apply_func, self, *args, how="full")
  File "/Users/harutaka/Desktop/projects/koalas/databricks/koalas/utils.py", line 255, in align_diff_series
    combined = combine_frames(this_series.to_frame(), *cols, how=how)
  File "/Users/harutaka/Desktop/projects/koalas/databricks/koalas/utils.py", line 116, in combine_frames
    raise ValueError("Cannot combine column argument because "
ValueError: Cannot combine column argument because it comes from a different dataframe
```

I got this error when I tried to compare two different series.

harupy: This works.

```python
>>> df = ks.DataFrame({
    'a': [1, 2, 3],
    'b': [4, 5, 6],
    'c': [7, 8, 9]
})

>>> df.a < df.b
0    True
1    True
2    True
Name: a, dtype: bool

>>> df.b < df.c
0    True
1    True
2    True
Name: b, dtype: bool

```

HyukjinKwon: For different dataframe, you can add the tests under `test_ops_on_diff_frames.py`. `compute.ops_on_diff_frames` option allows operations on different DataFrames.

HyukjinKwon:
Let me just merge - it looks pretty straightforward.

harupy:
@HyukjinKwon Thanks. I'll create a PR to add tests.

PR #1005, opened 2019-11-05
Add Series.between to the documentation

harupy:
I forgot to add `Series.between` to the doc in #997.


harupy:
@HyukjinKwon thanks

PR #1017, opened 2019-11-06
Implement Series.mask()

itholic:
This PR implement `mask` of `Series` 
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.mask.html#pandas.Series.mask)

```python
>>> s1 = ks.Series([0, 1, 2, 3, 4])
>>> s2 = ks.Series([100, 200, 300, 400, 500])
>>> s1.mask(s1 > 0).sort_index()
0    0.0
1    NaN
2    NaN
3    NaN
4    NaN
Name: 0, dtype: float64

>>> s1.mask(s1 > 1, 10).sort_index()
0     0
1     1
2    10
3    10
4    10
Name: 0, dtype: int64

>>> s1.mask(s1 > 1, s1 + 100).sort_index()
0      0
1      1
2    102
3    103
4    104
Name: 0, dtype: int64

>>> s1.mask(s1 > 1, s2).sort_index()
0      0
1      1
2    300
3    400
4    500
Name: 0, dtype: int64
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -3804,6 +3804,86 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
         return result
 
+    def mask(self, cond, other=np.nan):
+        """
+        Replace values where the condition is True.
+
+        Parameters
+        ----------
+        cond : boolean Series
+            Where cond is False, keep the original value. Where True,
+            replace with corresponding value from other.
+        other : scalar, Series
+            Entries where cond is True are replaced with corresponding value from other.
+
+        Returns
+        -------
+        Series
+
+        Examples
+        --------
+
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -3804,6 +3804,86 @@ def where(self, cond, other=np.nan):
 
         return result
 
+    def mask(self, cond, other=np.nan):
```
ueshin: Is this the same as `where` except for the opposite cond?
If so, how about just reusing it like `return self.where(~cond, other)`?

itholic: thanks for the comment!! what a gorgeous way it is. 👍

itholic: i just fixed it !
ah, i should have taken a look at functionalities once more before implement something.

PR #1018, opened 2019-11-06
Implement DataFrame.where() & DataFrame.mask()

### LINKED ISSUE #884: DataFrame.where
Implement `DataFrame.where`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.where.html

itholic:
Resolves #884 

This PR implement `where` of `DataFrame` 
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.where.html#pandas.DataFrame.where)
and `mask` of `DataFrame` (same as where except for the opposite cond)
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.mask.html#pandas.DataFrame.mask)

```python
>>> df1 = ks.DataFrame({'A': [0, 1, 2, 3, 4], 'B':[100, 200, 300, 400, 500]})
>>> df2 = ks.DataFrame({'A': [0, -1, -2, -3, -4], 'B':[-100, -200, -300, -400, -500]})
>>> df1
   A    B
0  0  100
1  1  200
2  2  300
3  3  400
4  4  500
>>> df2
   A    B
0  0 -100
1 -1 -200
2 -2 -300
3 -3 -400
4 -4 -500

>>> df1.where(df1 > 0).sort_index()
     A      B
0  NaN  100.0
1  1.0  200.0
2  2.0  300.0
3  3.0  400.0
4  4.0  500.0

>>> df1.where(df1 > 1, 10).sort_index()
    A    B
0  10  100
1  10  200
2   2  300
3   3  400
4   4  500

>>> df1.where(df1 > 1, df1 + 100).sort_index()
     A    B
0  100  100
1  101  200
2    2  300
3    3  400
4    4  500

>>> df1.where(df1 > 1, df2).sort_index()
   A    B
0  0  100
1 -1  200
2  2  300
3  3  400
4  4  500
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2150,6 +2150,118 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         return DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map))
 
+    def mask(self, cond, other=np.nan):
+        """
+        Replace values where the condition is True.
+
+        Parameters
+        ----------
+        cond : boolean DataFrame
+            Where cond is False, keep the original value. Where True,
+            replace with corresponding value from other.
+        other : scalar, DataFrame
+            Entries where cond is True are replaced with corresponding value from other.
+
+        Returns
+        -------
+        DataFrame
+
+        Examples
+        --------
+
+        >>> from databricks.koalas.config import set_option, reset_option
+        >>> set_option("compute.ops_on_diff_frames", True)
+        >>> df1 = ks.DataFrame({'A': [0, 1, 2, 3, 4], 'B':[100, 200, 300, 400, 500]})
+        >>> df2 = ks.DataFrame({'A': [0, -1, -2, -3, -4], 'B':[-100, -200, -300, -400, -500]})
+        >>> df1
+           A    B
+        0  0  100
+        1  1  200
+        2  2  300
+        3  3  400
+        4  4  500
+        >>> df2
+           A    B
+        0  0 -100
+        1 -1 -200
+        2 -2 -300
+        3 -3 -400
+        4 -4 -500
+
+        >>> df1.mask(df1 > 0).sort_index()
+             A   B
+        0  0.0 NaN
+        1  NaN NaN
+        2  NaN NaN
+        3  NaN NaN
+        4  NaN NaN
+
+        >>> df1.mask(df1 > 1, 10).sort_index()
+            A   B
+        0   0  10
+        1   1  10
+        2  10  10
+        3  10  10
+        4  10  10
+
+        >>> df1.mask(df1 > 1, df1 + 100).sort_index()
+             A    B
+        0    0  200
+        1    1  300
+        2  102  400
+        3  103  500
+        4  104  600
+
+        >>> df1.mask(df1 > 1, df2).sort_index()
+           A    B
+        0  0 -100
+        1  1 -200
+        2 -2 -300
+        3 -3 -400
+        4 -4 -500
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        tmp_cond_col_name = '__tmp_cond_col_{}__'
+        tmp_other_col_name = '__tmp_other_col_{}__'
+        kdf = self.copy()
+        for column in self._internal._data_columns:
+            kdf[tmp_cond_col_name.format(column)] = ~cond[column]
+            if isinstance(other, DataFrame):
+                kdf[tmp_other_col_name.format(column)] = other[column]
+            else:
+                kdf[tmp_other_col_name.format(column)] = other
+
+        # above logic make spark dataframe looks like below:
+        # +-----------------+---+---+------------------+-------------------+------------------+--...
+        # |__index_level_0__|  A|  B|__tmp_cond_col_A__|__tmp_other_col_A__|__tmp_cond_col_B__|__...
+        # +-----------------+---+---+------------------+-------------------+------------------+--...
+        # |                0|  0|100|              true|                  0|             false|  ...
+        # |                1|  1|200|             false|                 -1|             false|  ...
+        # |                3|  3|400|              true|                 -3|             false|  ...
+        # |                2|  2|300|             false|                 -2|              true|  ...
+        # |                4|  4|500|             false|                 -4|             false|  ...
+        # +-----------------+---+---+------------------+-------------------+------------------+--...
+
+        sdf = kdf._sdf
+        conditions = []
+        for column in self._internal._data_columns:
+            data_col_name = self._internal.column_name_for(column)
+            conditions.append(
+                F.when(sdf[tmp_cond_col_name.format(column)], sdf[data_col_name])
+                 .otherwise(sdf[tmp_other_col_name.format(column)]).alias(data_col_name))
+
+        index_column = self._internal.index_columns[0]
+        sdf = sdf.select(index_column, *conditions)
+        internal = self._internal.copy(
+            sdf=sdf,
+            index_map=self._internal.index_map,
+            data_columns=self._internal._data_columns,
+            column_index=self._internal.column_index,
+            column_index_names=None)
+
+        return DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map))
+
     @property
     def index(self):
         """The index (row labels) Column of the DataFrame.
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2112,18 +2112,64 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         3  3  400
         4  4  500
 
+        When the column name of cond is different from self, it treats all values are False
+        >>> cond = ks.DataFrame({'C': [0, -1, -2, -3, -4], 'D':[4, 3, 2, 1, 0]}) % 3 == 0
+        >>> cond
+               C      D
+        0   True  False
+        1  False   True
+        2  False  False
+        3   True  False
+        4  False   True
+
+        >>> df1.where(cond).sort_index()
+            A   B
+        0 NaN NaN
+        1 NaN NaN
+        2 NaN NaN
+        3 NaN NaN
+        4 NaN NaN
+
+        When the type of cond is Series, it just check boolean regardless of column name
+
+        >>> cond = ks.Series([1, 2]) > 1
+        >>> cond
+        0    False
+        1     True
+        Name: 0, dtype: bool
+
+        >>> df1.where(cond).sort_index()
+             A      B
+        0  NaN    NaN
+        1  1.0  200.0
+        2  NaN    NaN
+        3  NaN    NaN
+        4  NaN    NaN
+
         >>> reset_option("compute.ops_on_diff_frames")
         """
+        from databricks.koalas.series import Series
         tmp_cond_col_name = '__tmp_cond_col_{}__'
         tmp_other_col_name = '__tmp_other_col_{}__'
         kdf = self.copy()
-        for column in self._internal.data_columns:
-            kdf[tmp_cond_col_name.format(column)] = cond[column]
-            if isinstance(other, DataFrame):
-                kdf[tmp_other_col_name.format(column)] = other[column]
-            else:
-                kdf[tmp_other_col_name.format(column)] = other
+        if isinstance(cond, DataFrame):
+            for column in self._internal.data_columns:
+                kdf[tmp_cond_col_name.format(column)] = cond.get(column, False)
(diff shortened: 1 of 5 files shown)


databricks/koalas/frame.py
```diff
@@ -2038,6 +2038,230 @@ class  locomotion
...
+            conditions.append(
+                F.when(sdf[tmp_cond_col_name.format(column)], sdf[data_col_name])
+                 .otherwise(sdf[tmp_other_col_name.format(column)]).alias(data_col_name))
+
+        index_column = self._internal.index_columns[0]
+        sdf = sdf.select(index_column, *conditions)
+        internal = self._internal.copy(
+            sdf=sdf,
+            index_map=self._internal.index_map,
+            data_columns=self._internal._data_columns,
```
HyukjinKwon: `data_columns` usage was removed. We can remove it I guess.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -760,6 +760,56 @@ def test_xs(self):
         with self.assertRaisesRegex(KeyError, msg):
             kdf.xs(('mammal', 'dog', 'walks', 'foo'))
 
+    def test_where(self):
+        pdf1 = pd.DataFrame({'A': [0, 1, 2, 3, 4], 'B': [100, 200, 300, 400, 500]})
+        pdf2 = pd.DataFrame({'A': [0, -1, -2, -3, -4], 'B': [-100, -200, -300, -400, -500]})
+        kdf1 = ks.from_pandas(pdf1)
+        kdf2 = ks.from_pandas(pdf2)
+
+        set_option("compute.ops_on_diff_frames", True)
```
HyukjinKwon: I think we can add these tests at `OpsOnDiffFramesEnabledTest`

databricks/koalas/frame.py
```diff
@@ -2038,6 +2038,230 @@ class  locomotion
...
+
+        sdf = kdf._sdf
+        conditions = []
+        for column in self._internal._data_columns:
+            data_col_name = self._internal.column_name_for(column)
+            conditions.append(
+                F.when(sdf[tmp_cond_col_name.format(column)], sdf[data_col_name])
+                 .otherwise(sdf[tmp_other_col_name.format(column)]).alias(data_col_name))
+
+        index_column = self._internal.index_columns[0]
```
HyukjinKwon: Hm, what about the case of multi-index? Seems it won't work.

databricks/koalas/frame.py
```diff
@@ -2038,6 +2038,230 @@ class  locomotion
...
+        sdf = sdf.select(index_column, *conditions)
+        internal = self._internal.copy(
+            sdf=sdf,
+            index_map=self._internal.index_map,
+            data_columns=self._internal._data_columns,
+            column_index=self._internal.column_index,
+            column_index_names=None)
+
+        return DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map))
```
HyukjinKwon: Can we reuse `where` here?

HyukjinKwon:
Thanks for working on this!

databricks/koalas/frame.py
itholic: okay. pulling up-to-date codes and fix them to works properly

databricks/koalas/tests/test_dataframe.py
itholic: yeah i think so too.

databricks/koalas/frame.py
itholic: okay let me take a look at this. :)

databricks/koalas/frame.py
itholic: i just fixed logic to reuse `where`

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,204 @@ class  locomotion
...
+           A    B
+        0  0  100
+        1 -1  200
+        2  2  300
+        3  3  400
+        4  4  500
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        tmp_cond_col_name = '__tmp_cond_col_{}__'
```
HyukjinKwon: I would assert the given `cond` is a DataFrame

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,204 @@ class  locomotion
...
+        tmp_other_col_name = '__tmp_other_col_{}__'
+        kdf = self.copy()
+        for column in self._internal.data_columns:
+            kdf[tmp_cond_col_name.format(column)] = cond[column]
+            if isinstance(other, DataFrame):
+                kdf[tmp_other_col_name.format(column)] = other[column]
```
HyukjinKwon: What happen if the number of columns in `other` is different with the current DataFrame?

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,204 @@ class  locomotion
...
+        # |                4|  4|500|             false|                 -4|             false|  ...
+        # +-----------------+---+---+------------------+-------------------+------------------+--...
+
+        sdf = kdf._sdf
+        conditions = []
+        for column in self._internal.data_columns:
+            data_col_name = self._internal.column_name_for(column)
+            conditions.append(
+                F.when(sdf[tmp_cond_col_name.format(column)], sdf[data_col_name])
+                 .otherwise(sdf[tmp_other_col_name.format(column)]).alias(data_col_name))
```
HyukjinKwon: I would fix the indentation as below to make it easier to read.

```python
conditions.append(
    F.when(
        sdf[tmp_cond_col_name.format(column)], sdf[data_col_name]
    ).otherwise(
        sdf[tmp_other_col_name.format(column)]
    ).alias(data_col_name))
```

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,204 @@ class  locomotion
...
+        # +-----------------+---+---+------------------+-------------------+------------------+--...
+        # |                0|  0|100|              true|                  0|             false|  ...
+        # |                1|  1|200|             false|                 -1|             false|  ...
+        # |                3|  3|400|              true|                 -3|             false|  ...
+        # |                2|  2|300|             false|                 -2|              true|  ...
+        # |                4|  4|500|             false|                 -4|             false|  ...
+        # +-----------------+---+---+------------------+-------------------+------------------+--...
+
+        sdf = kdf._sdf
+        conditions = []
```
HyukjinKwon: I would name it `output`. It seems not a set of conditions but rather the output from the condition.

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,204 @@ class  locomotion
...
+                 .otherwise(sdf[tmp_other_col_name.format(column)]).alias(data_col_name))
+
+        index_columns = self._internal.index_columns
+        sdf = sdf.select(*index_columns, *conditions)
+        internal = self._internal.copy(
+            sdf=sdf,
+            index_map=self._internal.index_map,
+            column_index=self._internal.column_index,
+            column_scols=[scol_for(sdf, col) for col in self._internal.data_columns],
+            column_index_names=None)
```
HyukjinKwon: I think it will keep all the index information as are .. ? is `kdf.copy(sdf=sdf)` not enough?

HyukjinKwon:
Looks fine otherwise.

databricks/koalas/frame.py
itholic: that's right. i wasn't enough insight about copy of internal frame.
fixed them. thanks for the comment!

databricks/koalas/frame.py
itholic: looks way better, thanks!! 😄

databricks/koalas/frame.py
itholic: i just checked the pandas' behavior,

they just compare boolean value in order regardless of column name like below when the type of `cond` is `Series`

```python
>>> df
   A    B
0  0  100
1  1  200
2  2  300
3  3  400
4  4  500

>>> cond
0    False
1     True
dtype: bool

>>> df.where(cond)
     A      B
0  NaN    NaN
1  1.0  200.0
2  NaN    NaN
3  NaN    NaN
4  NaN    NaN
```

so i should have consider when type of `cond` is `Series` also.

now it works same as pandas.

databricks/koalas/frame.py
itholic: yeah i just checked the pandas',

when the number of columns in `other` is different with the self,

they just ignore additional columns or columns that have different name rather than raises error.

like this.

```python
>>> self
   A    B
0  0  100
1  1  200
2  2  300
3  3  400
4  4  500

>>> other
   A    B  C
0  0 -100  4
1 -1 -200  3
2 -2 -300  2
3 -3 -400  1
4 -4 -500  0

>>> self.where(cond, other)
   A    B
0  0 -100
1 -1 -200
2 -2  300
3  3 -400
4 -4 -500
```

and i just fixed make koalas works same as pandas like above example.

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,248 @@ class  locomotion
...
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        from databricks.koalas.series import Series
+        tmp_cond_col_name = '__tmp_cond_col_{}__'
+        tmp_other_col_name = '__tmp_other_col_{}__'
+        kdf = self.copy()
+        if isinstance(cond, DataFrame):
+            for column in self._internal.data_columns:
+                kdf[tmp_cond_col_name.format(column)] = cond.get(column, False)
```
HyukjinKwon: Can we deduplicate this two `for` loop? Seems this line is only the difference for both `if isinstance(cond, DataFrame)` and `if isinstance(other, DataFrame)`.

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,248 @@ class  locomotion
...
+                    sdf[tmp_cond_col_name.format(column)], sdf[data_col_name]
+                ).otherwise(
+                    sdf[tmp_other_col_name.format(column)]
+                ).alias(data_col_name))
+
+        index_columns = self._internal.index_columns
+        sdf = sdf.select(*index_columns, *output)
+        internal = self._internal.copy(sdf=sdf)
+
+        return DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map))
```
HyukjinKwon: Hm, why can't we `self._internal.copy(sdf=sdf)`?

itholic: yeah, i think we can just use copy here with `sdf` and `column_scols`  since i think maybe `self._internal.sdf` and `sdf` are not same identically.

databricks/koalas/frame.py
itholic: yes we can, thanks for the comment! :)

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,248 @@ class  locomotion
...
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        sdf = cond._internal.sdf
+        for col in cond._internal.data_columns:
+            sdf = sdf.withColumn(col, ~F.col(col))
+
+        internal = cond._internal.copy(
+            sdf=sdf,
+            index_map=cond._internal.index_map,
+            column_index=cond._internal.column_index,
```
HyukjinKwon: Can we remove this too?

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,248 @@ class  locomotion
...
+           A    B
+        0  0 -100
+        1  1 -200
+        2 -2 -300
+        3 -3 -400
+        4 -4 -500
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        sdf = cond._internal.sdf
```
HyukjinKwon: I think we should handle this case too when `cond` isn't a dataframe

databricks/koalas/frame.py
```diff
@@ -2042,6 +2042,248 @@ class  locomotion
...
+        for col in cond._internal.data_columns:
+            sdf = sdf.withColumn(col, ~F.col(col))
+
+        internal = cond._internal.copy(
```
HyukjinKwon: Shall we match this with:

```scala
        return DataFrame(self._internal.copy(
            sdf=sdf,
            column_scols=[scol_for(sdf, column) for column in self._internal.data_columns]))
```

?

itholic: thanks! i fixed that you commented 👍

HyukjinKwon:
Thanks, @itholic. Merged.

itholic:
@HyukjinKwon thanks :)

Issue #1010, opened 2019-11-05
name property of ks.Series([1,2,3]) is '0' not None


charlesdong1991:
Just notice a Series behaviour in a WIP PR,

```python
>>> s1 = ks.Series([1,2,3])
>>> s1.name
'0'
```
is this designed on purpose? i think in pandas this should be `None`.

HyukjinKwon:
I think it was because:

```python
>>> pd.DataFrame([1,2,3])
   0
0  1
1  2
2  3
```

However, for series, yes. I think we should fix. Maybe we should add an internal field name that's matched to `None` .. ?

cc @ueshin

itholic:
closed since it's resolved in #1712 .

PR #1019, opened 2019-11-06
[WIP] BUG: Fix name property output when name is not assigned in Series

charlesdong1991:
#1010

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -325,8 +325,7 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
             else:
                 if name is None:
                     self._has_no_name = True
-                else:
-                    self._has_no_name = False
+                self._has_no_name = False
                 s = pd.Series(
                     data=data, index=index, dtype=dtype, name=name, copy=copy, fastpath=fastpath)
             kdf = DataFrame(s)
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -323,13 +323,11 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
                 assert not fastpath
                 s = data
             else:
-                if name is None:
-                    self._has_no_name = True
-                self._has_no_name = False
                 s = pd.Series(
                     data=data, index=index, dtype=dtype, name=name, copy=copy, fastpath=fastpath)
             kdf = DataFrame(s)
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -323,6 +323,9 @@ def __init__(self, data=None, index=None, dtype=None, name=None, copy=False, fas
                 assert not fastpath
                 s = data
             else:
+                if name is None:
+                    self._has_no_name = True
+                self._has_no_name = False
```
ueshin: I don't think we should have this attribute. e.g.,

```py
>>> kser = ks.Series([1,2,3])
>>> kser.name = 'foo'
>>> kser
???  # maybe always None
```

How about just setting name for `self` after `IndexOpsMixin.__init__(...)`?

charlesdong1991: ahh, definitely better!!

databricks/koalas/series.py
```diff
@@ -327,6 +327,7 @@ def __init__(self, data=None, index=None, dtype=None, name=None, copy=False, fas
                     data=data, index=index, dtype=dtype, name=name, copy=copy, fastpath=fastpath)
             kdf = DataFrame(s)
             IndexOpsMixin.__init__(self, kdf._internal.copy(scol=kdf._internal.data_scols[0]), kdf)
+            self.name = name
```
ueshin: We need to think about the case:

```py
s = ks.Series(pd.Series([1,2,3], name='foo'))
```

maybe `self.name = s.name`?

databricks/koalas/series.py
```diff
@@ -891,7 +892,7 @@ def shape(self):
         return len(self),
 
     @property
-    def name(self) -> Union[str, Tuple[str, ...]]:
+    def name(self) -> Union[str, Tuple[str, ...], None]:
```
HyukjinKwon: I think it should be `Optional[Union[str, Tuple[str, ...]]`

HyukjinKwon: We can change the type hint for `@name.setter` below as well.

charlesdong1991:
ah, okay, thanks for both of your comments, will take a closer look later after work! @ueshin @HyukjinKwon

ueshin:
We can have a `None` as the name for `Series` since we have an actual Spark column in `column_scols` (previously `data_columns`), but we might need to check whether all APIs are working with it as the same as when we support `column_index`.

itholic:
closed since it's resolved in #1712 .

PR #1015, opened 2019-11-06
Implement groupby.expanding.sum, min, max, mean in Series and DataFrame

HyukjinKwon:
This PR implement groupby.expanding.sum, min, max, mean in Series and DataFrame:

```python
>>> import databricks.koalas as ks
>>> kser = ks.Series([2, 3, float("nan"), 10])
>>> kser.groupby(kser).expanding().sum()
0
3.0   1     3.0
2.0   0     2.0
10.0  3    10.0
Name: 0, dtype: float64
```

```python
>>> df = kser.to_frame()
>>> df.groupby(df['0']).expanding().sum()
          0
0
3.0  1   3.0
2.0  0   2.0
10.0 3  10.0
```

PR #1023, opened 2019-11-07
Keeps column index in groupby.expanding()

HyukjinKwon:
Small followup for #1015


HyukjinKwon:
I checked the tests passed manually. I will make it passed at https://github.com/databricks/koalas/pull/1022

PR #1025, opened 2019-11-07
Clean up and handle min_periods properly in rolling of Series and Frame

HyukjinKwon:
- Avoid to use `lag` which refers the value in the offset. `min_periods` refers the number of values.
- Use `row_number` instead to count the number of values properly as many as `min_periods`.
- Match the documentation with `Expanding`
- Fix ordering of defined APIs to match with `Expanding`.


HyukjinKwon:
cc @itholic since you're working on this. Let's match to `Expanding` implementation so that we can deduplicate both codes at `_RollingAndExpanding`.

itholic:
@HyukjinKwon okay i got it :)

HyukjinKwon:
I'm merging it. Still I am touching this code ..

PR #1026, opened 2019-11-07
Use Python 3.7 in Read the Docs build (with Conda)

HyukjinKwon:
Many packages seem not supported with Python 3.8 yet. This PR aims to set the Python version to 3.7 for now in Read the Docs build.

Note that we should switch back to pip once Spark 3.0 is released because we use Conda only because of JDK 8.

Current pip configuration in `.readthedocs.yml` seems being ignored due to Conda configuration.

Related to https://github.com/readthedocs/readthedocs.org/issues/6350

### DIFF
--- a/dev/readthedocs-environment.yml
+++ b/dev/readthedocs-environment.yml
@@ -5,9 +5,11 @@
 channels:
   - conda-forge
 dependencies:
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
I tested against another branch. Merging it.

HyukjinKwon:
cc @ueshin FYI ...

PR #1029, opened 2019-11-08
Fix comparison operators to treat NULL as False

### LINKED ISSUE #999: Applying a logistic operator on NaN returns None instead of Bool
```python
import pandas as pd
import numpy as np

##### pandas #####

pser = pd.Series([0, 1, 2, np.nan])
>>> pser
0    0.0
1    1.0
2    2.0
3    NaN
dtype: float64

>>> pser == 0
0     True
1    False
2    False
3    False  <- bool
dtype: bool

##### koalas #####

kser = ks.from_pandas(pser)
>>> kser
0    0.0
1    1.0
2    2.0
3    NaN
Name: 0, dtype: float64

>>> kser == 0
0     True
1    False
2    False
3     None  <- not bool
Name: 0, dtype: object
```

harupy:
This PR proposes to:
- Fix column comparison operators to treat NULL as False
- Resolve #999.

## pandas
pandas treats NULL as False
```python
>>> pser
0    0.0
1    1.0
2    2.0
3    NaN
dtype: float64

>>> pser == 0
0     True
1    False
2    False
3    False  <- bool
dtype: bool
```

## Koalas
Koalas currently treats NULL as NULL
```python
>>> kser
0    0.0
1    1.0
2    2.0
3    NaN
Name: 0, dtype: float64

# CURRENT
>>> kser == 0
0     True
1    False
2    False
3     None  <- not bool
Name: 0, dtype: object

# PROPOSED
>>> kser == 0
0     True
1    False
2    False
3    False  <- bool
Name: 0, dtype: bool
```

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -60,7 +60,7 @@ def _column_op(f):
             # check if `f` is a comparison operator
             comp_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
             is_comp_op = any(f == getattr(spark.Column, '__{}__'.format(comp_op))
-                            for comp_op in comp_ops)
+                             for comp_op in comp_ops)
 
             if is_comp_op:
                 filler = f == spark.Column.__ne__
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -197,7 +197,7 @@ class IndexOpsMixin(object):
     __pow__ = _column_op(spark.Column.__pow__)
     __rpow__ = _column_op(spark.Column.__rpow__)
 
-    # logistic operators
+    # comparison operators
     __eq__ = _column_op(spark.Column.__eq__)
     __ne__ = _column_op(spark.Column.__ne__)
     __lt__ = _column_op(spark.Column.__lt__)
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -61,7 +61,8 @@ def _column_op(f):
             log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
             is_log_op = any(f == getattr(spark.Column, f'__{log_op}__') for log_op in log_ops)
             if is_log_op:
-                scol = F.when(scol.isNull(), False).otherwise(scol)
+                filler = f != spark.Column.__ne__
+                scol = F.when(scol.isNull(), filler).otherwise(scol)
 
             return self._with_new_scol(scol)
         else:
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -57,13 +57,20 @@ def _column_op(f):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
 
-            # If f is a logistic operator, fill NULL with False
+            # If f is a logistic operator, fill NULL
             log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
             is_log_op = any(f == getattr(spark.Column, '__{}__'.format(log_op))
                             for log_op in log_ops)
+
             if is_log_op:
                 filler = f == spark.Column.__ne__
                 scol = F.when(scol.isNull(), filler).otherwise(scol)
+
+            elif f == spark.Column.__or__:
+                scol = F.when(self._scol.isNull() | scol.isNull(), False).otherwise(scol)
+
+            elif f == spark.Column.__and__:
+                scol = F.when(scol.isNull(), False).otherwise(scol)
 
             return self._with_new_scol(scol)
         else:
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -57,7 +57,7 @@ def _column_op(f):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
 
-            # If f is a logistic operator, fill NULL
+            # check if f is a logistic operator
             log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
             is_log_op = any(f == getattr(spark.Column, '__{}__'.format(log_op))
                             for log_op in log_ops)
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -61,7 +61,7 @@ def _column_op(f):
             log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
             is_log_op = any(f == getattr(spark.Column, f'__{log_op}__') for log_op in log_ops)
             if is_log_op:
-                filler = f != spark.Column.__ne__
+                filler = f == spark.Column.__ne__
                 scol = F.when(scol.isNull(), filler).otherwise(scol)
 
             return self._with_new_scol(scol)
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -59,7 +59,8 @@ def _column_op(f):
 
             # If f is a logistic operator, fill NULL with False
             log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
-            is_log_op = any(f == getattr(spark.Column, f'__{log_op}__') for log_op in log_ops)
+            is_log_op = any(f == getattr(spark.Column, '__{}__'.format(log_op))
+                            for log_op in log_ops)
             if is_log_op:
                 filler = f == spark.Column.__ne__
                 scol = F.when(scol.isNull(), filler).otherwise(scol)
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -57,6 +57,12 @@ def _column_op(f):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
 
+            # If f is a logistic operator, fill NULL with False
+            log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
+            is_log_op = any(f == getattr(spark.Column, f'__{log_op}__') for log_op in log_ops)
+            if is_log_op:
+                scol = F.when(scol.isNull(), False).otherwise(scol)
+
             return self._with_new_scol(scol)
         else:
             # Different DataFrame anchors
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -57,12 +57,12 @@ def _column_op(f):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
 
-            # check if f is a logistic operator
-            log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
-            is_log_op = any(f == getattr(spark.Column, '__{}__'.format(log_op))
-                            for log_op in log_ops)
+            # check if `f` is a comparison operator
+            comp_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
+            is_comp_op = any(f == getattr(spark.Column, '__{}__'.format(comp_op))
+                            for comp_op in comp_ops)
 
-            if is_log_op:
+            if is_comp_op:
                 filler = f == spark.Column.__ne__
                 scol = F.when(scol.isNull(), filler).otherwise(scol)
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -749,11 +749,11 @@ class DataFrame(_Frame, Generic[T]):
         ...                   index=['a', 'b', 'c', 'd'], columns=['a', 'b'])
 
         >>> df.eq(1)
-               a     b
-        a   True  True
-        b  False  None
-        c  False  True
-        d  False  None
+               a      b
+        a   True   True
+        b  False  False
+        c  False   True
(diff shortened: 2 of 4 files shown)


databricks/koalas/base.py
```diff
@@ -57,6 +57,12 @@ def wrapper(self, *args):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
 
+            # If f is a logistic operator, fill NULL with False
+            log_ops = ['eq', 'ne', 'lt', 'le', 'ge', 'gt']
+            is_log_op = any(f == getattr(spark.Column, f'__{log_op}__') for log_op in log_ops)
+            if is_log_op:
+                scol = F.when(scol.isNull(), False).otherwise(scol)
+
```
harupy: @HyukjinKwon @ueshin (cc @itholic @charlesdong1991 )
Not sure if this is the right implementation ...

itholic: maybe `scol = F.when(scol.isNull(), False).otherwise(True)` ?

itholic: Ah, and i think maybe it should be better to make test of this if you available  😄

harupy: @itholic Thanks! I haven't fixed the tests yet because I want to make sure first if this is the right implementation.

harupy:
One of the failed tests. It seems to be working.
```
_________________ [doctest] databricks.koalas.series.Series.eq _________________
532         ...                   index=['a', 'b', 'c', 'd'], columns=['a', 'b'])
533 
534         >>> df.a == 1
535         a     True
536         b    False
537         c    False
538         d    False
539         Name: a, dtype: bool
540 
541         >>> df.b.eq(1)
    @@ -1,5 +1,5 @@
    -a    True
    -b    None
    -c    True
    -d    None
    -Name: b, dtype: object
    +a     True
    +b    False
    +c     True
    +d    False
    +Name: b, dtype: bool
```
https://travis-ci.com/databricks/koalas/jobs/254303993#L2298

databricks/koalas/base.py
ueshin: Actually the problem is not that simple. E.g.,:

```py
>>> pser == 0
0     True
1    False
2    False
3    False
dtype: bool
>>> pser != 0
0    False
1     True
2     True
3     True
dtype: bool
```

and

```py
>>> ks.from_pandas(pser) == 0
0     True
1    False
2    False
3     None
Name: 0, dtype: object
>>> ks.from_pandas(pser) != 0
0    False
1     True
2     True
3     None
Name: 0, dtype: object
```

If we simply convert `None` to `False`, the `!=` case is not the same as pandas'.

We might need to add some logic one-by-one.

itholic: my pleasure :)

itholic: @ueshin 
ah, it looks way complicated than i thought. 😅 
thanks for the comment!

harupy: @ueshin Thanks! Is there any case where the current way doesn't work?

ueshin: Speaking of the behavior difference,

```py
>>> ~pd.Series([True, False, np.nan])
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pandas/core/generic.py", line 1465, in __invert__
    arr = operator.inv(com.values_from_object(self))
TypeError: bad operand type for unary ~: 'float'
>>> ~pd.Series([True, False, None])
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pandas/core/generic.py", line 1465, in __invert__
    arr = operator.inv(com.values_from_object(self))
TypeError: bad operand type for unary ~: 'NoneType'
```

whereas:

```py
>>> ~ks.from_pandas(pd.Series([True, False, np.nan]))
0    False
1     True
2     None
Name: 0, dtype: object
>>> ~ks.from_pandas(pd.Series([True, False, None]))
0    False
1     True
2     None
Name: 0, dtype: object
```

I'm not sure we should raise an error for this.

ueshin: This is funny.

```py
>>> pd.Series([True, True, np.nan]) | pd.Series([True, np.nan, True])
0     True
1     True
2    False
dtype: bool
```

ueshin: hm, let's focus only `['eq', 'ne', 'lt', 'le', 'ge', 'gt']` here.
I roughly checked the behavior and seems like `ne` is only an exception.

harupy: > This is funny.
> 
> ```python
> >>> pd.Series([True, True, np.nan]) | pd.Series([True, np.nan, True])
> 0     True
> 1     True
> 2    False
> dtype: bool
> ```
> 
> Why 1 and 2 are different result..

This should be reported to pandas 😅

ueshin: hmm

```py
>>> pser < np.nan
0    False
1    False
2    False
3    False
dtype: bool
>>> pser > np.nan
0    False
1    False
2    False
3    False
dtype: bool
>>>
>>> pser < None
0    False
1    False
2    False
3    False
dtype: bool
>>> pser > None
0    False
1    False
2    False
3    False
dtype: bool
```

whereas

```py
>>> ks.from_pandas(pser) < np.nan
0    True
1    True
2    True
3    None
Name: 0, dtype: object
>>> ks.from_pandas(pser) > np.nan
0    False
1    False
2    False
3     None
Name: 0, dtype: object
>>> ks.from_pandas(pser) < None
0    None
1    None
2    None
3    None
Name: 0, dtype: object
>>> ks.from_pandas(pser) > None
0    None
1    None
2    None
3    None
Name: 0, dtype: object
```

charlesdong1991: > This is funny.
> 
> ```python
> >>> pd.Series([True, True, np.nan]) | pd.Series([True, np.nan, True])
> 0     True
> 1     True
> 2    False
> dtype: bool
> ```
> 
> Why 1 and 2 are different result..

i think this is because python, `nan's` don't compare equal, and pandas uses this fact, so it will turn `False` no matter what you compare with `np.nan`, so in this case, 2 will return `False`, the 1 will return `True` is because, in `|` operator in Python, if the stuff on the left is already `True`, python will directly return `True` and not continue to compare with the stuff on the right of `|`.

harupy: @charlesdong1991 
Does Python `|` operator work that way?

harupy: @charlesdong1991 I wrote a simple test.
```python
import time

def sleep_10_seconds():
  time.sleep(10)
  return False

# does evaluates sleep_10_seconds
if True | sleep_10_seconds():
  print('yes')

# doesn't evaluates sleep_10_seconds
if True or sleep_10_seconds():
  print('yes')
```

charlesdong1991: @harupy i believe so.

the `|` operator will return the first object that evaluates to true or the last object in the expression, regardless of its truth value.

charlesdong1991: oh, i thought `|` is functionally equivalent to `or`

charlesdong1991: then it's a bit wired, and I need to take back what I said before 😅

harupy: What matters here is how pandas.Series' `|` operator work.

charlesdong1991: but the point is `|` is native python thing, not pandas thing.

In pandas, you could do `s1.eq(s2)` to compare, in the case @ueshin showed above, this is then
```python
>>>s1 = pd.Series([True, True, np.nan])
>>>s2 = pd.Series([True, np.nan, True])
>>>s1.eq(s2)
0     True
1    False
2    False
dtype: bool
```

charlesdong1991: maybe worth reporting an issue in pandas 😄

harupy: @ueshin @charlesdong1991 

https://github.com/pandas-dev/pandas/blob/master/pandas/tests/series/test_operators.py#L447-L452

```python
# True | np.nan => True
exp_or1 = pd.Series([True, True, True, False], index=list("ABCD"), name="x")
tm.assert_series_equal(s1 | s2, exp_or1)
# np.nan | True => np.nan, filled with False
exp_or = pd.Series([True, True, False, False], index=list("ABCD"), name="x")
tm.assert_series_equal(s2 | s1, exp_or)
```

charlesdong1991: ahh, then it behaves this way on purpose in pandas. thanks for looking at source code to clear things up @harupy

ueshin:
What about https://github.com/databricks/koalas/pull/1029#discussion_r343992874?
Should we also address here? cc @HyukjinKwon

databricks/koalas/base.py
ueshin: Let's discuss `&` and `|` and address if needed in a separate issue/PR.

harupy: I think `&` and `|` should be addressed in a separate PR.

harupy:
@HyukjinKwon @ueshin 
This change makes `koalas` behave differently from `pyspark`. Is that really ok?

ueshin:
Actually I'm not quite sure whether we should follow pandas' behavior here.
How about adding a config to switch the null semantics between pandas and SQL.

databricks/koalas/base.py
HyukjinKwon: Seems like pandas tries to follow the boolean operations rules in Python (`or` and `and`) - https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

```
# True | np.nan => True
# np.nan | True => np.nan
```

Then, ^ could make sense.

However, I am still kind of lost about "np.nan, filled with False" because `bool(np.nan)` is coerced to `True`, e.g.)

```python
>>> pd.Series([float("nan")]).astype("bool")
0    True
dtype: bool
```

HyukjinKwon:
How about a configuration between Python native and pandas?

HyukjinKwon:
https://github.com/databricks/koalas/pull/1029#discussion_r343992874 this behaivour at least looks consistent with Python native comparison.

HyukjinKwon:
I think it's fine to fix ... for now ...

harupy:
@HyukjinKwon 
> I think it's fine to fix ... for now ...

We don't need to fix this weird behavior for now?

HyukjinKwon:
I meant let's fix to be consistent with pandas/Python comparison for now since that's basically what Koalas needs to do.

harupy:
@HyukjinKwon ok, which one should I address, python or pandas?

HyukjinKwon:
For `['eq', 'ne', 'lt', 'le', 'ge', 'gt']`, my impression was that both are same. Are they different?

HyukjinKwon:
Let's stick to pandas for now if they are different ... maybe we should have a config as @ueshin said but I think it's fine for now ...

harupy:
@HyukjinKwon 
> For ['eq', 'ne', 'lt', 'le', 'ge', 'gt'], my impression was that both are same. Are they different?

No, they are same.

HyukjinKwon:
Let's stick to pandas for now.

harupy:
Is `logistic operators` correct name? It should be `comparison operators`, shouldn't it?
https://github.com/databricks/koalas/blob/e469d540d8a14059a2706d8f2fdc4afcca9be747/databricks/koalas/base.py#L185-L190

https://github.com/databricks/koalas/blob/master/databricks/koalas/base.py#L185

HyukjinKwon:
Yup, I think so.

harupy:
Fixed it

databricks/koalas/base.py
```diff
@@ -57,6 +57,21 @@ def wrapper(self, *args):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
 
+            # check if f is a logistic operator
```
HyukjinKwon: I would change the comment and variable names too :-).

harupy: Thanks! will fix them.

HyukjinKwon:
Merged to master.

HyukjinKwon:
Thanks, @harupy

harupy:
@HyukjinKwon thanks!

databricks/koalas/base.py
```diff
@@ -57,6 +57,21 @@ def wrapper(self, *args):
...
+
+            elif f == spark.Column.__or__:
+                scol = F.when(self._scol.isNull() | scol.isNull(), False).otherwise(scol)
+
+            elif f == spark.Column.__and__:
+                scol = F.when(scol.isNull(), False).otherwise(scol)
+
             return self._with_new_scol(scol)
         else:
             # Different DataFrame anchors
```
ueshin: Seems like we should apply the updates for this branch as well?

```py
>>> s1 = pd.Series([True, False, True], index=list("ABC"), name="x")
>>> s2 = pd.Series([True, True, False], index=list("ABD"), name="x")
>>> s1
A     True
B    False
C     True
Name: x, dtype: bool
>>> s2
A     True
B     True
D    False
Name: x, dtype: bool
>>> s1 | s2
A     True
B     True
C     True
D    False
Name: x, dtype: bool
>>> (ks.from_pandas(s1) | ks.from_pandas(s2)).sort_index()
A    True
B    True
C    True
D    None
Name: x, dtype: object
```

harupy: @ueshin Thank you for catching this. I'm working on this.

PR #1045, opened 2019-11-13
Fix different index level assignment when 'compute.ops_on_diff_frames' is enabled

HyukjinKwon:
```python
import pandas as pd
ks.options.compute.ops_on_diff_frames=True
pdf = pd.DataFrame({'b': [1,2,3], 'c': [4,5,6]})
columns = pd.MultiIndex.from_tuples([('a', 'b'), ('a', 'c')])
pdf.columns = columns
kdf = ks.DataFrame(pdf)
kdf['c'] = ks.Series([1,2,3])
kdf
```

Before:

```
   a     c
   b  c    NaN
0  1  4  1   1
1  2  5  1   2
2  3  6  1   3
```

After:

```
   a     c
   b  c
0  1  4  1
1  2  5  2
2  3  6  3
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7801,6 +7801,9 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             else:
                 key = [k if isinstance(k, tuple) else (k,) for k in key]
 
+            level = self._internal.column_index_level
+            key = [tuple(list(idx) + ([''] * (level - len(idx)))) for idx in key]
+
             def assign_columns(kdf, this_column_index, that_column_index):
                 assert len(key) == len(that_column_index)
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -7801,6 +7801,9 @@ def __setitem__(self, key, value):
             else:
                 key = [k if isinstance(k, tuple) else (k,) for k in key]
 
+            level = self._internal.column_index_level
+            key = [tuple(list(idx) + ([''] * (level - len(idx)))) for idx in key]
```
HyukjinKwon: Same as https://github.com/databricks/koalas/blob/master/databricks/koalas/frame.py#L3443-L3444

HyukjinKwon:
Merging it for today's release

PR #1050, opened 2019-11-17
Implements head() for DataFrameGroupBy & SeriesGroupBy

### LINKED ISSUE #879: GroupBy.head
Implement `GroupBy.head`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.GroupBy.head.html

itholic:
Resolves #879 
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.GroupBy.head.html#pandas.core.groupby.GroupBy.head)

```python
>>> df = ks.DataFrame({'a': [1, 1, 1, 1, 2, 2, 2, 3, 3, 3],
...                    'b': [2, 3, 1, 4, 6, 9, 8, 10, 7, 5],
...                    'c': [3, 5, 2, 5, 1, 2, 6, 4, 3, 6]},
...                   columns=['a', 'b', 'c'],
...                   index=[7, 2, 4, 1, 3, 4, 9, 10, 5, 6])
>>> df
    a   b  c
7   1   2  3
2   1   3  5
4   1   1  2
1   1   4  5
3   2   6  1
4   2   9  2
9   2   8  6
10  3  10  4
5   3   7  3
6   3   5  6

>>> df.groupby('a').head(2)
    a   b  c
7   1   2  3
2   1   3  5
10  3  10  4
5   3   7  3
3   2   6  1
4   2   9  2

>>> df.groupby('a')['b'].head(2)
7      2
2      3
10    10
5      7
3      6
4      9
Name: b, dtype: int64
```


databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,66 @@ def ffill(self, limit=None):
...
+        7      2
+        2      3
+        10    10
+        5      7
+        3      6
+        4      9
+        Name: b, dtype: int64
+        """
+        groupkeys = self._groupkeys
+        sdf = self._kdf._sdf.withColumn('order', monotonically_increasing_id())
```
ueshin: Shall we use `__order__` or something to avoid conflicts with user columns?

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,66 @@ def ffill(self, limit=None):
...
+        name = self._agg_columns[0]._internal.data_columns[0]
+        window = Window.partitionBy([s._scol for s in groupkeys]).orderBy(F.col('order'))
+        sdf = sdf.withColumn('order', F.row_number().over(window)).filter(F.col('order') <= n)
+        internal = _InternalFrame(
+            sdf=sdf,
+            index_map=self._kdf._internal.index_map,
+            column_scols=([scol_for(sdf, name)]
+                          if isinstance(self, SeriesGroupBy)
+                          else [scol_for(sdf, name) for name in self._kdf.columns]))
```
ueshin: `self._kdf._internal.data_columns` instead of `self._kdf.columns`?

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,66 @@ def ffill(self, limit=None):
...
+        2      3
+        10    10
+        5      7
+        3      6
+        4      9
+        Name: b, dtype: int64
+        """
+        groupkeys = self._groupkeys
+        sdf = self._kdf._sdf.withColumn('order', monotonically_increasing_id())
+        name = self._agg_columns[0]._internal.data_columns[0]
```
HyukjinKwon: Hm, does this cover the case when there's multiple groupby columns? (e.g., `groupby(by=["a", "b"]`)

itholic: yep, it can cover 😃 i just added test case

itholic: for more your information, this `name` here will be used when type of `self` is `SeriesGroupBy`, otherwise `name` is re-generated by `[scol_for(sdf, name) for name in self._kdf._internal.data_columns]` in line 1461

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,67 @@ def ffill(self, limit=None):
...
+        sdf = self._kdf._sdf.withColumn(tmp_col, monotonically_increasing_id())
+        name = self._agg_columns[0]._internal.data_columns[0]
+        window = Window.partitionBy([s._scol for s in groupkeys]).orderBy(F.col(tmp_col))
+        sdf = sdf.withColumn(tmp_col, F.row_number().over(window)).filter(F.col(tmp_col) <= n)
+        internal = _InternalFrame(
+            sdf=sdf,
+            index_map=self._kdf._internal.index_map,
+            column_scols=([scol_for(sdf, name)]
+                          if isinstance(self, SeriesGroupBy)
+                          else [scol_for(sdf, name) for name in self._kdf._internal.data_columns]))
```
itholic: @HyukjinKwon for more your information related with https://github.com/databricks/koalas/pull/1050#discussion_r347227629, that `name` will be used when type of `self` is `SeriesGroupBy`, otherwise the `name` is re-generated by `[scol_for(sdf, name) for name in self._kdf._internal.data_columns]` here

databricks/koalas/groupby.py
itholic: @ueshin Thanks for the comment!

databricks/koalas/groupby.py
HyukjinKwon: Let's move all related codes into where `name` variable exists. Looks difficult to follow.

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,67 @@ def ffill(self, limit=None):
...
+        7      2
+        2      3
+        10    10
+        5      7
+        3      6
+        4      9
+        Name: b, dtype: int64
+        """
+        groupkeys = self._groupkeys
+        tmp_col = '__order__'
```
HyukjinKwon: I would name it `__natural_order__`

HyukjinKwon: Actually, can't we just use `row_number`?

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,67 @@ def ffill(self, limit=None):
...
+        sdf = self._kdf._sdf.withColumn(tmp_col, monotonically_increasing_id())
+        name = self._agg_columns[0]._internal.data_columns[0]
+        window = Window.partitionBy([s._scol for s in groupkeys]).orderBy(F.col(tmp_col))
+        sdf = sdf.withColumn(tmp_col, F.row_number().over(window)).filter(F.col(tmp_col) <= n)
```
HyukjinKwon: `monotonically_increasing_id()` here wouldn't work if there are multiple partitions because `monotonically_increasing_id()` isn't sequential:

```
scala> spark.range(100).repartition(200).select(monotonically_increasing_id()).show()
+-----------------------------+
|monotonically_increasing_id()|
+-----------------------------+
|                 103079215104|
|                 111669149696|
|                 120259084288|
...
```

HyukjinKwon: I think we can just use `row_number`

itholic: doesn't function `head` pick the top N values without sort the order?

if we use `row_number`, i think we have to use it with `orderBy`,

then, the existing order is sorted so that we couldn't get the proper result?

For example, we have dataframe like this.

```python
>>> df = ks.DataFrame({'a': [1, 1, 1, 1, 2, 2, 2, 3, 3, 3],
...                    'b': [2, 3, 1, 4, 6, 9, 8, 10, 7, 5],
...                    'c': [3, 5, 2, 5, 1, 2, 6, 4, 3, 6]},
...                   columns=['a', 'b', 'c'],
...                   index=[7, 2, 4, 1, 3, 4, 9, 10, 5, 6])

>>> df
    a   b  c
7   1   2  3
2   1   3  5
4   1   1  2
1   1   4  5
3   2   6  1
4   2   9  2
9   2   8  6
10  3  10  4
5   3   7  3
6   3   5  6
```

we expect the result of `df.groupby('a').head(2)` like below.

```python
>>> df.groupby('a').head(2)
    a   b  c
7   1   2  3
2   1   3  5
3   2   6  1
4   2   9  2
10  3  10  4
5   3   7  3
```

but if we use `row_number`, maybe the result will be shown like below.

```python
   a  b  c
4  1  1  2
7  1  2  3
6  3  5  6
5  3  7  3
3  2  6  1
9  2  8  6
```

is this result seems okay??

HyukjinKwon: It's already sorted by `monotonically_increasing_id` so using `row_number` won't make much differences. We cannot use `F.col(tmp_col) <= n` as is anyway because `tmp_col` will be bigger than `n`.

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,67 @@ def ffill(self, limit=None):
...
+        5      7
+        3      6
+        4      9
+        Name: b, dtype: int64
+        """
+        groupkeys = self._groupkeys
+        tmp_col = '__order__'
+        sdf = self._kdf._sdf.withColumn(tmp_col, monotonically_increasing_id())
+        name = self._agg_columns[0]._internal.data_columns[0]
+        window = Window.partitionBy([s._scol for s in groupkeys]).orderBy(F.col(tmp_col))
```
HyukjinKwon: Can we don't use `tmp_col` in this please but do `.orderBy(F.col(monotonically_increasing_id()))` directly?

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,67 @@ def ffill(self, limit=None):
...
+        3      6
+        4      9
+        Name: b, dtype: int64
+        """
+        groupkeys = self._groupkeys
+        tmp_col = '__row_number__'
+        sdf = self._kdf._sdf
+        window = Window.partitionBy([s._scol for s in groupkeys]) \
+                       .orderBy(F.monotonically_increasing_id())
+        sdf = sdf.withColumn(tmp_col, F.row_number().over(window)).filter(F.col(tmp_col) <= n)
```
HyukjinKwon: I think we dont need the temp column. You can just use row_number at filter. We should avoid to use the temp column to avoid the name conflict although this a reserved column name in Koalas.

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,67 @@ def ffill(self, limit=None):
...
+        internal = _InternalFrame(
+            sdf=sdf,
+            index_map=self._kdf._internal.index_map,
+            column_scols=([scol_for(sdf, self._agg_columns[0]._internal.data_columns[0])]
```
HyukjinKwon: Can we move this up? Making all into single line seems difficult to read.

databricks/koalas/groupby.py
itholic: i have tried, but we can't use `F.row_number()` directly at filter in this case, since we have to use window function for partition by group which is not supported in filter clause.

```python
>>> sdf.filter(F.row_number().over(window))
pyspark.sql.utils.AnalysisException: 'It is not allowed to use window functions inside WHERE and HAVING clauses;'
```

so i tried like below, but they need also temporal column.
```python
>>> sdf.select("*", F.row_number().over(window).alias('__row_num__')).filter(F.col('__row_num__') < 5)
```

so finally, i think there is only way to avoid using `tmp_col` is using `F.row_number` in `when` clause, after then filter out the null value with `dropna()` which we're doing latest PR.

```python
sdf.select("*", F.when(F.row_number().over(window) <= n, True)).dropna()
```

what are you think?

ueshin: Shall we use the temp column if we can't use `F.row_number()` directly at filter?
It's hard to read.

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,70 @@ def ffill(self, limit=None):
...
+
+        if isinstance(self, DataFrameGroupBy):
+            internal = self._kdf._internal.copy(
+                sdf=sdf,
+                column_scols=[scol_for(sdf, col) for col in self._kdf._internal.data_columns])
+        else:
+            internal = self._kdf._internal.copy(
+                sdf=sdf,
+                column_index=[(self._kser.name,)],
+                column_scols=[scol_for(sdf, self._kser.name)])
```
ueshin: `self._kser._internal.data_columns[0]` instead of `self._kser.name`?
Those are different concepts.
The one is internal metadata, the other is external data. Please distinguish them.

databricks/koalas/tests/test_groupby.py
```diff
@@ -810,6 +810,15 @@ def test_idxmin(self):
...
                        kdf.groupby(('x', 'a')).idxmin(skipna=False).sort_index())
 
+    def test_head(self):
+        pdf = pd.DataFrame({'a': [1, 1, 2, 2, 3],
+                            'b': [1, 2, 3, 4, 5],
+                            'c': [5, 4, 3, 2, 1]}, columns=['a', 'b', 'c'])
+        kdf = ks.from_pandas(pdf)
+
+        self.assert_eq(pdf.groupby(['a', 'b']).head(2),
+                       kdf.groupby(['a', 'b']).head(2).sort_index())
```
ueshin: Could you add more tests? e.g., multi-index, multi-index columns, tuple named Series, large n, small n, negative n, random order index, etc.

databricks/koalas/groupby.py
itholic: Thanks for the comment!
although they have same value, we should distinguish them to keep concept.
am i understood right?

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,70 @@ def ffill(self, limit=None):
...
+        window = Window.partitionBy([s._scol for s in groupkeys]) \
+                       .orderBy(F.monotonically_increasing_id())
+        sdf = sdf.select("*", F.when(F.row_number().over(window) <= n, True)).dropna()
+
+        if isinstance(self, DataFrameGroupBy):
+            internal = self._kdf._internal.copy(
+                sdf=sdf,
+                column_scols=[scol_for(sdf, col) for col in self._kdf._internal.data_columns])
+        else:
+            internal = self._kdf._internal.copy(
```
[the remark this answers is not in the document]
itholic: thanks for the comment!
i'm not pretty sure about why do we use explicitly `scol=None` ?
maybe just `self._kser._internal.copy(sdf=sdf)` is enough?

databricks/koalas/tests/test_groupby.py
itholic: it sounds good, i'll add 😄

itholic: pandas seems doesn't support groupby for tuple-named Series like below.

```python
>>> pser = pd.Series([1, 2, 3, 4, 5], name=('i', 'am', 'a', 'series'))
>>> pser.groupby(pser.name).head(2)
KeyError: ('i', 'am', 'a', 'series')
```

but it works at koalas, but result is not correct.

```python
>>> kser = ks.Series([1, 2, 3, 4, 5], name=('i', 'am', 'a', 'series'))
>>> kser.groupby(kser.name).head(2)
4    5
0    1
2    3
1    2
3    4
Name: (i, am, a, series), dtype: int64
```

shall we make it work? or raises `KeyError` like pandas?

ueshin: It's not the correct usage anyway.

```py
>>> import pandas as pd
>>> pser = pd.Series([1, 2, 3, 4, 5], name='x')
>>> pser.groupby(pser.name).head(2)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pandas/core/generic.py", line 7632, in groupby
    observed=observed, **kwargs)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 2110, in groupby
    return klass(obj, by, **kwds)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pandas/core/groupby/groupby.py", line 360, in __init__
    mutated=self.mutated)
  File "/Users/ueshin/workspace/databricks-koalas/miniconda/envs/databricks-koalas_3.6/lib/python3.6/site-packages/pandas/core/groupby/grouper.py", line 578, in _get_grouper
    raise KeyError(gpr)
KeyError: 'x'
```

Koalas currently allows it but we might need to block it?

databricks/koalas/groupby.py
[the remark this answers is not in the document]
ueshin: We shouldn't use the internal with non-None scol for DataFrame.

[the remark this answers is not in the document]
itholic: @ueshin thanks!

ueshin:
Actually I'm feeling something strange with this implementation.
We need to address `scol` for `SeriesGroupBy`.
e.g.,

```py
>>> pdf.b.cumsum().groupby(pdf.a).head(2)
7      2
2      5
3     16
4     25
10    43
5     50
Name: b, dtype: int64
```

whereas (reorder by hand for comparison):

```py
>>> kdf.b.cumsum().groupby(kdf.a).head(2)
7     27
2      3
3      9
4     18
10    37
5     25
Name: b, dtype: int64
```

databricks/koalas/tests/test_groupby.py
itholic: @ueshin thanks for giving an opinion.

i also think we better block it also like pandas.

i'll fix them.

itholic:
@ueshin thanks 👍 . let me check :)

databricks/koalas/tests/test_groupby.py
itholic: i just open an issue(#1095 ) to handle this.

after quickly solve that, i'll come back here and go on.

HyukjinKwon:
Hm, that seems to be an issue in `cumsum`:

```python
>>> pdf.b.cumsum().sort_index()
1     10
2      5
3     16
4      6
4     25
5     50
6     55
7      2
9     33
10    43
Name: b, dtype: int64
>>> kdf.b.cumsum().sort_index()
1      4
2      7
3     13
4     14
4     23
5     30
6     35
7     37
9     45
10    55
Name: b, dtype: int64
```

HyukjinKwon:
Ah, so it's https://github.com/databricks/koalas/pull/1113

HyukjinKwon:
Okay, I tested that the results are same on the top of #1113.

databricks/koalas/groupby.py
```diff
@@ -1401,6 +1401,68 @@ def ffill(self, limit=None):
...
+        3      6
+        4      9
+        Name: b, dtype: int64
+        """
+        groupkeys = self._groupkeys
+        tmp_col = '__row_number__'
+        sdf = self._kdf._sdf
+        window = Window.partitionBy([s._scol for s in groupkeys]) \
+                       .orderBy(F.monotonically_increasing_id())
+        sdf = sdf.withColumn(tmp_col, F.row_number().over(window)).filter(F.col(tmp_col) <= n)
```
HyukjinKwon: Can we drop the temp column after this?

itholic: ah, it should be better. i've just fixed

PR #1052, opened 2019-11-18
Let's use 'kser' rather than 'ks' or 's' as variable name for Series 

itholic:
Since the `ks` is used as koalas' import naming convention we recommended like `from databricks.koalas as ks`

Let's don't use `ks` as variable name for Series, use `kser` instead for unity & readability of our codes and to avoid confusing.

Plus, i think maybe we better also avoid using of just `s` as possible for more readability.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2464,8 +2464,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         1  False   True  False
         """
         kdf = self.copy()
-        for name, ks in kdf.iteritems():
-            kdf[name] = ks.isnull()
+        for name, kser in kdf.iteritems():
+            kdf[name] = kser.isnull()
         return kdf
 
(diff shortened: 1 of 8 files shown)


itholic:
@ueshin @HyukjinKwon Could you please let me know your opinion after confirm this PR when you guys available? 😄

ueshin:
Shall we fix some more?
- `frame.py`: There are two `for name, ks in kdf.iteritems():`
- `test_dataframe.py`: There are some `right_ks`.

itholic:
@ueshin, @HyukjinKwon  Thanks for the comment! i just fix the missing ones.

ueshin:
Shall we fix too?

> `test_dataframe.py`: There are some `right_ks`.

itholic:
@ueshin Thanks for finding! :)

HyukjinKwon:
Merged to master.

PR #1054, opened 2019-11-19
Implements idxmax()/idxmin() for DataFrame

### LINKED ISSUE #1043: DataFrame.idxmax
Implement `DataFrame.idxmax`

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.idxmax.html

itholic:
Resolves #1043 

```python
>>> kdf = ks.DataFrame({'a': [1, 2, 3, 2],
...                     'b': [4.0, 2.0, 3.0, 1.0],
...                     'c': [300, 200, 400, 200]})
>>> kdf
   a    b    c
0  1  4.0  300
1  2  2.0  200
2  3  3.0  400
3  2  1.0  200

>>> kdf.idxmax()
a    2
b    0
c    2
Name: 0, dtype: int64
```

For Multi-column Index

```python
>>> kdf = ks.DataFrame({'a': [1, 2, 3, 2],
...                     'b': [4.0, 2.0, 3.0, 1.0],
...                     'c': [300, 200, 400, 200]})
>>> kdf.columns = pd.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
>>> kdf
   a    b    c
   x    y    z
0  1  4.0  300
1  2  2.0  200
2  3  3.0  400
3  2  1.0  200

>>> kdf.idxmax()
(a, x)    2
(b, y)    0
(c, z)    2
Name: 0, dtype: int64
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7522,7 +7522,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         return self.columns
 
     # TODO: axis = 1
-    def idxmax(self, axis=0, skipna=True):
+    def idxmax(self, axis=0):
         """
         Return index of first occurrence of maximum over requested axis.
         NA/null values are excluded.
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7572,24 +7572,28 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         2  3  3.0  400
         3  2  1.0  200
 
-        >>> kdf.idxmax()
-        (a, x)    2
-        (b, y)    0
-        (c, z)    2
+        >>> kdf.idxmax().sort_index()
+        a  x    2
+        b  y    0
+        c  z    2
         Name: 0, dtype: int64
         """
         from databricks.koalas.series import Series
         sdf = self._sdf
         max_cols = map(lambda x: F.max(x).alias(x), self._internal.data_columns)
         sdf_max = sdf.select(*max_cols)
+        # `sdf_max` looks like below
+        # +------+------+------+
+        # |(a, x)|(b, y)|(c, z)|
+        # +------+------+------+
+        # |     3|   4.0|   400|
+        # +------+------+------+
 
-        col_idxmax_dict = dict()
-        for column_name, max_val in zip(sdf_max.columns, sdf_max.head()):
-            idx_val = sdf.select(self._internal.index_columns) \
-                         .where(F.col(column_name) == max_val).head()[0]
-            col_idxmax_dict[column_name] = idx_val
+        max_vals = [sdf.select(self._internal.index_columns)
+                       .where(F.col(column_name) == max_val).head()[0]
+                    for column_name, max_val in zip(sdf_max.columns, sdf_max.head())]
 
-        return Series(col_idxmax_dict)
+        return Series(max_vals, index=self.columns)
 
     # TODO: axis = 1
     def idxmin(self, axis=0):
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -7521,6 +7521,142 @@ def keys(self):
...
+        (a, x)    2
+        (b, y)    0
+        (c, z)    2
+        Name: 0, dtype: int64
+        """
+        from databricks.koalas.series import Series
+        sdf = self._sdf
+        col_idxmax_dict = dict()
+        for column in self._internal.data_columns:
+            max_val = sdf.select(F.max(column)).head()[0]
```
HyukjinKwon: Can we avoid to execute multiple jobs? Seems it will execute multiple jobs as many as its data columns.

itholic: just fix the logic (2n -> n + 1)

databricks/koalas/frame.py
```diff
@@ -7521,6 +7521,146 @@ def keys(self):
...
+        (c, z)    2
+        Name: 0, dtype: int64
+        """
+        from databricks.koalas.series import Series
+        sdf = self._sdf
+        max_cols = map(lambda x: F.max(x).alias(x), self._internal.data_columns)
+        sdf_max = sdf.select(*max_cols)
+
+        col_idxmax_dict = dict()
+        for column_name, max_val in zip(sdf_max.columns, sdf_max.head()):
```
HyukjinKwon: We should deduplicate the logic with `Series.idx` such as https://github.com/databricks/koalas/pull/678 (although I don't think we need a separate private method like `_idx`).

Can we document that this will requires to sort for each column in a `... note ::`?

databricks/koalas/frame.py
```diff
@@ -7521,6 +7521,146 @@ def keys(self):
...
+        0  1  4.0  300
+        1  2  2.0  200
+        2  3  3.0  400
+        3  2  1.0  200
+
+        >>> kdf.idxmax()
+        (a, x)    2
+        (b, y)    0
+        (c, z)    2
+        Name: 0, dtype: int64
```
ueshin: This result is not correct.

```py
>>> pdf = pd.DataFrame({'a': [1, 2, 3, 2], 'b': [4.0, 2.0, 3.0, 1.0], 'c': [300, 200, 400, 200]})
>>> pdf.columns = pd.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
>>> pdf
   a    b    c
   x    y    z
0  1  4.0  300
1  2  2.0  200
2  3  3.0  400
3  2  1.0  200
>>> pdf.idxmax()
a  x    2
b  y    0
c  z    2
dtype: int64
```

The index should be a multi-index.

itholic: thanks i've fixed :)

databricks/koalas/frame.py
```diff
@@ -7764,6 +7764,148 @@ def keys(self):
...
+        # `sdf_max` looks like below
+        # +------+------+------+
+        # |(a, x)|(b, y)|(c, z)|
+        # +------+------+------+
+        # |     3|   4.0|   400|
+        # +------+------+------+
+
+        max_vals = [sdf.select(self._internal.index_columns)
+                       .where(F.col(column_name) == max_val).head()[0]
+                    for column_name, max_val in zip(sdf_max.columns, sdf_max.head())]
```
HyukjinKwon: My concern is that this is too expensive. It will requires to run jobs as many as columns + 1.

I think you can
1. create a condition with `column = max value` with OR.
2. collect everything (because the number of rows with max values are usually small in general). We might have to add a `.. note::` in the docstring.
3. Perform idmax or idmin against the result from 2.

itholic: if given `kdf` looks like below,

```python
>>> kdf
   a    b    c
   x    y    z
0  1  4.0  300
1  2  2.0  200
2  3  3.0  400
3  2  1.0  200
```

it's `idxmax()` should be 

```python
```

HyukjinKwon: I think you can just call `toPandas` and call `idxmax` or `idxmin` against it.

itholic: @HyukjinKwon ah, yes i got it. thanks for the comment :)

PR #1056, opened 2019-11-19
Implements transpose & add property T for Index/MultiIndex

itholic:
property `T` for Index & MultiIndex
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.T.html#pandas.Index.T)

i think `T` for Index/MultiIndex just returns itselves in pandas like below.

```python
>>> idx = pd.Index(['a', 'b', 'c'])
>>> idx

>>> idx.T

>>> idx is idx.T
True
```

For MultiIndex

```python
>>> midx = pd.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
>>> midx
MultiIndex([('a', 'x'),
            ('b', 'y'),
            ('c', 'z')],
           )

>>> midx.T
MultiIndex([('a', 'x'),
            ('b', 'y'),
            ('c', 'z')],
           )

>>> midx is midx.T
True
```


databricks/koalas/indexes.py
```diff
@@ -120,6 +120,37 @@ def size(self) -> int:
         """
         return len(self._kdf)  # type: ignore
 
+    @property
+    def T(self):
```
HyukjinKwon: Can you implement `transpose` and reuse it for `T`? `    T = property(transpose)`.

databricks/koalas/missing/indexes.py
```diff
@@ -30,7 +30,6 @@ def unsupported_property(property_name, deprecated=False, reason=""):
 class _MissingPandasLikeIndex(object):
 
     # Properties
-    T = unsupported_property('T')
```
HyukjinKwon: Should remove in `_MissingPandasLikeMultiIndex:` too

databricks/koalas/indexes.py
itholic: ah, it would be better. okay i go for it

databricks/koalas/missing/indexes.py
itholic: thanks for the comment!

databricks/koalas/indexes.py
itholic: ah, i just realized that pandas has no `transpose` for `Index`.
then maybe would it be better to keep this behavior??

HyukjinKwon: I think pandas has it ..

```python
>>> import pandas as pd
>>> pd.DataFrame({'a': [1,2,3]}).index.transpose()
RangeIndex(start=0, stop=3, step=1)
```

itholic: oops... then maybe they forgot to upload this to their docs.

스크린샷 2019-11-19 오후 3 50 51

thanks i'm going to add it !

docs/source/reference/indexing.rst
```diff
@@ -23,6 +23,8 @@ Properties
    Index.names
    Index.ndim
    Index.empty
+   Index.T
+   Index.transpose
```
HyukjinKwon: Ah, then let's exclude `transpose` in the API documentation to be consistent.

itholic: okay i just excluded them 👍

HyukjinKwon:
ah, sorry can you resolve conflicts @itholic ?

itholic:
@HyukjinKwon sure, i did 😄

Issue #873, opened 2019-10-03
Index.to_numpy


ueshin:
Implement `Index.to_numpy`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.to_numpy.html

ueshin:
This was resolved by #1058.

PR #1058, opened 2019-11-19
Implement Index.to_numpy

guyao:
#873  Implement Index.to_numpy
Since we already have `Index.to_pandas`, just directly use it.


databricks/koalas/indexes.py
```diff
@@ -145,6 +145,21 @@ def to_pandas(self) -> pd.Index:
 
     toPandas = to_pandas
 
+    def to_numpy(self, dtype=None, copy=False):
+        """
+        A NumPy ndarray representing the values in this Index or MultiIndex.
```
HyukjinKwon: Can you add a note:

```
        .. note:: This method should only be used if the resulting NumPy ndarray is expected
```

Also, can you please port related docs (e.g., Parameters, Returns) to here? (see https://github.com/pandas-dev/pandas/blob/v0.25.3/pandas/core/base.py#L916-L1010)

HyukjinKwon: We should also remove https://github.com/databricks/koalas/blob/master/databricks/koalas/missing/indexes.py#L100 and https://github.com/databricks/koalas/blob/master/databricks/koalas/missing/indexes.py#L100

Also, add this new API into https://github.com/databricks/koalas/blob/master/docs/source/reference/indexing.rst

HyukjinKwon:
Thanks for this PR @guyao. cc @itholic as well since you're working on Index related APIs.

itholic:
nice working @guyao , and thanks for the cc @HyukjinKwon 😃

databricks/koalas/indexes.py
guyao: Thanks!  Will do.

databricks/koalas/indexes.py
```diff
@@ -144,6 +144,36 @@ def to_pandas(self) -> pd.Index:
...
+        numpy.ndarray
+
+        Examples
+        --------
+        >>> ks.Series([1, 2, 3, 4]).index.to_numpy()
+        array([0, 1, 2, 3])
+        >>> ks.DataFrame({'a': ['a', 'b', 'c']}, index=[[1, 2, 3], [4, 5, 6]]).index.to_numpy()
+        array([(1, 4), (2, 5), (3, 6)], dtype=object)
+        """
+        return self.to_pandas().to_numpy(dtype=dtype, copy=copy)
```
HyukjinKwon: Seems like pandas 0.23 does not have this API. Can you check pandas's to_numpy and mimic it here?

If that's not possible or complicated, we should skip the doctests and document Koalas with pands 0.23 does not support this API.

guyao: Thanks for pointing out! Yes pandas add this api since `0.24`. 

It is not complicated to mimic. But personally I think some part of code is not stable. I see some WIP PR in the comment. 
https://github.com/pandas-dev/pandas/blob/v0.25.3/pandas/core/base.py#L916-L1010
Currently if index.dtype is datetime and parameter dtype is none, it would set it as object.

If call pandas directly we could have consistent behaviors with pandas. 

If mimic the it, I would do 

```python
        result = np.asarray(self.to_pandas()._values, dtype=dtype)
        if copy:
            result = result.copy()
        return result
```


Any suggestion?

HyukjinKwon: Thanks for pointing it out. Hm, let's just copy as is for now. I'm periodically checking the existing APIs and maybe I would be able to sync it again when pandas happens to have some diff later here.

HyukjinKwon:
Looks good otherwise if tests pass

HyukjinKwon:
Thanks @guyao. Merged to master.

PR #1059, opened 2019-11-19
Explicitly block Index ^ MultiIndex for Index.symmetric_difference

itholic:
Related with https://github.com/databricks/koalas/pull/953#issuecomment-555368991 and https://github.com/databricks/koalas/pull/953#issuecomment-555376690


itholic:
the reason of codecov failure is we skip the doctest related with MiltiIndex since it would be shown different result depends on python version.

PR #1060, opened 2019-11-20
Implements Top-level function to_numeric()

### LINKED ISSUE #1044: Top-level Pandas conversion function - to_numeric
Implement `pandas.to_numeric`

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html

itholic:
Resolves #1044 
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html#pandas.to_numeric)

i think we can directly use pandas one, except the case of `Series`

```python
>>> kser = ks.Series(['1.0', '2', '-3'])
>>> kser
0    1.0
1      2
2     -3
Name: 0, dtype: object

>>> ks.to_numeric(kser)
0    1.0
1    2.0
2   -3.0
Name: 0, dtype: float32

If given Series contains invalid value to cast float, just cast it to `np.nan`

>>> kser = ks.Series(['apple', '1.0', '2', '-3'])
>>> kser
0    apple
1      1.0
2        2
3       -3
Name: 0, dtype: object

>>> ks.to_numeric(kser)
0    NaN
1    1.0
2    2.0
3   -3.0
Name: 0, dtype: float32

Also support for list, tuple, np.array, or a single numeric type

>>> ks.to_numeric(['1.0', '2', '-3'])
array([ 1.,  2., -3.])

>>> ks.to_numeric(('1.0', '2', '-3'))
array([ 1.,  2., -3.])

>>> ks.to_numeric(np.array(['1.0', '2', '-3']))
array([ 1.,  2., -3.])

>>> ks.to_numeric(1.0)
1.0
```

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -1915,7 +1915,7 @@ def to_numeric(arg):
     3   -3.0
     Name: 0, dtype: float32
 
-    Also support for list, tuple, np.array, or a single numeric type
+    Also support for list, tuple, np.array, or a scalar
 
     >>> ks.to_numeric(['1.0', '2', '-3'])
     array([ 1.,  2., -3.])
@@ -1926,7 +1926,7 @@ def to_numeric(arg):
     >>> ks.to_numeric(np.array(['1.0', '2', '-3']))
     array([ 1.,  2., -3.])
 
-    >>> ks.to_numeric(1.0)
+    >>> ks.to_numeric('1.0')
     1.0
     """
     if isinstance(arg, Series):
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -1942,9 +1942,9 @@ def to_numeric(arg):
     """
     if isinstance(arg, Series):
         sdf = arg._internal.sdf
-        name = arg.name
+        name = arg._internal.data_columns[0]
         sdf = sdf.withColumn(name, sdf[name].cast('float'))
-        column_scols = [sdf[name]]
+        column_scols = [scol_for(sdf, name)]
         return Series(arg._internal.copy(
             sdf=sdf,
             column_scols=column_scols,
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -44,7 +44,7 @@ from databricks.koalas.series import Series, _col
 __all__ = ["from_pandas", "range", "read_csv", "read_delta", "read_table", "read_spark_io",
            "read_parquet", "read_clipboard", "read_excel", "read_html", "to_datetime",
            "get_dummies", "concat", "melt", "isna", "isnull", "notna", "notnull",
-           "read_sql_table", "read_sql_query", "read_sql", "read_json", "merge"]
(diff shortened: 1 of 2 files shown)


databricks/koalas/namespace.py
```diff
@@ -1863,6 +1863,85 @@ def merge(obj, right: 'DataFrame', how: str = 'inner',
...
+    >>> ks.to_numeric(np.array(['1.0', '2', '-3']))
+    array([ 1.,  2., -3.])
+
+    >>> ks.to_numeric('1.0')
+    1.0
+    """
+    if isinstance(arg, Series):
+        sdf = arg._internal.sdf
+        name = arg.name
+        sdf = sdf.withColumn(name, sdf[name].cast('float'))
```
HyukjinKwon: I think we shouldn't always cast it to `float`. We can:

1. Disallow `errors='ignore'` case because it looks this is the only case `object` type is allowed.
2. Infer schema from the small sample (see `Series.apply` as an example)
3. Cast it to the inferred type as you did. Alternatively we can also use `pandas_udf` to allow to throw an exception by given `errors` parameters.

databricks/koalas/namespace.py
```diff
@@ -1863,6 +1863,96 @@ def merge(obj, right: 'DataFrame', how: str = 'inner',
...
+
+    >>> ks.to_numeric(np.array(['1.0', '2', '-3']))
+    array([ 1.,  2., -3.])
+
+    >>> ks.to_numeric('1.0')
+    1.0
+    """
+    if isinstance(arg, Series):
+        sdf = arg._internal.sdf
+        name = arg.name
```
ueshin: We should use `arg._internal.data_columns[0]` instead of `arg.name`.

databricks/koalas/namespace.py
```diff
@@ -1863,6 +1863,96 @@ def merge(obj, right: 'DataFrame', how: str = 'inner',
...
+        sdf = sdf.withColumn(name, sdf[name].cast('float'))
+        column_scols = [sdf[name]]
```
ueshin: `scol_for(sdf, name)`.

databricks/koalas/namespace.py
```diff
@@ -1863,6 +1863,96 @@ def merge(obj, right: 'DataFrame', how: str = 'inner',
...
+    .. note:: this API executes the function once to infer the type which is
+         potentially expensive, for instance, when the dataset is created after
+         aggregations or sorting.
+
+         To avoid this, specify return type in ``func``, for instance, as below:
+
+         >>> def square(x) -> np.int32:
+         ...     return x ** 2
+
+         Koalas uses return type hint and does not try to infer the type.
```
ueshin: Do we need this?

databricks/koalas/namespace.py
ueshin: What's the conclusion?
Seems like we just cast to float so far?

ueshin: If we just simply cast to float, then we should:

```py
return arg._with_new_scol(arg._internal.scol.cast('float'))
```

?

PR #1061, opened 2019-11-20
Modify doctest of Series.rank method

fwani:
I found the same doctest for Dataframe.rank() and Series.rank().

- DataFrame.rank
https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.DataFrame.rank.html#databricks.koalas.DataFrame.rank

- Series.rank
https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.Series.rank.html#databricks.koalas.Series.rank

I think the doctest of Series.rank should be changed.

So, I modified it.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -2918,47 +2918,56 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
         Examples
         --------
-        >>> df = ks.DataFrame({'A': [1, 2, 2, 3], 'B': [4, 3, 2, 1]}, columns= ['A', 'B'])
-        >>> df
-           A  B
-        0  1  4
-        1  2  3
-        2  2  2
-        3  3  1
+        >>> s = ks.Series([1, 2, 2, 3], name='A')
+        >>> s
+        0    1
+        1    2
+        2    2
+        3    3
+        Name: A, dtype: int64
 
-        >>> df.rank().sort_index()
-             A    B
-        0  1.0  4.0
-        1  2.5  3.0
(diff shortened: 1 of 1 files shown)


databricks/koalas/series.py
```diff
@@ -2918,47 +2918,56 @@ def rank(self, method='average', ascending=True):
...
-        3  4.0  1.0
+        >>> s = ks.Series([1, 2, 2, 3], name='A')
+        >>> s
+        0    1
+        1    2
+        2    2
+        3    3
+        Name: A, dtype: int64
+
+        >>> s.rank()
```
itholic: Don't we need to sort the result anymore?
Because i think koalas doesn't guarantee the index order internally although now all tests are passed.
How about cc @ueshin @HyukjinKwon

itholic: For more information, `rank` is sorting the order internally like below.

스크린샷 2019-11-21 오전 8 57 30

`_scol` means spark column, since koalas manage the data on spark internally.

```python
>>> kser = ks.Series([1, 2, 3, 4, 5])
>>> kser._scol
Column<b'0'>
>>> type(kser._scol)
<class 'pyspark.sql.column.Column'>
```

fwani: Thank you for explanation.

And I think we need sorting to rank at DataFrame.
But as you know in Series, the sorting operation is already applied in rank method.

itholic:
It seems good to me

HyukjinKwon:
Merged, thanks.

PR #1062, opened 2019-11-21
Add Property iat for DataFrame & Series

itholic:
iat for DataFrame & Series
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.iat.html#pandas.DataFrame.iat

For DataFrame

```python
>>> df = ks.DataFrame([[0, 2, 3], [0, 4, 1], [10, 20, 30]],
...                   columns=['A', 'B', 'C'])
>>> df
    A   B   C
0   0   2   3
1   0   4   1
2  10  20  30

>>> df.iat[1, 2]
1
```

For Series

```python
>>> kser = ks.Series([1, 2, 3], index=[10, 20, 30])
>>> kser
10    1
20    2
30    3
Name: 0, dtype: int64
>>> kser.iat[1]
2
```

Another examples are in `tests/test_indexing.py` including MultiIndex & MultiIndex columns tests.


databricks/koalas/indexing.py
```diff
@@ -181,6 +181,105 @@ def __getitem__(self, key):
...
+        if not (isinstance(col_sel, int) or
+                (isinstance(col_sel, tuple) and all(isinstance(col, str) for col in col_sel))):
+            raise ValueError('iAt based indexing on multi-index can only have integer values')
+        if isinstance(col_sel, int):
+            if col_sel > len(self._internal.data_columns):
+                raise KeyError(name_like_string(col_sel))
+            col_sel = (self._internal.data_columns[col_sel],)
+
+        tmp_col = '__natural_order__'
+        window = Window.orderBy(F.monotonically_increasing_id())
```
HyukjinKwon: I think this is too inefficient and overkill since it will move all the data into a single partition to only have the natural order and its number.

Can you try this way?

https://github.com/databricks/koalas/blob/5012eb39eb3938b56fe330ecb67a2a193b9e630e/databricks/koalas/internal.py#L474-L508

1. Run and get number for each partition in the first job
2. Fine the row/column and filter out everything in a second job

itholic: i fixed by creating distributed index what you recommended way to solve partition issue.

databricks/koalas/indexing.py
```diff
@@ -181,6 +182,108 @@ def __getitem__(self, key):
...
+            if col_sel > len(self._internal.data_columns):
+                raise KeyError(name_like_string(col_sel))
+            col_sel = (self._internal.data_columns[col_sel],)
+
+        sdf = self._internal.sdf.select(self._internal.data_columns)
+        set_option("compute.default_index_type", "distributed-sequence")
+        try:
+            sdf = _InternalFrame.attach_default_index(sdf)
+        finally:
+            reset_option("compute.default_index_type")
```
ueshin: We can't just reset the option. e.g., what if a user already set the option?

databricks/koalas/indexing.py
```diff
@@ -181,6 +182,108 @@ def __getitem__(self, key):
...
+                raise TypeError("Use Series.iat like .iat[row_integer_position]")
+            row_sel = key
+            col_sel = 0
+
+        if len(self._internal.index_map) == 1:
+            if is_list_like(row_sel):
+                raise ValueError(
+                    'iAt based indexing on a single index can only have a single value')
+        if not (isinstance(col_sel, int) or
+                (isinstance(col_sel, tuple) and all(isinstance(col, str) for col in col_sel))):
```
ueshin: Do we need to check this? e.g., does `kdf.iat[0, ('x', 'y')]` work for multi-index columns?

itholic: ah, you're right.

we can just check whether the type of `col_sel` is `int` or not

thanks for the catch!!

databricks/koalas/indexing.py
itholic: ah, thanks for the catch!

and are you working at #1077 to manage something like this case? 😄

HyukjinKwon:
@itholic, can you make the tests passed at Python 3.5 too?

itholic:
@HyukjinKwon it's wired that those tests failing at python3.5.6 in travisCI are working well in my local python3.5.6

스크린샷 2019-12-02 오후 5 25 09

let me check a little more

HyukjinKwon:
Can you test it against Spark 2.3? Seems the Spark version matters.

databricks/koalas/indexing.py
```diff
@@ -181,6 +182,109 @@ def __getitem__(self, key):
...
+            if col_sel > len(self._internal.data_columns):
+                raise KeyError(name_like_string(col_sel))
+            col_sel = (self._internal.data_columns[col_sel],)
+
+        sdf = self._internal.sdf.select(self._internal.data_columns)
+
+        option_origin = get_option("compute.default_index_type")
+        set_option("compute.default_index_type", "distributed-sequence")
+        try:
+            sdf = _InternalFrame.attach_default_index(sdf)
```
HyukjinKwon: I think you can change `attach_default_index` as below:

```python
def attach_default_index(sdf, index_type=None)
    if index_type is None:
        index_type = get_option("compute.default_index_type")
    ...
```

and use `_InternalFrame.attach_default_index(sdf, index_type="distributed-sequence")` here

databricks/koalas/indexing.py
```diff
@@ -181,6 +182,109 @@ def __getitem__(self, key):
...
+        finally:
+            set_option("compute.default_index_type", option_origin)
+        cond = F.col(SPARK_INDEX_NAME_FORMAT(0)) == row_sel
+        pdf = sdf.where(cond).select(*col_sel).toPandas()
+
+        if len(pdf) < 1:
+            raise KeyError(name_like_string(row_sel))
+
+        values = pdf.iloc[:, 0].values
```
HyukjinKwon: Maybe you want to do `.iloc[0, 0]`

HyukjinKwon:
Seems fine in general if tests pass

databricks/koalas/indexing.py
```diff
@@ -181,6 +182,103 @@ def __getitem__(self, key):
...
+        if not isinstance(col_sel, int):
+            raise ValueError('iAt based indexing on multi-index can only have integer values')
+        if isinstance(col_sel, int):
+            if col_sel > len(self._internal.data_columns):
+                raise KeyError(name_like_string(col_sel))
+            col_sel = (self._internal.data_columns[col_sel],)
+
+        sdf = self._internal.sdf.select(self._internal.data_columns)
+
+        sdf = _InternalFrame.attach_default_index(sdf, default_index_type="distributed-sequence")
```
itholic: at this point, `sdf` looks like in pyspark2.4.

```
+---+-----------------+
|  a|__index_level_0__|
+---+-----------------+
|  0|                1|
|  4|                5|
|  2|                3|
|  3|                4|
|  6|                7|
|  5|                6|
|  7|                8|
|  8|                9|
|  1|                2|
+---+-----------------+
```

but in 2.3, it switched.

```
+-----------------+---+
|__index_level_0__|  a|
+-----------------+---+
|                1|  0|
|                5|  4|
|                3|  2|
|                4|  3|
|                7|  6|
|                6|  5|
|                8|  7|
|                9|  8|
|                2|  1|
+-----------------+---+
```

databricks/koalas/indexing.py
```diff
@@ -181,6 +182,103 @@ def __getitem__(self, key):
...
+                    'iAt based indexing on a single index can only have a single value')
+        if not isinstance(col_sel, int):
+            raise ValueError('iAt based indexing on multi-index can only have integer values')
+        if isinstance(col_sel, int):
+            if col_sel > len(self._internal.data_columns):
+                raise KeyError(name_like_string(col_sel))
+            col_sel = (self._internal.data_columns[col_sel],)
+
+        sdf = self._internal.sdf.select(self._internal.data_columns)
+
```
guyao: Should we retain the original index ? 
IMO code here discard the original index and attach a new `distributed-sequence` one. 

For example, If the index type is `distributed-sequence`, the sdf might looks like that. 

(SPARK_PARTITION_ID is the spark partition number, I create this spark.dataframe by creating a rdd and do repartition into 2)
```
+-----------------+--------------------+-----+
|__index_level_0__|SPARK_PARTITION_ID()|value|
+-----------------+--------------------+-----+
|                0|                   0|    2|
|                1|                   0|    6|
|                2|                   0|    7|
|       8589934592|                   1|    1|
|       8589934593|                   1|    3|
|       8589934594|                   1|    4|
|       8589934595|                   1|    5|
|       8589934596|                   1|    8|
+-----------------+--------------------+-----+
```

If I was the user, when I want to select value of `Row(8589934592,1,1)`, I would do `df.iat[8589934592, 2]` rather than `df.iat[3, 2]`. Because in distributed scenario the order is not important (and does not guarantee continuous) , I would directly use the index I see which is `8589934592`.

itholic: hello @guyao ,

Thanks for your kindness review :)

maybe in that case what you're considering about,

i think we can use `AtIndexer` rather than `iAtIndexer` like below.

```python
>>> kdf  # koalas DataFrame
            value
0               2
1               6
2               7
8589934592      1
8589934593      3
8589934594      4
8589934595      5
8589934596      8
>>>
>>> kdf.at[8589934593, 'value']
3
```

Above all, the more fundamental reason is that we need to follow pandas' behavior like below.

```python
>>> pdf  # pandas DataFrame
            value
0               2
1               6
2               7
8589934592      1
8589934593      3
8589934594      4
8589934595      5
8589934596      8
>>> pdf.iat[4, 0]  # pandas `iat`, we should follow this behavior.
3

>>> kdf  # koalas DataFrame, as same as pandas' above
            value
0               2
1               6
2               7
8589934592      1
8589934593      3
8589934594      4
8589934595      5
8589934596      8
>>> kdf.iat[4, 0]  # koalas `iat` which should work same with pandas' `iat`
3
>>> kdf.at[]
```

guyao: Got it, make sense. 
Thanks for explanation.

guyao: My concern is that I am not sure whether Spark will guarantee any order.

HyukjinKwon:
@itholic can you fix the test failures?

itholic:
@HyukjinKwon it seems they have different behavior between pyspark 2.3 & 2.4 when they create default index by `distributed-sequence` option.

2.3
```
+-----------------+---+
|__index_level_0__|  a|
+-----------------+---+
|                1|  0|
|                5|  4|
|                3|  2|
|                4|  3|
|                7|  6|
|                6|  5|
|                8|  7|
|                9|  8|
|                2|  1|
+-----------------+---+
```

2.4
```
+-----------------+---+
|__index_level_0__|  a|
+-----------------+---+
|                0|  1|
|                4|  5|
|                2|  3|
|                3|  4|
|                6|  7|
|                5|  6|
|                7|  8|
|                8|  9|
|                1|  2|
+-----------------+---+
```

2.4 is correct one.

and i think maybe this problem comes from `pandas_udf` around below.

스크린샷 2019-12-10 오후 2 41 17

(`attach_default_index` in `internal.py`)

databricks/koalas/internal.py
```diff
@@ -506,6 +507,9 @@ def attach_default_index(sdf):
             columns = [f.name for f in return_schema]
 
             # 3. Group by partition id and assign each range.
+            return_schema = StructType(
```
HyukjinKwon: This can be removed - it's duplicated as above.

databricks/koalas/indexing.py
```diff
@@ -131,6 +132,103 @@ def __getitem__(self, key):
...
+            assert self._is_series, type(self._kdf_or_kser)
+            if not isinstance(key, int) and len(key) != 1:
+                raise TypeError("Use Series.iat like .iat[row_integer_position]")
+            row_sel = key
+            col_sel = 0
+
+        if len(self._internal.index_map) == 1:
+            if is_list_like(row_sel):
+                raise ValueError(
+                    'iAt based indexing on a single index can only have a single value')
```
HyukjinKwon: `iAt` -> `iat`

HyukjinKwon:
cc @rxin and @thunterdb FYI.

databricks/koalas/indexing.py
```diff
@@ -131,6 +132,103 @@ def __getitem__(self, key):
...
+        if not isinstance(col_sel, int):
+            raise ValueError('iAt based indexing on multi-index can only have integer values')
+        if isinstance(col_sel, int):
+            if col_sel > len(self._internal.data_columns):
+                raise KeyError(name_like_string(col_sel))
+            col_sel = (self._internal.data_columns[col_sel],)
+
+        sdf = self._internal.sdf.select(self._internal.data_columns)
+
+        sdf = _InternalFrame.attach_default_index(sdf, default_index_type="distributed-sequence")
```
HyukjinKwon: Ha .. I don't particularly like this to get a global sequence and position .. but I guess we have no way.

itholic: yeah i totally agree with you, it looks a little hacky and unstable 😅 
but i seems no other way for now, let's try to find the solution soon.

databricks/koalas/indexing.py
```diff
@@ -131,6 +132,103 @@ def __getitem__(self, key):
...
+        return isinstance(self._kdf_or_kser, DataFrame)
+
+    @property
+    def _is_series(self):
+        from databricks.koalas.series import Series
+        return isinstance(self._kdf_or_kser, Series)
+
+    @property
+    def _internal(self):
+        return self._kdf_or_kser._internal
```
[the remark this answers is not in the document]
itholic: ah, now they're inherited from `IndexerLike`. thanks!! :)

itholic:
@ueshin 

> Btw, what's the difference between `iat[n]` and `iloc[n]`?

`iat[n]` and `iloc[n]` are work exactly same.

more strictly speaking, `iat[n]` and `iloc[n]` are work exactly same for '`Series`'.

for `DataFrame`, `iat` needs two arguments, row/column pair, but `iloc` can be used with only one argument, row, like the below. (it returns `Series`)

```python
>>> pdf
   a  b
0  1  4
1  2  5
3  3  6
5  4  3
6  5  2
8  6  1
9  7  0
9  8  0
9  9  0

>>> pdf.iat[3]
...
TypeError: _get_value() missing 1 required positional argument: 'col'

>>> pdf.iloc[3]
a    4
b    3
Name: 5, dtype: int64
```



> If the difference is none or small enough, we can consolidate the logic to `iloc[n]` and delegate to it?

IMO, it seems possible if appropriate and sufficient consideration is given to a particular case (`DataFrame`).

Let me check this a bit more, and if it's possible, i'll consolidate them in this PR.

ueshin:
As for DataFrames, are `pdf.iat[n, m]` and `pdf.iloc[n, m]` the same then?

ueshin:
Btw, I guess `at` and `loc` is the same thing. I'm working on it.

itholic:
@ueshin 

> As for DataFrames, are `pdf.iat[n, m]` and `pdf.iloc[n, m]` the same then?

correct. they are same :)

ueshin:
@itholic FYI: as for `at` and `loc`, seems like the return value is not identical when the key is duplicated. (`np.array` vs. `Series`).
I'd leave it as is for now.
`iat` and `iloc` should work with the delegation, please go ahead.

itholic:
@ueshin 

> @itholic FYI: as for `at` and `loc`, seems like the return value is not identical when the key is duplicated. (`np.array` vs. `Series`).
> I'd leave it as is for now.
> `iat` and `iloc` should work with the delegation, please go ahead.

Thanks for noticing me :)

---

Anyway, i've found some problem while i working on this.

assuming that we have DataFrame like the below

```python
>>> pser  # pandas dataframe
7    10
6    20
5    30
4    40
3    50
2    60
1    70
dtype: int64

>>> kser  # koalas dataframe
7    10
6    20
5    30
4    40
3    50
2    60
1    70
Name: 0, dtype: int64
```

for given the above frames, below example works differently from pandas.

```python
>>> pser.to_frame().loc[6:4]
    0
6  20
5  30
4  40

>>> kser.to_frame().loc[6:4]
Empty DataFrame
Columns: [0]
Index: []
```

First, i'm going to fix this and will get back to here.

itholic:
cc @ueshin 

i just consolidated the logic `iat[n]` & `iloc[n]` (for Seires, or `iat[n, m]` & `iloc[n, m]` for DataFrame) at https://github.com/databricks/koalas/pull/1062/commits/60c96b396d7eedef5bf1b4236286b22e00871fa1, but not pretty sure if it is what you intended. 😅

would you please check this if when you available?

databricks/koalas/indexing.py
```diff
@@ -131,6 +132,89 @@ def __getitem__(self, key):
...
+
+    >>> kser.iat[1]
+    2
+    """
+    def __init__(self, kdf_or_kser):
+        from databricks.koalas.frame import DataFrame
+        from databricks.koalas.series import Series
+        assert isinstance(kdf_or_kser, (DataFrame, Series)), \
+            'unexpected argument type: {}'.format(type(kdf_or_kser))
+        self._kdf_or_kser = kdf_or_kser
```
ueshin: I don't think we need this?

databricks/koalas/indexing.py
```diff
@@ -131,6 +132,89 @@ def __getitem__(self, key):
                           or len(values) > 1) else values[0]
 
 
+class iAtIndexer(_IndexerLike):
```
ueshin: Shall we name `IAtIndexer`? or we might want to rename `ILocIndexer` to `iLocIndexer`.

databricks/koalas/indexing.py
```diff
@@ -131,6 +132,89 @@ def __getitem__(self, key):
...
+            assert self._is_series, type(self._kdf_or_kser)
+            if not isinstance(key, int) and len(key) != 1:
+                raise TypeError("Use Series.iat like .iat[row_integer_position]")
+            row_sel = key
+            col_sel = 0
+
+        if len(self._internal.index_map) == 1:
+            if is_list_like(row_sel):
+                raise ValueError(
+                    'iat based indexing on a single index can only have a single value')
```
ueshin: Why do we need this?

databricks/koalas/indexing.py
itholic: pandas is using under score as prefix for the name of 'Indexers' classes like `_iAtIndexer` or `_iLocIndexer` like the below.

스크린샷 2020-01-12 오후 2 26 55

so, although i don't think it is not a big deal, how about we also add the under score as prefix to name of Indexers classes like they're doing???

itholic:
> Actually what I intended was to call `iloc[n]` or `iloc[n, m]` in `iAtIndexer`.
> 
> ```python
> class iAtIndexer:
>     def __getitem__():
>         if self._is_df:
>             ...
>             return self._kdf_or_kser.iloc[rows_sel, cols_sel]
>         else:
>             ...
>             return self._kdf_or_kser.iloc[rows_sel]
> 
> ...
> 
> class ILocIndexer:
>     def _select_rows(self, rows_sel):
>         ...
>         elif isinstance(rows_sel, int):
>             # complete the implementation
> ```

ah, i got it. let me rearrange again.

thanks :)

databricks/koalas/indexing.py
itholic: ah, yeah it is inherited from `_IndexerLike `.

thanks for the comment! :)

databricks/koalas/indexing.py
itholic: we need this since iAt based indexing can only have integer indexers like below.

```python
>>> pdf
   a  b
0  1  4
1  2  5
3  3  6
5  4  3
6  5  2
8  6  1
9  7  0
9  8  0
9  9  0
>>>
>>> pdf.iat[(1, 1), 1]
...
ValueError: iAt based indexing can only have integer indexers
```
but the error message seems not proper.

i'll fix the message

databricks/koalas/indexing.py
itholic: @ueshin 

could you check this discussion when you available??

HyukjinKwon:
Let me merge this one for now.

databricks/koalas/indexing.py
nikilp: Probably hacky solutions should pertain to the specific versions giving incorrect results.

PR #1065, opened 2019-11-22
Fix Series.idxmax() & idxmin() to make result properly

itholic:
if given series like below,

```python
>>> kser = ks.Series([1, 100, 1, 100, 1, 100])
>>> kser
0      1
1    100
2      1
3    100
4      1
5    100
```

And if we want to get an index of maximum value, we can use `idxmax()`

the pandas doc says, "If multiple values equal the maximum, the 'first' row label with that value is returned",  (Actually, our doc have missed the part, 'first')

스크린샷 2019-11-22 오후 12 09 16

So in this case, it should returns '1' which is the first row label of those maximum values.

but the existing behavior returns random label like this.

```python
>>> kser.idxmax()
5
>>> kser.idxmax()
5
>>> kser.idxmax()
1
>>> kser.idxmax()
5
>>> kser.idxmax()
3
>>> kser.idxmax()
3
>>> kser.idxmax()
1
```

so i think it should be better fixed.

(those descriptions above are also applied to `idxmin()` exactly same)

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -3165,9 +3165,9 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         # desc_nulls_(last|first) is used via Py4J directly because
         # it's not supported in Spark 2.3.
         if skipna:
-            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_last()))
+            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_last()), *index_scols)
         else:
-            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_first()))
+            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_first()), *index_scols)
         results = sdf.select([scol] + index_scols).take(1)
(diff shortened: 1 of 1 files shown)


ueshin:
Sorry for the late response.

Actually the result is still not correct:

```py
>>> pser = pd.Series([1, 100, 1, 100, 1, 100], index=[10, 3, 5, 2, 1, 8])
>>> pser
10      1
3     100
5       1
2     100
1       1
8     100
dtype: int64
>>> pser.idxmax()
3
>>> pser.idxmin()
10
```

whereas:

```py
>>> ks.from_pandas(pser).idxmax()
2
>>> ks.from_pandas(pser).idxmin()
1
```

HyukjinKwon:
Thanks Takuya. Reverted for now. @itholic I think we should better test more closely next time.

itholic:
@ueshin thanks for the comment!
ah, it seems that they return the `head(1)` rather than `mathematically min/max` when multiple values equal the max/min.
i'll create the new PR to fix this more properly soon. 👍

PR #1068, opened 2019-11-22
Remove fillna from Series.between

harupy:
Remove `fillna` from `Series.between` because comparison operators fill `NULL` with a boolean value (#1029).

https://github.com/databricks/koalas/blob/fe93ac1e6247d5f1ee658d90a887eacbb5ffe655/databricks/koalas/series.py#L740-L744


HyukjinKwon:
Looks fine. Did you mean #1029 instead of #1045 :-)?

harupy:
@HyukjinKwon Thanks for the catch. I fixed the comment.

PR #1069, opened 2019-11-23
Make corr return koalas.DataFrame

harupy:
This PR proposes to make `corr` return `koalas.DataFrame`.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1056,7 +1056,7 @@ class DataFrame(_Frame, Generic[T]):
 
           * `min_periods` argument is not supported
         """
-        return corr(self, method)
+        return ks.from_pandas(corr(self, method))
 
     def iteritems(self) -> Iterable:
         """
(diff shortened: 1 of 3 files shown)


harupy:
@HyukjinKwon 
Thanks for merging. I'm planning to add a new argument named `as_pandas` (`False` by default) to `koalas.DataFrame.corr` in another PR. The idea is if this argument is `True`, return `pandas.DataFrame` which is more suitable for python visualization libraries. What do you think?

HyukjinKwon:
I think users still can just call `to_pandas` on the returned Koalas DataFrame from `corr`. Or is there another case I missed?

harupy:
> I think users still can just call to_pandas on the returned Koalas DataFrame from corr.

Yes, but when a user wants `pandas.DataFrame`, `ks.from_pandas(corr_pdf)` is unnecessary process.

HyukjinKwon:
I think it's fine. Otherwise we should add such argument to all APIs ...

harupy:
@HyukjinKwon You're right. I started thinking this is no necessary...

Issue #1003, opened 2019-11-05
Triage Top Missing APIs to explicitly don't support or implement.


HyukjinKwon:
@ueshin wrote and ran some notebooks to analyze statistics from missing API calls as below. We might need to explicitly don't support with a proper workaround in its exception message and/or implement it first.

- [x] Series.\_\_iter\_\_(self)
- [x] DataFrame.apply(self, func, axis=0, broadcast=None, raw=False, reduce=None, result_type=None, args=(), **kwds) #1259
- [x] Series.values
- [x] Index.values
- [x] DataFrame.info(self, verbose=None, buf=None, max_cols=None, memory_usage=None, null_counts=None) - https://github.com/databricks/koalas/issues/872
- [x] Index.to_numpy(self, dtype=None, copy=False)
- [x] DataFrame.iterrows(self)
- [x] DataFrame.unstack(self, level=-1, fill_value=None)  #1295
- [x] DataFrame.rename(self, mapper=None, index=None, columns=None, axis=None, copy=True, inplace=False, level=None)
- [ ] DataFrame.tail(self, n=5)
- [x] Index.\_\_iter\_\_(self)
- [x] DataFrame.where(self, cond, other=nan, inplace=False, axis=None, level=None, errors='raise', try_cast=False, raise_on_error=None)
- [x] DataFrame.values
- [x] DataFrameGroupBy.head(self, n=5)
- [x] DataFrame.query(self, expr, inplace=False, **kwargs)
- [x] DataFrame.take(self, indices, axis=0, is_copy=True, **kwargs)  #1292
- [x] DataFrame.take(self, indices, axis=0, convert=None, is_copy=True, **kwargs)  #1292
- [x] Index.tolist(self)
- [ ] SeriesGroupBy.unique
- [x] Series.keys
- [x] Series.replace

itholic:
i think `Series.__iter__(self)` is quite widely & frequently used by data scientists.

so in my opinion, how about make it available for now with proper warning message although it has risk of memory issue when large-size dataset?

now it raises `PandasNotImplementedError` like below,

```python
>>> for a in kser:
...     a
...
Traceback (most recent call last):
databricks.koalas.exceptions.PandasNotImplementedError: The method `pd.Series.__iter__()` is not implemented. If you want to collect your data as an NumPy array, use 'to_numpy()' instead.
```
[...]

Issue #885, opened 2019-10-03
DataFrame.iterrows


ueshin:
Implement `DataFrame.iterrows`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.iterrows.html

ueshin:
I don't think we will support this kind of APIs. cc @HyukjinKwon

PR #1070, opened 2019-11-24
Implement Dataframe.iterrows using spark.dataframe.toLocalIterator

guyao:
#885 
Inspired by https://github.com/databricks/koalas/issues/1003#issuecomment-556908359

the data type yield from iterator is `pd.Series`. I assume each row is small enough.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1111,6 +1111,8 @@ class DataFrame(_Frame, Generic[T]):
         data : pandas.Series
             The data of the row as a Series.
 
+        Returns
+        -------
         it : generator
             A generator that iterates over the rows of the frame.
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1111,8 +1111,6 @@ class DataFrame(_Frame, Generic[T]):
         data : pandas.Series
             The data of the row as a Series.
 
-        Returns
-        -------
         it : generator
             A generator that iterates over the rows of the frame.
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1101,13 +1101,48 @@ class DataFrame(_Frame, Generic[T]):
         return list((col_name, self[col_name]) for col_name in cols)
 
     def iterrows(self):
-        # TODO: document
+        """
+        Iterate over DataFrame rows as (index, Series) pairs.
+        Yields
+        ------
+        index : label or tuple of label
+            The index of the row. A tuple for a `MultiIndex`.
+        data : pandas.Series
+            The data of the row as a Series.
+        it : generator
+            A generator that iterates over the rows of the frame.
+
+        Notes
+        -----
+        1. Because ``iterrows`` returns a Series for each row,
+           it does **not** preserve dtypes across the rows (dtypes are
+           preserved across columns for DataFrames). For example,
+           >>> df = pd.DataFrame([[1, 1.5]], columns=['int', 'float'])
+           >>> row = next(df.iterrows())[1]
+           >>> row
+           int      1.0
+           float    1.5
+           Name: 0, dtype: float64
+           >>> print(row['int'].dtype)
+           float64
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -1100,6 +1100,56 @@ def iteritems(self) -> Iterable:
         cols = list(self.columns)
         return list((col_name, self[col_name]) for col_name in cols)
 
+    def iterrows(self):
+        """
+        Iterate over DataFrame rows as (index, Series) pairs.
+        Yields
```
HyukjinKwon: Looks you should add some spaces between the doc and `Yields`.

databricks/koalas/frame.py
```diff
@@ -1100,6 +1100,56 @@ def iteritems(self) -> Iterable:
...
+            The data of the row as a Series.
+        it : generator
+            A generator that iterates over the rows of the frame.
+
+        Notes
+        -----
+        1. Because ``iterrows`` returns a Series for each row,
+           it does **not** preserve dtypes across the rows (dtypes are
+           preserved across columns for DataFrames). For example,
+           >>> df = pd.DataFrame([[1, 1.5]], columns=['int', 'float'])
```
HyukjinKwon: here too

databricks/koalas/frame.py
```diff
@@ -1100,6 +1100,56 @@ def iteritems(self) -> Iterable:
...
+           float    1.5
+           Name: 0, dtype: float64
+           >>> print(row['int'].dtype)
+           float64
+           >>> print(df['int'].dtype)
+           int64
+           To preserve dtypes while iterating over the rows, it is better
+           to use :meth:`itertuples` which returns namedtuples of the values
+           and which is generally faster than ``iterrows``.
+        2. You should **never modify** something you are iterating over.
```
HyukjinKwon: and here too

databricks/koalas/frame.py
ueshin: This is not addressed yet.

guyao: Thanks, fixed!

databricks/koalas/frame.py
```diff
@@ -1100,6 +1100,59 @@ def iteritems(self) -> Iterable:
...
+           >>> row = next(df.iterrows())[1]
+           >>> row
+           int      1.0
+           float    1.5
+           Name: 0, dtype: float64
+           >>> print(row['int'].dtype)
+           float64
+           >>> print(df['int'].dtype)
+           int64
+           To preserve dtypes while iterating over the rows, it is better
```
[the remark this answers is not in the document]
guyao: Thanks, fixed! 
Added multi index test in `test/test_dataframe`

databricks/koalas/frame.py
```diff
@@ -1110,6 +1110,7 @@ def iterrows(self):
             The index of the row. A tuple for a `MultiIndex`.
         data : pandas.Series
             The data of the row as a Series.
+
```
ueshin: nit: should revert this change?

guyao: I saw an empty line between `data` and `it` in pandas doc 
https://github.com/pandas-dev/pandas/blob/v0.25.3/pandas/core/frame.py#L860-L910

This function returns a generator `it` and items are `index` and `data`. That might be the reason an empty line here

ueshin: I see, then maybe we should add `Returns` section?
```
        Returns
        -------
        it : generator
            ...
```

guyao: Fixed!

ueshin: ah, sorry, seems like we can't use both `Yields` and `Returns` at the same time.

```
ValueError: Docstring contains both a Returns and Yields section.
```

Could you revert the last commit?

guyao: oh interesting. 
sorry I should try to build the doc to test
branch updated

ueshin:
Thanks! merging.

PR #1051, opened 2019-11-18
Implements pct_change() for DataFrame

itholic:
Resolves #878 
(https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.DataFrame.pct_change.html)

```
>>> df = ks.DataFrame({
...     'FR': [4.0405, 4.0963, 4.3149],
...     'GR': [1.7246, 1.7482, 1.8519],
...     'IT': [804.74, 810.01, 860.13]},
...     index=['1980-01-01', '1980-02-01', '1980-03-01'])
>>> df
                FR      GR      IT
1980-01-01  4.0405  1.7246  804.74
1980-02-01  4.0963  1.7482  810.01
1980-03-01  4.3149  1.8519  860.13

>>> df.pct_change()
                  FR        GR        IT
1980-01-01       NaN       NaN       NaN
1980-02-01  0.013810  0.013684  0.006549
1980-03-01  0.053365  0.059318  0.061876
```

You can set periods to shift for forming percent change

```
>>> df.pct_change(2)
                  FR        GR       IT
1980-01-01       NaN       NaN      NaN
1980-02-01       NaN       NaN      NaN
1980-03-01  0.067912  0.073814  0.06883
```

PR #1071, opened 2019-11-24
Implement pct_change() for Series

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.pct_change.html#pandas.Series.pct_change

```python
>>> kser = ks.Series([90, 91, 85])
>>> kser
0    90
1    91
2    85
Name: 0, dtype: int64

>>> kser.pct_change()
0         NaN
1    0.011111
2   -0.065934
Name: 0, dtype: float64

>>> kser.pct_change(periods=2)
0         NaN
1         NaN
2   -0.055556
Name: 0, dtype: float64
"""
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4303,14 +4303,14 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
         for column_name in self._internal.data_columns:
             prev_row = F.lag(F.col(column_name), periods).over(window)
-            sdf = sdf.withColumn(column_name, (F.col(column_name) - prev_row) / prev_row)
+            sdf = sdf.withColumn(column_name, (scol_for(sdf, column_name) - prev_row) / prev_row)
 
         internal = _InternalFrame(
             sdf=sdf,
             index_map=self._internal.index_map,
             column_index=self._internal.column_index)
 
-        return Series(internal)
+        return _col(DataFrame(internal))
 
     def _cum(self, func, skipna, part_cols=()):
         # This is used to cummin, cummax, cumsum, etc.
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4279,36 +4279,39 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         Examples
         --------
 
-        >>> kser = ks.Series([90, 91, 85])
+        >>> kser = ks.Series([90, 91, 85], index=[2, 4, 1])
         >>> kser
-        0    90
-        1    91
-        2    85
+        2    90
+        4    91
+        1    85
         Name: 0, dtype: int64
 
         >>> kser.pct_change()
-        0         NaN
-        1    0.011111
-        2   -0.065934
+        2         NaN
+        4    0.011111
+        1   -0.065934
+        Name: 0, dtype: float64
+
+        >>> kser.sort_index().pct_change()
+        1         NaN
+        2    0.058824
+        4    0.011111
         Name: 0, dtype: float64
 
         >>> kser.pct_change(periods=2)
-        0         NaN
-        1         NaN
-        2   -0.055556
+        2         NaN
+        4         NaN
+        1   -0.055556
         Name: 0, dtype: float64
         """
         sdf = self._internal._sdf
-        window = Window.orderBy(self._internal.index_columns).rowsBetween(-periods, -periods)
+        window = Window.orderBy(F.monotonically_increasing_id()).rowsBetween(-periods, -periods)
 
         for column_name in self._internal.data_columns:
-            prev_row = F.lag(F.col(column_name), periods).over(window)
+            prev_row = F.lag(scol_for(sdf, column_name), periods).over(window)
             sdf = sdf.withColumn(column_name, (scol_for(sdf, column_name) - prev_row) / prev_row)
 
-        internal = _InternalFrame(
-            sdf=sdf,
-            index_map=self._internal.index_map,
-            column_index=self._internal.column_index)
+        internal = self._internal.copy(sdf=sdf)
 
         return _col(DataFrame(internal))
 
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4304,18 +4304,12 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         1   -0.055556
         Name: 0, dtype: float64
         """
-        sdf = self._internal._sdf
+        scol = self._internal.scol
+
         window = Window.orderBy(F.monotonically_increasing_id()).rowsBetween(-periods, -periods)
+        prev_row = F.lag(scol, periods).over(window)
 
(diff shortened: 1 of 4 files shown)


itholic:
i'm going to integrated the logic with `DataFrame.pct_change()` after this and #1051 will be merged.

databricks/koalas/series.py
```diff
@@ -4258,6 +4258,60 @@ def xs(self, key, level=None):
...
+        for column_name in self._internal.data_columns:
+            prev_row = F.lag(F.col(column_name), periods).over(window)
+            sdf = sdf.withColumn(column_name, (F.col(column_name) - prev_row) / prev_row)
+
+        internal = _InternalFrame(
+            sdf=sdf,
+            index_map=self._internal.index_map,
+            column_index=self._internal.column_index)
+
+        return Series(internal)
```
ueshin: This loses the anchor. We should use `_col(DataFrame(internal))`.

databricks/koalas/series.py
```diff
@@ -4258,6 +4258,60 @@ def xs(self, key, level=None):
...
+        1         NaN
+        2   -0.055556
+        Name: 0, dtype: float64
+        """
+        sdf = self._internal._sdf
+        window = Window.orderBy(self._internal.index_columns).rowsBetween(-periods, -periods)
+
+        for column_name in self._internal.data_columns:
+            prev_row = F.lag(F.col(column_name), periods).over(window)
+            sdf = sdf.withColumn(column_name, (F.col(column_name) - prev_row) / prev_row)
```
ueshin: `scol_for(sdf, column_name)` instead of `F.col(column_name)`?

databricks/koalas/series.py
itholic: @ueshin ,thanks for the comment! 😸 

i have a question since i'm not pretty sure about something,

if we using `copy()` when make `internal` rather than `_InternalFrame`, then can use `Series(internal)`directly??

databricks/koalas/series.py
itholic: ah yes, i think we can. thanks for the comment!

databricks/koalas/series.py
ueshin: Nope, we should create `DataFrame` first whenever we modify `sdf`.

itholic: @ueshin 
ah, the matter was a `sdf`!
thanks very much for the explanation. now i totally got it. 👍

ueshin: Sorry, I mean, `Series(internal)` is not a proper constructor. We need to specify `anchor`.
`_col(DataFrame(internal))` is for hiding the construction of the `Series`.

itholic: @ueshin 
ah, thanks for replying :) 
then how about `Series(internal, anchor=self)` ??

ueshin: `self` is not `DataFrame`.

itholic: @ueshin thanks!! 😄

databricks/koalas/series.py
```diff
@@ -4258,6 +4258,60 @@ def xs(self, key, level=None):
...
+        0         NaN
+        1         NaN
+        2   -0.055556
+        Name: 0, dtype: float64
+        """
+        sdf = self._internal._sdf
+        window = Window.orderBy(self._internal.index_columns).rowsBetween(-periods, -periods)
+
+        for column_name in self._internal.data_columns:
+            prev_row = F.lag(F.col(column_name), periods).over(window)
```
ueshin: `scol_for(sdf, column_name)` instead of `F.col(column_name)`?

databricks/koalas/series.py
```diff
@@ -4258,6 +4258,60 @@ def xs(self, key, level=None):
...
+        Name: 0, dtype: float64
+
+        >>> kser.pct_change(periods=2)
+        0         NaN
+        1         NaN
+        2   -0.055556
+        Name: 0, dtype: float64
+        """
+        sdf = self._internal._sdf
+        window = Window.orderBy(self._internal.index_columns).rowsBetween(-periods, -periods)
```
ueshin: `index_scols` instead of `index_columns`?

ueshin: Btw, seems like we should order by `F.monotonically_increasing_id()` instead of index?

```py
>>> import pandas as pd
>>> pser = pd.Series([90, 91, 85], index=[2,4,1])
>>> pser
2    90
4    91
1    85
dtype: int64
>>> pser.pct_change()
2         NaN
4    0.011111
1   -0.065934
dtype: float64
>>> pser.sort_index().pct_change()
1         NaN
2    0.058824
4    0.011111
dtype: float64

>>> ks.from_pandas(pser).pct_change()
1         NaN
2    0.058824
4    0.011111
Name: 0, dtype: float64
```

databricks/koalas/series.py
itholic: thanks for the comment :)
at this point, i'm really apologize that i have a question again 😅 

(maybe this question is more related with spark)

for now, when we select from spark dataframe, there are 3 way i recognized.
(let's assume that column name is 'koalas')

1. `sdf.select('koalas')`
2. `sdf.select(F.col('koalas'))`
3. `sdf.select(scol_for(sdf, 'koalas'))`

i know how they are different, but I wonder how it differs in terms of performance. (Perhaps this is why we prefer the 3th way?)

ueshin: The performances are the same.
`scol_for` is introduced for supporting column names including `.` (#217, #490).

databricks/koalas/tests/test_series.py
```diff
@@ -909,3 +909,33 @@ def test_keys(self):
...
+    def test_pct_change(self):
+        kser = ks.Series([90, 91, 85], index=[2, 4, 1])
+        pser = kser.to_pandas()
+
+        self.assert_eq(repr(kser.pct_change(periods=-1)),
+                       repr(pser.pct_change(periods=-1)))
+        self.assert_eq(repr(kser.pct_change(periods=-100000000)),
+                       repr(pser.pct_change(periods=-100000000)))
+        self.assert_eq(repr(kser.pct_change(periods=100000000)),
+                       repr(pser.pct_change(periods=100000000)))
```
ueshin: Why are these using `repr()`?

databricks/koalas/series.py
itholic: @ueshin ah, now i have understood after checking that issues. thanks!

databricks/koalas/tests/test_series.py
itholic: Because they raise `AssertionError` even they looks like same result.

스크린샷 2019-12-03 오전 10 10 44

but i just changed them to use `almost=True` rather than `repr`

itholic:
https://github.com/databricks/koalas/pull/1071#pullrequestreview-325810948

@ueshin It looks really cool. thanks for spending your time for this. 👍

PR #1075, opened 2019-11-25
Allow Series.__getitem__ to take boolean Series

### LINKED ISSUE #1073: Series.__getitem__ doesn't work when Series is given
```python
import pandas as pd

pser = pd.Series([1, 2, 3])

>>> pser[pser == 1]
0    1
dtype: int64

kser = ks.from_pandas(pser)

>>> kser[kser == 1]

  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/databricks/koalas/series.py", line 4188, in __getitem__
    for level, index in enumerate(key)]
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/databricks/koalas/series.py", line 4188, in <listcomp>
    for level, index in enumerate(key)]
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/pyspark/sql/column.py", line 115, in _
    njc = getattr(self._jc, name)(jc)
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/py4j/java_gateway.py", line 1248, in __call__
    args_command, temp_args = self._build_args(*args)
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/py4j/java_gateway.py", line 1212, in _build_args
    (new_args, temp_args) = self._get_args(args)
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/py4j/java_gateway.py", line 1199, in _get_args
    temp_arg = converter.convert(arg, self.gateway_client)
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/py4j/java_collections.py", line 512, in convert
    for key in object.keys():
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/databricks/koalas/indexes.py", line 492, in __iter__
    return _MissingPandasLikeIndex.__iter__(self)
  File "/Users/harutaka/.pyenv/versions/anaconda3-5.3.1/envs/kp-dev-env/lib/python3.6/site-packages/databricks/koalas/missing/__init__.py", line 24, in unsupported_function
    reason=reason)
databricks.koalas.exceptions.PandasNotImplementedError: The method `pd.Index.__iter__()` is not implemented. If you want to collect your data as an NumPy array, use 'to_numpy()' instead.
```

harupy:
Resolves #1073

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4383,11 +4383,17 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         return len(self.to_dataframe())
 
     def __getitem__(self, key):
+        if isinstance(key, Series):
+            bcol = key._scol.cast(BooleanType())
+            sdf = self._internal.sdf.filter(bcol)
+            internal = self._internal.copy(sdf=sdf)
+            return _col(DataFrame(internal))
+
         if not isinstance(key, tuple):
             key = (key,)
         if len(self._internal._index_map) < len(key):
             raise KeyError("Key length ({}) exceeds index depth ({})"
-                           .format(len(key), len(self._internal.index_map)))
+                        .format(len(key), len(self._internal.index_map)))
 
         cols = (self._internal.index_scols[len(key):] +
                 [self._internal.scol_for(self._internal.column_index[0])])
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4393,7 +4393,7 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
             key = (key,)
         if len(self._internal._index_map) < len(key):
             raise KeyError("Key length ({}) exceeds index depth ({})"
-                        .format(len(key), len(self._internal.index_map)))
+                           .format(len(key), len(self._internal.index_map)))
 
         cols = (self._internal.index_scols[len(key):] +
                 [self._internal.scol_for(self._internal.column_index[0])])
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4215,11 +4215,10 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         return len(self.to_dataframe())
 
     def __getitem__(self, key):
-        if isinstance(key, Series):
-            bcol = key._scol.cast(BooleanType())
-            sdf = self._internal.sdf.filter(bcol)
-            internal = self._internal.copy(sdf=sdf)
-            return _col(DataFrame(internal))
+        if isinstance(key, Series) and isinstance(key.spark_type, BooleanType):
+            self._kdf["__temp_col__"] = key
+            sdf = self._kdf._sdf.filter(F.col("__temp_col__")).drop("__temp_col__")
+            return _col(DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map)))
 
         if not isinstance(key, tuple):
             key = (key,)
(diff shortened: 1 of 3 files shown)


harupy:
ping @HyukjinKwon @ueshin

databricks/koalas/series.py
```diff
@@ -4383,6 +4383,12 @@ def __len__(self):
         return len(self.to_dataframe())
 
     def __getitem__(self, key):
+        if isinstance(key, Series):
```
HyukjinKwon: @harupy, can we also check if the given Series's type is `bool` to prevent such cases:

```python
>>> pd.DataFrame({'a': [1,2,3]})[pd.Series(['a'])]
   a
0  1
1  2
2  3
```

databricks/koalas/series.py
```diff
@@ -4383,6 +4383,12 @@ def __len__(self):
...
+            bcol = key._scol.cast(BooleanType())
+            sdf = self._internal.sdf.filter(bcol)
```
HyukjinKwon: Also, ideally we should also consider about different Series from different DataFrames.

We can do it via using existing logic that handles such series from different DataFrames:

```python
// To allows 'compute.ops_on_diff_frames', it assigns the given column to the current Frame.
self._kdf["__temp_col__"] = key
sdf = self._internal.sdf.filter(F.col("__temp_col__"))
internal = self._internal.copy(sdf=sdf)
...
```

^^ It's not tested so should be double checked.

databricks/koalas/series.py
harupy: will fix it

databricks/koalas/series.py
harupy: Thanks for the comment. Does the code above work for both different and same dataframes?

HyukjinKwon: Yeah, I think so because I used such similar trick (e.g., https://github.com/databricks/koalas/commit/7070bc6e29dd2ae628b51efa2972112e6b833bff#diff-5ba644c40e914732959b9b4867f30c53R2158)

HyukjinKwon: We could add one test under `OpsOnDiffFramesEnabledTest` (e.g., `kser[ks.Series([True])]`)

HyukjinKwon: Oh! wait, can we reuse `where`?

harupy: Let me try

harupy: `self.to_dataframe().where(...)` didn't work. It just fills `False` with `NaN`.

databricks/koalas/series.py
```diff
@@ -4215,6 +4215,11 @@ def __len__(self):
         return len(self.to_dataframe())
 
     def __getitem__(self, key):
+        if isinstance(key, Series) and isinstance(key.spark_type, BooleanType):
+            self._kdf["__temp_col__"] = key
+            sdf = self._kdf._sdf.filter(F.col("__temp_col__")).drop("__temp_col__")
+            return _col(DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map)))
```
HyukjinKwon: hm, actually it should set column index. In `where` too .. I see some related problems .. let me fix it separately.

PR #1077, opened 2019-11-26
Add option_context.

ueshin:


docs/source/reference/general_functions.rst
```diff
@@ -14,6 +14,7 @@ Working with options
     reset_option
     get_option
     set_option
+    option_context
```
HyukjinKwon: I think we should document at `docs/source/user_guide/options.rst` too. Seems pandas has it too (https://pandas.pydata.org/pandas-docs/stable/user_guide/options.html).

Maybe we can fix some instances at "Operations on different DataFrames" too .. ?

HyukjinKwon: There seems one more instance in 10 mins too.

ueshin: I don't think we should add to 10 mins. It's an example for spark conf.

PR #1078, opened 2019-11-26
Fix idxmax() / idxmin() for Series work properly

itholic:
Reopen of #1065

fix more properly with considering of more examples commented at https://github.com/databricks/koalas/pull/1065#issuecomment-558303388

```python
>>> pser = pd.Series([1, 100, None, 100, 1, 100], index=[10, 3, 5, 2, 1, 8])
>>> kser = ks.from_pandas(pser)
>>>
>>> pser.idxmax()
3
>>> kser.idxmax()
3
>>> pser.idxmin()
10
>>> kser.idxmin()
10
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -3165,9 +3165,9 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         # desc_nulls_(last|first) is used via Py4J directly because
         # it's not supported in Spark 2.3.
         if skipna:
-            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_last()))
+            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_last()), *index_scols)
         else:
-            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_first()))
+            sdf = sdf.orderBy(Column(scol._jc.desc_nulls_first()), *index_scols)
         results = sdf.select([scol] + index_scols).take(1)
         if len(results) == 0:
             raise ValueError("attempt to get idxmin of an empty sequence")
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -3158,34 +3158,47 @@ def idxmax(self, skipna=True):
...
+        2     100
+        1       1
+        8     100
+        Name: 0, dtype: int64
+
+        >>> s.idxmax()
+        3
         """
+        if len(self) == 0:
+            raise ValueError("attempt to get idxmin of an empty sequence")
```
ueshin: Why moved here?

itholic: Ah, at the first time i thought it can reduce the cost since `if len(self) == 0` then we can skip all process after this line,

but now i'm thinking maybe i missed case that if the size of `self` is extremely big, it can rather reduce the performance.

i'll move this logic to more proper place.

HyukjinKwon:
Can you make tests passed at Python 3.5 too @itholic?

databricks/koalas/series.py
```diff
@@ -3250,28 +3265,43 @@ def idxmin(self, skipna=True):
...
+        >>> s.idxmin()
+        10
         """
         sdf = self._internal._sdf
         scol = self._scol
         index_scols = self._internal.index_scols
-        # asc_nulls_(list|first)is used via Py4J directly because
-        # it's not supported in Spark 2.3.
         if skipna:
-            sdf = sdf.orderBy(Column(scol._jc.asc_nulls_last()))
```
HyukjinKwon: @itholic the test failure against Spark 2.3.x is because you replaced this `Column(scol._jc.asc_nulls_last())` -> `scol.asc_nulls_last()`.

HyukjinKwon: > asc_nulls_(list|first)is used via Py4J directly because it's not supported in Spark 2.3.

ueshin:
What do you think about https://github.com/databricks/koalas/pull/1078#pullrequestreview-323984727?

> I guess we should just use F.monotonically_increasing_id() instead of index_scols in the previous PR #1065?

itholic:
@ueshin Ah! sorry i got missed that comment. 😢 

It makes sense. Thanks for the comment i fix it!! 👍

ueshin:
Thanks! merging.

itholic:
Thanks :)

PR #1086, opened 2019-12-01
Implement MultiIndex.levshape

RainFung:
Get MultiIndex unique value count in every level.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -821,7 +821,7 @@ class MultiIndex(Index):
                     ('c', 'z')],
                    )
 
-        >>> midx.levshape
+        >>> midx.levshape # doctest: +SKIP
         (3,3)
         """
         result = self._internal._sdf\
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -845,7 +845,7 @@ class MultiIndex(Index):
                     ('c', 'z')],
                    )
 
-        >>> midx.levshape # doctest: +SKIP
+        >>> midx.levshape
         (3, 3)
         """
         internal = self._internal
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -832,25 +832,9 @@ class MultiIndex(Index):
         raise PandasNotImplementedError(class_name='pd.MultiIndex', property_name='name')
 
     @property
-    def levshape(self, i=-2, j=-1):
+    def levshape(self):
         """
         A tuple with the length of each level.
-        Swap level i with level j.
-        Calling this method does not change the ordering of the values.
-
-        Parameters
-        ----------
-        i : int, str, default -2
-            First level of index to be swapped. Can pass level name as string.
-            Type of parameters can be mixed.
-        j : int, str, default -1
-            Second level of index to be swapped. Can pass level name as string.
-            Type of parameters can be mixed.
-
-        Returns
-        -------
-        MultiIndex
-            A new MultiIndex.
 
         Examples
         --------
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -807,6 +807,27 @@ class MultiIndex(Index):
     def name(self, name: str) -> None:
         raise PandasNotImplementedError(class_name='pd.MultiIndex', property_name='name')
 
+    @property
+    def levshape(self) -> tuple:
+        """
+        A tuple with the length of each level.
+
+        Examples
+        --------
+        >>> midx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
+        >>> midx  # doctest: +SKIP
+        MultiIndex([('a', 'x'),
+                    ('b', 'y'),
+                    ('c', 'z')],
+                   )
+
+        >>> midx.levshape
+        (3,3)
+        """
+        result = self._internal._sdf\
+            .agg(*(F.countDistinct(F.col(c)) for c in self._internal.index_columns)).collect()
+        return tuple(v for v in result[0])
+
     def to_pandas(self) -> pd.MultiIndex:
         """
         Return a pandas MultiIndex.
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -846,11 +846,11 @@ class MultiIndex(Index):
                    )
 
         >>> midx.levshape # doctest: +SKIP
-        (3,3)
+        (3, 3)
         """
-        result = self._internal._sdf\
-            .agg(*(F.countDistinct(F.col(c)) for c in self._internal.index_columns)).collect()
-        return tuple(v for v in result[0])
+        internal = self._internal
+        result = internal._sdf.agg(*(F.countDistinct(c) for c in internal.index_scols)).collect()[0]
+        return tuple(result)
 
     def to_pandas(self) -> pd.MultiIndex:
         """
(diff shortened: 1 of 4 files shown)


databricks/koalas/indexes.py
```diff
@@ -807,6 +807,27 @@ def name(self) -> str:
...
+        --------
+        >>> midx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
+        >>> midx  # doctest: +SKIP
+        MultiIndex([('a', 'x'),
+                    ('b', 'y'),
+                    ('c', 'z')],
+                   )
+
+        >>> midx.levshape
+        (3,3)
```
itholic: i think we skip this also.

Seems that it have different print out behavior depends on python version.

`(3,3)` for 3.6, but `(3, 3)` for 3.5 & 3.7

RainFung: Ok, it's a common issue in tuple.

RainFung:
Tuple issue caused by python version need help.

databricks/koalas/indexes.py
```diff
@@ -832,6 +832,27 @@ def name(self) -> str:
...
+        Examples
+        --------
+        >>> midx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
+        >>> midx  # doctest: +SKIP
+        MultiIndex([('a', 'x'),
+                    ('b', 'y'),
+                    ('c', 'z')],
+                   )
+
+        >>> midx.levshape
```
HyukjinKwon: Can you skip this test then? Seems the results varies due to Python version as pointed out. You can do `  # doctest: +SKIP`

databricks/koalas/indexes.py
```diff
@@ -831,6 +831,27 @@ def name(self) -> str:
     def name(self, name: str) -> None:
         raise PandasNotImplementedError(class_name='pd.MultiIndex', property_name='name')
 
+    @property
+    def levshape(self) -> tuple:
```
HyukjinKwon: Seems you should remove `levshape` in `_MissingPandasLikeMultiIndex`

RainFung:
@itholic  Can you help me solve the unknown failing problem? Thanks.

databricks/koalas/missing/indexes.py
```diff
@@ -119,7 +119,7 @@ class _MissingPandasLikeMultiIndex(object):
 
     # Properties
     is_all_dates = unsupported_property('is_all_dates')
-    levshape = unsupported_property('levshape')
+    shape = unsupported_property('shape')
```
itholic: why do you add this?? i thnik it is already implemented ?

```python
>>> index = Index([1, 2, 3])
>>> index.shape
(3,)
```

not pretty sure, but i guess maybe that is the reason of build failure  :)

itholic: in the above question, "method for checking the python version" means like that?

```python
>>> import sys
>>> sys.version
'3.7.2 (v3.7.2:9a3ffc0492, Dec 24 2018, 02:44:43) \n[Clang 6.0 (clang-600.0.57)]'
```

databricks/koalas/indexes.py
ueshin: What's going on here?
I don't think the tuple string is not changed throughout 3.5 - 3.7.

```py
Python 3.5.6 |Anaconda, Inc.| (default, Aug 26 2018, 16:30:03)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> (3,3)
(3, 3)
```

```py
Python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 13:42:17)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> (3,3)
(3, 3)
```

```py
Python 3.7.5 (default, Oct 25 2019, 10:52:18)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> (3,3)
(3, 3)
```

databricks/koalas/tests/test_indexes.py
```diff
@@ -150,6 +151,14 @@ def test_multi_index_names(self):
         with self.assertRaises(PandasNotImplementedError):
             kidx.name = 'renamed'
 
+    def test_multi_index_levshape(self):
+        pidx = pd.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2)])
+        kidx = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2)])
+        if sys.version[0:3] == '3.6':
+            self.assertEqual(pidx.levshape, kidx.levshape)
+        else:
+            self.assertEqual((3, 3), kidx.levshape)
```
ueshin: Why do we need to special handle 3.6?

```py
Python 3.6.8 |Anaconda, Inc.| (default, Dec 29 2018, 19:04:46)
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> pd.MultiIndex.from_tuples([('x', 'a'), ('x', 'b')]).levshape
(1, 2)
```

```py
Python 3.7.5 (default, Oct 25 2019, 10:52:18)
[Clang 4.0.1 (tags/RELEASE_401/final)] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pandas as pd
>>> pd.MultiIndex.from_tuples([('x', 'a'), ('x', 'b')]).levshape
(1, 2)
```

seems same?

databricks/koalas/indexes.py
ueshin: It should be `(3, 3)` and we should not skip it.

databricks/koalas/indexes.py
```diff
@@ -831,6 +831,27 @@ def name(self) -> str:
...
+        MultiIndex([('a', 'x'),
+                    ('b', 'y'),
+                    ('c', 'z')],
+                   )
+
+        >>> midx.levshape # doctest: +SKIP
+        (3,3)
+        """
+        result = self._internal._sdf\
+            .agg(*(F.countDistinct(F.col(c)) for c in self._internal.index_columns)).collect()
```
ueshin: `self._internal.index_scols` instead of `self._internal.index_columns`?

databricks/koalas/indexes.py
```diff
@@ -831,6 +831,27 @@ def name(self) -> str:
...
+        return tuple(v for v in result[0])
```
ueshin: `tuple(result[0])` should work?

databricks/koalas/indexes.py
ueshin: ah, I see, the travis tests for the py3.6 environment are wrong.

ueshin: Submitted a PR #1107.

itholic: Ah, that was a problem. Thanks for fixing @ueshin :+1:

databricks/koalas/indexes.py
```diff
@@ -831,6 +831,27 @@ def name(self) -> str:
...
+        Examples
+        --------
+        >>> midx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
+        >>> midx  # doctest: +SKIP
+        MultiIndex([('a', 'x'),
+                    ('b', 'y'),
+                    ('c', 'z')],
+                   )
+
+        >>> midx.levshape # doctest: +SKIP
```
ueshin: Could you remove `# doctest: +SKIP`? It should work.

ueshin:
Thanks! merging.

PR #1087, opened 2019-12-01
Document MultiIndex.nlevels

RainFung:


databricks/koalas/indexes.py
```diff
@@ -807,6 +807,25 @@ def name(self) -> str:
     def name(self, name: str) -> None:
         raise PandasNotImplementedError(class_name='pd.MultiIndex', property_name='name')
 
+    @property
+    def nlevels(self) -> int:
```
itholic: i think we already implemented at #945 ?

now `MultiIndex` is using `Index`'s `nlevels` directly, since `MultiIndex` inherits `Index` and they have exactly same logic.

docs/source/reference/indexing.rst
```diff
@@ -96,6 +96,7 @@ MultiIndex Properties
    MultiIndex.names
    MultiIndex.ndim
    MultiIndex.T
+   MultiIndex.nlevels
```
itholic: But this addition seems good to me.

i think we can merge this PR with this only, except in `indexes.py`

RainFung:
Can you check the Travis CI problem?

itholic:
@RainFung Okay, let's figure out how can we handle this problem together :)

HyukjinKwon:
Let me merge this one - I don't think it's related to the build failure.

PR #1091, opened 2019-12-01
implement index.droplevel & multiindex.droplevel

RainFung:
I wonder know how to implement droplevel both in index and multiindex like Pandas api.
https://pandas.pydata.org/pandas-docs/stable/reference/indexing.html#index


databricks/koalas/indexes.py
```diff
@@ -587,6 +587,67 @@ def copy(self, name=None):
...
+                    ('y', 2)],
+                   names=['level2', 'level3'])
+        """
+        if not isinstance(level, (tuple, list)):
+            level = [level]
+
+        for l in level:
+            if isinstance(l, int) and (l > self.nlevels - 1):
+                raise IndexError("Too many levels: Index has only {} levels, not {}"
+                                 .format(self.nlevels, l))
```
itholic: maybe

`.format(self.nlevels, l + 1))` rather than `.format(self.nlevels, l))` ??

because pandas are doing like this.

```python
>>> pidx
MultiIndex([('a', 'x', 1),
            ('b', 'y', 2)],
           names=['level1', 'level2', 'level3'])
>>> pidx.droplevel(4)
...
packages/pandas/core/indexes/multi.py", line 1303, in _get_level_number
    "not %d" % (self.nlevels, level + 1)
IndexError: Too many levels: Index has only 3 levels, not 5
```

databricks/koalas/tests/test_indexes.py
```diff
@@ -319,3 +319,22 @@ def test_multiindex_from_arrays(self):
...
 
         self.assert_eq(pidx, kidx)
+
+    def test_multiindex_droplevel(self):
+        pidx = pd.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2)],
+                                         names=['level1', 'level2', 'level3'])
+        kidx = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2)],
+                                         names=['level1', 'level2', 'level3'])
+        with self.assertRaisesRegex(IndexError,
+                                    "Too many levels: Index has only 3 levels, not 4"):
```
itholic: according to https://github.com/databricks/koalas/pull/1091/files#r352351964,

i think this should be better changed to `"Too many levels: Index has only 3 levels, not 5"`

databricks/koalas/indexes.py
```diff
@@ -587,6 +587,67 @@ def copy(self, name=None):
...
+            raise ValueError(
+                "Cannot remove {} levels from an index with {} "
+                "levels: at least one level must be "
+                "left.".format(len(level), self.nlevels)
+            )
+
+        index_columns = [v for k, v in enumerate(self._internal.index_columns)
+                         if ((k not in level) & (v not in level))]
+        internal = _InternalFrame(sdf=self._internal._sdf.select(index_columns),
+                                  index_map=[(i, (i,)) for i in index_columns])
```
itholic: let's consider about `kidx = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2)])` also ?

the `index_map` should be like below.
```
[('__index_level_0__', None), ('__index_level_1__', None), ('__index_level_2__', None)]
```

but maybe now we generate
```
[('__index_level_0__', ('__index_level_0__',)), ('__index_level_1__', ('__index_level_1__',)), ('__index_level_2__', ('__index_level_2__',))]
```

RainFung: image
Maybe you missed the index `names`.

RainFung:
The code passed the lastest Travis CI test, but failed in old version. It's so confused how to deal with the py version problem.

databricks/koalas/indexes.py
itholic: yeah, i didn't use `names` intentionally since i think we should consider this case also.

when `names` is not specified, maybe we should make index_map like below?

```python
index_map=[(i, None) for i in index_columns]
```

RainFung: Yes, you are right. Now `name` is not supported well.

HyukjinKwon:
cc @ueshin

databricks/koalas/indexes.py
```diff
@@ -587,6 +587,66 @@ def copy(self, name=None):
...
+            raise ValueError(
+                "Cannot remove {} levels from an index with {} "
+                "levels: at least one level must be "
+                "left.".format(len(level), self.nlevels)
+            )
+
+        index_columns = [v for k, v in enumerate(self._internal.index_columns)
+                         if ((k not in level) & (v not in level))]
+        internal = _InternalFrame(sdf=self._internal._sdf.select(index_columns),
+                                  index_map=[(i, None) for i in index_columns])
```
ueshin: We should preserve the index names if the index are named.

```py
>>> midx
MultiIndex(levels=[['a', 'x'], ['b', 'y'], [1, 2]],
           codes=[[0, 1], [0, 1], [0, 1]],
           names=['l1', 'l2', 'l3'])
>>> midx.droplevel(0)
MultiIndex(levels=[['b', 'y'], [1, 2]],
           codes=[[0, 1], [0, 1]])
```

whereas

```py
>>> midx.to_pandas().droplevel(0)
MultiIndex(levels=[['b', 'y'], [1, 2]],
           codes=[[0, 1], [0, 1]],
           names=['l2', 'l3'])
```

databricks/koalas/indexes.py
```diff
@@ -587,6 +587,66 @@ def copy(self, name=None):
...
+                   )
+        """
+        if not isinstance(level, (tuple, list)):
+            level = [level]
+
+        for l in level:
+            if isinstance(l, int) and (l > self.nlevels - 1):
+                raise IndexError("Too many levels: Index has only {} levels, not {}"
+                                 .format(self.nlevels, l+1))
+            if isinstance(l, str) and (l not in self._internal.index_columns):
```
ueshin: `self._internal.index_names` instead of `self._internal.index_columns`?

```py
>>> midx = ks.DataFrame({'a': ['a', 'b']}, index=[['a', 'x'], ['b', 'y'], [1, 2]]).index
>>> midx.names = ['l1', 'l2', 'l3']
>>> midx
MultiIndex(levels=[['a', 'x'], ['b', 'y'], [1, 2]],
           codes=[[0, 1], [0, 1], [0, 1]],
           names=['l1', 'l2', 'l3'])
>>> midx.droplevel('l1')
  File "/Users/ueshin/workspace/databricks-koalas/worktrees/work/databricks/koalas/indexes.py", line 631, in droplevel
    raise KeyError("Level {} not found".format(l))
KeyError: 'Level l1 not found'
```

databricks/koalas/indexes.py
```diff
@@ -587,6 +587,66 @@ def copy(self, name=None):
...
+                raise KeyError("Level {} not found".format(l))
+
```
ueshin: Need to add a check to verify the tupled name is valid?

```py
>>> midx
MultiIndex(levels=[['a', 'x'], ['b', 'y'], [1, 2]],
           codes=[[0, 1], [0, 1], [0, 1]],
           names=[('x', 'l1'), 'l2', 'l3'])
>>> midx.droplevel([('x', 'l1')])
MultiIndex(levels=[['a', 'x'], ['b', 'y'], [1, 2]],
           codes=[[0, 1], [0, 1], [0, 1]])
```

databricks/koalas/indexes.py
itholic: ah sorry, @RainFung 

what i meant at https://github.com/databricks/koalas/pull/1091#discussion_r352353249,

we should consider the both case when index has its own name or not.

your existing code in previous comment could handle only the case index has own name.

so we should consider both case.

databricks/koalas/indexes.py
```diff
@@ -587,6 +587,66 @@ def copy(self, name=None):
...
+        >>> midx.droplevel(0) # doctest: +SKIP
+        MultiIndex([('b', 1),
+                    ('y', 2)]
+                   )
+        """
+        if not isinstance(level, (tuple, list)):
+            level = [level]
+
+        for l in level:
```
itholic: not to be necessarily fixed, just IMO,  it should better avoid `l` for naming as possible.

because It is often confused with the number `1` , and can impair readability. 😄

databricks/koalas/indexes.py
```diff
@@ -641,6 +641,71 @@ def copy(self, name=None):
...
+                "left.".format(len(level), nlevels)
+            )
+        int_level = [n if isinstance(n, int) else names.index(n) for n in level]
+        index_columns = self._internal.index_columns.copy()
+        index_columns = [index_columns[c] for c in range(0, nlevels)
+                         if c not in int_level]
+
+        index_map = self._internal.index_map.copy()
+        index_map = [index_map[c] for c in range(0, nlevels) if c not in int_level]
+        result = _InternalFrame(sdf=self._internal._sdf.select(index_columns), index_map=index_map)
```
ueshin: What if all the index columns are dropped?

databricks/koalas/indexes.py
```diff
@@ -641,6 +641,71 @@ def copy(self, name=None):
...
+        if len(index_columns) > 1:
+            result = MultiIndex(DataFrame(result))
+        else:
+            result = Index(DataFrame(result))
+        return result
```
ueshin:
```py
return DataFrame(result).index
```

should work?

databricks/koalas/indexes.py
```diff
@@ -641,6 +641,71 @@ def copy(self, name=None):
...
+        if len(level) >= nlevels:
+            raise ValueError(
+                "Cannot remove {} levels from an index with {} "
+                "levels: at least one level must be "
+                "left.".format(len(level), nlevels)
+            )
+        int_level = [n if isinstance(n, int) else names.index(n) for n in level]
+        index_columns = self._internal.index_columns.copy()
+        index_columns = [index_columns[c] for c in range(0, nlevels)
+                         if c not in int_level]
```
ueshin: We should use `self._internal.index_scols` instead of `self._internal.index_columns`.

HyukjinKwon:
Let me take this over.

HyukjinKwon:
Let's track it in https://github.com/databricks/koalas/pull/1340.

PR #1096, opened 2019-12-03
Add a basic NumPy ufunc compatibility

HyukjinKwon:
This PR implements `__array_ufunc__`  (see https://docs.scipy.org/doc/numpy/reference/ufuncs.html#output-type-determination) to allow some of basic ufunc can run against Koalas Series and Index (some dunder APIs).

```python
>>> import numpy as np
>>> kdf = ks.range(10)
>>> kdf = np.add(kdf.id, kdf.id)
>>> type(kdf)
<class 'databricks.koalas.series.Series'>
>>> kdf
0     0
1     2
2     4
3     6
4     8
5    10
6    12
7    14
8    16
9    18
Name: id, dtype: int64
```

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -24,7 +24,6 @@ from typing import Union, Callable, Any
 import numpy as np
 import pandas as pd
 from pandas.api.types import is_list_like
-from pandas.core import ops
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Window
 from pyspark.sql.types import DoubleType, FloatType, LongType, StringType, TimestampType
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Tests passed

databricks/koalas/base.py
```diff
@@ -224,6 +289,16 @@ def __rfloordiv__(self, other):
...
     __ror__ = _column_op(spark.Column.__ror__)
 
+    # NDArray Compat
+    def __array_ufunc__(self, ufunc: Callable, method: str, *inputs: Any, **kwargs: Any):
+        result = maybe_dispatch_ufunc_to_dunder_op(
+            self, ufunc, method, *inputs, **kwargs)
+        if result is not NotImplemented:
+            return result
+        else:
+            # TODO: support more APIs?
```
ueshin: Maybe we can delegate to pandas UDF?

databricks/koalas/base.py
```diff
@@ -36,6 +36,71 @@
...
+
+    if method == "__call__" and op_name in special and kwargs.get("out") is None:
+        if isinstance(inputs[0], type(self)):
+            name = "__{}__".format(op_name)
+            return getattr(self, name, not_implemented)(inputs[1])
+        else:
+            name = flipped.get(op_name, "__r{}__".format(op_name))
+            return getattr(self, name, not_implemented)(inputs[0])
+    else:
+        return NotImplemented
```
ueshin: We will be able to add more functions supported in Spark natively.

ueshin:
Thanks! I'd merge this for now as a basic of NumPy ufunc compatibility.

databricks/koalas/base.py
HyukjinKwon: Oh that's nice suggestion. Let me investigate a bit more about this. It will only work when the output is n to n but I'm sure there will be the case.

ueshin: Yeah, if we can delegate to pandas UDF in a general way, we can take time to add more Spark native functions like `np.sqrt` or `np.log`.

PR #1100, opened 2019-12-03
Avoid to return a different DataFrame from Series where (and multiple small bug fixes)

HyukjinKwon:
**1..** We can avoid operations on different DataFrames. See below:

```python
>>> kdf = ks.DataFrame({'a': [1,2,3]})
>>> kdf.a.where(kdf.a > 1)
  File "/.../koalas/databricks/koalas/series.py", line 3916, in where
    kdf['__tmp_cond_col__'] = cond
  File "/.../koalas/databricks/koalas/frame.py", line 8089, in __setitem__
    kdf = align_diff_frames(assign_columns, self, value, fillna=False, how="left")
  File "/.../koalas/databricks/koalas/utils.py", line 187, in align_diff_frames
    combined = combine_frames(this, that, how=how)
  File "/.../koalas/databricks/koalas/utils.py", line 122, in combine_frames
    "Cannot combine the series or dataframe because it comes from a different dataframe. "
ValueError: Cannot combine the series or dataframe because it comes from a different dataframe. In order to allow this operation, enable 'compute.ops_on_diff_frames' option.
```

`kdf.a` is from the same DataFrame but currently it requires `compute.ops_on_diff_frames` to be enabled.

This is because `self.to_frame()` creates another DataFrame and it loses it's anchor DataFrame; therefore, it became a different DataFrame.

**2..** We should set `column_index` and `column_index_names`. Otherwise, output DataFrames will lose column indexes (e.g., multi-index columns).

**3..** When we should use `self._kdf` in `Series` API implementation, we should keep check if it matters at output data column because, if we apply some APIs that do not need to create another DataFrame (which is preferred), it will ignore the applied column. See the example below:

```python
ser = df.a.round()  # `kdf` is as is but it only replaces `scol`.
ser._kdf.filter(...).show()  # `scol` is missing in the output
```

**4..** Currently, `compute.ops_on_diff_frames` should only be enabled in `OpsOnDiffFramesEnabledTest`. This option is disabled by default and discouraged due to performance issue.

It was mistakenly enabled at `SeriesTest`. I removed it back.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -3916,8 +3916,10 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
         assert isinstance(cond, Series)
 
-        should_try_ops_on_diff_frame = \
-            cond._kdf is not self._kdf or isinstance(other, Series) and other._kdf is not self._kdf
+        # We should check the DataFrame from both `cond` and `other`.
+        should_try_ops_on_diff_frame = (
+                cond._kdf is not self._kdf or
+                (isinstance(other, Series) and other._kdf is not self._kdf))
 
         if should_try_ops_on_diff_frame:
             # Try to perform it with 'compute.ops_on_diff_frame' option.
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -3918,8 +3918,8 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
         # We should check the DataFrame from both `cond` and `other`.
         should_try_ops_on_diff_frame = (
-                cond._kdf is not self._kdf or
-                (isinstance(other, Series) and other._kdf is not self._kdf))
+            cond._kdf is not self._kdf or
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
cc @itholic and @harupy FYI

databricks/koalas/series.py
```diff
@@ -3912,28 +3912,47 @@ def where(self, cond, other=np.nan):
...
-        condition = F.when(sdf['__tmp_cond_col__'], sdf[data_col_name]) \
-                     .otherwise(sdf['__tmp_other_col__']).alias(data_col_name)
-        sdf = sdf.select(index_column, condition)
-        result = _col(ks.DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map)))
 
-        return result
+        assert isinstance(cond, Series)
+
+        should_try_ops_on_diff_frame = \
+            cond._kdf is not self._kdf or isinstance(other, Series) and other._kdf is not self._kdf
```
ueshin: Could you use parentheses to clarify the range of `or` and `and` conditions?

PR #1102, opened 2019-12-03
Implement Index.fillna

RainFung:
Implement Index.fillna via spark.DataFrame.fillna


databricks/koalas/indexes.py
```diff
@@ -360,6 +360,36 @@ def rename(self, name: Union[str, Tuple[str, ...]], inplace: bool = False):
...
+        >>> ki
+        Float64Index([1.0, 2.0, nan], dtype='float64')
+
+        >>> ki.fillna(0)
+        Float64Index([1.0, 2.0, 0.0], dtype='float64')
+        """
+        if not isinstance(value, (float, int, str, bool)):
+            raise TypeError("Unsupported type %s" % type(value))
+        sdf = self._internal.sdf.fillna(value)
+        result = Index(DataFrame(_InternalFrame(sdf=sdf, index_map=self._internal.index_map)))
```
itholic: i think it will be lost name of index.

```python
>>> import numpy as np

>>> kidx = ks.Index([1.0, 2.0, np.nan], name='koalas')
>>> kidx.fillna(10000)
Float64Index([1.0, 2.0, 10000.0], dtype='float64')
```

i think you can

```python
DataFrame(self._kdf._internal.copy(sdf=sdf)).index
```

databricks/koalas/tests/test_indexes.py
```diff
@@ -319,3 +319,12 @@ def test_multiindex_from_arrays(self):
         kidx = ks.MultiIndex.from_arrays(arrays)
 
         self.assert_eq(pidx, kidx)
+
+    def test_index_fillna(self):
+        pidx = pd.DataFrame({'a': ['a', 'b', 'c']}, index=[1, 2, None]).index
+        kidx = ks.DataFrame({'a': ['a', 'b', 'c']}, index=[1, 2, None]).index
+
+        self.assert_eq(pidx.fillna(0), kidx.fillna(0))
+
```
itholic: and let's add test with name here around.

for example,

```python
pidx = pidx.rename('name')
kidx = pidx.rename('name')

self.assert_eq(pidx.fillna(0), kidx.fillna(0))
```

ueshin:
LGTM.

ueshin:
Thanks! merging.

PR #1107, opened 2019-12-04
Add __doc__ for replaced property when logging is enabled.

ueshin:
When logging is enabled, docs for replaced properties didn't exist, and as a side-effect, doctests for the properties were not run.


ueshin:
I'd merge this for now.

PR #1108, opened 2019-12-05
Set the pyarrow version upper bound again.

ueshin:
Seems still unstable with PyArrow 0.15. Let me set the upper bound again for now.


ueshin:
I'd merge this for now since the tests passed.

PR #1116, opened 2019-12-11
Fix value_counts() to work properly when dropna is True

itholic:
This PR Resolves comment https://github.com/databricks/koalas/pull/949#discussion_r356364770

```python
>>> kdf
   a    b
0  1  NaN
1  2  1.0
2  3  NaN

>>> kdf.a
0    1
1    2
2    3
Name: a, dtype: int64

>>> kdf.a.value_counts()
2    1
3    1
1    1
Name: a, dtype: int64
```

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -943,7 +943,8 @@ class IndexOpsMixin(object):
         from databricks.koalas.series import Series, _col
         if bins is not None:
             raise NotImplementedError("value_counts currently does not support bins")
-        # if Series from DataFrame, we should've specified a proper subset.
+
+        # if given Series is a subset of DataFrame, we should've specified a proper subset.
         if self._kdf._internal is not self._internal:
             self = _col(self.to_frame())
         if dropna:
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -947,10 +947,10 @@ class IndexOpsMixin(object):
         if dropna:
             sdf_dropna = self._internal._sdf.select(self._scol).dropna()
         else:
-            sdf_dropna = self._internal._sdf
+            sdf_dropna = self._internal._sdf.select(self._scol)
         index_name = SPARK_INDEX_NAME_FORMAT(0)
         column_name = self._internal.data_columns[0]
-        sdf = sdf_dropna.groupby(sdf_dropna[column_name].alias(index_name)).count()
+        sdf = sdf_dropna.groupby(scol_for(sdf_dropna, column_name).alias(index_name)).count()
         if sort:
             if ascending:
                 sdf = sdf.orderBy(F.col('count'))
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -949,7 +949,8 @@ class IndexOpsMixin(object):
         else:
             sdf_dropna = self._internal._sdf
         index_name = SPARK_INDEX_NAME_FORMAT(0)
-        sdf = sdf_dropna.groupby(sdf_dropna[self.name].alias(index_name)).count()
+        column_name = self._internal.data_columns[0]
+        sdf = sdf_dropna.groupby(sdf_dropna[column_name].alias(index_name)).count()
         if sort:
             if ascending:
                 sdf = sdf.orderBy(F.col('count'))
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -944,15 +944,12 @@ class IndexOpsMixin(object):
         if bins is not None:
             raise NotImplementedError("value_counts currently does not support bins")
 
-        # if given Series is a subset of DataFrame, we should've specified a proper subset.
-        if self._kdf._internal is not self._internal:
-            self = _col(self.to_frame())
         if dropna:
-            sdf_dropna = self._internal._sdf.dropna()
+            sdf_dropna = self._internal._sdf.select(self._scol).dropna()
         else:
             sdf_dropna = self._internal._sdf
         index_name = SPARK_INDEX_NAME_FORMAT(0)
-        sdf = sdf_dropna.groupby(self._scol.alias(index_name)).count()
+        sdf = sdf_dropna.groupby(sdf_dropna[self.name].alias(index_name)).count()
         if sort:
             if ascending:
                 sdf = sdf.orderBy(F.col('count'))
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -20,7 +20,9 @@ Base and utility classes for Koalas objects.
 
 from functools import wraps
 from typing import Union, Callable, Any
+from distutils.version import LooseVersion
 
+import pyspark
 import numpy as np
 import pandas as pd
 from pandas.api.types import is_list_like
@@ -33,7 +35,7 @@ from databricks import koalas as ks  # For running doctests and reference resolu
 from databricks.koalas import numpy_compat
 from databricks.koalas.internal import _InternalFrame, SPARK_INDEX_NAME_FORMAT
 from databricks.koalas.typedef import pandas_wraps, spark_type_to_pandas_dtype
-from databricks.koalas.utils import align_diff_series, scol_for, validate_axis
+from databricks.koalas.utils import align_diff_series, scol_for, validate_axis, default_session
 from databricks.koalas.frame import DataFrame
 
 
@@ -941,6 +943,11 @@ class IndexOpsMixin(object):
         Name: koalas, dtype: int64
         """
         from databricks.koalas.series import Series, _col
+        if LooseVersion(pyspark.__version__) < LooseVersion("2.4") and \
+                default_session().conf.get("spark.sql.execution.arrow.enabled") == "true":
+            raise RuntimeError("if you're using pyspark < 2.4, set conf "
+                               "'spark.sql.execution.arrow.enabled' to 'false' "
+                               "for using this function")
         if bins is not None:
             raise NotImplementedError("value_counts currently does not support bins")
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/base.py
```diff
@@ -944,6 +944,9 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
         if bins is not None:
             raise NotImplementedError("value_counts currently does not support bins")
 
+        # if given Series is a subset of DataFrame, we should've specified a proper subset.
+        if self._kdf._internal is not self._internal:
+            self = _col(self.to_frame())
```
ueshin: This doesn't fix all the problems.

```py
>>> kser = ks.Series([1,2,3], index=[2,None,5])
>>> kser
2.0    1
NaN    2
5.0    3
Name: 0, dtype: int64
>>> kser.value_counts()
1    1
3    1
Name: 0, dtype: int64
```

whereas:

```py
>>> kser.to_pandas().value_counts()
3    1
2    1
1    1
Name: 0, dtype: int64
```

itholic: @ueshin thanks for the review. fixed & added tests cases :)

databricks/koalas/base.py
```diff
@@ -945,11 +945,12 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
...
 
         if dropna:
-            sdf_dropna = self._internal._sdf.dropna()
+            sdf_dropna = self._internal._sdf.select(self._scol).dropna()
         else:
             sdf_dropna = self._internal._sdf
         index_name = SPARK_INDEX_NAME_FORMAT(0)
-        sdf = sdf_dropna.groupby(self._scol.alias(index_name)).count()
+        column_name = self._internal.data_columns[0]
+        sdf = sdf_dropna.groupby(sdf_dropna[column_name].alias(index_name)).count()
```
ueshin: Although I don't remember why we changed it, maybe we can go back to the previous implementation?
https://github.com/databricks/koalas/pull/949/files#diff-f38e425003958ac0d1bd6ffa418f2613L1892-L1897
Do you remember why we changed to use `dropna`?

databricks/koalas/tests/test_series.py
```diff
@@ -269,6 +269,32 @@ def test_value_counts(self):
...
+                       pdf.a.value_counts(normalize=True), almost=True)
+        self.assert_eq(kdf.a.value_counts(ascending=True),
+                       pdf.a.value_counts(ascending=True), almost=True)
+        self.assert_eq(kdf.a.value_counts(normalize=True, dropna=False),
+                       pdf.a.value_counts(normalize=True, dropna=False), almost=True)
+        self.assert_eq(kdf.a.value_counts(ascending=True, dropna=False),
+                       pdf.a.value_counts(ascending=True, dropna=False), almost=True)
+
+        # Series with NaN index
+        pser = pd.Series([1, 2, 3], index=[2, None, 5])
```
ueshin: Could you also add some more tests for index is:

- `pd.MultiIndex.from_tuples([('x', 'a'), ('x', 'b'), ('y', 'c')])`
- `pd.MultiIndex.from_tuples([('x', 'a'), ('x', None), ('y', 'c')])`
- `pd.MultiIndex.from_tuples([('x', 'a'), None, ('y', 'c')])`

also for each case, `kser.index.value_counts(...)`

databricks/koalas/base.py
ueshin: Maybe this is the reason. https://github.com/databricks/koalas/pull/949#issuecomment-548185858

ueshin: `scol_for(sdf_dropna, column_name)` instead of `sdf_dropna[column_name]`?

itholic: yes you're right. thanks for reminding me either :)

databricks/koalas/tests/test_series.py
itholic: sure, i'll add.
thanks for reviewing 👍

databricks/koalas/base.py
```diff
@@ -941,15 +943,21 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
         Name: koalas, dtype: int64
         """
         from databricks.koalas.series import Series, _col
+        if LooseVersion(pyspark.__version__) < LooseVersion("2.4") and \
+                default_session().conf.get("spark.sql.execution.arrow.enabled") == "true":
+            raise RuntimeError("if you're using pyspark < 2.4, set conf "
+                               "'spark.sql.execution.arrow.enabled' to 'false' "
+                               "for using this function")
```
ueshin: What's this for?

databricks/koalas/tests/test_series.py
```diff
@@ -269,6 +287,122 @@ def test_value_counts(self):
...
+                       pser.index.value_counts(normalize=True), almost=True)
+        self.assert_eq(kser.index.value_counts(ascending=True),
+                       pser.index.value_counts(ascending=True), almost=True)
+        self.assert_eq(kser.index.value_counts(normalize=True, dropna=False),
+                       pser.index.value_counts(normalize=True, dropna=False), almost=True)
+        self.assert_eq(kser.index.value_counts(ascending=True, dropna=False),
+                       pser.index.value_counts(ascending=True, dropna=False), almost=True)
+
+        # Series with MultiIndex some of index is NaN.
+        # This test only available for pandas >= 0.24.
```
ueshin: Why are these only for `pandas >= 0.24`?

databricks/koalas/tests/test_series.py
```diff
@@ -244,6 +246,13 @@ def test_nunique(self):
         self.assertEqual(ks.Series(range(100)).nunique(approx=True, rsd=0.01), 100)
 
     def test_value_counts(self):
+        set_arrow_conf = False
+        if LooseVersion(pyspark.__version__) < LooseVersion("2.4") and \
+                default_session().conf.get("spark.sql.execution.arrow.enabled") == "true":
+            default_session().conf.set("spark.sql.execution.arrow.enabled", "false")
+            set_arrow_conf = True
```
ueshin: Why do we need this?
If we need this, could you make sure to set back to the original conf?

databricks/koalas/base.py
itholic: A build test in travisCI has failed when python 3.5. (https://travis-ci.com/databricks/koalas/jobs/266345663) like below.

스크린샷 2019-12-13 오전 10 33 27

so i checked, and it seems there was an issue that pyspark2.3 doesn't support manage `spark.sql.execution.arrow.enabled` automatically for struct type. (maybe pyspark2.4 manage this conf automatically)

So, this for raising proper exception message about that.

databricks/koalas/tests/test_series.py
itholic: because `pandas < 0.24` doesn't support `None` for `MultiIndex` like below way.

```python
>>> pd.__version__
'0.23.0'
>>> pser.index = pd.MultiIndex.from_tuples([('x', 'a'), None, ('y', 'c')])
...
TypeError: object of type 'NoneType' has no len()
```

so i think this test should be ran on `pandas >= 0.24`

databricks/koalas/tests/test_series.py
itholic: we need this to test at `pyspark2.3` same reason with https://github.com/databricks/koalas/pull/1116#discussion_r357451741,

i fixed the test logic to make sure setting back to original conf value.

Thanks for the review :)

databricks/koalas/base.py
ueshin: I see. The limitation is only when `MultiIndex`, right? Then we should throw only when `MultiIndex`.

itholic: @ueshin ah, that makes sense. okay i'll fix it. thanks! :)

databricks/koalas/tests/test_series.py
itholic: For now, i've fixed it to the current implementation.

databricks/koalas/tests/test_series.py
```diff
@@ -269,6 +281,132 @@ def test_value_counts(self):
...
+            default_session().conf.set("spark.sql.execution.arrow.enabled", "false")
+            try:
+                self._test_value_counts()
+            finally:
+                default_session().conf.set("spark.sql.execution.arrow.enabled", "true")
+            self.assertRaises(
+                RuntimeError,
+                lambda: ks.MultiIndex.from_tuples([('x', 'a'), ('x', 'b')]).value_counts())
+        else:
+            self._test_value_counts()
```
ueshin: Could you disable arrow execution only when pyspark < 2.4 and for MultiIndex?

databricks/koalas/base.py
```diff
@@ -941,15 +943,23 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
         Name: koalas, dtype: int64
         """
         from databricks.koalas.series import Series, _col
+        from databricks.koalas.indexes import MultiIndex
+        if LooseVersion(pyspark.__version__) < LooseVersion("2.4") and \
+                default_session().conf.get("spark.sql.execution.arrow.enabled") == "true" and \
+                isinstance(self, MultiIndex):
+            raise RuntimeError("if you're using pyspark < 2.4, set conf "
+                               "'spark.sql.execution.arrow.enabled' to 'false' "
+                               "for using this function with MultiIndex")
```
ueshin: Btw, I think we should do this only in `MultiIndex` by overriding.

itholic: ah, that seems better. thanks for the advice!

databricks/koalas/tests/test_series.py
itholic: okay i will 👍

HyukjinKwon:
Let me make a followup.

itholic:
@HyukjinKwon ah, thanks! can you cc me after finished??

PR #1117, opened 2019-12-11
Implement Index.drop

RainFung:
Implement Index.drop by using spark isin API


databricks/koalas/indexes.py
```diff
@@ -580,6 +580,33 @@ def unique(self, level=None):
...
+        >>> index = ks.Index([1, 2, 3])
+        >>> index
+        Int64Index([1, 2, 3], dtype='int64')
+
+        >>> index.drop([1])
+        Int64Index([2, 3], dtype='int64')
+        """
+        if not isinstance(labels, (tuple, list)):
+            labels = [labels]
+        sdf = self._internal.sdf[~self._internal.index_scols[0].isin(labels)]
```
itholic: i think maybe you can use below code to cover Index/MultiIndex cases :)

```python
sdf = self._internal.sdf
conds = []
for index_scol, label in zip(self._internal.index_scols, labels):
    conds.append(index_scol == label)

cond = ~reduce(lambda x, y: x & y, conds)
sdf.where(cond)
```

HyukjinKwon: Yeah, @RainFung. I think this comment is correct. Can you update please accordingly?

HyukjinKwon: ping @RainFung

databricks/koalas/tests/test_indexes.py
```diff
@@ -391,3 +408,4 @@ def test_multiindex_rename(self):
         pidx = pidx.rename(['my', 'name', 'is'])
         kidx = kidx.rename(['my', 'name', 'is'])
         self.assert_eq(pidx, kidx)
+
```
HyukjinKwon: Can we remove this line?

PR #1124, opened 2019-12-12
Implement DataFrame.info

### LINKED ISSUE #872: DataFrame.info
Implement `DataFrame.info`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.info.html
`memory usage` is not needed for Koalas.

HyukjinKwon:
This PR proposes `DataFrame.info`.

```python
>>> ks.range(100).info()
<class 'databricks.koalas.frame.DataFrame'>
Index: 100 entries, 0 to 99
Data columns (total 1 columns):
id    100 non-null int64
```

Resolves #872

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8088,7 +8088,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         Prints information of all columns:
 
-        >>> df.info(verbose=True)
+        >>> df.info(verbose=True)  # doctest: +SKIP
         <class 'databricks.koalas.frame.DataFrame'>
         Index: 5 entries, 0 to 4
         Data columns (total 3 columns):
@@ -8100,7 +8100,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         Prints a summary of columns count and its dtypes but not per column
         information:
 
-        >>> df.info(verbose=False)
+        >>> df.info(verbose=False)  # doctest: +SKIP
         <class 'databricks.koalas.frame.DataFrame'>
         Index: 5 entries, 0 to 4
         Columns: 3 entries, int_col to float_col
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8036,6 +8036,100 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         return ks.from_pandas(pdf.idxmin())
 
+    def info(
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -8036,6 +8036,100 @@ def idxmin(self, axis=0):
...
+        dtypes: float64(1), int64(1), object(1)
+
+        Pipe output of DataFrame.info to buffer instead of sys.stdout, get
+        buffer content and writes to a text file:
+
+        >>> import io
+        >>> buffer = io.StringIO()
+        >>> df.info(buf=buffer)
+        >>> s = buffer.getvalue()
+        >>> path = "/tmp/test"
```
ueshin: We have a fixture named `path`?

databricks/koalas/frame.py
```diff
@@ -8036,6 +8036,100 @@ def idxmin(self, axis=0):
...
+        >>> with open('%s/info/foo.txt' % path, "w",
+        ...           encoding="utf-8") as f:
+        ...     _ = f.write(s)
+        """
+        # To avoid pandas' existing config affects Koalas.
+        # TODO: should we have corresponding Koalas configs?
```
ueshin: Yeah, we should in a separate PR.

databricks/koalas/frame.py
```diff
@@ -8036,6 +8036,99 @@ def idxmin(self, axis=0):
...
+        with pd.option_context(
+                'display.max_info_columns', sys.maxsize,
+                'display.max_info_rows', sys.maxsize):
+            try:
+                self._data = self  # hack to use pandas' info as is.
+                return pd.DataFrame.info(
+                    self, verbose=verbose, buf=buf, max_cols=max_cols,
+                    memory_usage=False, null_counts=null_counts)
```
ueshin: I'm wondering how many Spark jobs are kicked.

HyukjinKwon: It seems launches 4 jobs. `_summary` twice (for index, and columns) and `len` twice. `len` can be saved once but I guess it's fine for now.

ueshin: Does it depend on the column number?

HyukjinKwon: I just checked it too. Nope, it doesn't.

ueshin: ah, sounds great.

PR #1128, opened 2019-12-13
Support 'compute.ops_on_diff_frames' for NumPy ufunc compat in Series

HyukjinKwon:
This PR proposes to support 'ops_on_diff_frames' option for NumPy ufunc compat in Series

```python
>>> import numpy as np
>>> a = ks.range(10).id
>>> b = ks.range(10, 20).id
>>> ks.set_option('compute.ops_on_diff_frames', True)
>>> np.arctan2(a, b)
0    0.000000
1    0.090660
2    0.165149
3    0.226799
4    0.278300
5    0.321751
6    0.358771
7    0.390607
8    0.418224
9    0.442374
Name: id, dtype: float64
```

### DIFF
--- a/databricks/koalas/tests/test_numpy_compat.py
+++ b/databricks/koalas/tests/test_numpy_compat.py
@@ -17,6 +17,7 @@ import numpy as np
 import pandas as pd
 
 from databricks import koalas as ks
+from databricks.koalas import set_option, reset_option
 from databricks.koalas.numpy_compat import unary_np_spark_mappings, binary_np_spark_mappings
 from databricks.koalas.testing.utils import ReusedSQLTestCase, SQLTestUtils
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_numpy_compat.py
```diff
@@ -103,3 +107,19 @@ def test_np_spark_compat(self):
                         np_func(pdf.a, 1), np_func(kdf.a, 1), almost=True)
                 except Exception as e:
                     raise AssertionError("Test in '%s' function was failed." % np_name) from e
+
+        # Test only top 5 for now. 'compute.ops_on_diff_frames' option increases too much time.
+        try:
+            set_option('compute.ops_on_diff_frames', True)
+            for np_name, spark_func in list(binary_np_spark_mappings.items())[:5]:
```
HyukjinKwon: BTW, I checked all the tests pass for its complete list.

ueshin: I checked it in my local as well.

ueshin:
Thanks! merging.

PR #1129, opened 2019-12-13
Support arithmetic and comparison APIs on itself in DataFrame

HyukjinKwon:
This PR fixes the existing DataFrame's arithmetic and comparison APIs to take the same DataFrame as its argument. See below:

```python
>>> kdf = ks.range(10)
>>> kdf + kdf
   id
0   0
1   2
2   4
3   6
4   8
5  10
6  12
7  14
8  16
9  18
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -109,6 +109,12 @@ circle          1      361
 triangle        4      181
 rectangle       5      361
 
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
I'm pretty sure on this. Let me merge this one.

PR #1130, opened 2019-12-13
'isna' type functions should return proper message for MultiIndex

itholic:
In pandas, `pd.MultiIndex.isna()`, `pd.MultiIndex.isnull()`, `pd.MultiIndex.notna()`, `pd.MultiIndex.notnull()` return same error message that `isna is not defined for MultiIndex` like below.

```python
>>> pidx = pd.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2)])

>>> pidx.isna()
...
NotImplementedError: isna is not defined for MultiIndex

>>> pidx.isnull()
...
NotImplementedError: isna is not defined for MultiIndex

>>> pidx.notna()
...
NotImplementedError: isna is not defined for MultiIndex

>>> pidx.notnull()
...
NotImplementedError: isna is not defined for MultiIndex
```

i think we'd better mimic them for our functions.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -604,7 +604,7 @@ class IndexOpsMixin(object):
         """
         from databricks.koalas.indexes import MultiIndex
         if isinstance(self, MultiIndex):
-            raise NotImplementedError("isna is not defined for MultiIndex")
+            raise NotImplementedError("notna is not defined for MultiIndex")
         return (~self.isnull()).rename(self.name)
 
     notna = notnull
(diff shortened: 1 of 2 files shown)


databricks/koalas/base.py
```diff
@@ -599,6 +602,9 @@ def notnull(self):
         >>> ser.rename("a").to_frame().set_index("a").index.notna()
         Index([True, True, False], dtype='object', name='a')
         """
+        from databricks.koalas.indexes import MultiIndex
+        if isinstance(self, MultiIndex):
+            raise NotImplementedError("isna is not defined for MultiIndex")
```
HyukjinKwon: Let's just fix `isna` to `notnull` here.

itholic: at the first time, i just mimic pandas', but i think also it is better. thanks!

PR #1131, opened 2019-12-14
Explicitly don't support Index.duplicated

RainFung:
This PR proposes to explicitly unsupport Index.duplicated.


RainFung:
@itholic  Hello Lee, Can you give some advice?

itholic:
Sure, let me take a look a little while :)

databricks/koalas/indexes.py
```diff
@@ -456,6 +457,47 @@ def to_series(self, name: Union[str, Tuple[str, ...]] = None) -> Series:
...
+        0    True
+        1   False
+        2    True
+        Name: duplicated, dtype: bool
+        """
+        index_scols = self._internal.index_scols
+        sdf = self._internal.sdf
+        window = Window.partitionBy(index_scols).rowsBetween(Window.unboundedPreceding,
+                                                             Window.unboundedFollowing)
+        sdf = sdf.withColumn('duplicated', F.count(index_scols[0]).over(window) > 1)
```
itholic: after this the order will be shuffled,
so i think how about we make one more extra column to keep natural order like below.
(and use more complex column name for 'duplicated' to avoid duplication with user column)
```python
sdf = sdf.withColumn('__natural_order__', F.monotonically_increasing_id())
sdf = sdf.withColumn('__duplicated__', F.count(index_scols[0]).over(window) > 1)
sdf = sdf.select('__duplicated__').orderBy('__natural_order__')
```

and now we returns `Series`, not `Index`,


(actually i'm not pretty sure is it right way that we returns `Index` rather than `np.ndarray`)

```python
internal = _InternalFrame(sdf=sdf, index_map=[('__duplicated__', None)])
```

RainFung: Yes, there are some functions which we need return result in natural order. But there is no extra column like `__index_level_0__`  in `MultiIndex`.

databricks/koalas/indexes.py
```diff
@@ -456,6 +457,47 @@ def to_series(self, name: Union[str, Tuple[str, ...]] = None) -> Series:
...
+
+        Examples
+        --------
+        By default, for each set of duplicated values, all duplicates is set to True:
+
+        >>> idx = ks.Index(['lama', 'cow', 'lama'])
+        >>> idx.duplicated()
+        0    True
+        1   False
+        2    True
```
HyukjinKwon: The results seems different from pandas':

```python
>>> i = pd.Index(['lama', 'cow', 'lama'])
>>> i
Index(['lama', 'cow', 'lama'], dtype='object')
>>> i.duplicated()
array([False, False,  True])
```

databricks/koalas/indexes.py
```diff
@@ -456,6 +457,47 @@ def to_series(self, name: Union[str, Tuple[str, ...]] = None) -> Series:
                                          column_index_names=None),
                       anchor=kdf)
 
+    def duplicated(self, keep=False):
```
HyukjinKwon: Let's explicitly don't support this since it returns an array which is supposed to be big. Users can easily workaround by `DataFrame.duplicated()` API:

```python
>>> kdf
   id  col
0   0    2
1   1    2
2   2    2
3   3    2
4   4    2
5   5    2
6   6    2
7   7    2
8   8    2
9   9    2
>>> kdf[['col']].duplicated()
0    False
1     True
2     True
3     True
4     True
5     True
6     True
7     True
8     True
9     True
Name: 0, dtype: bool
```

We can include this in the error message.

HyukjinKwon: I think we can just explicitly unsupport this API as its return type is NumPy arrow and the data size seems expected to be large.
We can just say users can easily work around by `DataFrame.deduplicated()` API.

```python
>>> kdf.index.to_series().to_frame().duplicated()
0    False
1    False
2    False
3    False
4    False
5    False
6    False
7    False
8    False
9    False
Name: 0, dtype: bool
```

I think we shouldn't implement `Series.duplicated()` as well. 

Can you refer https://github.com/databricks/koalas/pull/744 and fix it accordingly?

databricks/koalas/missing/indexes.py
```diff
@@ -100,6 +99,9 @@ class _MissingPandasLikeIndex(object):
     get_values = unsupported_function('get_values', deprecated=True)
     item = unsupported_function('item', deprecated=True)
     contains = unsupported_function('contains', deprecated=True)
+    duplicated = unsupported_property(
+        'duplicated', reason="'duplicated' API returns np.ndarray and the data size is so large."
+                             "You can just use DataFrame.deduplicated instead")
```
HyukjinKwon: Looks it should be added into `_MissingPandasLikeMultiIndex` too.

HyukjinKwon: Can you also add the message into `databricks/koalas/missing/common.py` and reuse it? Looks good otherwise.

HyukjinKwon: you can call common here instead to deduplicate `duplicated  = common. duplicated(unsupported_property)`

RainFung: Thanks for your kind remind. Fix it already.

databricks/koalas/missing/series.py
```diff
@@ -128,6 +127,9 @@ class _MissingPandasLikeSeries(object):
     ptp = unsupported_function('ptp', deprecated=True)
     argmax = unsupported_function('argmax', deprecated=True)
     argmin = unsupported_function('argmin', deprecated=True)
+    duplicated = unsupported_property(
+        'duplicated', reason="'duplicated' API returns np.ndarray and the data size is so large."
+                             "You can just use DataFrame.deduplicated instead")
```
HyukjinKwon: here too

databricks/koalas/indexes.py
ueshin: oh wait.
Seems like `Series.duplicated()` returns `Series`. We can support it.
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.duplicated.html

HyukjinKwon: Yeah, I missed it. @RainFung, can you implement this one?

ueshin: maybe we can implement it in a separate PR. just revert the updates for Series missing file.

RainFung: OK，  i'll fix it today

ueshin:
Did we made `Series.duplicated` deprecated?
What about https://github.com/databricks/koalas/pull/1131#discussion_r359501550 ?

PR #1136, opened 2019-12-17
Implement dot() for Series

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.dot.html#pandas.Series.dot

```python
>>> s = ks.Series([0, 1, 2, 3])
>>> other = ks.Series([-1, 2, -3, 4])

>>> s.dot(other)
8

>>> s @ other
8

>>> df = ks.DataFrame([[0, 1], [-2, 3], [4, -5], [6, 7]])
>>> s.dot(df)
0    24
1    14
Name: 0, dtype: int64
```


databricks/koalas/series.py
```diff
@@ -4194,6 +4194,70 @@ def pct_change(self, periods=1):
...
+        1    14
+        Name: 0, dtype: int64
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        if repr(self.index) != repr(other.index):
+            raise ValueError("matrices are not aligned")
+
+        if isinstance(other, DataFrame):
+            idx_val_dict = {col_name: (self * other[col_name]).sum() for col_name in other}
```
HyukjinKwon: This will cause join (and shuffle) as many as the number of columns ... you can unsupport this case for now.

databricks/koalas/series.py
```diff
@@ -4194,6 +4194,70 @@ def pct_change(self, periods=1):
...
+        >>> s @ other
+        8
+
+        >>> df = ks.DataFrame([[0, 1], [-2, 3], [4, -5], [6, 7]])
+        >>> s.dot(df)
+        0    24
+        1    14
+        Name: 0, dtype: int64
+
+        >>> reset_option("compute.ops_on_diff_frames")
```
HyukjinKwon: and let's make the example without `compute.ops_on_diff_frames` (e.g., s @ s)

databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -574,6 +574,41 @@ def test_multi_index_column_assignment_frame(self):
...
+
+        kser_other = ks.Series([90, 91, 85], index=[1, 2, 4])
+        with self.assertRaisesRegex(ValueError, "matrices are not aligned"):
+            kser.dot(kser_other)
+
+        kser_other = ks.Series([90, 91, 85, 100], index=[2, 4, 1, 0])
+        with self.assertRaisesRegex(ValueError, "matrices are not aligned"):
+            kser.dot(kser_other)
+
+        # with DataFram is not supported for now since performance issue,
```
HyukjinKwon: `DataFram` -> `DataFrame`

databricks/koalas/series.py
```diff
@@ -4194,6 +4194,62 @@ def pct_change(self, periods=1):
...
+        --------
+        >>> s = ks.Series([0, 1, 2, 3])
+
+        >>> s.dot(s)
+        14
+
+        >>> s @ s
+        14
+        """
+        if repr(self.index) != repr(other.index):
```
HyukjinKwon: I think we shouldn't use `repr`. It will fail if, for instance, index name is different, no?

databricks/koalas/series.py
```diff
@@ -4194,6 +4194,62 @@ def pct_change(self, periods=1):
...
+            raise ValueError("matrices are not aligned")
+
+        if isinstance(other, DataFrame):
+            raise ValueError(
+                "Series.dot() is currently not supported with DataFrame since "
+                "it will cause expansive calculation as many as the number "
+                "of columns of DataFrame")
+
+        elif isinstance(other, Series):
```
HyukjinKwon: What happens if `other` is not `DataFrame` or `Series`? e.g., `int`.

databricks/koalas/series.py
HyukjinKwon: I think you could do `(kdf.index.to_series() != kdf.index.to_series()).any()`

HyukjinKwon: Otherwise, this comparison will fail even if the name of index is different.

databricks/koalas/tests/test_ops_on_diff_frames.py
itholic: thanks. i'll take a more closer look next time 😅

databricks/koalas/series.py
itholic: it raises `IndexError` in pandas like below

```python
>>> pser
0    0
1    1
2    2
3    3
Name: 0, dtype: int64

>>> pser.dot(2)
...
IndexError: tuple index out of range
```

i'll fix this for working same in koalas.

databricks/koalas/series.py
itholic: ah, thanks for the advice! okay i'll manage this

HyukjinKwon:
@itholic, can you rebase this PR? seems fine.

itholic:
@HyukjinKwon Rebase & test finished :)

databricks/koalas/series.py
```diff
@@ -4335,6 +4335,63 @@ def pct_change(self, periods=1):
...
+        >>> s = ks.Series([0, 1, 2, 3])
+
+        >>> s.dot(s)
+        14
+
+        >>> s @ s
+        14
+        """
+        if len(self.index) != len(other.index):
+            raise ValueError("matrices are not aligned")
```
HyukjinKwon: Hm, does this only check the length? I think it also checks if the values in index are same or not. Also, we don't have to perform count if both are same Series.

itholic: I agree that it is better to check if both series is identically same,

but don't think we have to compare the values of indexes.

If the length is same, we can just compute the dot product regardless of their values.

itholic: That's right, thanks for the comment!

databricks/koalas/series.py
```diff
@@ -4335,6 +4335,80 @@ def pct_change(self, periods=1):
...
+        """
+        Compute the dot product between the Series and the columns of other.
+
+        This method computes the dot product between the Series and another
+        one, or the Series and each columns of a DataFrame.
+
+        It can also be called using `self @ other` in Python >= 3.5.
+
+        .. note:: The behaviour of this API is slightly different from pandas
+            since its original implementation requires quite expensive operation
```
HyukjinKwon: `quite expensive operation` -> `an expensive operation`

databricks/koalas/series.py
```diff
@@ -4335,6 +4335,80 @@ def pct_change(self, periods=1):
...
+            (It requires sorting & comparing for all index values which is quite
+            expensive and may cause out-of-memory with large dataset).
+            pandas raises ValueError if all values of index are not same, while
+            Koalas just ignoring them as NaN.
```
HyukjinKwon: This API is slightly different from pandas when indexes from both are not aligned. To match with pandas', it requires to read the whole data for, for example, counting. pandas raises an exception; however, Koalas just proceeds and performs by ignoring mismatches with NaN permissively.

itholic:
@HyukjinKwon Thanks for the review, i fixed them :D

PR #1137, opened 2019-12-17
Set the upper-bound for pandas.

ueshin:
Pandas 1.0.0 will be released soon and it says it will only support Python 3.6.1 and higher, and also it might contain behavior changes.
We will still support Python 3.5 for a while, we should set the upper-bound for pandas for now to confirm Koalas works with it.


ueshin:
If we use Python <3.6.1, pip/conda command will just use `0.x` versions?

HyukjinKwon:
> If we use Python <3.6.1, pip/conda command will just use 0.x versions? 

Yea, that's right up to my knowledge.

PR #1146, opened 2019-12-19
Add a hidden column `__natural_order__`.

ueshin:
Adding a hidden column `__natural_order__` for ordering the rows as it is, especially for window-like operations like `cumxxx`.


databricks/koalas/frame.py
```diff
@@ -1919,7 +1920,7 @@ def transform(self, func):
                 applied.append(wrapped(self[idx]).rename(idx))
 
         sdf = self._sdf.select(
-            self._internal.index_scols + [c._scol for c in applied])
+            self._internal.index_scols + [c._scol for c in applied] + list(HIDDEN_COLUMNS))
```
HyukjinKwon: Do we need to select the hidden columns here explicitly? I suspect it was selected since they are map-like operations (?). Maybe we should just don't do this in all code base if I am not mistaken ...

HyukjinKwon:
I am merging this for now. I will think about how to clean up too ... we can address the comments together later.

databricks/koalas/frame.py
ueshin: Actually in most cases we can remove, but in some cases we can't, e.g., right after `_cum`, and window-like functions after #1151 is merged.

PR #1159, opened 2019-12-30
Fix .loc to work properly when using 'slice'

### LINKED ISSUE #1158: .loc behavior when using 'slice'
I test as shown in the documentation below. 
I found something strange.

https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.Series.loc.html
image

pandas :
'sidewinder' is at the end of the list, so it prints normally.

```python
>>> pdf = pd.DataFrame([[1, 2], [4, 5], [7, 8]], 
...             index=['cobra', 'viper', 'sidewinder'],  
...             columns=['max_speed', 'shield'])
>>> pdf.loc['cobra':'viper', 'max_speed']
cobra    1
viper    4
Name: max_speed, dtype: int64

```

but koalas prints 'sidewinder'

```python
>>> kdf = ks.DataFrame([[1, 2], [4, 5], [7, 8]],
...              index=['cobra', 'viper', 'sidewinder'],
...              columns=['max_speed', 'shield'])
>>> kdf.loc['cobra':'viper', 'max_speed']
cobra         1
viper         4
sidewinder    7
Name: max_speed, dtype: int64

>>> kdf.loc['cobra':, 'max_speed']
cobra         1
viper         4
sidewinder    7
Name: max_speed, dtype: int64

>>> kdf.loc['sidewinder':, 'max_speed']
viper         4
sidewinder    7
Name: max_speed, dtype: int64

>>> kdf.loc['viper':, 'max_speed']
viper    4
Name: max_speed, dtype: int64
```

Input is done in the order ['cobra', 'viper', 'sidewinder'], 
but koals appears to be recognized as ['cobra', 'sidewinder', 'viper'].

itholic:
Resolve #1158 

```python
>>> kdf
            max_speed  shield
cobra               1       2
viper               4       5
sidewinder          7       8

>>> kdf.loc['cobra':'viper', 'max_speed']
cobra    1
viper    4
Name: max_speed, dtype: int64

>>> kdf.to_pandas().loc['cobra':'viper', 'max_speed']
cobra    1
viper    4
Name: max_speed, dtype: int64
```

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -426,14 +426,14 @@ class LocIndexer(_LocIndexerLike):
                         start = rows_sel.start
                         start_order_column = index_column._scol
                 if stop is None and rows_sel.stop is not None:
-                    if not (index.is_monotonic_decreasing or index.is_monotonic_increasing):
+                    if not (index.is_monotonic_increasing or index.is_monotonic_decreasing):
                         raise KeyError(rows_sel.stop)
                     else:
                         stop = rows_sel.stop
                         stop_order_column = index_column._scol
 
                 # we don't use StringType since we're using `__natural_order__` for comparing
-                if index_data_type is StringType:
+                if isinstance(index_data_type, StringType):
                     index_data_type = LongType()
 
                 # if start and stop are same, just get all start(or stop) values
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -290,13 +290,12 @@ class LocIndexer(_LocIndexerLike):
     Slice with labels for row and single label for column. As mentioned
     above, note that both the start and stop of the slice are included.
 
-    Also note that the row for 'sidewinder' is included since 'sidewinder'
-    is between 'cobra' and 'viper'.
+    Also note that the row for 'sidewinder' is not included since 'sidewinder'
+    is not between 'cobra' and 'viper'.
 
     >>> df.loc['cobra':'viper', 'max_speed']
-    cobra         1
-    viper         4
-    sidewinder    7
+    cobra    1
+    viper    4
     Name: max_speed, dtype: int64
 
     Conditional that returns a boolean Series
@@ -400,16 +399,23 @@ class LocIndexer(_LocIndexerLike):
                 # If slice is None - select everything, so nothing to do
                 return None, None
             elif len(self._internal.index_columns) == 1:
-                start = rows_sel.start
-                stop = rows_sel.stop
-
+                sdf = self._kdf_or_kser._internal.sdf
                 index_column = self._kdf_or_kser.index.to_series()
-                index_data_type = index_column.spark_type
+
+                # get natural order from '__natural_order__' from start to stop
+                # based on index_columns to keep natural order.
+                start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
+                           .where(index_column._scol == rows_sel.start) \
+                           .first()[0]
+                stop = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
+                          .where(index_column._scol == rows_sel.stop) \
+                          .first()[0]
+
                 cond = []
                 if start is not None:
-                    cond.append(index_column._scol >= F.lit(start).cast(index_data_type))
+                    cond.append(sdf[NATURAL_ORDER_COLUMN_NAME] >= F.lit(start))
                 if stop is not None:
-                    cond.append(index_column._scol <= F.lit(stop).cast(index_data_type))
+                    cond.append(sdf[NATURAL_ORDER_COLUMN_NAME] <= F.lit(stop))
 
                 if len(cond) > 0:
                     return reduce(lambda x, y: x & y, cond), None
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -233,7 +233,6 @@ class LocIndexer(_LocIndexerLike):
 
     .. note:: Note that contrary to usual python slices, **both** the
         start and the stop are included, and the step of the slice is not allowed.
-        In addition, with a slice, Koalas works as a filter between the range.
 
     .. note:: With a list or array of labels for row selection,
         Koalas behaves as a filter without reordering by the labels.
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -290,9 +290,6 @@ class LocIndexer(_LocIndexerLike):
     Slice with labels for row and single label for column. As mentioned
     above, note that both the start and stop of the slice are included.
 
-    Also note that the row for 'sidewinder' is not included since 'sidewinder'
-    is not between 'cobra' and 'viper'.
-
     >>> df.loc['cobra':'viper', 'max_speed']
     cobra    1
     viper    4
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -22,7 +22,7 @@ from functools import reduce
 from pandas.api.types import is_list_like
 from pyspark import sql as spark
 from pyspark.sql import functions as F
-from pyspark.sql.types import BooleanType
+from pyspark.sql.types import BooleanType, StringType
 from pyspark.sql.utils import AnalysisException
 
 from databricks.koalas.internal import _InternalFrame, HIDDEN_COLUMNS, NATURAL_ORDER_COLUMN_NAME
@@ -401,21 +401,35 @@ class LocIndexer(_LocIndexerLike):
             elif len(self._internal.index_columns) == 1:
                 sdf = self._kdf_or_kser._internal.sdf
                 index_column = self._kdf_or_kser.index.to_series()
+                index_data_type = index_column.spark_type
+                start = rows_sel.start
+                stop = rows_sel.stop
 
                 # get natural order from '__natural_order__' from start to stop
-                # based on index_columns to keep natural order.
-                start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
-                           .where(index_column._scol == rows_sel.start) \
-                           .first()[0]
-                stop = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
-                          .where(index_column._scol == rows_sel.stop) \
-                          .first()[0]
-
+                # to keep natural order when type of rows are StringType
+                if isinstance(index_data_type, StringType):
+                    start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
+                               .where(index_column._scol == start) \
+                               .first()
+                    stop = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
+                              .where(index_column._scol == stop) \
+                              .first()
+                    order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
+                    if start is not None:
+                        start = start[0]
+                    else:
+                        raise KeyError(rows_sel.start)
+                    if stop is not None:
+                        stop = stop[0]
+                    else:
+                        raise KeyError(rows_sel.stop)
+                else:
+                    order_column = index_column._scol
                 cond = []
                 if start is not None:
-                    cond.append(sdf[NATURAL_ORDER_COLUMN_NAME] >= F.lit(start))
+                    cond.append(order_column >= F.lit(start))
                 if stop is not None:
-                    cond.append(sdf[NATURAL_ORDER_COLUMN_NAME] <= F.lit(stop))
+                    cond.append(order_column <= F.lit(stop))
 
                 if len(cond) > 0:
                     return reduce(lambda x, y: x & y, cond), None
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -22,7 +22,7 @@ from functools import reduce
 from pandas.api.types import is_list_like
 from pyspark import sql as spark
 from pyspark.sql import functions as F
-from pyspark.sql.types import BooleanType, StringType
+from pyspark.sql.types import BooleanType, StringType, LongType
 from pyspark.sql.utils import AnalysisException
 
 from databricks.koalas.internal import _InternalFrame, HIDDEN_COLUMNS, NATURAL_ORDER_COLUMN_NAME
@@ -417,19 +417,24 @@ class LocIndexer(_LocIndexerLike):
                     order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
                     if start is not None:
                         start = start[0]
+                    elif rows_sel.start is None:
+                        pass
                     else:
                         raise KeyError(rows_sel.start)
                     if stop is not None:
                         stop = stop[0]
+                    elif rows_sel.stop is None:
+                        pass
                     else:
                         raise KeyError(rows_sel.stop)
+                    index_data_type = LongType()
                 else:
                     order_column = index_column._scol
                 cond = []
                 if start is not None:
-                    cond.append(order_column >= F.lit(start))
+                    cond.append(order_column >= F.lit(start).cast(index_data_type))
                 if stop is not None:
-                    cond.append(order_column <= F.lit(stop))
+                    cond.append(order_column <= F.lit(stop).cast(index_data_type))
 
                 if len(cond) > 0:
                     return reduce(lambda x, y: x & y, cond), None
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -408,24 +408,21 @@ class LocIndexer(_LocIndexerLike):
                 # get natural order from '__natural_order__' from start to stop
                 # to keep natural order when type of rows are StringType
                 if isinstance(index_data_type, StringType):
-                    start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
-                               .where(index_column._scol == start) \
-                               .first()
-                    stop = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
-                              .where(index_column._scol == stop) \
-                              .first()
+                    start_and_stop = (
+                        sdf.select(index_column._scol, NATURAL_ORDER_COLUMN_NAME)
+                           .where((index_column._scol == start) | (index_column._scol == stop))
+                           .collect())
+
+                    start = [row[1] for row in start_and_stop if row[0] == start]
+                    start = start[0] if len(start) > 0 else None
+
+                    stop = [row[1] for row in start_and_stop if row[0] == stop]
+                    stop = stop[0] if len(stop) > 0 else None
+
                     order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
-                    if start is not None:
-                        start = start[0]
-                    elif rows_sel.start is None:
-                        pass
-                    else:
+                    if start is None and rows_sel.start is not None:
                         raise KeyError(rows_sel.start)
-                    if stop is not None:
-                        stop = stop[0]
-                    elif rows_sel.stop is None:
-                        pass
-                    else:
+                    if stop is None and rows_sel.stop is not None:
                         raise KeyError(rows_sel.stop)
                     index_data_type = LongType()
                 else:
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -420,7 +420,7 @@ class LocIndexer(_LocIndexerLike):
                 # if index order is not monotonic increasing or decreasing
                 # and specified values don't exist in index, raise KeyError
                 if start is None and rows_sel.start is not None:
-                    if not (index.is_monotonic_decreasing or index.is_monotonic_increasing):
+                    if not (index.is_monotonic_increasing or index.is_monotonic_decreasing):
                         raise KeyError(rows_sel.start)
                     else:
                         start = rows_sel.start
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -404,29 +404,33 @@ class LocIndexer(_LocIndexerLike):
                 index_data_type = index_column.spark_type
                 start = rows_sel.start
                 stop = rows_sel.stop
+                order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
 
                 # get natural order from '__natural_order__' from start to stop
-                # to keep natural order when type of rows are StringType
+                # to keep natural order.
+                start_and_stop = (
+                    sdf.select(index_column._scol, NATURAL_ORDER_COLUMN_NAME)
+                       .where((index_column._scol == start) | (index_column._scol == stop))
+                       .collect())
+
+                start = [row[1] for row in start_and_stop if row[0] == start]
+                start = start[0] if len(start) > 0 else None
+
+                stop = [row[1] for row in start_and_stop if row[0] == stop]
+                stop = stop[0] if len(stop) > 0 else None
+
                 if isinstance(index_data_type, StringType):
-                    start_and_stop = (
-                        sdf.select(index_column._scol, NATURAL_ORDER_COLUMN_NAME)
-                           .where((index_column._scol == start) | (index_column._scol == stop))
-                           .collect())
-
-                    start = [row[1] for row in start_and_stop if row[0] == start]
-                    start = start[0] if len(start) > 0 else None
-
-                    stop = [row[1] for row in start_and_stop if row[0] == stop]
-                    stop = stop[0] if len(stop) > 0 else None
-
-                    order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
+                    index_data_type = LongType()
                     if start is None and rows_sel.start is not None:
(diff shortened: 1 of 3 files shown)


databricks/koalas/indexing.py
```diff
@@ -400,16 +399,23 @@ def _select_rows(self, rows_sel):
...
-
+                sdf = self._kdf_or_kser._internal.sdf
                 index_column = self._kdf_or_kser.index.to_series()
-                index_data_type = index_column.spark_type
+
+                # get natural order from '__natural_order__' from start to stop
+                # based on index_columns to keep natural order.
+                start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
+                           .where(index_column._scol == rows_sel.start) \
+                           .first()[0]
```
HyukjinKwon: What happens if the given `start` and `stop` don't exist?

itholic: thanks for the comment

i'll check this with adding more test case related with it.

databricks/koalas/indexing.py
```diff
@@ -400,16 +399,23 @@ def _select_rows(self, rows_sel):
...
+                stop = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
+                          .where(index_column._scol == rows_sel.stop) \
+                          .first()[0]
```
HyukjinKwon: I think you can use case-when and avoid launching a job execution here. For instance,

```python
natural_order = F.when(
    index_column._scol == rows_sel.stop, F.col(NATURAL_ORDER_COLUMN_NAME)
).otherwise(F.lit(None))
cond.append(natural_order.isNotNull() & sdf[NATURAL_ORDER_COLUMN_NAME] <= natural_order)
```

Not tested but I guess this works?

itholic: thanks for the comment!

oops i just checked this comment after using another way.

how about now implementation because we have to catch `KeyError` when given index name is not included in actual index when its type is `StringType`?

for example, when type of index is numeric, it is not raise KeyError like below.

```python
>>> pdf
   month  year  sale
0      1  2012    55
1      4  2014    40
2      7  2013    84
3     10  2014    31

>>> pdf.loc[0:1000]
   month  year  sale
0      1  2012    55
1      4  2014    40
2      7  2013    84
3     10  2014    31
```

but when the type of index is string, it raises KeyError like below.

```python
>>> pdf
            max_speed  shield
cobra               1       2
viper               4       5
sidewinder          7       8

>>> pdf['cobra':'hello']
...
KeyError: 'hello'
```

HyukjinKwon: That's all good but problem is `first()` triggers each job and we should avoid it.
Currently, it looks going to launch 3 jobs. It's better to make the API in single pass.

itholic: okay i see. let me take manage this. i'll try the way you recommended 👍

beobest2: also, please check date data type
I think it is same issue 

```python
>>> dates = pd.date_range('20130101', periods=6)
>>> pdf = pd.DataFrame(np.random.randn(6, 4), index=dates, columns=list('ABCD'))
>>> pdf
                   A         B         C         D
2013-01-01  0.322155 -0.851969 -1.501208 -0.660413
2013-01-02 -0.388889  0.508557 -1.215688  0.761596
2013-01-03 -1.698188 -0.006648  2.111597 -1.404037
2013-01-04 -0.850982  1.449845 -0.259124  0.096576
2013-01-05  0.147729  0.604907  0.414314  0.125392
2013-01-06  0.016597 -0.102719 -0.965090  0.823600
>>> kdf = ks.from_pandas(pdf)
>>> pdf.loc[dates[0]:dates[1], 'D']
2013-01-01   -0.660413
2013-01-02    0.761596
Freq: D, Name: D, dtype: float64
>>> kdf.loc[dates[0]:dates[1], 'D']
```

itholic: yeah it also fixed with existing changes

```
>>> kdf.loc[dates[0]:dates[1], 'D']
2013-01-01   -1.033028
2013-01-02    0.763634
Name: D, dtype: float64
```

itholic: @HyukjinKwon 

hmm... i tried to fix, but i think it is hard to avoid using `first()` since we have to know the start & stop number exactly as an integer to make proper condition.

for example, it doesn't work properly like below since `natural_order` is also condition, not a number. (it always return the result only including `start` or `stop` row itself, not as a range)

```
>>> natural_order
Column<b'CASE WHEN (__index_level_0__ = viper) THEN __natural_order__ ELSE NULL END'>
>>> sdf.show()
+-----------------+---------+------+-----------------+
|__index_level_0__|max_speed|shield|__natural_order__|
+-----------------+---------+------+-----------------+
|            cobra|        1|     2|      25769803776|
|            viper|        4|     5|      60129542144|
|       sidewinder|        7|     8|      94489280512|
+-----------------+---------+------+-----------------+

>>> sdf.where(sdf[NATURAL_ORDER_COLUMN_NAME] <= natural_order).show()
+-----------------+---------+------+-----------------+
|__index_level_0__|max_speed|shield|__natural_order__|
+-----------------+---------+------+-----------------+
|            viper|        4|     5|      60129542144|
+-----------------+---------+------+-----------------+
```

itholic:
hmm.. i have no idea why codecov has failed although it seems that tests are covering all codes

databricks/koalas/indexing.py
```diff
@@ -400,16 +399,42 @@ def _select_rows(self, rows_sel):
...
+                index_data_type = index_column.spark_type
                 start = rows_sel.start
                 stop = rows_sel.stop
 
-                index_column = self._kdf_or_kser.index.to_series()
-                index_data_type = index_column.spark_type
+                # get natural order from '__natural_order__' from start to stop
+                # to keep natural order when type of rows are StringType
+                if isinstance(index_data_type, StringType):
+                    start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
```
HyukjinKwon: @itholic, I mean the changes like this:

```diff
@@ -408,24 +395,21 @@ class LocIndexer(_LocIndexerLike):
                 # get natural order from '__natural_order__' from start to stop
                 # to keep natural order when type of rows are StringType
                 if isinstance(index_data_type, StringType):
-                    start = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
-                               .where(index_column._scol == start) \
-                               .first()
-                    stop = sdf.select(NATURAL_ORDER_COLUMN_NAME) \
-                              .where(index_column._scol == stop) \
-                              .first()
-                    order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
-                    if start is not None:
+                    start_and_stop = sdf.select(index_column._scol, NATURAL_ORDER_COLUMN_NAME) \
+                               .where((index_column._scol == start) | (index_column._scol == stop)) \
+                               .collect()
+
+                    start = [row[1] for row in start_and_stop if row[0] == start]
+                    if len(start) > 0:
                         start = start[0]
-                    elif rows_sel.start is None:
-                        pass
-                    else:
-                        raise KeyError(rows_sel.start)
-                    if stop is not None:
+                    stop = [row[1] for row in start_and_stop if row[0] == stop]
+                    if len(stop) > 0:
                         stop = stop[0]
-                    elif rows_sel.stop is None:
-                        pass
-                    else:
+
+                    order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
+                    if start is None and rows_sel.start is not None:
+                        raise KeyError(rows_sel.start)
+                    if stop is None and rows_sel.stop is not None:
                         raise KeyError(rows_sel.stop)
                     index_data_type = LongType()
                 else:
```

HyukjinKwon: Ah, also add a comment that we call `collect()` assuming the frame does not have the records that have same start and end indexes a lot (otherwise it will cause out-of-memory easily)

itholic: ah, that makes sense.
i though that you told like let's don't use job like `first()` or `collect()` like.
thanks for the comment with kindness example!

databricks/koalas/indexing.py
```diff
@@ -400,16 +399,39 @@ def _select_rows(self, rows_sel):
...
+                    stop = stop[0] if len(stop) > 0 else None
+
+                    order_column = sdf[NATURAL_ORDER_COLUMN_NAME]
+                    if start is None and rows_sel.start is not None:
+                        raise KeyError(rows_sel.start)
+                    if stop is None and rows_sel.stop is not None:
+                        raise KeyError(rows_sel.stop)
+                    index_data_type = LongType()
+                else:
+                    order_column = index_column._scol
```
HyukjinKwon: @itholic, I think we should always use the natural order regardless of types. For example:

```python
>>> pd.DataFrame({'a': [1,2,3]}, index=[3,2,1]).loc[3:2, 'a']
3    1
2    2
```

but

```python
>>> ks.DataFrame({'a': [1,2,3]}, index=[3,2,1]).loc[3:2, 'a']
Series([], Name: a, dtype: int64)
```

itholic: ah, i missed that case, let me manage it soon
thanks !

itholic: i've just fixed :)

databricks/koalas/tests/test_indexing.py
```diff
@@ -222,6 +222,21 @@ def test_loc(self):
 
         self.assert_eq(kdf.loc[1000:], pdf.loc[1000:])
         self.assert_eq(kdf.loc[-2000:-1000], pdf.loc[-2000:-1000])
+        self.assert_eq(kdf.loc[-2000:-1000], pdf.loc[-2000:-1000])
```
ueshin: What's this line for? the same as the previous line?

databricks/koalas/indexing.py
```diff
@@ -400,16 +399,43 @@ def _select_rows(self, rows_sel):
...
+                       .where((index_column._scol == start) | (index_column._scol == stop))
+                       .collect())
+
+                start = [row[1] for row in start_and_stop if row[0] == start]
+                start = start[0] if len(start) > 0 else None
+
+                stop = [row[1] for row in start_and_stop if row[0] == stop]
+                stop = stop[0] if len(stop) > 0 else None
+
+                if isinstance(index_data_type, StringType):
```
ueshin: What's wrong with `StringType`?

databricks/koalas/indexing.py
```diff
@@ -400,16 +399,43 @@ def _select_rows(self, rows_sel):
...
+                    index_data_type = LongType()
+                    if start is None and rows_sel.start is not None:
+                        raise KeyError(rows_sel.start)
+                    if stop is None and rows_sel.stop is not None:
+                        raise KeyError(rows_sel.stop)
+                else:
+                    if start is None and stop is None:
+                        start = rows_sel.start
+                        stop = rows_sel.stop
+                        order_column = index_column._scol
```
ueshin: This might cause unexpected behavior changes between whether the slice hit or not.
E.g.,

```py
>>> kdf
    max_speed  shield
10          1       2
30          4       5
20          7       8
>>> kdf.loc[10:20, 'max_speed']
10    1
30    4
20    7
Name: max_speed, dtype: int64
>>> kdf.loc[1:21, 'max_speed']
10    1
20    7
Name: max_speed, dtype: int64
```

databricks/koalas/indexing.py
```diff
@@ -290,13 +290,12 @@ class LocIndexer(_LocIndexerLike):
     Slice with labels for row and single label for column. As mentioned
     above, note that both the start and stop of the slice are included.
 
-    Also note that the row for 'sidewinder' is included since 'sidewinder'
-    is between 'cobra' and 'viper'.
+    Also note that the row for 'sidewinder' is not included since 'sidewinder'
+    is not between 'cobra' and 'viper'.
```
ueshin: We may remove this line if the behavior is same as pandas'.

databricks/koalas/indexing.py
ueshin: Btw, seems like pandas throws exceptions when the specified values don't exist or are duplicated.

```py
>>> pdf
    max_speed  shield
10          1       2
30          4       5
20          7       8
30          9      10

>>> pdf.loc[1:20]
...
KeyError: 1

>>> pdf.loc[10:30]
...
KeyError: 'Cannot get right slice bound for non-unique label: 30'
```

Now that we collect the data, I guess we can check the values and follow such behaviors.

databricks/koalas/indexing.py
itholic: @ueshin thanks for the comment :)

As described at https://github.com/databricks/koalas/issues/1158#issuecomment-569570936 ,

the comparison with `StringType` couldn't keep the natural order since it literally compare the index by alphabetically. (that's why the issue #1158 is occured)

so we replace `StringType` into natural order using column `__natural_order__`.

For example, assuming that we have `sdf` like below,

```
>>> kdf._internal.sdf.show()
+-----------------+---------+------+-----------------+
|__index_level_0__|max_speed|shield|__natural_order__|
+-----------------+---------+------+-----------------+
|            cobra|        1|     2|      25769803776|
|            viper|        4|     5|      60129542144|
|       sidewinder|        7|     8|      94489280512|
+-----------------+---------+------+-----------------+
```

and if the given key is `['cobra':'viper']`, we replace it into `[25769803776:60129542144]`.

that's why we're checking and casting `StringType` into `LongType` here.

databricks/koalas/indexing.py
itholic: @ueshin thank you for confirming! i'll fix them properly soon 😄

databricks/koalas/indexing.py
ueshin: Now we are trying to use `__natural_order__` to calculate actual start and stop regardless of the data type. I don't think we consider the data type here again. WDYT?

itholic: @ueshin 

Ah, i got what you meant.

yes i totally agree with you and this part should be corrected.

but FYI, i finally remember that the very first reason why i check the `StringType` here.

at the first time, i thought the behavior of `loc` is slightly different between when `StringType` and `LongType` because of the below example.

```python
>>> pdf1
   max_speed  shield
A          1       2
C          4       5
B          7       8

>>> pdf2
    max_speed  shield
10          1       2
20          4       5
30          7       8

>>> pdf1.loc['A':'Z']
...
ValueError: index must be monotonic increasing or decreasing

During handling of the above exception, another exception occurred:

...
KeyError: 'Z'

>>> pdf2.loc[10:10000]
    max_speed  shield
10          1       2
20          4       5
30          7       8
```

if the given key of slice, `start` or `stop`, is not included list of index, i thought that the `StringType` raise `KeyError` while `LongType` works well, but it was wrong.

this different was not come from the type of index, but `ValueError: index must be monotonic increasing or decreasing`.

And the cases that index has not index as monotonic increasing or decreasing order is missing our test cases for `loc`, i had have mistaken.

i'll make it work properly soon.

thanks for the comment! :)

databricks/koalas/indexing.py
itholic: > Btw, seems like pandas throws exceptions when the specified values don't exist or are duplicated.

you're right, but more strictly say, they throws exceptions when the specified values don't exist or are duplicated, and the index order is not monotonically increasing or decreasing. 

```python
>>> pdf
    max_speed  shield
10          1       2
30          4       5
20          7       8

>>> pdf.loc[1:20]
...
ValueError: index must be monotonic increasing or decreasing
...
...
KeyError: 1
```

i think it seems not related with duplicated

```python
>>> pdf
    max_speed  shield
10          1       2
20          4       5
20          7       8
20          3       2
30          9      10
30          2       6
>>>
>>> pdf.loc[1:20]
    max_speed  shield
10          1       2
20          4       5
20          7       8
20          3       2
```

ueshin: I see. It's mixed of the monotonic-increasing and the mis-hit and the duplication ..

ueshin:
btw, we can remove line 236?
> In addition, with a slice, Koalas works as a filter between the range.

databricks/koalas/indexing.py
itholic: Yeah it is a bit complicated :confused:

itholic:
Resolves #1173

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,56 @@ def _select_rows(self, rows_sel):
...
+                        start_order_column = index_column._scol
+                if stop is None and rows_sel.stop is not None:
+                    if not (index.is_monotonic_decreasing or index.is_monotonic_increasing):
+                        raise KeyError(rows_sel.stop)
+                    else:
+                        stop = rows_sel.stop
+                        stop_order_column = index_column._scol
+
+                # we don't use StringType since we're using `__natural_order__` for comparing
+                if index_data_type is StringType:
```
HyukjinKwon: `isinstance(index_data_type, StringType)`? I am not sure how it works because `index_data_type` will be an instance of a data type, and `index_data_type is StringType` doesn't seem compare properly. Can we have a test?

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,56 @@ def _select_rows(self, rows_sel):
                 # If slice is None - select everything, so nothing to do
                 return None, None
             elif len(self._internal.index_columns) == 1:
+                sdf = self._kdf_or_kser._internal.sdf
+                index = self._kdf_or_kser.index
+                index_column = index.to_series()
+                index_data_type = index_column.spark_type
                 start = rows_sel.start
                 stop = rows_sel.stop
```
HyukjinKwon: Can we remove this and use `rows_sel.start` and `rows_sel.stop`? Seems confusing what `start` and `stop` have inside because it's later changed to the position in the natural order.

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,56 @@ def _select_rows(self, rows_sel):
...
+                start = [row[1] for row in start_and_stop if row[0] == start]
+                start = start[0] if len(start) > 0 else None
+
+                stop = [row[1] for row in start_and_stop if row[0] == stop]
+                stop = stop[0] if len(stop) > 0 else None
+
+                # if index order is not monotonic increasing or decreasing
+                # and specified values don't exist in index, raise KeyError
+                if start is None and rows_sel.start is not None:
+                    if not (index.is_monotonic_decreasing or index.is_monotonic_increasing):
```
HyukjinKwon: Let's do `index.is_monotonic_increasing or index.is_monotonic_decreasing`. Arguably, index is often monotonically increasing, and if that's the case we can avoid executing `is_monotonic_decreasing` by short-circuiting.

databricks/koalas/indexing.py
itholic: ah.. yes i seems must be fixed.

i have tested before commit, but maybe it was not considering enough cases.

thanks!

databricks/koalas/indexing.py
itholic: > Arguably, index is often monotonically increasing, 

yes that makes sense. thanks!

HyukjinKwon: Can you change here too?

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,56 @@ def _select_rows(self, rows_sel):
...
+                        stop = rows_sel.stop
+                        stop_order_column = index_column._scol
+
+                # we don't use StringType since we're using `__natural_order__` for comparing
+                if isinstance(index_data_type, StringType):
+                    index_data_type = LongType()
+
+                # if start and stop are same, just get all start(or stop) values
+                if start == stop:
+                    return index_column._scol == F.lit(rows_sel.start).cast(index_data_type), None
```
HyukjinKwon: Why do we need to use `LongType()` when this is a `StringType()`? `rows_sel.start` is a string and `index_column._scol` is a string column.

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,56 @@ def _select_rows(self, rows_sel):
...
+                start_and_stop = (
+                    sdf.select(index_column._scol, NATURAL_ORDER_COLUMN_NAME)
+                       .where((index_column._scol == start) | (index_column._scol == stop))
+                       .collect())
+
+                start = [row[1] for row in start_and_stop if row[0] == start]
+                start = start[0] if len(start) > 0 else None
+
+                stop = [row[1] for row in start_and_stop if row[0] == stop]
+                stop = stop[0] if len(stop) > 0 else None
```
HyukjinKwon: To cover duplicated, you can use `stop[-1]` I think.

HyukjinKwon:
```python
>>> pd.DataFrame([1, 2, 3], index=[1, 2, 2]).loc[:2]
   0
1  1
2  2
2  3
```
vs
```python
>>> ks.DataFrame([1, 2, 3], index=[1, 2, 2]).loc[:2]
   0
1  1
2  2
```

databricks/koalas/indexing.py
HyukjinKwon: Oh, okay, there are two cases where this condition can be hit.

1. One is both data exist in the frame so `start` and `stop` become actual indexes
2. If they don't exist, it takes the value as is (the case I am talking about now).

In second case, I think we don't need to do such comparison.

HyukjinKwon:
Looks fine otherwise.

databricks/koalas/indexing.py
itholic: @HyukjinKwon 

Exactly. thanks for the comment! i'm fixing now.  :)

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,56 @@ def _select_rows(self, rows_sel):
...
+                        stop = rows_sel.stop
+                        stop_order_column = index_column._scol
+
+                # if start and stop are same, just get all start(or stop) values
+                if start == stop:
+                    return index_column._scol == F.lit(rows_sel.start).cast(index_data_type), None
+
+                # we don't use StringType since we're using `__natural_order__` for comparing
+                if isinstance(index_data_type, StringType):
+                    index_data_type = LongType()
```
HyukjinKwon: I think this logic itself is incorrect. `start_order_column` can be either `index_column._scol` or `__natual_column__`. So, when the type is string, it can be either a string type or long type.

```python
>>> pd.DataFrame([1, 2, 3], index=['a', 'b', 'd']).loc['a':'z']
   0
a  1
b  2
d  3
```

```python
>>> ks.DataFrame([1, 2, 3], index=['a', 'b', 'd']).loc['a':'z']
Empty DataFrame
Columns: [0]
Index: []
```

HyukjinKwon: I think you should:


```python
                # Assume we use the natural order by default.
                start_order_column_type = LongType()
                stop_order_column_type = LongType()

                # if index order is not monotonic increasing or decreasing
                # and specified values don't exist in index, raise KeyError
                if start is None and rows_sel.start is not None:
                    if not (index.is_monotonic_decreasing or index.is_monotonic_increasing):
                        raise KeyError(rows_sel.start)
                    else:
                        start = rows_sel.start
                        start_order_column = index_column._scol
                        start_order_column_type = index_data_type
                if stop is None and rows_sel.stop is not None:
                    if not (index.is_monotonic_decreasing or index.is_monotonic_increasing):
                        raise KeyError(rows_sel.stop)
                    else:
                        stop = rows_sel.stop
                        stop_order_column = index_column._scol
                        stop_order_column_type = index_data_type

                # if start and stop are same, just get all start(or stop) values
                if start == stop:
                    return index_column._scol == F.lit(rows_sel.start).cast(index_data_type), None

                cond = []
                if start is not None:
                    cond.append(start_order_column >= F.lit(start).cast(start_order_column_type))
                if stop is not None:
                    cond.append(stop_order_column <= F.lit(stop).cast(stop_order_column_type))
```

HyukjinKwon:
Thanks, merged.

databricks/koalas/indexing.py
```diff
@@ -400,16 +395,58 @@ def _select_rows(self, rows_sel):
...
+                # and specified values don't exist in index, raise KeyError
+                if start is None and rows_sel.start is not None:
+                    if not (index.is_monotonic_increasing or index.is_monotonic_decreasing):
+                        raise KeyError(rows_sel.start)
+                    else:
+                        start = rows_sel.start
+                        start_order_column = index_column._scol
+                        start_order_column_type = index_data_type
+                if stop is None and rows_sel.stop is not None:
+                    if not (index.is_monotonic_increasing or index.is_monotonic_decreasing):
```
ueshin: TODO: if both start and stop don't hit, both `index.is_monotonic_increasing` and `index.is_monotonic_decreasing` run twice. we should avoid the overhead.

itholic: @ueshin 

thanks for the comment. okay i'll keep an eye on this to resolve.

ueshin: Submitted #1174.

itholic: thanks for fixing this!! resolved. :)

PR #1160, opened 2019-12-30
Remove the orphan scenario-based test alone

HyukjinKwon:
This PR removes the orphan scenario test alone in CI. Such test cases take a while to pass. Rather than doing it in the regular CI PR build, it should rather just do it periodically somewhere else.
To iterate dev loop quicker, I think we should focus on adding unittest or scoped functional test for now (as we have been doing it so far)


HyukjinKwon:
cc @ueshin

ueshin:
Thanks! merging.

PR #1162, opened 2020-01-02
Implements argmin()/argmax() for Index/MultiIndex

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.argmax.html#pandas.Index.argmax

```python
>>> kidx = ks.Index([10, 9, 8, 7, 100, 5, 4, 3, 100, 3])
>>> kidx
Int64Index([10, 9, 8, 7, 100, 5, 4, 3, 100, 3], dtype='int64')

>>> kidx.argmax()
4
>>> kidx.argmin()
7
```

and not support for MultiIndex like pandas.

```python
>>> midx = ks.MultiIndex([[100, 200, 300],
...                       [50, 60, 70]],
...                       [1, 1, 1, 1, 1, 2, 1, 2, 2]])
>>> midx.argmax()
...
TypeError: reduction operation 'argmax' not allowed for this dtype
>>> midx.argmin()
...
TypeError: reduction operation 'argmin' not allowed for this dtype
```


databricks/koalas/indexes.py
```diff
@@ -958,6 +958,105 @@ def max(self):
...
+        >>> kidx
+        Int64Index([10, 9, 8, 7, 100, 5, 4, 3, 100, 3], dtype='int64')
+
+        >>> kidx.argmax()
+        4
+        """
+        max_idx = self.max()
+        sdf = self._internal.sdf
+        sdf = sdf.select(self.to_series()._scol.alias('__index_value__'))
+        # sdf here looks like below
```
HyukjinKwon: I think we don't need this comment.

databricks/koalas/indexes.py
```diff
@@ -958,6 +958,105 @@ def max(self):
...
+        >>> kidx = ks.Index([10, 9, 8, 7, 100, 5, 4, 3, 100, 3])
+        >>> kidx
+        Int64Index([10, 9, 8, 7, 100, 5, 4, 3, 100, 3], dtype='int64')
+
+        >>> kidx.argmax()
+        4
+        """
+        max_idx = self.max()
+        sdf = self._internal.sdf
+        sdf = sdf.select(self.to_series()._scol.alias('__index_value__'))
```
HyukjinKwon: `self.to_series()._scol` -> `self._scol`

HyukjinKwon: Seems also we don't need this temp column. You can just call directly against `self._scol`.

databricks/koalas/indexes.py
```diff
@@ -1358,6 +1457,12 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
 
     value_counts.__doc__ = IndexOpsMixin.value_counts.__doc__
 
+    def argmax(self):
+        raise TypeError("reduction operation 'argmax' not allowed for this dtype")
```
HyukjinKwon: Just for clarification, it's matched to pandas' multi-index right?

itholic: yes you're right. :)

```python
>>> pmidx = pd.MultiIndex.from_tuples([('a', 'x'), ('b', 'y')])
>>> pmidx.argmax()
...
TypeError: reduction operation 'argmax' not allowed for this dtype
```

databricks/koalas/indexes.py
itholic: it is same reason with https://github.com/databricks/koalas/pull/1165#discussion_r362698852

databricks/koalas/indexes.py
```diff
@@ -961,6 +961,91 @@ def max(self):
...
+        # |                2|              8|
+        # |                3|              7|
+        # |                6|              4|
+        # |                5|              5|
+        # |                7|              3|
+        # |                8|            100|
+        # |                1|              9|
+        # +-----------------+---------------+
+
+        sdf = sdf.where(scol_for(sdf, '__index_value__') == max_idx)
```
HyukjinKwon: @itholic, can we do this in a single job like the below?

```python
sdf.orderBy(
    F.col("__index_value__").asc(), F.col("__natural_order__").desc()
).select("__index_level_0__").first()[0]
```

HyukjinKwon: BTW, it's just an example. Use, for instance, `NATURAL_ORDER_COLUMN_NAME`.

itholic: i seems possible. thanks for the comment i just fixed!

PR #1163, opened 2020-01-02
Implements append() for Index/MultiIndex

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.append.html#pandas.Index.append

```python
>>> kidx = ks.Index([10, 5, 0, 5, 10, 5, 0, 10])
>>> kidx
Int64Index([10, 5, 0, 5, 10, 5, 0, 10], dtype='int64')

>>> kidx.append(kidx)
Int64Index([10, 5, 0, 5, 10, 5, 0, 10, 10, 5, 0, 5, 10, 5, 0, 10], dtype='int64')
```

Support for MiltiIndex

```python
>>> kidx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y')])
>>> kidx  # doctest: +SKIP
MultiIndex([('a', 'x'),
            ('b', 'y')],
           )

>>> kidx.append(kidx)  # doctest: +SKIP
MultiIndex([('a', 'x'),
            ('b', 'y'),
            ('a', 'x'),
            ('b', 'y')],
           )
```


databricks/koalas/indexes.py
```diff
@@ -958,6 +958,56 @@ def max(self):
...
+                    ('b', 'y')],
+                   )
+        """
+        if type(self) is not type(other):
+            raise NotImplementedError(
+                "append() between Index & MultiIndex currently is not supported")
+
+        sdf_self = self._internal.sdf
+        sdf_other = other._internal.sdf
+        sdf_appended = sdf_self.union(sdf_other)
```
HyukjinKwon: @itholic can you also check when the name of indexes are different, and when they are index / multi-index?

itholic: > can you also check when the name of indexes are different

added test cases & fix related logic

> and when they are index / multi-index?

i think we currently hard to support since we can't make index with various types like below pandas' case.

```python
>>> pidx1 = pd.Index([10, 5, 0, 5, 10, 5, 0, 10])
>>> pidx2 = pd.MultiIndex.from_tuples([('a', 'x'), ('b', 'y')])
>>> pidx1.append(pidx2)
Index([10, 5, 0, 5, 10, 5, 0, 10, ('a', 'x'), ('b', 'y')], dtype='object')
```

ueshin: > pandas doesn't keep the name of index, so we're also following that behavior

Actually pandas does keep the name of index for multi-index.

```py
>>> pmidx1
MultiIndex([('a', 'x', 1),
            ('b', 'y', 2),
            ('c', 'z', 3)],
           names=['x', 'y', 'z'])
>>> pmidx2
MultiIndex([('a', 'x', 1),
            ('b', 'y', 2),
            ('c', 'z', 3)],
           names=['p', 'q', 'r'])
>>> pmidx1.append(pmidx2)
MultiIndex([('a', 'x', 1),
            ('b', 'y', 2),
            ('c', 'z', 3),
            ('a', 'x', 1),
            ('b', 'y', 2),
            ('c', 'z', 3)],
           names=['x', 'y', 'z'])
```

databricks/koalas/indexes.py
```diff
@@ -961,6 +961,56 @@ def max(self):
...
```
ueshin: This doesn't follow the actual index:

```py
>>> kdf1
   a  b
a  1  4
b  2  5
c  3  6
>>> kdf2
   a   d
x  7  10
y  8  11
z  9  12

>>> kdf1.set_index('a').index.append(kdf2.set_index('d').index)
Int64Index([1, 2, 3, 7, 8, 9], dtype='int64')
```

whereas:

```py
>>> pdf1.set_index('a').index.append(pdf2.set_index('d').index)
Int64Index([1, 2, 3, 10, 11, 12], dtype='int64')
```

databricks/koalas/indexes.py
itholic: @ueshin 

opps i don't understand why they only keep the `names` for `MultiIndex` while not `name` for `Index` ! (just joke i'll manage it 😄 )

Thanks for checking this out!!

ueshin: > i don't understand why they only keep the `names` for `MultiIndex` while not keep `name` for `Index` !

Agree. I prefer to keep the name even for `Index` .. Let's follow pandas for now. 😺

itholic: @ueshin 

yeah let's just keep following pandas 👍  

i just fixed it :)

databricks/koalas/indexes.py
itholic: @ueshin thanks for checking this out, i just fixed & added related test cases :)

databricks/koalas/indexes.py
```diff
@@ -961,6 +961,62 @@ def max(self):
...
+
+        # names should be kept when MultiIndex, but Index wouldn't keep its name.
+        if isinstance(self, MultiIndex):
+            index_map = self._internal.index_map
+        else:
+            index_map = [(idx_col, None) for idx_col in self._internal.index_columns]
+
+        internal = _InternalFrame(
+            sdf=sdf_appended.select(self._internal.index_scols),
```
ueshin: Do we need to select again?

itholic: it seems you're right. thanks! :)

PR #1168, opened 2020-01-04
DataFrameGroupBy.describe

### LINKED ISSUE #1166: DataFrameGroupBy.describe

deepyaman:
Close #1166 

Manual test:

```
>>> df = ks.DataFrame({'a': [1, 1, 3], 'b': [4, 5, 6], 'c': [7, 8, 9]})
20/01/04 18:16:35 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Using Spark's default log4j profile: org/apache/spark/log4j-defaults.properties
Setting default log level to "WARN".
To adjust logging level use sc.setLogLevel(newLevel). For SparkR, use setLogLevel(newLevel).
>>> df.groupby('a').describe()
20/01/04 18:17:09 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
20/01/04 18:17:10 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
20/01/04 18:17:11 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
      b                                        c                                   
  count mean       std min 25% 50% 75% max count mean       std min 25% 50% 75% max
a                                                                                  
1     2  4.5  0.707107   4   4   4   5   5     2  7.5  0.707107   7   7   7   8   8
3     1  6.0       NaN   6   6   6   6   6     1  9.0       NaN   9   9   9   9   9
```

TODO:

- [x] Fix formatting (line length over 100, etc.)
- [ ] Add docstring
- [x] Add unit tests
- [x] Reorder percentiles

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -232,29 +232,6 @@ class GroupBy(object):
                               column_scols=[scol_for(sdf, col) for col in data_columns],
                               index_map=index_map)
 
-    def describe(self):
-        kdf = self.agg(["count", "mean", "std", "min", "quartiles", "max"]).reset_index()
-
-        # Split "quartiles" columns into first, second, and third quartiles.
-        for label, content in kdf.iteritems():
-            if label[1] == "quartiles":
-                exploded = ks.DataFrame(content.tolist())
-                exploded.columns = [(label[0], "25%"), (label[0], "50%"), (label[0], "75%")]
-                kdf = kdf.drop(label).join(exploded)
-
-        # Reindex the DataFrame to reflect initial grouping and agg columns.
-        input_groupnames = [s.name for s in self._groupkeys]
-        kdf.set_index([(key, "") for key in input_groupnames], inplace=True)
-        kdf.index.names = input_groupnames
-
-        # Reorder columns lexicographically by agg column followed by stats.
-        agg_cols = (col.name for col in self._agg_columns)
-        stats = ["count", "mean", "std", "min", "25%", "50%", "75%", "max"]
-        kdf = kdf[list(product(agg_cols, stats))]
-
-        # Cast columns to ``"float64"`` to match `pandas.DataFrame.groupby`.
-        return kdf.astype("float64")
-
     def count(self):
         """
         Compute count of group, excluding missing values.
@@ -1966,6 +1943,29 @@ class DataFrameGroupBy(GroupBy):
                                       column_scols=[scol_for(sdf, c._internal.data_columns[0])
                                                     for c in applied])
         return DataFrame(internal)
+
+    def describe(self):
+        kdf = self.agg(["count", "mean", "std", "min", "quartiles", "max"]).reset_index()
+
+        # Split "quartiles" columns into first, second, and third quartiles.
+        for label, content in kdf.iteritems():
+            if label[1] == "quartiles":
+                exploded = ks.DataFrame(content.tolist())
+                exploded.columns = [(label[0], "25%"), (label[0], "50%"), (label[0], "75%")]
+                kdf = kdf.drop(label).join(exploded)
+
+        # Reindex the DataFrame to reflect initial grouping and agg columns.
+        input_groupnames = [s.name for s in self._groupkeys]
+        kdf.set_index([(key, "") for key in input_groupnames], inplace=True)
+        kdf.index.names = input_groupnames
+
+        # Reorder columns lexicographically by agg column followed by stats.
+        agg_cols = (col.name for col in self._agg_columns)
+        stats = ["count", "mean", "std", "min", "25%", "50%", "75%", "max"]
+        kdf = kdf[list(product(agg_cols, stats))]
+
+        # Cast columns to ``"float64"`` to match `pandas.DataFrame.groupby`.
+        return kdf.astype("float64")
 
 
 class SeriesGroupBy(GroupBy):
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -210,6 +210,12 @@ class GroupBy(object):
                 if aggfunc == "nunique":
                     reordered.append(
                         F.expr('count(DISTINCT `{0}`) as `{1}`'.format(name, data_col)))
+
+                # Implement "quartiles" aggregate function for ``describe``.
+                elif aggfunc == "quartiles":
+                    reordered.append(
+                        F.expr('percentile_approx(`{0}`, array(0.25, 0.5, 0.75)) as `{1}`'.format(name, data_col)))
+
                 else:
                     reordered.append(F.expr('{1}(`{0}`) as `{2}`'.format(name, aggfunc, data_col)))
         sdf = sdf.groupby(*groupkey_cols).agg(*reordered)
@@ -223,6 +229,23 @@ class GroupBy(object):
                               column_index=column_index,
                               column_scols=[scol_for(sdf, col) for col in data_columns],
                               index_map=index_map)
+
+    def describe(self):
+        kdf = self.agg(["count", "mean", "std", "min", "quartiles", "max"]).reset_index()
+
+        # Split "quartiles" columns into first, second, and third quartiles.
+        for label, content in kdf.iteritems():
+            if label[1] == "quartiles":
+                exploded = ks.DataFrame(content.tolist())
+                exploded.columns = [(label[0], "25%"), (label[0], "50%"), (label[0], "75%")]
+                kdf = kdf.drop(label).join(exploded)
+
+        # Reindex the DataFrame to reflect initial grouping and agg columns.
+        input_groupnames = [s.name for s in self._groupkeys]
+        kdf.set_index([(key, "") for key in input_groupnames], inplace=True)
+        kdf.index.names = input_groupnames
+
+        return kdf
 
     def count(self):
         """
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -215,7 +215,8 @@ class GroupBy(object):
                 # Implement "quartiles" aggregate function for ``describe``.
                 elif aggfunc == "quartiles":
                     reordered.append(
-                        F.expr('percentile_approx(`{0}`, array(0.25, 0.5, 0.75)) as `{1}`'.format(name, data_col)))
+                        F.expr('percentile_approx(`{0}`, array(0.25, 0.5, 0.75)) as `{1}`'.format(
+                            name, data_col)))
 
                 else:
                     reordered.append(F.expr('{1}(`{0}`) as `{2}`'.format(name, aggfunc, data_col)))
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -22,6 +22,7 @@ import sys
 import inspect
 from collections import Callable, OrderedDict, namedtuple
 from functools import partial
+from itertools import product
 from typing import Any, List, Tuple, Union
 
 import numpy as np
@@ -245,7 +246,10 @@ class GroupBy(object):
         kdf.set_index([(key, "") for key in input_groupnames], inplace=True)
         kdf.index.names = input_groupnames
 
-        return kdf
+        # Reorder columns lexicographically by agg column followed by stats.
+        agg_cols = (col.name for col in self._agg_columns)
+        stats = ["count", "mean", "std", "min", "25%", "50%", "75%", "max"]
+        return kdf[list(product(agg_cols, stats))]
 
     def count(self):
         """
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -249,7 +249,11 @@ class GroupBy(object):
         # Reorder columns lexicographically by agg column followed by stats.
         agg_cols = (col.name for col in self._agg_columns)
         stats = ["count", "mean", "std", "min", "25%", "50%", "75%", "max"]
-        return kdf[list(product(agg_cols, stats))]
+        order = list(product(agg_cols, stats))
+        kdf = kdf[order]
+
+        # Cast columns to ``"float64"`` to match `pandas.DataFrame.groupby`.
+        return kdf.astype({col: "float64" for col in order})
 
     def count(self):
         """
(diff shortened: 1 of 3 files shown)


itholic:
hmm it seems that the result is not same with pandas??

```python
>>> kdf = ks.DataFrame({'a': [1, 1, 3], 'b': [4, 5, 6], 'c': [7, 8, 9]})
>>> pdf = kdf.to_pandas()

>>> kdf.groupby('a').describe()
      b                            c                          b           c
  count mean       std min max count mean       std min max 25% 50% 75% 25% 50% 75%
a
1     2  4.5  0.707107   4   5     2  7.5  0.707107   7   8   4   4   5   7   7   8
3     1  6.0       NaN   6   6     1  9.0       NaN   9   9   6   6   6   9   9   9

>>> pdf.groupby('a').describe()
      b                                               c
  count mean       std  min   25%  50%   75%  max count mean       std  min   25%  50%   75%  max
a
1   2.0  4.5  0.707107  4.0  4.25  4.5  4.75  5.0   2.0  7.5  0.707107  7.0  7.25  7.5  7.75  8.0
3   1.0  6.0       NaN  6.0  6.00  6.0  6.00  6.0   1.0  9.0       NaN  9.0  9.00  9.0  9.00  9.0
```

first of all, i think we better add an unittest to `tests/test_groupby.py` even just very simple case like below.

```python
def test_describe(self):
    kdf = ks.DataFrame({'a': [1, 1, 3], 'b': [4, 5, 6], 'c': [7, 8, 9]})
    pdf = kdf.to_pandas()

    self.assert_eq(kdf.groupby('a').describe.sort_index(),
                   pdf.groupby('a').describe.sort_index())
``` 

and thanks for the contribution, let's make it work together 👍

databricks/koalas/groupby.py
```diff
@@ -224,6 +230,23 @@ def _spark_groupby(kdf, func, groupkeys):
                               column_scols=[scol_for(sdf, col) for col in data_columns],
                               index_map=index_map)
 
+    def describe(self):
+        kdf = self.agg(["count", "mean", "std", "min", "quartiles", "max"]).reset_index()
+
+        # Split "quartiles" columns into first, second, and third quartiles.
+        for label, content in kdf.iteritems():
+            if label[1] == "quartiles":
+                exploded = ks.DataFrame(content.tolist())
```
itholic: this line seems a little danger since it can potentially raise memory issue such like OOM,
(because `tolist()` loads all the data into the single driver's memory.)

so i think maybe we can use `content.to_frame()`, or find another way.

or we can simply just don't support quartiles for now since memory issue, and note it to docstring.

deepyaman:
> hmm it seems that the result is not same with pandas??

Good catch. I assume you're referring to the cases where the values are different moreso than the differences in decimal representation or column order.

Under the hood, pandas `describe` uses `quantile`, letting `interpolation` default to `'linear'`. You'd actually get the same quartiles if you specified `'nearest'` instead:

```
>>> pdf.groupby('a').quantile([0.25, 0.5, 0.75], interpolation='nearest')
        b  c
a           
1 0.25  4  7
  0.50  4  7
  0.75  5  8
3 0.25  6  9
  0.50  6  9
  0.75  6  9
```

That being said, I don't think you can even guarantee that the implementation would match the pandas result specifying `interpolation` in all cases, since I used `percentile_approx` to handle larger data.

Would it make sense to reimplement some sort of linear interpolation, or is documenting that the quantiles aren't interpolated sufficient? I'm not aware of any sort of interpolated percentile calculation in Spark, but I could be wrong.

Happy to add a test tomorrow or day after, probably based on what think the expected (interpolation) behavior should be?

deepyaman:
> since the one of main purpose of Koalas is make our API works same with pandas as much as possible, this part should be made clear very first.

Speaking of keeping the APIs as similar as possible, do you think the fact that I've added a `"quartiles"` aggregation function that doesn't exist in pandas is okay? That's actually something I was concerned about.

If not, I could use your help in determining how else to do it. My first thought would be to create a function very similar to `_spark_groupby`, but just with support for that "quartiles" aggregation function. Then could implement `DataFrameGroupBy.quantile` (calling the new function), and use the new `quantile` in `describe`.

itholic:
@deepyaman 

thanks for the explanation :)

ah sorry i just meant that the 'shape' of the result looks not same with pandas.

for example, `DataFrameGroupBy.describe` for Koalas looks like below:
(columns `b` and `c` are repeated twice.)

```python
>>> kdf.groupby('a').describe()
      b                            c                          b           c
  count mean       std min max count mean       std min max 25% 50% 75% 25% 50% 75%
a
1     2  4.5  0.707107   4   5     2  7.5  0.707107   7   8   4   4   5   7   7   8
3     1  6.0       NaN   6   6     1  9.0       NaN   9   9   6   6   6   9   9   9
```

whereas for pandas looks like below:
(columns `b` and `c` appear only once.)

```python
>>> pdf.groupby('a').describe()
      b                                               c
  count mean       std  min   25%  50%   75%  max count mean       std  min   25%  50%   75%  max
a
1   2.0  4.5  0.707107  4.0  4.25  4.5  4.75  5.0   2.0  7.5  0.707107  7.0  7.25  7.5  7.75  8.0
3   1.0  6.0       NaN  6.0  6.00  6.0  6.00  6.0   1.0  9.0       NaN  9.0  9.00  9.0  9.00  9.0
```

> Happy to add a test tomorrow or day after, probably based on what think the expected (interpolation) behavior should be?

i think the idea of now implementation that using `percentile_approx` looks good.

and maybe it is better just adding the note to docstring why the result is slightly different from pandas as you commented. (to handle larger data)

itholic:
@deepyaman 

> Speaking of keeping the APIs as similar as possible, do you think the fact that I've added a "quartiles" aggregation function that doesn't exist in pandas is okay? That's actually something I was concerned about.

yes you're right.

although it seems okay to me, but maybe we better discuss it with other maintainers who have more insight of those kind of functionalities.

cc @ueshin @HyukjinKwon

deepyaman:
> ah sorry i just meant that the 'shape' of the result looks not same with pandas.

Got it. I'm aware and can fix that relatively easily. Included it on the list of TODOs for this PR in the initial message:

> TODO:
> 
> * [ ]  Fix formatting (line length over 100, etc.)
> * [ ]  Add docstring
> * [ ]  Add unit tests
> * [ ]  Reorder percentiles

"Reorder percentiles" probably wasn't the best description. Anyway, just wanted to see if there were more fundamental issues first. :)

databricks/koalas/groupby.py
ueshin: Yeah, we can just get items from the "quartiles" column.

deepyaman: @itholic @ueshin Hey! Sorry it's taking me a while to make the requested changes. I figured out a way to refactor this using `_column_op`; will it work? I'm still using `to_numpy`, though, only because Koalas doesn't allow constructing a DataFrame from a column-Series mapping yet.

itholic: @deepyaman 

Hi Deepyaman! thanks for your continued efforts here.

Basically, for handling DataFrame of Koalas efficiently,

we usually use internal spark DataFrame (`sdf` in short, and you can get by `kdf._sdf` or `kdf._internal.sdf`)

I made some another way of implementation using `sdf` for you below.
(I can't say that this is a perfect & good quality code, but maybe it will help you to understand Koalas' internal processing even just a bit)

```python
    def describe(self):
        kdf = self.agg(["count", "mean", "std", "min", "max", "quartiles"]).reset_index()
        formatted_percentiles = ["25%", "50%", "75%"]
        sdf = kdf._sdf
        group_key_names = [groupkey.name for groupkey in self._groupkeys]

        quartiles_columns = []
        for data_column in self._kdf._internal.data_columns:
            if data_column not in group_key_names:
                quartiles_columns.append((data_column, 'quartiles'))
        # `quartiles_columns` here looks like the below
        # [('b', 'quartiles'), ('c', 'quartiles')]

        # add columns (b, 25%), (b, 50%) ... (c, 50%), (c, 75%) to `sdf`
        for col_name, quartiles_column in quartiles_columns:
            for i, percentile in enumerate(formatted_percentiles):
                sdf = sdf.withColumn(
                    name_like_string((col_name, percentile)),
                    F.col(name_like_string((col_name, quartiles_column))).getItem(i))
        # so, `sdf` here looks like the below
        # +-----------------+-----+ ... +-----------------+--------+--------+--------+--------+--------+--------+
        # |__index_level_0__|(a, )| ... |__natural_order__|(b, 25%)|(b, 50%)|(b, 75%)|(c, 25%)|(c, 50%)|(c, 75%)|
        # +-----------------+-----+ ... +-----------------+--------+--------+--------+--------+--------+--------+
        # |                0|    1| ... |     592705486848|       4|       4|       5|       7|       7|       8|
        # |                1|    3| ... |     919123001344|       6|       6|       6|       9|       9|       9|
        # +-----------------+-----+ ... +-----------------+--------+--------+--------+--------+--------+--------+

        # make the column list what we want to select from `sdf`
        columns = []
        for col_name, _ in quartiles_columns:
            for func in ["count", "mean", "std", "min", "max", "25%", "50%", "75%"]:
                columns.append((col_name, func))

        name_like_string_columns = [name_like_string(col) for col in columns]
        internal = _InternalFrame(
            sdf=sdf.select(*self._kdf._internal.index_columns, *name_like_string_columns),
            index_map=self._kdf._internal.index_map)

        idx = pd.MultiIndex.from_tuples(columns)
        # `idx` here looks like the below
        # MultiIndex([('b', 'count'),
        #             ('b',  'mean'),
        #             ('b',   'std'),
        #             ('b',   'min'),
        #             ('b',   'max'),
        #             ('b',   '25%'),
        #             ('b',   '50%'),
        #             ('b',   '75%'),
        #             ('c', 'count'),
        #             ('c',  'mean'),
        #             ('c',   'std'),
        #             ('c',   'min'),
        #             ('c',   'max'),
        #             ('c',   '25%'),
        #             ('c',   '50%'),
        #             ('c',   '75%')],
        #            )

        result = DataFrame(internal)
        result.columns = idx

        return result
```

the not implementation invokes several times 

```python
>>> kdf.groupby('a').describe()
2020-01-13 17:08:31 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
2020-01-13 17:08:31 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
2020-01-13 17:08:32 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
2020-01-13 17:08:32 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
2020-01-13 17:08:32 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
2020-01-13 17:08:33 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
2020-01-13 17:08:33 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
      b                                             c
  count mean       std  min  25%  50%  75%  max count mean       std  min  25%  50%  75%  max
a
1   2.0  4.5  0.707107  4.0  4.0  4.0  5.0  5.0   2.0  7.5  0.707107  7.0  7.0  7.0  8.0  8.0
3   1.0  6.0       NaN  6.0  6.0  6.0  6.0  6.0   1.0  9.0       NaN  9.0  9.0  9.0  9.0  9.0
```

we can reduce them via handling internal frame properly like the below.

```python
>>> kdf.groupby('a').describe()
2020-01-13 17:23:15 WARN  WindowExec:66 - No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
      b                                        c
  count mean       std min max 25% 50% 75% count mean       std min max 25% 50% 75%
0     2  4.5  0.707107   4   5   4   4   5     2  7.5  0.707107   7   8   7   7   8
1     1  6.0       NaN   6   6   6   6   6     1  9.0       NaN   9   9   9   9   9
```

deepyaman: @itholic Thank you for the feedback. I'll try to rewrite it following your suggestions above and get back to you.

itholic: @deepyaman My pleasure :) Hope it helps you!

databricks/koalas/groupby.py
```diff
@@ -1936,6 +1946,34 @@ def _shift(self, periods, fill_value):
...
+        # Split "quartiles" columns into first, second, and third quartiles.
+        for label, content in kdf.iteritems():
+            if label[1] == "quartiles":
+                exploded = ks.DataFrame(
+                    {
+                        (label[0], x): _column_op(itemgetter(i))(content).to_numpy()
+                        for i, x in enumerate(formatted_percentiles)
+                    }
+                )
+                kdf = kdf.drop(label).join(exploded)
```
HyukjinKwon: `join` is expensive .. we should avoid. Actually I like @itholic's suggestion in that way.

HyukjinKwon:
I am going to just merge. @itholic can you make a PR to address the comments you pointed out?

HyukjinKwon:
We should also fix docstring.

HyukjinKwon:
Thanks @deepyaman for finding this issue and working on this.

itholic:
@HyukjinKwon okay i'll make PR soon.
@deepyaman Thank you for your efforts!

deepyaman:
@itholic @HyukjinKwon Sorry, I don't get much time to work on these things during the week. I didn't see a PR with the docstring/code updates yet; I started by adding the docstring, if that's OK. I'll add in @itholic's code shortly, too.

itholic:
@deepyaman 

Don't worry and keep going :)

i also didn't check more about this since it's crazy busy last week 😅 

i'll check #1202 soon, and thanks for the contribution!!

PR #1172, opened 2020-01-08
Add support scalar values as row_sel for `loc` indexer.

### LINKED ISSUE #383: .loc behavior for scalar values
Assuming a simple DataFrame `kdf = ks.DataFrame({'A': [1,2,3]})`, running `kdf.loc[1]` results in the following exception:
```SparkPandasNotImplementedError: Cannot use a scalar value for row selection with Spark. You are trying to use pandas function .loc[..., ...], use spark function select, where```

However, `kdf.loc[1:1]` behaves the same as in pandas and returns the wanted column but as a DataFrame while `.loc[1]` would return a Series.

The `LocIndexer` docs state the following:
```
Allowed inputs are:
    - A single label, e.g. ``5`` or ``'a'``, (note that ``5`` is
      interpreted as a *label* of the index, and **never** as an
      integer position along the index) for column selection.
```

Can someone point out the rationale behind this behavior? My only guess would be that the exception tries to keep users from using the label as an integer position, which is not even the point of `.loc` but rather `.iloc`, right?

### PR DESCRIPTION
**Pull request by ueshin**
Resolves #383.

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -25,7 +25,7 @@ from pyspark.sql import functions as F
 from pyspark.sql.types import BooleanType
 from pyspark.sql.utils import AnalysisException
 
-from databricks.koalas.internal import _InternalFrame, HIDDEN_COLUMNS, NATURAL_ORDER_COLUMN_NAME
+from databricks.koalas.internal import _InternalFrame, NATURAL_ORDER_COLUMN_NAME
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: databricks/koalas/indexing.py [hunk 1]
```diff
@@ -269,9 +283,9 @@ class LocIndexer(_LocIndexerLike):
     A single label for row selection is not allowed.
```
HyukjinKwon: seems we should fix this doc too.

PR #1182, opened 2020-01-09
[WIP] Support MultiIndex for `loc` indexer with slice as `row_sel`.

itholic:
Resolves #1175 

```python
>>> kser
x  a    1
   b    2
y  c    3
   d    4
z  e    5
Name: 0, dtype: int64

>>> kser.loc['y':]
y  c    3
   d    4
z  e    5
Name: 0, dtype: int64

>>> kser.loc[:'y']
x  a    1
   b    2
y  c    3
   d    4
Name: 0, dtype: int64

>>> kser.loc[('x', 'b'):]
x  b    2
y  c    3
   d    4
z  e    5
Name: 0, dtype: int64

>>> kser.loc[:('y', 'c')]
x  a    1
   b    2
y  c    3
Name: 0, dtype: int64

>>> kser.loc[('x', 'b'):('y', 'c')]
x  b    2
y  c    3
Name: 0, dtype: int64

>>> kser.loc['x':('y', 'c')]
x  a    1
   b    2
y  c    3
Name: 0, dtype: int64

>>> kser.loc[('x', 'b'):'y']
x  b    2
y  c    3
   d    4
Name: 0, dtype: int64
```

PR #1183, opened 2020-01-09
is_monotonic* for MultiIndex

itholic:
The existing implementation only support for `Series`, `Index`, but not `MultiIndex`

so fix this for also supporting `MultiIndex` like pandas.

- is_monotonic_increasing

```python
>>> midx = ks.MultiIndex.from_tuples(
... [('x', 'a'), ('x', 'b'), ('y', 'c'), ('y', 'd'), ('z', 'e')])
>>> midx
MultiIndex([('x', 'a'),
            ('x', 'b'),
            ('y', 'c'),
            ('y', 'd'),
            ('z', 'e')],
           )
>>> midx.is_monotonic_increasing
True
>>> midx.is_monotonic_decreasing
False
```

- is_monotonic_decreasing

```python
>>> midx = ks.MultiIndex.from_tuples(
... [('z', 'e'), ('z', 'd'), ('y', 'c'), ('y', 'b'), ('x', 'a')])
>>> midx
MultiIndex([('z', 'a'),
            ('z', 'b'),
            ('y', 'c'),
            ('y', 'd'),
            ('x', 'e')],
           )
>>> midx.is_monotonic_increasing
False
>>> midx.is_monotonic_decreasing
True
```


itholic:
cc @ueshin @HyukjinKwon 

FYI: this property(`is_monotonic*`) will be used in #1182 for handling exception logic.

so if available, would you please review this PR first?

ueshin:
Shall we use overriding?

itholic:
@ueshin 

> Shall we use overriding?

Ah, It seems better. I'm going to move this on to `MultiIndex` class

Thanks for the comment :)

databricks/koalas/indexes.py
```diff
@@ -1312,6 +1312,112 @@ def levshape(self):
         result = internal._sdf.agg(*(F.countDistinct(c) for c in internal.index_scols)).collect()[0]
         return tuple(result)
 
+    @property
+    def is_monotonic(self):
```
ueshin: We don't need to override this too. Just overriding `_is_monotonic` should be fine.

ueshin: Let's have the doctests in the super class.

databricks/koalas/indexes.py
```diff
@@ -1312,6 +1312,112 @@ def levshape(self):
...
+        """
+        return self._is_monotonic().all()
+
+    def _is_monotonic(self):
+        col = self._scol
+        window = Window.orderBy(NATURAL_ORDER_COLUMN_NAME).rowsBetween(-1, -1)
+        cond = (col >= F.lag(col, 1).over(window)) & col.isNotNull()
+
+        internal = _InternalFrame(
+            sdf=self._internal.sdf.select(cond))
```
ueshin: This cause an unneeded default index attaching. Could you avoid it?

databricks/koalas/indexes.py
```diff
@@ -1312,6 +1312,112 @@ def levshape(self):
...
+        """
+        return self._is_monotonic_decreasing().all()
+
+    def _is_monotonic_decreasing(self):
+        col = self._scol
+        window = Window.orderBy(NATURAL_ORDER_COLUMN_NAME).rowsBetween(-1, -1)
+        cond = (col <= F.lag(col, 1).over(window)) & col.isNotNull()
+
+        internal = _InternalFrame(
+            sdf=self._internal.sdf.select(cond))
```

databricks/koalas/indexes.py
```diff
@@ -1312,6 +1312,112 @@ def levshape(self):
...
+                   )
+        >>> midx.is_monotonic
+        False
+        """
+        return self._is_monotonic().all()
+
+    def _is_monotonic(self):
+        col = self._scol
+        window = Window.orderBy(NATURAL_ORDER_COLUMN_NAME).rowsBetween(-1, -1)
+        cond = (col >= F.lag(col, 1).over(window)) & col.isNotNull()
```
ueshin: Do we need `& col.isNotNull()`?

databricks/koalas/tests/test_indexes.py
```diff
@@ -524,3 +524,145 @@ def test_argmax(self):
                 TypeError,
                 "reduction operation 'argmax' not allowed for this dtype"):
             kidx.argmax()
+
+    def test_monotonic(self):
+        # test monotonic_increasing & monotonic_decreasing for MultiIndex
+        kmidx = ks.MultiIndex.from_tuples(
+            [('w', 'a'), ('x', 'b'), ('y', 'c'), ('z', 'd')])
+        pmidx = kmidx.to_pandas()
```
ueshin: Could you add comments to describe each example is increasing or decreasing or none to be easy to read each case?

ueshin: Also could you add some cases including `None`?

ueshin: Btw, can we use just a loop? Seems like each test is the same shape.

itholic: @ueshin ah, yes i'm working to make this shorten :)

databricks/koalas/indexes.py
itholic: thanks for the comment :)

i've added `index_map` explicitly

itholic:
while working here, i found something.

1. pandas only concern about the order of first index of MultiIndex
```python
>>> pidx
MultiIndex([(1.0, 100),
            (2.0, 200),
            (nan, 300),
            (4.0, 400),
            (4.0, 500)],
           )
>>> pidx.is_monotonic_increasing
False

>>> pidx
MultiIndex([(1, 100.0),
            (2, 200.0),
            (3,   nan),
            (4, 400.0),
            (4, 500.0)],
           )
>>> pidx.is_monotonic_increasing
True
```

2. pandas ignore the last null value of `MultiIndex` unlike `Index`
```python
>>> pidx
Index([1, 2, 3, 4, None], dtype='object')
>>> pidx.is_monotonic_increasing
False

>>> pidx
MultiIndex([(1.0, 100),
            (2.0, 200),
            (3.0, 300),
            (4.0, 400),
            (nan, 500)],
           )
>>> pidx.is_monotonic_increasing
True
```

databricks/koalas/indexes.py
itholic: let me check is more since this matter is related with https://github.com/databricks/koalas/pull/1183#issuecomment-572849529

databricks/koalas/tests/test_indexes.py
```diff
@@ -524,3 +524,66 @@ def test_argmax(self):
...
+        datas.append([(-5, 'e'), (-4, 'd'), (-3, 'c'), (-2, 'b'), (-1, 'a')])
+        datas.append([(-5, 'e'), (-3, 'd'), (-2, 'c'), (-4, 'b'), (-1, 'a')])
+        datas.append([(-5, 'e'), (-4, 'c'), (-3, 'b'), (-2, 'd'), (-1, 'a')])
+
+        # None type tests
+        # TODO: pandas returns `True` commented cases below
+        # since they ignore the `None` positioned last.
+        datas.append([(None, 100), (2, 200), (3, 300), (4, 400), (5, 500)])
+        datas.append([(1, 100), (2, 200), (None, 300), (4, 400), (5, 500)])
+        # datas.append([(1, 100), (2, 200), (3, 300), (4, 400), (None, 500)])
```
itholic: we have some tricky problem here.

if `MultiIndex` has `None` in a row at the last, we should ignore them.

for example.

```python
>>> pidx
MultiIndex([(1.0, 100),
            (2.0, 200),
            (3.0, 300),
            (4.0, 400),
            (nan, 500)],
           )
>>> pidx.is_monotonic
True

>>> pidx
MultiIndex([(1.0, 100),
            (2.0, 200),
            (3.0, 300),
            (nan, 400),
            (nan, 500)],
           )
>>> pidx.is_monotonic
True

>>> pidx
MultiIndex([(1.0, 100),
            (2.0, 200),
            (nan, 300),
            (nan, 400),
            (nan, 500)],
           )
>>> pidx.is_monotonic
True
```

but to this, i think we have to traverse all data and it might possibly quite expensive.

cc @ueshin @HyukjinKwon could you share the idea to solve this if when you available??

databricks/koalas/indexes.py
```diff
@@ -1312,6 +1312,28 @@ def levshape(self):
...
         return tuple(result)
 
+    def _is_monotonic(self):
+        col = self._internal.index_scols[0]
+        window = Window.orderBy(NATURAL_ORDER_COLUMN_NAME).rowsBetween(-1, -1)
+        cond = (col >= F.lag(col, 1).over(window)) & col.isNotNull()
+
+        internal = _InternalFrame(
+            sdf=self._internal.sdf.select(cond),
+            index_map=None)
```
ueshin: `index_map=None` invokes attaching default index.

ueshin:
> 1. pandas only concern about the order of first index of MultiIndex

It does concern about the whole index.

```py
>>> pd.MultiIndex.from_tuples([(1,100),(2,200),(3,None),(4,400),(4,500)]).is_monotonic
True
>>> pd.MultiIndex.from_tuples([(1,100),(2,200),(3,None),(4,500),(4,400)]).is_monotonic
False
```

> 2. pandas ignore the last null value of `MultiIndex` unlike `Index`
>
> if `MultiIndex` has `None` in a row at the last position, we should ignore them when `is_monotonic_increasing`.
> they doesn't apply the same logic to `is_monotonic_decreasing` like the below. they doesn't ignore `None` for `is_monotonic_decreasing` unlike `is_monotonic_increasing`.

It's just handling null value as the biggest number.

```py
>>> pd.MultiIndex.from_tuples([(1,100),(2,200),(3,300),(4,400),(None,500)]).is_monotonic
True
>>> pd.MultiIndex.from_tuples([(None,100),(2,200),(3,300),(4,400),(5,500)]).is_monotonic
False
>>> pd.MultiIndex.from_tuples([(5,100),(4,200),(3,300),(2,400),(None,500)]).is_monotonic_decreasing
False
>>> pd.MultiIndex.from_tuples([(None,100),(4,200),(3,300),(2,400),(1,500)]).is_monotonic_decreasing
True
```

databricks/koalas/indexes.py
ueshin: Shall we keep the index columns to avoid attaching default index?

databricks/koalas/tests/test_indexes.py
```diff
@@ -524,3 +524,66 @@ def test_argmax(self):
...
                 "reduction operation 'argmax' not allowed for this dtype"):
             kidx.argmax()
+
+    def test_monotonic(self):
+        # test monotonic_increasing & monotonic_decreasing for MultiIndex
+        datas = []
+
+        # increasing / decreasing ordered each index level with string
+        datas.append([('w', 'a'), ('x', 'b'), ('y', 'c'), ('z', 'd')])
+        datas.append([('w', 'd'), ('x', 'c'), ('y', 'b'), ('z', 'a')])
```
ueshin: What if

- `('x', 'd'), ('y', 'c'), ('y', 'b'), ('z', 'a')]`
- `('x', 'd'), ('y', 'b'), ('y', 'c'), ('z', 'a')]`

?

databricks/koalas/tests/test_indexes.py
```diff
@@ -524,3 +524,66 @@ def test_argmax(self):
...
+        datas.append([(5, 500), (4, 400), (3, 300), (2, 200), (1, 100)])
+        # mixed order each index level with integer
+        datas.append([(1, 500), (3, 400), (2, 300), (4, 200), (5, 100)])
+        datas.append([(1, 100), (2, 300), (3, 200), (4, 400), (5, 500)])
+
+        # negative tests
+        datas.append([(-1, -100), (-2, -200), (-3, -300), (-4, -400), (-5, -500)])
+        datas.append([(-1, -500), (-2, -400), (-3, -300), (-4, -200), (-5, -100)])
+        datas.append([(-5, -100), (-4, -200), (-3, -300), (-2, -400), (-1, -500)])
+        datas.append([(-5, -500), (-4, -400), (-3, -300), (-2, -200), (-1, -100)])
```
ueshin: Do we need these?

databricks/koalas/tests/test_indexes.py
```diff
@@ -524,3 +524,66 @@ def test_argmax(self):
...
+        datas.append([(5, 100), (4, 200), (3, 300), (2, 400), (1, None)])
+        datas.append([(None, None), (2, 200), (3, 300), (4, 400), (5, 500)])
+        datas.append([(1, 100), (2, 200), (None, None), (4, 400), (5, 500)])
+        # datas.append([(1, 100), (2, 200), (3, 300), (4, 400), (None, None)])
+        datas.append([(-5, None), (-4, None), (-3, None), (-2, None), (-1, None)])
+        datas.append([(None, 'e'), (None, 'c'), (None, 'b'), (None, 'd'), (None, 'a')])
+        # datas.append([(None, None), (None, None), (None, None), (None, None), (None, None)])
+
+        for i, data in enumerate(datas):
+            print(i)
```
ueshin: Please don't forget to remove this before we merge this.

itholic:
> > 1. pandas only concern about the order of first index of MultiIndex
> 
> It does concern about the whole index.
> 
> ```python
> >>> pd.MultiIndex.from_tuples([(1,100),(2,200),(3,None),(4,400),(4,500)]).is_monotonic
> True
> >>> pd.MultiIndex.from_tuples([(1,100),(2,200),(3,None),(4,500),(4,400)]).is_monotonic
> False
> ```
> 
> > 1. pandas ignore the last null value of `MultiIndex` unlike `Index`
> > 
> > if `MultiIndex` has `None` in a row at the last position, we should ignore them when `is_monotonic_increasing`.
> > they doesn't apply the same logic to `is_monotonic_decreasing` like the below. they doesn't ignore `None` for `is_monotonic_decreasing` unlike `is_monotonic_increasing`.
> 
> It doesn't ignore null value but consider as the biggest value.
> 
> ```python
> >>> pd.MultiIndex.from_tuples([(1,100),(2,200),(3,300),(4,400),(None,500)]).is_monotonic
> True
> >>> pd.MultiIndex.from_tuples([(None,100),(2,200),(3,300),(4,400),(5,500)]).is_monotonic
> False
> >>> pd.MultiIndex.from_tuples([(5,100),(4,200),(3,300),(2,400),(None,500)]).is_monotonic_decreasing
> False
> >>> pd.MultiIndex.from_tuples([(None,100),(4,200),(3,300),(2,400),(1,500)]).is_monotonic_decreasing
> True
> ```

Thank you for correcting my misunderstanding.

databricks/koalas/tests/test_indexes.py
itholic: i'll add them to test cases. thanks!

itholic: hmm.. i think we should consider grouped previous level of index for each level of index >= 2.

itholic:
wow it's really tricky.

pandas treats `None` as the largest value only for numeric type.

otherwise (for string type), they returns `False` for `None`.

**AND** they returns `True` if whole values of index 'partition' are `None`.

i think my explantation above seems not enough,

so let me show you some example below.

```python
pd.MultiIndex.from_tuples([('x', 'd'), ('y', None), ('y', 'c'), ('z', 'a')]).is_monotonic_increasing
False
pd.MultiIndex.from_tuples([('x', 'd'), ('y', 'b'), ('y', None), ('z', 'a')]).is_monotonic_increasing
False
pd.MultiIndex.from_tuples([('x', 'd'), ('y', None), ('y', None), ('z', 'a')]).is_monotonic_increasing
True
```

itholic:
> Shall we use overriding?

i think maybe we better separate the logic again since the way of implementation changed a lot from `Index.is_monotonic*`

ueshin:
- for bool type

```py
>>> pd.MultiIndex.from_tuples([('x', True), ('y', True), ('y', None), ('z', True)]).is_monotonic_increasing
True
>>> pd.MultiIndex.from_tuples([('x', True), ('y', None), ('y', True), ('z', True)]).is_monotonic_increasing
True
>>> pd.MultiIndex.from_tuples([('x', True), ('y', None), ('y', None), ('z', True)]).is_monotonic_increasing
True
```

how inconsistent the behavior is ..

ueshin:
@itholic Please check itholic/koalas#1.

itholic:
@ueshin a ha.. there are many things to consider more. 😅 

okay i'll check that PR soon, thanks for helping me solve this problem 👍

PR #1185, opened 2020-01-09
Set index_map explicitly to avoid unconsciously attaching the default index.

ueshin:


ueshin:
I'd merge this now. Please feel free to leave comments if any.

PR #1187, opened 2020-01-10
Add Index.to_frame.

ueshin:


databricks/koalas/tests/test_indexes.py
```diff
@@ -76,6 +76,9 @@ def test_to_series(self):
         self.assert_eq(kidx.to_series(), pidx.to_series())
         self.assert_eq(kidx.to_series(name='a'), pidx.to_series(name='a'))
 
+        # FIXME: the index values are not addressed the change. (#1190)
+        # self.assert_eq((kidx + 1).to_series(), (pidx + 1).to_series())
```
ueshin: Filed #1190.

databricks/koalas/indexes.py
```diff
@@ -1312,6 +1383,92 @@ def levshape(self):
...
+
+        sdf = self._internal.sdf.select([scol.alias(name_like_string(idx))
+                                         for scol, idx in zip(self._internal.index_scols, name)]
+                                        + [NATURAL_ORDER_COLUMN_NAME])
+
+        if index:
+            index_map = [(name_like_string(idx), n)
+                         for idx, n in zip(name, self._internal.index_names)]
+        else:
+            index_map = None  # type: ignore
```
itholic: I'm not pretty sure,

but how about `index_map = [(SPARK_INDEX_NAME_FORMAT(0), None)]`??

then maybe can we avoid `attach_default_index` ?

ueshin: What's the `SPARK_INDEX_NAME_FORMAT(0)` here then?

itholic: ah, i was mistaken. sorry for the confusion. 😞

PR #1192, opened 2020-01-13
Add some examples to describe private vars via __init__ in DataFrame

HyukjinKwon:
It removes the (outdated) documentations for private variables. Instead, it adds some examples against the constructor. So we don't forget to update the doctests as well.


ueshin:
Thanks! merging.

PR #1193, opened 2020-01-14
Fix reset_index with the default index is "distributed-sequence".

ueshin:


databricks/koalas/frame.py
```diff
@@ -2722,36 +2722,32 @@ def rename(index):
...
         else:
-            columns = [name for _, name in new_index_map] + self._internal.column_index
+            column_index = [name for _, name in new_index_map] + self._internal.column_index
+
+        internal = self._internal.copy(
+            sdf=sdf,
+            index_map=index_map,
+            column_index=column_index,
+            column_scols=([scol_for(sdf, name_like_string(name)) for _, name in new_index_map]
+                          + [scol_for(sdf, col) for col in self._internal.data_columns]))
```
ueshin: Essentially this line is the fix.

Issue #1194, opened 2020-01-15
Support pandas>=1.0.0


itholic:
~~As pandas 1.0.0 will be released soon, we need prepare for it.~~

As pandas 1.0.0 has released, we need support it.

so I have created some to-do list below.

- [x] update the list of APIs that are newly added .
- [x] update the list of APIs that are deprecated.
- [x] update the list of APIs that are removed.
- [x] update the name of APIs that are renamed.
- [x] list up & update the implementation of existing Koalas APIs that have incompatible with pandas>=1.0.0.
- [x] update `requirements-dev.txt` for pandas>=1.0.0
- [x] update `.travis.yml` for pandas>=1.0.0
- [x] update GitHub Actions for pandas>=1.0.0

itholic:
cc @ueshin @HyukjinKwon 

If those lists look fine, let me take this work.

itholic:
Please let me know if there are more things should be considered for supporting pandas 1.0.0., i'll add to list.

HyukjinKwon:
Yeah, please go ahead.

ueshin:
Btw, will we drop or keep `pandas<1.0`?
Dropping the older versions might be easy, but I don't think we should drop them until most of pandas users move to `pandas>=1.0`.
@itholic Could you roughly think about the design and estimate how more efforts we will need if we keep them?

itholic:
I have roughly estimate how more efforts we will need if we keep `pandas < 1.0.0` until most of users moved to `pandas >= 1.0.0`.

For now, i think we can handle most of them with proper warning or error message for each pandas version user.

and i think it can be categorized in some ways. (of course might we need more consideration for more detail cases)

### 1. newly added elements in pandas 1.0.0
- i think we can just simply add them to support `pandas < 1.0.0` and `>= 1.0.0` both. (of course, for pandas <1.0.0 users it would be better to print the appropriate error message like "this property (or method) is supported since `pandas >= 1.0.0`")
### 2. removed elements in pandas 1.0.0
- i think maybe we can show `deprecated warning` properly (and since we already haven't considered implementation for those deprecated elements, we don't need to worry too much about this)
### 3. renamed elements
<img width="330" alt="스크린샷 2020-01-19 오후 2 27 30" src="https://user-images.githubusercontent.com/44108233/72675467-d97d6c80-3ac7-11ea-9e5b-b08334253240.png">

- this part is a little considerable, IMO, I think maybe we can support both versions right now like below.
```python
[...]

PR #1197, opened 2020-01-16
Support pandas >=1.0.0

itholic:
Related with #1194 

This PR is for updating newly added & deprecated & deleted APIs on pandas 1.0.0 to keep compatible with pandas 1.0.0 to be released soon, based on [What’s new in 1.0.0](https://pandas.pydata.org/pandas-docs/version/1.0.0/whatsnew/v1.0.0.html)

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -73,7 +73,7 @@ jobs:
         include:
           - python-version: 3.6
             spark-version: 2.4.5
-            pandas-version: 0.24.2
+            pandas-version: [0.24.2, 1.0.1]
             pyarrow-version: 0.13.0
             logger: databricks.koalas.usage_logging.usage_logger
           - python-version: 3.7
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -17,7 +17,7 @@ jobs:
       # The name of the directory '.cache' is for Travis CI. Once we remove Travis CI,
       # we should download Spark to a directory with a different name to prevent confusion.
       SPARK_CACHE_DIR: /home/runner/.cache/spark-versions
-      PANDAS_VERSION: 0.23.4
+      PANDAS_VERSION: 0.24.2
       PYARROW_VERSION: 0.10.0
       # DISPLAY=0.0 does not work in Github Actions with Python 3.5. Here we work around wtih xvfb-run
       PYTHON_EXECUTABLE: xvfb-run python
@@ -73,13 +73,9 @@ jobs:
         include:
           - python-version: 3.6
             spark-version: 2.4.5
-            pandas-version: 0.24.2
+            pandas-version: 0.25.3
             pyarrow-version: 0.13.0
             logger: databricks.koalas.usage_logging.usage_logger
-          - python-version: 3.7
-            spark-version: 2.4.5
-            pandas-version: 0.25.3
-            pyarrow-version: 0.14.1
           - python-version: 3.7
             spark-version: 2.4.5
             pandas-version: 1.0.1
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -70,18 +70,20 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7]
-        spark-version: [2.4.5]
-        pandas-version: [0.24.2, 0.25.3, 1.0.0, 1.0.1]
-        pyarrow-version: [0.13.0, 0.14.1]
-        exclude:
+        include:
           - python-version: 3.6
+            spark-version: 2.4.5
+            pandas-version: 0.24.2
+            pyarrow-version: 0.13.0
+            logger: databricks.koalas.usage_logging.usage_logger
+          - python-version: 3.7
+            spark-version: 2.4.5
             pandas-version: 0.25.3
-          - python-version: 3.6
             pyarrow-version: 0.14.1
           - python-version: 3.7
-            pandas-version: 0.24.2
-          - python-version: 3.7
-            pyarrow-version: 0.13.0
+            spark-version: 2.4.5
+            pandas-version: 1.0.1
+            pyarrow-version: 0.14.1
     env:
       PYTHON_VERSION: ${{ matrix.python-version }}
       SPARK_VERSION: ${{ matrix.spark-version }}
@@ -90,6 +92,7 @@ jobs:
       # `QT_QPA_PLATFORM` for resolving 'QXcbConnection: Could not connect to display :0.0'
       DISPLAY: 0.0
       QT_QPA_PLATFORM: offscreen
+      KOALAS_USAGE_LOGGER: ${{ matrix.logger }}
       # The name of the directory '.cache' is for Travis CI. Once we remove Travis CI,
       # we should download Spark to a directory with a different name to prevent confusion.
       SPARK_CACHE_DIR: /home/runner/.cache/spark-versions
@@ -133,13 +136,10 @@ jobs:
         conda config --prepend envs_dirs $HOME/miniconda/envs
         conda activate test-environment
         export SPARK_HOME="$SPARK_CACHE_DIR/spark-$SPARK_VERSION-bin-hadoop2.7"
-        # Set live logger before run test only for python 3.6
-        if [ $PYTHON_VERSION == 3.6 ]; then
-            export KOALAS_USAGE_LOGGER=databricks.koalas.usage_logging.usage_logger
-        fi
         ./dev/lint-python
         ./dev/pytest
     # Push the results back to codecov
     - name: Update Codecov
       run: bash <(curl -s https://codecov.io/bash)
+
 
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -70,25 +70,18 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7]
-        include:
+        spark-version: [2.4.5]
+        pandas-version: [0.24.2, 0.25.3, 1.0.0, 1.0.1]
+        pyarrow-version: [0.13.0, 0.14.1]
+        exclude:
           - python-version: 3.6
-            spark-version: 2.4.5
-            pandas-version: 0.24.2
-            pyarrow-version: 0.13.0
-            logger: databricks.koalas.usage_logging.usage_logger
-          - python-version: 3.7
-            spark-version: 2.4.5
             pandas-version: 0.25.3
-            pyarrow-version: 0.14.1
-          # tmp test for pandas>=1.0.0
           - python-version: 3.6
-            spark-version: 2.4.5
-            pandas-version: 1.0.1
-            pyarrow-version: 0.13.0
-          - python-version: 3.7
-            spark-version: 2.4.5
-            pandas-version: 1.0.1
             pyarrow-version: 0.14.1
+          - python-version: 3.7
+            pandas-version: 0.24.2
+          - python-version: 3.7
+            pyarrow-version: 0.13.0
     env:
       PYTHON_VERSION: ${{ matrix.python-version }}
       SPARK_VERSION: ${{ matrix.spark-version }}
@@ -97,7 +90,6 @@ jobs:
       # `QT_QPA_PLATFORM` for resolving 'QXcbConnection: Could not connect to display :0.0'
       DISPLAY: 0.0
       QT_QPA_PLATFORM: offscreen
-      KOALAS_USAGE_LOGGER: ${{ matrix.logger }}
       # The name of the directory '.cache' is for Travis CI. Once we remove Travis CI,
       # we should download Spark to a directory with a different name to prevent confusion.
       SPARK_CACHE_DIR: /home/runner/.cache/spark-versions
@@ -141,6 +133,10 @@ jobs:
         conda config --prepend envs_dirs $HOME/miniconda/envs
         conda activate test-environment
         export SPARK_HOME="$SPARK_CACHE_DIR/spark-$SPARK_VERSION-bin-hadoop2.7"
+        # Set live logger before run test only for python 3.6
+        if [ $PYTHON_VERSION == 3.6 ]; then
+            export KOALAS_USAGE_LOGGER=databricks.koalas.usage_logging.usage_logger
+        fi
         ./dev/lint-python
         ./dev/pytest
     # Push the results back to codecov
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,7 @@ install:
       conda config --env --add pinned_packages python=$TRAVIS_PYTHON_VERSION && \
       conda config --env --add pinned_packages pandas==$PANDAS_VERSION && \
       conda config --env --add pinned_packages pyarrow==$PYARROW_VERSION && \
-      conda install -c conda-forge/label/rc --yes pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION && \
+      conda install -c conda-forge --yes pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION && \
       conda install -c conda-forge --yes --freeze-installed --file requirements-dev.txt && \
       conda list;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,17 +38,6 @@ matrix:
         - SPARK_VERSION=2.4.5
         - PANDAS_VERSION=0.25.3
         - PYARROW_VERSION=0.14.1
-    - python: 3.7
-      dist: xenial
-      sudo: true
-      jdk:
-        - oraclejdk8
-      env:
-        - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
-        - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
-        - SPARK_VERSION=2.4.5
-        - PANDAS_VERSION=1.0.1
-        - PYARROW_VERSION=0.14.1
 
 before_install:
   - ./dev/download_travis_dependencies.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,17 +6,15 @@ cache:
 
 matrix:
   include:
-    # commented belows for now because pandas 1.0.0 doesn't support python3.5
-    # we should have discussion about how we handle this. maybe just drop python 3.5?
-    # - python: "3.5"
-    #   jdk:
-    #     - oraclejdk8
-    #   env:
-    #     - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
-    #     - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
-    #     - SPARK_VERSION=2.3.4
-    #     - PANDAS_VERSION=0.23.4
-    #     - PYARROW_VERSION=0.10.0
+    - python: "3.5"
+      jdk:
+        - oraclejdk8
+      env:
+        - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
+        - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
+        - SPARK_VERSION=2.3.4
+        - PANDAS_VERSION=0.23.4
+        - PYARROW_VERSION=0.10.0
     - python: "3.6"
       jdk:
         - oraclejdk8
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ matrix:
         - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
         - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
         - SPARK_VERSION=2.3.4
-        - PANDAS_VERSION=0.23.4
+        - PANDAS_VERSION=1.0.0rc0
         - PYARROW_VERSION=0.10.0
     - python: "3.6"
       jdk:
@@ -22,7 +22,7 @@ matrix:
         - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
         - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
         - SPARK_VERSION=2.4.4
-        - PANDAS_VERSION=0.24.2
+        - PANDAS_VERSION=1.0.0rc0
         - PYARROW_VERSION=0.13.0
         - KOALAS_USAGE_LOGGER='databricks.koalas.usage_logging.usage_logger'
     # When we upgrade Spark 3.0 for this, we could try to just use JDK 11 at 'xenial'.
@@ -36,7 +36,7 @@ matrix:
         - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
         - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
         - SPARK_VERSION=2.4.4
-        - PANDAS_VERSION=0.25.3
+        - PANDAS_VERSION=1.0.0rc0
         - PYARROW_VERSION=0.14.1
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ cache:
 
 matrix:
   include:
-    # pandas>=1.0.0 doesn't support python3.5,
-    # so we just leave the existing PANDAS_VERSION for python3.5 as it is.
     - python: "3.5"
       jdk:
         - oraclejdk8
@@ -24,7 +22,7 @@ matrix:
         - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
         - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
         - SPARK_VERSION=2.4.5
-        - PANDAS_VERSION=1.0.1
+        - PANDAS_VERSION=0.24.2
         - PYARROW_VERSION=0.13.0
         - KOALAS_USAGE_LOGGER='databricks.koalas.usage_logging.usage_logger'
     # When we upgrade Spark 3.0 for this, we could try to just use JDK 11 at 'xenial'.
@@ -38,21 +36,8 @@ matrix:
         - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
         - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
         - SPARK_VERSION=2.4.5
-        - PANDAS_VERSION=1.0.1
+        - PANDAS_VERSION=0.25.3
         - PYARROW_VERSION=0.14.1
-
-    # For testing, we temporarily test pandas<1.0.0 and pandas>=1.0.0 both for python3.6 and 3.7.
-    # TODO: So, we should remove below tests before PR #1197 is merged.
-    - python: "3.6"
-      jdk:
-        - oraclejdk8
-      env:
-        - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
-        - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
-        - SPARK_VERSION=2.4.5
-        - PANDAS_VERSION=0.24.2
-        - PYARROW_VERSION=0.13.0
-        - KOALAS_USAGE_LOGGER='databricks.koalas.usage_logging.usage_logger'
     - python: 3.7
       dist: xenial
       sudo: true
@@ -62,9 +47,8 @@ matrix:
         - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
         - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
         - SPARK_VERSION=2.4.5
-        - PANDAS_VERSION=0.25.3
+        - PANDAS_VERSION=1.0.1
         - PYARROW_VERSION=0.14.1
-    # TODO: Again, never forget to remove above tests before PR #1197 is merged.
 
 before_install:
   - ./dev/download_travis_dependencies.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,17 @@ cache:
 
 matrix:
   include:
-    - python: "3.5"
-      jdk:
-        - oraclejdk8
-      env:
-        - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
-        - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
-        - SPARK_VERSION=2.3.4
-        - PANDAS_VERSION=0.23.4
-        - PYARROW_VERSION=0.10.0
+    # commented belows for now because pandas 1.0.0 doesn't support python3.5
+    # we should have discussion about how we handle this. maybe just drop python 3.5?
+    # - python: "3.5"
+    #   jdk:
+    #     - oraclejdk8
+    #   env:
+    #     - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
+    #     - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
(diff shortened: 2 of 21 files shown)


itholic:
While i'm here, i'll keep checking the pandas release note and updating this PR in order to be compatible with Pandas 1.0.0 as soon as possible.

databricks/koalas/frame.py
```diff
@@ -7964,13 +7960,15 @@ def info(
         ...           encoding="utf-8") as f:
         ...     _ = f.write(s)
         >>> with open('%s/info.txt' % path) as f:
-        ...     f.readlines()  # doctest: +NORMALIZE_WHITESPACE, +ELLIPSIS
-        [...databricks.koalas.frame.DataFrame...,
+        ...     f.readlines()  # doctest: +SKIP
+        ["<class 'databricks.koalas.frame.DataFrame'>\\n",
```
itholic: skip this for supporting pandas<1.0.0 & pandas>=1.0.0 both since their behavior are different depend on pandas version.

databricks/koalas/frame.py
```diff
@@ -7964,13 +7960,15 @@ def info(
         ...           encoding="utf-8") as f:
         ...     _ = f.write(s)
         >>> with open('%s/info.txt' % path) as f:
-        ...     f.readlines()  # doctest: +NORMALIZE_WHITESPACE, +ELLIPSIS
-        [...databricks.koalas.frame.DataFrame...,
+        ...     f.readlines()  # doctest: +SKIP
```
itholic: skip this for supporting pandas<1.0.0 & pandas>=1.0.0 both since their behavior are different depend on pandas version.

itholic:
Although there are several works remain should be solved, i think this PR ready for getting review.

cc @ueshin @HyukjinKwon

.travis.yml
```diff
@@ -6,23 +6,25 @@ cache:
...
+    # we should have discussion about how we handle this. maybe just drop python 3.5?
+    # - python: "3.5"
+    #   jdk:
+    #     - oraclejdk8
+    #   env:
+    #     - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
+    #     - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
+    #     - SPARK_VERSION=2.3.4
+    #     - PANDAS_VERSION=0.23.4
+    #     - PYARROW_VERSION=0.10.0
```
itholic: we need discussion about these.

databricks/koalas/generic.py
```diff
@@ -297,12 +297,12 @@ def cumprod(self, skipna: bool = True):
         """
         return self._apply_series_op(lambda kser: kser._cumprod(skipna))  # type: ignore
 
+    # TODO: Since this method has removed pandas >= 1.0.0,
+    # so we should be block this from using it directly via external API
     def get_dtype_counts(self):
         """
```
itholic: Although `get_dtype_counts` has removed since `pandas>=1.0.0`, we still need this for `DataFrame.info`.

itholic:
We're testing for pandas<1.0.0 and pandas>=1.0.0 both for python3.6 and python3.7 only temporarily in this PR, and must removed tests for old version of pandas before this PR is merged.
(FYI, python3.5 is not supported for pandas1.0.0, and we have to discussion about this)

https://github.com/databricks/koalas/blob/61753e16236f5a904e6580cb16af25bc61f106e5/.travis.yml#L7-L67

databricks/koalas/window.py
```diff
@@ -1420,37 +1439,37 @@ def count(self):
         Examples
         --------
         >>> s = ks.Series([2, 2, 3, 3, 3, 4, 4, 4, 4, 5, 5])
-        >>> s.groupby(s).expanding(3).count().sort_index()  # doctest: +NORMALIZE_WHITESPACE
+        >>> s.groupby(s).expanding(3).count().sort_index()  # doctest: +SKIP
```
itholic: skip this for supporting pandas<1.0.0 & pandas>=1.0.0 both since their behavior are different depend on pandas version.

databricks/koalas/window.py
```diff
@@ -1420,37 +1439,37 @@ def count(self):
...
-           10    2.0
+        5  9     NaN
+           10    NaN
         Name: 0, dtype: float64
 
         For DataFrame, each expanding count is computed column-wise.
 
         >>> df = ks.DataFrame({"A": s.to_numpy(), "B": s.to_numpy() ** 2})
-        >>> df.groupby(df.A).expanding(2).count().sort_index()  # doctest: +NORMALIZE_WHITESPACE
+        >>> df.groupby(df.A).expanding(2).count().sort_index()  # doctest: +SKIP
```
itholic: ditto

databricks/koalas/frame.py
ueshin: Is the updated one from `pandas>=1.0`?

databricks/koalas/groupby.py
```diff
@@ -563,8 +563,11 @@ def diff(self, periods=1):
         5    NaN
         Name: a, dtype: float64
         """
-        return self._apply_series_op(
-            lambda sg: sg._kser._diff(periods, part_cols=sg._groupkeys_scols))
+        if self._kdf.index.is_monotonic or self._kdf.index.is_monotonic_decreasing:
```
ueshin: Shall we use the similar logic to what we did in `loc` indexer?
This can invoke Spark job twice.

ueshin: Btw, why do you know are these related to index ordering?

ueshin: Seems like it's not related to index ordering?

```py
>>> pd.__version__
>>> df = pd.DataFrame({'a': [1,2,3,4,5,6], 'b': [1,1,2,3,5,8], 'c': [1,4,9,16,25,36]}, index=[10,0,20,30,5,15])
'1.0.0'
>>> df
    a  b   c
10  1  1   1
0   2  1   4
20  3  2   9
30  4  3  16
5   5  5  25
15  6  8  36
>>> df.groupby(['b']).diff().sort_index()
      a    c
0   1.0  3.0
5   NaN  NaN
10  NaN  NaN
15  NaN  NaN
20  NaN  NaN
30  NaN  NaN
```

If you saw such a behavior, it could be the different reason.
Could you share the code we can reproduce?

ueshin: Ahh, I found the reproducible example:

```py
>>> df = pd.DataFrame({'a': [1, 2, 3, 4, 5, 6], 'b': [1, 1, 2, 3, 5, 8], 'c': [1, 4, 9, 16, 25, 36]}, index=np.random.rand(6))
>>> df.groupby(['b']).diff().sort_index()
...
```

Actually I suspect it's a pandas' bug.

itholic: thanks for the researching.

and i also tested several examples like the below,

```python
>>> df = pd.DataFrame({'a': [1,2,3,4,5,6], 'b': [1,1,2,3,5,8], 'c': [1,4,9,16,25,36]}, index=[0.4, 0.9, 0.0, 0.6, 0.5, 0.1])
>>> df.groupby(['b']).diff()
       a    c
0.4  NaN  NaN
0.9  1.0  3.0
0.0  NaN  NaN
0.6  NaN  NaN
0.5  NaN  NaN
0.1  NaN  NaN
>>> df = pd.DataFrame({'a': [1,2,3,4,5,6], 'b': [1,1,2,3,5,8], 'c': [1,4,9,16,25,36]}, index=[0.46, 0.97, 0.00, 0.69, 0.55, 0.15])
>>> df.groupby(['b']).diff()
        a    c
0.46  NaN  NaN
0.97  1.0  3.0
0.00  NaN  NaN
0.69  NaN  NaN
0.55  NaN  NaN
0.15  NaN  NaN
>>> df = pd.DataFrame({'a': [1,2,3,4,5,6], 'b': [1,1,2,3,5,8], 'c': [1,4,9,16,25,36]}, index=[0.465, 0.970, 0.004, 0.690, 0.551, 0.155])
>>> df.groupby(['b']).diff()
...
```

so we can guess it might be pandas' bug when index is not monotonic and when float-type index has more than 3 decimal points.

itholic: oops.. but the below pandas example doesn't raise exception even the float-type index has more than 3 decimal points.

```python
>>> df = pd.DataFrame({'a': [1,2,3,4,5,6], 'b': [1,1,2,3,5,8], 'c': [1,4,9,16,25,36]}, index=[0.46101, 0.9700, 0.0000, 0.6900, 0.5500, 0.1500])
>>> df.groupby(['b']).diff()
           a    c
0.46101  NaN  NaN
0.97000  1.0  3.0
0.00000  NaN  NaN
0.69000  NaN  NaN
0.55000  NaN  NaN
0.15000  NaN  NaN
```

hmm.... i think it's totally not reasonable.

databricks/koalas/frame.py
itholic: @ueshin Not explicitly mentioned in the official 1.0.0 documentation, but it seems like that because in pandas 0.25.3 (which is the latest version before 1.0.0) API reference,

스크린샷 2020-02-07 오후 3 39 17

and pandas 1.0.0 API reference.

스크린샷 2020-02-07 오후 3 39 29

ueshin: I see, skipping this is fine.
btw, we should rather update the previous examples skipped above?

databricks/koalas/groupby.py
ueshin: I'd say we can just ignore this behavior for now.
I mean, just revert the changes around here.
WDYT? cc @HyukjinKwon

itholic: > I mean, just revert the changes around here.

@ueshin yeah i totally agree with that since this implementation seems to be too overkilling.

databricks/koalas/frame.py
itholic: okay, then I'll come back here to change this later when we totally migrate to pandas 1.0.0 with dropping the lower versions in the future.

databricks/koalas/groupby.py
itholic: then i'm keeping this for now until @HyukjinKwon response about this.

itholic:
Just FYI:
Now pandas 1.0.1 has been released (https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.0.1.html).

i'll keep following to maintain this PR up-to-date.

requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 # Dependencies in Koalas. When you update don't forget to update setup.py and install.rst in docs.
-pandas>=0.23.2,<1.0
+pandas>=0.23.2,<=1.0.1
```
ueshin: Maybe we can just remove the upper bound?

databricks/koalas/generic.py
itholic: we decided keep this at `Deprecated functions`

requirements-dev.txt
itholic: @ueshin thanks for the comment!

requirements-dev.txt
```diff
@@ -1,5 +1,5 @@
 # Dependencies in Koalas. When you update don't forget to update setup.py and install.rst in docs.
-pandas>=0.23.2,<=1.0.1
+pandas>=0.23.2,<=1.0.*
```
ueshin: I mean, just

```
pandas>=0.23.2
```

should be fine?

ueshin: btw, don't forget to also update `setup.py`.

itholic: ah, yeah i was a bit confused 😹
And just updated `setup.py` also. thanks!

databricks/koalas/groupby.py
HyukjinKwon: +1 for https://github.com/databricks/koalas/pull/1197/files#r376736123

.travis.yml
```diff
@@ -22,11 +24,35 @@ matrix:
...
+        - oraclejdk8
+      env:
+        - PATH=$(echo "$PATH" | sed -e 's/:\/usr\/local\/lib\/jvm\/openjdk11\/bin//')
+        - JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-amd64
+        - SPARK_VERSION=2.4.5
+        - PANDAS_VERSION=1.0.1
+        - PYARROW_VERSION=0.14.1
+
+    # For testing, we temporarily test pandas<1.0.0 and pandas>=1.0.0 both for python3.6 and 3.7.
+    # TODO: So, we should remove below tests before PR #1197 is merged.
```
HyukjinKwon: @itholic can you file separate issue for each todo?

itholic: yes i will. thanks for the review :)

.github/workflows/master.yml
```diff
@@ -70,16 +70,18 @@ jobs:
...
-            logger: databricks.koalas.usage_logging.usage_logger
-          - python-version: 3.7
-            spark-version: 2.4.5
             pandas-version: 0.25.3
+          - python-version: 3.6
             pyarrow-version: 0.14.1
+          - python-version: 3.7
+            pandas-version: 0.24.2
+          - python-version: 3.7
+            pyarrow-version: 0.13.0
```
HyukjinKwon: @itholic, I think it's too many. Let's just match with Travis CI's cases for now.

databricks/koalas/generic.py
```diff
@@ -297,12 +297,11 @@ def cumprod(self, skipna: bool = True):
...
         return self._apply_series_op(lambda kser: kser._cumprod(skipna))  # type: ignore
 
+    # TODO: Although this has removed pandas >= 1.0.0, but we're keeping this as deprecated
     def get_dtype_counts(self):
         """
         Return counts of unique dtypes in this object.
 
-        .. deprecated:: 0.14.0
```
HyukjinKwon: We can still deprecate and convert this function to private function only for `DataFrame.info`.  Let's just deprecate and match it to pandas'

databricks/koalas/missing/frame.py
```diff
@@ -87,29 +87,32 @@ class _MissingPandasLikeDataFrame(object):
...
     to_stata = unsupported_function('to_stata')
     to_timestamp = unsupported_function('to_timestamp')
+    to_markdown = unsupported_function('to_markdown')
     truncate = unsupported_function('truncate')
     tshift = unsupported_function('tshift')
     tz_convert = unsupported_function('tz_convert')
     tz_localize = unsupported_function('tz_localize')
     unstack = unsupported_function('unstack')
 
     # Deprecated functions
```
HyukjinKwon: I don't think we should necessarily classify the APIs deprecated before/after 1.0.0. You could just merge both.

databricks/koalas/namespace.py
```diff
@@ -27,6 +27,7 @@
 import numpy as np
 import pandas as pd
```
HyukjinKwon: This space seems not needed per PEP8 import order.

.github/workflows/master.yml
itholic: ah, i just temporarily added them for checking this PR valid for all version of pandas. i'll restore them after test passed. thanks!! :)

databricks/koalas/window.py
HyukjinKwon: How is it different?

HyukjinKwon: We should match the behaviours to pandas 1.0.0 always, and make the tests working with pandas 0.2X.X.

databricks/koalas/tests/test_stats.py
```diff
@@ -81,8 +83,9 @@ def test_axis_on_dataframe(self):
             self.assert_eq(kdf.count(axis=1), pdf.count(axis=1))
             self.assert_eq(kdf.var(axis=1), pdf.var(axis=1))
             self.assert_eq(kdf.std(axis=1), pdf.std(axis=1))
-            self.assert_eq(kdf.max(axis=1), pdf.max(axis=1))
-            self.assert_eq(kdf.min(axis=1), pdf.min(axis=1))
+            # TODO: `max` & `min` with `axis=1` now raise error in pandas 1.0.0
```
HyukjinKwon: Can you fix Koalas API to throw an exception consistently?

databricks/koalas/missing/frame.py
itholic: maybe it makes easy to move to pandas>=1.0.0 with dropping all the lower versions of pandas in the future??

it's not just separated into based on before and after, but whether it has been removed in pandas>=1.0.0 or not.

databricks/koalas/tests/test_series_conversion.py
```diff
@@ -54,5 +54,6 @@ def test_to_latex(self):
         self.assert_eq(kser.to_latex(sparsify=False), pser.to_latex(sparsify=False))
         self.assert_eq(kser.to_latex(index_names=False), pser.to_latex(index_names=False))
         self.assert_eq(kser.to_latex(bold_rows=True), pser.to_latex(bold_rows=True))
-        self.assert_eq(kser.to_latex(encoding='ascii'), pser.to_latex(encoding='ascii'))
+        # Error in pandas - ValueError: buf is not a file name and encoding is specified.
+        # self.assert_eq(kser.to_latex(encoding='ascii'), pser.to_latex(encoding='ascii'))
```
HyukjinKwon: Can we consistently throw an exception in Koalas too?

databricks/koalas/tests/test_rolling.py
```diff
@@ -137,6 +135,7 @@ def test_groupby_rolling_sum(self):
         self._test_groupby_rolling_func("sum")
 
     def test_groupby_rolling_std(self):
+        # TODO: `std` now raise error in pandas 1.0.0
```
HyukjinKwon: Can we fix Koalas to throw an exception too?

databricks/koalas/missing/frame.py
HyukjinKwon: We should always keep this file consistent to 1.0.0. You can remove here the methods if that's removed in the latest pandas.

databricks/koalas/generic.py
```diff
@@ -297,6 +297,7 @@ def cumprod(self, skipna: bool = True):
         """
         return self._apply_series_op(lambda kser: kser._cumprod(skipna))  # type: ignore
 
+    # TODO: Although this has removed pandas >= 1.0.0, but we're keeping this as deprecated
```
HyukjinKwon: Can you remove this comment too?

HyukjinKwon:
Okay, there have been some discussions if we should match the current Koalas to panads' 1.0.0 behaviours (regardless of the actual pandas versions being used). I think we should match .. but okay we can do it separately. Let's clean up and I will merge.

itholic:
Now we're testing using both of `Travis CI` and `GitHub Actions`.

And each matched python-pandas version like the below.

- Travis CI
  - python3.5 with pandas 0.23.4
  - python3.6 with pandas 0.24.2
  - python3.7 with pandas 0.25.3
- GitHub Actions
  - python3.5 with pandas 0.24.2
  - python3.6 with pandas 0.25.3
  - python3.7 with pandas 1.0.1

databricks/koalas/missing/frame.py
itholic: hmm. i have removed them but restore again with inspiring by https://github.com/databricks/koalas/issues/1194#issuecomment-583063664

especially.

> removed elements
> * If we already implemented, I'd prefer to keep it and show the deprecation message.
> * If it's in the missing list, I'd prefer to keep it but move it to deprecation list.


okay, but now i think the both are somewhat make sense, so let's make discussion together with @ueshin in more detail at follow-up pr will be created.

databricks/koalas/tests/test_series_plot.py
```diff
@@ -73,9 +73,9 @@ def test_bar_plot(self):
         pdf = self.pdf1
         kdf = self.kdf1
 
-        ax1 = pdf['a'].plot("bar", colormap='Paired')
+        ax1 = pdf['a'].plot(kind="bar", colormap='Paired')
```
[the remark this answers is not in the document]
itholic: It has changed in pandas >=1.0.0 :joy_cat: 

Written at [What's new in 1.0.0](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.0.0.html)

> Series.plot() no longer accepts positional arguments, pass keyword arguments instead (GH30003)

databricks/koalas/window.py
itholic: @ueshin 

> What's the difference, btw?

- pandas < 1.0.0
```python
>>> s = pd.Series([2, 2, 3, 3, 3, 4, 4, 4, 4, 5, 5])
>>> s.groupby(s).expanding(3).count().sort_index()
2  0     1.0
   1     2.0
3  2     1.0
   3     2.0
   4     3.0
4  5     1.0
   6     2.0
   7     3.0
   8     4.0
5  9     1.0
   10    2.0
dtype: float64
```

- pandas >= 1.0.0
```python
>>> s.groupby(s).expanding(3).count().sort_index()
2  0     NaN
   1     NaN
3  2     NaN
   3     NaN
   4     3.0
4  5     NaN
   6     NaN
   7     3.0
   8     4.0
5  9     NaN
   10    NaN
dtype: float64
```

databricks/koalas/window.py
```diff
@@ -49,8 +52,20 @@ def _apply_as_series_or_frame(self, func):
             "to handle the index and columns of output.")
 
     def count(self):
+        # Rolling and Expanding have different behavior for `count` in pandas 1.0.0
+        # for now, we consider both pandas < 1.0.0 and pandas >= 1.0.0
         def count(scol):
             return F.count(scol).over(self._window)
+
+        if LooseVersion(pd.__version__) >= LooseVersion('1.0.0'):
+            if isinstance(self, (Expanding, ExpandingGroupby)):
```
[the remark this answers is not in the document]
itholic: okay i'll fix them in the follow-ups right after thins PR is cleared. :)

Thanks for the comment!

databricks/koalas/tests/test_series_plot.py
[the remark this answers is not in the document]
ueshin: Got it, thanks!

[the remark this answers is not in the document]
itholic: No problem :)

databricks/koalas/window.py
itholic: @HyukjinKwon @ueshin 

I think now it is time to make the purpose of "Support pandas 1.0.0" clear.

Which of the following is exactly our purpose?

1. Make the Koalas behavior match to `pandas 1.0.0` while fixing & passing existing tests
2. Make the Koalas behavior match to `pandas 1.0.0` while keeping & skipping existing tests
3. Make the Koalas behavior match to `installed pandas` while selectively keeping & skipping depends on version of pandas.

ueshin: My opinion is 3), based on what I commented at https://github.com/databricks/koalas/issues/1194#issuecomment-583063664.
We need more feedback and let me ask some people about it.

Btw, the behavior of `s.groupby(s).expanding(3).count()` with `pandas<1.0` above seems like a bug of pandas' .. ? (https://github.com/pandas-dev/pandas/pull/30923)
If so, we can just follow the new behavior?

HyukjinKwon: Hm, i thought 1. might be a better way because pandas is just a dependency in Koalas after all, and we match our behaviours to pandas, assuming the latest pandas has a corrector behaviours..

ueshin: Let me correct a little, my opinion is still basically 3), but if the behavior changes have reasonable reasons.

As for this case, the behavior with `pandas<1.0` seems like a bug, so I think we can just fix it.

Then, are there any other behavior differences?
If no, we can fix it and merge this.

If there are still behavior differences and the differences have reasonable reasons, we should discuss one-by-one.

itholic: @HyukjinKwon @ueshin 

Thanks for the comments :)

> Then, are there any other behavior differences?

`DataFrame.info` has changed as we have discussed at https://github.com/databricks/koalas/pull/1197/files#r373335746.

I think it's a good example to talk about this.
(they just changed the behavior of printing-out, not for fixing bug)

here is an example.

- pandas < 1.0.0
```python
>>> df.info()
<class 'pandas.core.frame.DataFrame'>
Int64Index: 4 entries, 0 to 3
Data columns (total 4 columns):
A    4 non-null int64
B    3 non-null float64
C    1 non-null float64
D    4 non-null int64
dtypes: float64(2), int64(2)
```

- pandas >= 1.0.0
```
>>> df.info()
<class 'pandas.core.frame.DataFrame'>
Int64Index: 4 entries, 0 to 3
Data columns (total 4 columns):
 #   Column  Non-Null Count  Dtype
---  ------  --------------  -----
 0   A       4 non-null      int64
 1   B       3 non-null      float64
 2   C       1 non-null      float64
 3   D       4 non-null      int64
dtypes: float64(2), int64(2)
```

itholic: And i also thought we're following 1, since if we should follow the behavior of the version of installed pandas, why just do this for < 1.0.0 and >= 1.0.0 , not for every single version of pandas?

ueshin: I meant the major version. It might contain breaking changes.

itholic: @ueshin 

> I meant the major version. It might contain breaking changes.

Ah, got it.

Then, just let discussion about `DataFrame.info()` for now, should we check the installed major pandas version and modify it to work differently depend on that version?? 🧐

HyukjinKwon:
@itholic, I had a offline discussion with @ueshin. His concern is basically user's expectation which is a good user-oriented point; however, just to reduce the overhead at the current moment, we delayed how to handle this later.

Here's the conclusion we reached at this moment:

- Match to the Koalas behaviours to the latest of pandas.
- Do not match to the Koalas behaviours to the latest pandas when it is practically difficult, when Koalas can reasonably reuse pandas' APIs, or  whenever it is more reasonable.
- Do not match to the Koalas behaviours to the latest pandas when the new behaviours look obviously a bug.

itholic:
@HyukjinKwon @ueshin 

Thanks for the conclusion :D

Okay, so I'll check again whole changes one last time based on that concept.

HyukjinKwon:
Sorry, @itholic, I removed my last comments for nits but it removed yours together. Ignore my last comments.

itholic:
@HyukjinKwon ah, it's okay. 👍

.github/workflows/master.yml
```diff
@@ -138,3 +138,4 @@ jobs:
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
 
+
```
HyukjinKwon: Shall we remove this new line?

databricks/koalas/generic.py
```diff
@@ -297,6 +297,7 @@ def cumprod(self, skipna: bool = True):
...
     def get_dtype_counts(self):
```
HyukjinKwon: @itholic, you can simply rename `get_dtype_counts` -> `_get_dtype_counts`

databricks/koalas/generic.py
HyukjinKwon: This comment doesn't say why we keep it.

databricks/koalas/tests/test_dataframe_conversion.py
```diff
@@ -181,7 +181,9 @@ def test_to_latex(self):
         self.assert_eq(kdf.to_latex(sparsify=False), pdf.to_latex(sparsify=False))
         self.assert_eq(kdf.to_latex(index_names=False), pdf.to_latex(index_names=False))
         self.assert_eq(kdf.to_latex(bold_rows=True), pdf.to_latex(bold_rows=True))
-        self.assert_eq(kdf.to_latex(encoding='ascii'), pdf.to_latex(encoding='ascii'))
+        # TODO: error in pandas >= 1.0.0:
+        # `ValueError: buf is not a file name and encoding is specified.`
+        # self.assert_eq(kdf.to_latex(encoding='ascii'), pdf.to_latex(encoding='ascii'))
```
HyukjinKwon: Can we fix this case to match with pandas 1.0.0?

HyukjinKwon: If pandas 1.0.0 behaviour looks like a bug, we don't need to add `TODO` here.

databricks/koalas/tests/test_groupby.py
```diff
@@ -748,9 +747,11 @@ def test_shift(self):
...
         self.assert_eq(kdf.groupby(['a', 'b'])['c'].shift().sort_index(),
                        pdf.groupby(['a', 'b'])['c'].shift().sort_index(), almost=True)
-        self.assert_eq(kdf.groupby(['b'])[['a', 'c']].shift(periods=-1, fill_value=0).sort_index(),
-                       pdf.groupby(['b'])[['a', 'c']].shift(periods=-1, fill_value=0).sort_index(),
-                       almost=True)
+        # TODO: seems like a pandas' bug when fill_value is not None?
+        # self.assert_eq(
+        #     kdf.groupby(['b'])[['a', 'c']].shift(periods=-1, fill_value=0).sort_index(),
+        #     pdf.groupby(['b'])[['a', 'c']].shift(periods=-1, fill_value=0).sort_index(),
+        #     almost=True)
```
HyukjinKwon: This one, same. We can just test when pandas version is not 1.0.0. Can you describe why it looks like a bug in pandas?

databricks/koalas/tests/test_indexes.py
```diff
@@ -777,39 +777,40 @@ def test_monotonic(self):
         datas.append([(-5, 'e'), (-4, 'c'), (-3, 'b'), (-2, 'd'), (-1, 'a')])
 
         # None type tests (None type is treated as the largets value)
-        datas.append([(None, 100), (2, 200), (3, 300), (4, 400), (5, 500)])
+        # TODO: the commented tests below should be uncommented after fixing for pandas >= 1.0.0
```
HyukjinKwon: Is it difficult to patch to pandas' 1.0.0 behaviours?

databricks/koalas/window.py
```diff
@@ -49,8 +52,20 @@ def _apply_as_series_or_frame(self, func):
             "to handle the index and columns of output.")
 
     def count(self):
+        # Rolling and Expanding have different behavior for `count` in pandas 1.0.0
+        # for now, we consider both pandas < 1.0.0 and pandas >= 1.0.0
```
HyukjinKwon: So it was decided to match to pandas' 1.0.0 right?

HyukjinKwon:
Okay, let's merge this and address my comments in a followup @itholic. Can you just resolve conflicts?

databricks/koalas/tests/test_dataframe_conversion.py
itholic: yeah, this rule has changed from pandas 1.0.0 like the below.
(refered from https://pandas.pydata.org/docs/whatsnew/v1.0.0.html)

> Added encoding argument to DataFrame.to_string() for non-ascii text (GH28766)

we can just change to do this will be tested only when pandas>=1.0.0

databricks/koalas/window.py
itholic: correct, now we can remove this line.

Thanks !

databricks/koalas/tests/test_groupby.py
itholic: - 0.24.0 <= pandas < 1.0.0
```python
>>> pdf.groupby(['a', 'b']).shift(periods=-1, fill_value=0).sort_index()
          a  b   c
0.061934  2  2   9
0.348846  0  0   0
0.417891  2  2  16
0.442334  3  4  36
0.453143  3  3  25
0.523323  2  2  16
0.523910  2  2  16
0.574407  0  0   0
0.587845  0  0   0
0.639976  3  4  36
0.670847  1  1   1
0.675415  1  1   4
0.798390  0  0   0
0.832881  2  2   9
0.853427  1  1   4
0.856181  1  1   4
0.904274  3  3  25
0.997218  1  1   1
```

- pandas >= 1.0.0
```python
>>> pdf.groupby(['a', 'b']).shift(periods=-1, fill_value=0).sort_index()
...
KeyError: 0.0
```

itholic: > We can just test when pandas version is not 1.0.0

thanks! i fixed with following this way

databricks/koalas/tests/test_indexes.py
itholic: yeah its changing looks a bit complicated, let me try to fix this follow-up PR

itholic:
> Okay, let's merge this and address my comments in a followup @itholic. Can you just resolve conflicts?

yeah, i've resolved nit comments and will address rest ones at follow-up.

thanks for the comments !

databricks/koalas/window.py
```diff
@@ -51,6 +54,16 @@ def _apply_as_series_or_frame(self, func):
     def count(self):
         def count(scol):
             return F.count(scol).over(self._window)
+
+        if LooseVersion(pd.__version__) >= LooseVersion('1.0.0'):
```
HyukjinKwon: @itholic, given our discussion, we should match it to the latest behaviour. I think we don't have to check the pandas version.

HyukjinKwon:
Okay, I am going to merge this to make it rolling. @itholic, can you do all todos in the next followup?

itholic:
@HyukjinKwon yup. i'm working on that.

PR #1198, opened 2020-01-16
Fix `DataFrame.idxmin/idxmax`.

ueshin:


ueshin:
Yeah, it is. I'll do it next time.

PR #1199, opened 2020-01-16
Add utility methods to `_InternalFrame`.

ueshin:
Added `with_new_columns` and `with_filter` utilities to `_InternalFrame` and changed the usage in `DataFrame`. We can replace with the methods in other places but I'd leave it to the following PRs.


ueshin:
FYI: #1198 should be reviewed and merged prior to this one.

ueshin:
> ~We can replace with the methods in other places but I'd leave it to the following PRs.~

I applied to the other places where we can simply apply.

Issue #1200, opened 2020-01-17
Using nested dicts leaks pyspark.sql.Row abstraction


chunyang:
Please consider the example below. Starting with an initial pandas DataFrame, a round trip through Spark DataFrame will preserve the nested dict fields, but a round trip through koalas DataFrame will expose `pyspark.sql.Row` objects.

```python
import databricks.koalas as ks
import pandas as pd
import pyspark

pdf = pd.DataFrame({"A": [{"B": "C"}, {"D": "C"}]})

# Spark DataFrame preserves nested dict structure
spark = pyspark.sql.SparkSession.builder.master("local[*,1]").getOrCreate()
sdf = spark.createDataFrame(pdf)
sdf.show()
# +--------+
# |       A|
# +--------+
# |[B -> C]|
# |[D -> C]|
# +--------+
sdf.toPandas()
#             A
# 0  {'B': 'C'}
# 1  {'D': 'C'}
sdf.toPandas().iloc[0, 0]
# {'B': 'C'}
sdf
# DataFrame[A: map<string,string>]

# Koalas DataFrame converts nested dict structure to nested pyspark.sql.Row
kdf = ks.DataFrame(pdf)
kdf
#            A                                                                    
# 0  (C, None)
# 1  (None, C)
kdf.to_pandas()
#            A
# 0  (C, None)
# 1  (None, C)
kdf.to_pandas().iloc[0, 0]
# Row(B='C', D=None)
kdf._internal.spark_df
# DataFrame[A: struct<B:string,D:string>]
```

This seems to be due to different code paths for schema inference in Spark vs Koalas?
Spark: https://github.com/apache/spark/blob/v3.0.0-preview2/python/pyspark/sql/session.py#L787
Koalas: https://github.com/databricks/koalas/blob/v0.25.0/databricks/koalas/typedef.py#L170

koalas version: 0.25.0
pandas version: 0.25.3
pyspark version: 3.0.0.dev0

ueshin:
Firstly, Koalas doesn't support nested types yet, and I'm not sure it will or not. If you want to handle nested types, you should just use PySpark.

Then ..

Yeah, actually the type inference is different from Spark's and we rely on Arrow so far, and Arrow doesn't support map type inference or conversion yet. I'm not sure whether it's correct to infer this case as map type, though.

As for returning `Row` object when `to_pandas()`, we should enable arrow execution within Koalas. I'll submit a PR.

itholic:
[...]

PR #1201, opened 2020-01-17
Force to enable Arrow execution mode.

ueshin:
This is part of #1200.

Currently we don't care about whether arrow execution mode is enabled or not, but it might cause the behavior difference between enabled or not. We should always use arrow execution mode.

### DIFF
--- a/databricks/koalas/utils.py
+++ b/databricks/koalas/utils.py
@@ -19,6 +19,7 @@ Commonly used utils in Koalas.
 
 import functools
 from collections import OrderedDict
+from contextlib import contextmanager
 from distutils.version import LooseVersion
 from typing import Callable, Dict, List, Tuple, Union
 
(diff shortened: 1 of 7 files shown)


ueshin:
Some tests will fail with Spark 2.3 environment.

databricks/koalas/utils.py
```diff
@@ -288,6 +289,32 @@ def default_session(conf=None):
     return builder.getOrCreate()
```
HyukjinKwon: @ueshin, just out of curiosity, why don't we just enable it here?

HyukjinKwon: I mean, 

```python
session.conf.set(session.conf.get('spark.sql.execution.arrow.enabled', True))
```

So, if users disable it explicitly, just disable. Otherwise, enable it by default.

ueshin: That works too, but I didn't want to make unexpected behavior changes out of Koalas. Also we can focus on the Arrow enabled behavior. WDYT?

HyukjinKwon:
Offline discussed with @ueshin.

I am not sure yet if we should fix it by enabling it. There is still an issue as pointed out - we don't know if a dictionary -> a struct is correct or not. This might be something Spark <> Arrow might have to fix.

ueshin:
ok, I'd close this for now.

PR #1205, opened 2020-01-21
Add property is_all_dates for Index/MultiIndex

itholic:
https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.is_all_dates.html#pandas.Index.is_all_dates

```python
>>> idx = ks.Index([datetime(2019, 1, 1, 0, 0, 0), datetime(2019, 2, 3, 0, 0, 0)])
>>> idx
DatetimeIndex(['2019-01-01', '2019-02-03'], dtype='datetime64[ns]', freq=None)

>>> idx.is_all_dates
True

>>> idx = ks.Index([datetime(2019, 1, 1, 0, 0, 0), None])
>>> idx
DatetimeIndex(['2019-01-01', 'NaT'], dtype='datetime64[ns]', freq=None)

>>> idx.is_all_dates
True

>>> idx = ks.Index([0, 1, 2])
>>> idx
Int64Index([0, 1, 2], dtype='int64')
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1225,6 +1225,40 @@ class Index(IndexOpsMixin):
                 names = self_names
         return self.rename(name=names, inplace=inplace)
 
+    @property
+    def is_all_dates(self):
+        """
+        Return if all data types of the index are datetime.
+        remember that since Koalas does not support multiple data types in an index,
+        so it returns True if any type of data is datetime.
+
+        Examples
+        --------
+        >>> from datetime import datetime
+
+        >>> idx = ks.Index([datetime(2019, 1, 1, 0, 0, 0), datetime(2019, 2, 3, 0, 0, 0)])
+        >>> idx
+        DatetimeIndex(['2019-01-01', '2019-02-03'], dtype='datetime64[ns]', freq=None)
+
+        >>> idx.is_all_dates
+        True
+
+        >>> idx = ks.Index([datetime(2019, 1, 1, 0, 0, 0), None])
+        >>> idx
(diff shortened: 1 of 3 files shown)


itholic:
Since pandas doesn't have a proper description for this property,

I guessed its purpose based on its name and several test cases.

databricks/koalas/indexes.py
```diff
@@ -1225,6 +1225,40 @@ def set_names(self, names, level=None, inplace=False):
...
+        True
+
+        >>> idx = ks.Index([0, 1, 2])
+        >>> idx
+        Int64Index([0, 1, 2], dtype='int64')
+
+        >>> idx.is_all_dates
+        False
+        """
+        return True if repr(self).startswith("DatetimeIndex") else False
```
HyukjinKwon: @itholic, this will triggers to collect all data when `repr` is called. I think you can just check `self. spark_type()` is a `DateType` or not

itholic: @HyukjinKwon thanks i just fixed :)

databricks/koalas/indexes.py
```diff
@@ -1225,6 +1225,40 @@ def set_names(self, names, level=None, inplace=False):
...
+        True
+
+        >>> idx = ks.Index([0, 1, 2])
+        >>> idx
+        Int64Index([0, 1, 2], dtype='int64')
+
+        >>> idx.is_all_dates
+        False
+        """
+        return True if isinstance(self.spark_type, TimestampType) else False
```
HyukjinKwon: You can just return `isinstance(self.spark_type, TimestampType)` right away I think.

itholic: thanks! fixed :)

PR #1202, opened 2020-01-18
Create docstring for ``DataFrameGroupBy.describe`` and reimplement leveraging `_InternalFrame`

deepyaman:
Close #1166

PR #1217, opened 2020-01-23
Add more test cases and improve error message in DataFrameGroupBy.describe

itholic:
following up #1202 

- add more test cases
- raise proper message for string type


itholic:
after this release, i'm going to fix this for supporting string type also

PR #1220, opened 2020-01-23
Fix identical and equals for the comparison between the same object.

ueshin:
`identical` and `equals` should be `True` for the comparison between the same object.

```py
>>> import pandas as pd
>>> idx = pd.Index(['a', 'b', 'c'])
>>> midx = pd.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
>>> midx.names = ("nameA", "nameB")

>>> idx.identical(idx)
True
>>> midx.identical(midx)
True

>>> idx.equals(idx)
True
>>> midx.equals(midx)
True
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -199,7 +199,7 @@ class Index(IndexOpsMixin):
         For Index
 
         >>> idx.identical(idx)
-        False
+        True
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Oops, thanks for fixing this. @ueshin.

PR #1221, opened 2020-01-23
Make attach_default_index generate the index column name.

ueshin:
The usage of `_InternalFrame.attach_default_index` is spreading out of `_InternalFrame`, and we might not be able to control the index column name.
We should make the function generate the index column name instead of the fixed name.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -454,14 +454,13 @@ class _InternalFrame(object):
         assert isinstance(sdf, spark.DataFrame)
 
         if index_map is None:
-            # Here is when Koalas DataFrame is created directly from Spark DataFrame.
-            assert not any(SPARK_INDEX_NAME_PATTERN.match(name) for name in sdf.schema.names), \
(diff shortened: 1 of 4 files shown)


databricks/koalas/internal.py
```diff
@@ -454,14 +454,13 @@ def __init__(self, sdf: spark.DataFrame,
         assert isinstance(sdf, spark.DataFrame)
 
         if index_map is None:
-            # Here is when Koalas DataFrame is created directly from Spark DataFrame.
-            assert not any(SPARK_INDEX_NAME_PATTERN.match(name) for name in sdf.schema.names), \
+            assert not any(SPARK_INDEX_NAME_PATTERN.match(name) for name in sdf.columns), \
                 "Index columns should not appear in columns of the Spark DataFrame. Avoid " \
-                "index colum names [%s]." % SPARK_INDEX_NAME_PATTERN
+                "index column names [%s]." % SPARK_INDEX_NAME_PATTERN
```
ueshin: I don't think we still need this check, but I'd leave it as an assertion for now.

PR #1223, opened 2020-01-24
Add attach_distributed_sequence_column with SPARK_DEFAULT_INDEX_NAME in InternalFrame

HyukjinKwon:
This PR adds:

- `attach_distributed_sequence_column` for the case when we need a sequence in an API.
- Use `SPARK_DEFAULT_INDEX_NAME` instead of `SPARK_INDEX_NAME_FORMAT(0)` for default column name.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -48,7 +48,7 @@ from databricks import koalas as ks  # For running doctests and reference resolu
 from databricks.koalas.utils import validate_arguments_and_invoke_function, align_diff_frames
 from databricks.koalas.generic import _Frame
 from databricks.koalas.internal import (_InternalFrame, HIDDEN_COLUMNS, NATURAL_ORDER_COLUMN_NAME,
-                                        SPARK_INDEX_NAME_FORMAT)
+                                        SPARK_INDEX_NAME_FORMAT, SPARK_DEFAULT_INDEX_NAME)
 from databricks.koalas.missing.frame import _MissingPandasLikeDataFrame
 from databricks.koalas.ml import corr
 from databricks.koalas.utils import column_index_level, name_like_string, scol_for, validate_axis
@@ -2696,8 +2696,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             # Now, new internal Spark columns are named as same as index name.
             new_index_map = [(column, name) for column, name in new_index_map]
 
-            sdf, index_column = _InternalFrame.attach_default_index(sdf)
-            index_map = [(index_column, None)]
+            sdf = _InternalFrame.attach_default_index(sdf)
+            index_map = [(SPARK_DEFAULT_INDEX_NAME, None)]
 
         if drop:
             new_index_map = []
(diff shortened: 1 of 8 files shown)


databricks/koalas/internal.py
```diff
@@ -516,68 +517,107 @@ def attach_default_index(sdf, default_index_type=None):
...
+        +--------+---+
+        |       0|  a|
+        |       1|  b|
+        |       2|  c|
+        +--------+---+
+        """
+
+        scols = [scol_for(sdf, column) for column in sdf.columns]
+
+        # 1. Calculates counts per each partition ID. `counts` here is, for instance,
```
HyukjinKwon: It just moved the function as is

databricks/koalas/frame.py
```diff
@@ -2696,8 +2696,8 @@ def rename(index):
             # Now, new internal Spark columns are named as same as index name.
             new_index_map = [(column, name) for column, name in new_index_map]
 
-            sdf, index_column = _InternalFrame.attach_default_index(sdf)
-            index_map = [(index_column, None)]
+            sdf = _InternalFrame.attach_default_index(sdf)
```
ueshin: Actually I'd prefer to let the function generate the column name.

databricks/koalas/indexes.py
```diff
@@ -1218,11 +1218,9 @@ def argmax(self):
...
         4
         """
-        sdf = self._internal.sdf
-        sdf = sdf.select(self._scol.alias('__index_value__'))
-
-        sdf, index_column = _InternalFrame.attach_default_index(
-            sdf, default_index_type='distributed-sequence')
+        sdf = self._internal.sdf.select(self._scol)
+        sequence_col = "__distributed_sequence_column__"
+        sdf = _InternalFrame.attach_distributed_sequence_column(sdf, column_name=sequence_col)
```

HyukjinKwon: Here's actually it adds a column that has a sequence with a temp column. Maybe it's better to have a method that generates a name automatically but that would make the debug harder with an arbitrary name.

Using index column `__index_level_0__` as a temp column wouldn't be a good idea. If we want to generate the names for temp columns in general, we might have to do it separately.

databricks/koalas/frame.py
HyukjinKwon: I think it's better to have the same default index name, rather than a name that can be potentially different. If we worry about the name `__index_level_0__` being conflicted, seems fine here:

```python
>>> ks.range(10).reset_index().explain(True)
```

```
...
== Physical Plan ==
*(4) Project [(_we0#21 - 1) AS __index_level_0__#19, index#14, id#0L]
+- Window [row_number() windowspecdefinition(_w0#20L ASC NULLS FIRST, specifiedwindowframe(RowFrame, unboundedpreceding$(), currentrow$())) AS _we0#21], [_w0#20L ASC NULLS FIRST]
   +- *(3) Sort [_w0#20L ASC NULLS FIRST], false, 0
      +- *(3) Project [(_we0#5 - 1) AS index#14, id#0L, monotonically_increasing_id() AS _w0#20L]
         +- Window [row_number() windowspecdefinition(_w0#4L ASC NULLS FIRST, specifiedwindowframe(RowFrame, unboundedpreceding$(), currentrow$())) AS _we0#5], [_w0#4L ASC NULLS FIRST]
            +- *(2) Sort [_w0#4L ASC NULLS FIRST], false, 0
               +- Exchange SinglePartition, true, [id=#32]
                  +- *(1) Project [id#0L, monotonically_increasing_id() AS _w0#4L]
                     +- *(1) Range (0, 10, step=1, splits=12)
```

HyukjinKwon:
Discussed offline with @ueshin. Merged for now.

PR #1224, opened 2020-01-24
Select the series correctly in SeriesGroupBy APIs

HyukjinKwon:
Can be tested with the codes below:

```python
import numpy as np
import pandas as pd

kdf = ks.DataFrame({'cust_id':['a', 'a', 'a', 'b', 'b'],
                   'sales': [100, 200, 300, 400, 500],
                   'days':[12.1,13.1,14.1,78.1,87.2]})

kdf.groupby('cust_id')['days'].apply(lambda x: x.ewm(alpha=0.5, adjust=False).mean())


pdf = kdf.to_pandas()
pdf.groupby('cust_id')['days'].apply(lambda x: x.ewm(alpha=0.5, adjust=False).mean())
```

```
0    12.10
1    12.60
2    13.35
3    78.10
4    82.65
Name: days, dtype: float64
```

Basic idea is that, it creates a DataFrame from Series to reuse existing implementations, and resolves the columns by name.

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -2053,9 +2053,9 @@ class SeriesGroupBy(GroupBy):
 
     @property
     def _kdf(self) -> DataFrame:
-        # TODO: if names from _kser and _groupkeys are name, grouping key is just ignored.
-        #    it can be a problem when both series have the same name but different operations.
-        series = [self._kser] + [s for s in self._groupkeys if s.name != self._kser.name]
+        # TODO: Currently cannot handle the case when the values in current series
+        #  and groupkeys series are different but only their names are same.
+        series = [self._kser] + [s for s in self._groupkeys if not s._equals(self._kser)]
         return DataFrame(self._kser._kdf._internal.with_new_columns(series))
 
     @property
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -44,7 +44,7 @@ from databricks.koalas.missing.groupby import _MissingPandasLikeDataFrameGroupBy
     _MissingPandasLikeSeriesGroupBy
 from databricks.koalas.series import Series, _col
 from databricks.koalas.config import get_option
-from databricks.koalas.utils import column_index_level, scol_for
+from databricks.koalas.utils import column_index_level, scol_for, name_like_string
 from databricks.koalas.window import RollingGroupby, ExpandingGroupby
 
 # to keep it the same as pandas
@@ -182,8 +182,10 @@ class GroupBy(object):
         else:
             agg_cols = [col.name for col in self._agg_columns]
             func_or_funcs = OrderedDict([(col, func_or_funcs) for col in agg_cols])
-
-        kdf = DataFrame(GroupBy._spark_groupby(self._kdf, func_or_funcs, self._groupkeys))
+        index_map = [(SPARK_INDEX_NAME_FORMAT(i), s._internal.column_index[0])
+                     for i, s in enumerate(self._groupkeys)]
+        kdf = DataFrame(GroupBy._spark_groupby(
+            self._kdf, func_or_funcs, self._groupkeys_scols, index_map))
         if not self._as_index:
             kdf = kdf.reset_index()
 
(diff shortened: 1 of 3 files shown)


databricks/koalas/groupby.py
```diff
@@ -2031,28 +2036,26 @@ def __getattr__(self, item: str) -> Any:
...
+        return Series._fillna(self._kser, *args, **kwargs, part_cols=self._groupkeys_scols)
 
     def _shift(self, periods, fill_value):
-        groupkey_scols = [s._scol for s in self._groupkeys]
-        return Series._shift(self._kser, periods, fill_value, part_cols=groupkey_scols)
+        return Series._shift(self._kser, periods, fill_value, part_cols=self._groupkeys_scols)
 
     @property
     def _kdf(self) -> DataFrame:
-        return self._kser._kdf
```
HyukjinKwon: The fix is here.

databricks/koalas/groupby.py
```diff
@@ -2016,6 +2012,15 @@ class SeriesGroupBy(GroupBy):
     def __init__(self, kser: Series, by: List[Series], as_index: bool = True):
         self._kser = kser
         self._groupkeys = by
```
HyukjinKwon: and here.

databricks/koalas/groupby.py
```diff
@@ -2016,6 +2013,15 @@ class SeriesGroupBy(GroupBy):
...
+        # TODO: this class resolves the groupkeys and agg_columns always by columns names
+        #   e.g., F.col("..."). This is because of the limitation of `SeriesGroupBy`
+        #   implementation, which reuses the implementation in `GroupBy`.
+        #   `SeriesGroupBy` creates another DataFrame and
+        #   internal IDs of the columns become different. Maybe we should refactor the whole
+        #   class in the future.
+        self._groupkeys_scols = [F.col(name_like_string(s.name)) for s in self._groupkeys]
+        self._agg_columns_scols = [F.col(name_like_string(s.name)) for s in self._agg_columns]
```
ueshin: `F.col(s._internal.data_columns[0])` instead of `F.col(name_like_string(s.name))`?

databricks/koalas/groupby.py
```diff
@@ -2031,28 +2037,26 @@ def __getattr__(self, item: str) -> Any:
...
 
     def _shift(self, periods, fill_value):
-        groupkey_scols = [s._scol for s in self._groupkeys]
-        return Series._shift(self._kser, periods, fill_value, part_cols=groupkey_scols)
+        return Series._shift(self._kser, periods, fill_value, part_cols=self._groupkeys_scols)
 
     @property
     def _kdf(self) -> DataFrame:
-        return self._kser._kdf
+        # TODO: if names from _kser and _groupkeys are name, grouping key is just ignored.
```
ueshin: nit: `_groupkeys are same`?

databricks/koalas/groupby.py
```diff
@@ -2031,28 +2037,26 @@ def __getattr__(self, item: str) -> Any:
...
+        #    it can be a problem when both series have the same name but different operations.
+        series = [self._kser] + [s for s in self._groupkeys if s.name != self._kser.name]
```
ueshin: How about using `s._equals(self._kser)` which compares the `._scol._jc`s?

HyukjinKwon: Ah, thank you. Nice.

databricks/koalas/groupby.py
HyukjinKwon: I switched but realised that some tests fail with that change. Seems like there's some cases when internal column names and series names are different:

```python
>>> (ks.range(10).id + 1)._internal.data_columns
['(id + 1)']
>>> (ks.range(10).id + 1).name
'id'
```

Seems we use:

```python
    @property
    def _kdf(self) -> DataFrame:
        series = [self._kser] + [s for s in self._groupkeys if not s._equals(self._kser)]
        return DataFrame(self._kser._kdf._internal.with_new_columns(series))
```

which always aliases via using `column_index`. So, it seems guaranteed to have the internal Spark column names same as `s.name`. I think using `name` here for this workaround is correct.

databricks/koalas/groupby.py
HyukjinKwon: In that case, seems it fails because the same names exist in the same DataFrame `AnalysisException: Reference 'b' is ambiguous, could be: b, b.;` ... but yeah I think it's better to fail fast for now rather than returning a wrong result.

databricks/koalas/tests/test_groupby.py
```diff
@@ -734,8 +734,8 @@ def test_apply(self):
                        pdf.groupby("b").apply(lambda x: x + 1).sort_index())
         self.assert_eq(kdf.groupby(['a', 'b']).apply(lambda x: x * x).sort_index(),
                        pdf.groupby(['a', 'b']).apply(lambda x: x * x).sort_index())
-        self.assert_eq(kdf.groupby(['b'])['a'].apply(lambda x: x).sort_index(),
-                       pdf.groupby(['b'])['a'].apply(lambda x: x).sort_index())
+        self.assert_eq(kdf.groupby(['b'])['b'].apply(lambda x: x).sort_index(),
+                       pdf.groupby(['b'])['b'].apply(lambda x: x).sort_index())
```
HyukjinKwon: So, we will not support this case for now, which I think it's fine (?).

ueshin:
Thanks! merging.

sushmit86:
Is this issue fixed?

HyukjinKwon:
Yup, this will be available for the next week's release.

PR #1225, opened 2020-01-24
Move iat implementation to iloc.

ueshin:
This moves `iat `implementation to `iloc` and adds the infrastructure to support other types of access via `iloc`.


ueshin:
I'd merge this for now. Please feel free to comment on this if any.

PR #1229, opened 2020-01-27
Fix `_InternalFrame.with_new_columns` not to rename columns.

ueshin:
We should not have renamed the columns in `_InternalFrame.with_new_columns`, otherwise the new column names could be unexpectable.

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -2018,8 +2018,8 @@ class SeriesGroupBy(GroupBy):
         #   `SeriesGroupBy` creates another DataFrame and
         #   internal IDs of the columns become different. Maybe we should refactor the whole
         #   class in the future.
-        self._groupkeys_scols = [F.col(name_like_string(s.name)) for s in self._groupkeys]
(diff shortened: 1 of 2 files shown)


databricks/koalas/groupby.py
```diff
@@ -2018,8 +2018,8 @@ def __init__(self, kser: Series, by: List[Series], as_index: bool = True):
         #   `SeriesGroupBy` creates another DataFrame and
         #   internal IDs of the columns become different. Maybe we should refactor the whole
         #   class in the future.
-        self._groupkeys_scols = [F.col(name_like_string(s.name)) for s in self._groupkeys]
-        self._agg_columns_scols = [F.col(name_like_string(s.name)) for s in self._agg_columns]
+        self._groupkeys_scols = [F.col(s._internal.data_columns[0]) for s in self._groupkeys]
+        self._agg_columns_scols = [F.col(s._internal.data_columns[0]) for s in self._agg_columns]
```
HyukjinKwon: I think now we can enable the tests back (https://github.com/databricks/koalas/pull/1224#discussion_r370972255). I will make a small followup.

Issue #1219, opened 2020-01-23
Koalas does not support http/https scheme for ks.read_csv(URL)


dmatrix:
It seems like Koalas does not support full URL schemes as pandas does for reading files. For example, this code snippet will throw an exception:

import databricks.koalas as ks
import pandas as pd

ks_df = ks.read_csv("http://bit.ly/uforeports")
...
java.io.IOException: No FileSystem for scheme: http

The workaround is to read in as pandas and convert it as follows:
ufo = pd.read_csv("http://bit.ly/uforeports")
ks_df = ks_ufo = ks.from_pandas(ufo)

But doing so does not preserve the ordering of the index on axis=0.

that is doing ufo.head(2) and ks_df.head(2) return different values.
ufo.head(2) returns the top two rows, while ks_df.head(2) returns the last two values, as if did a tail.

dmatrix:
Here is the output of a different row order:

<img width="856" alt="Screen Shot 2020-01-24 at 8 54 17 AM" src="https://user-images.githubusercontent.com/1117597/73087444-5cffda80-3e87-11ea-9ef3-4f04b4947659.png">

patryk-oleniuk:
I am also missing that behavior of koalas

ueshin:
Here's two things:

- read files over `http` protocol
  - so far, Koalas relies on Spark DataFrameReader which doesn't support `http` protocol. We need some workaround to support it.
- `head` doesn't show the expected head records.
  - Koalas doesn't guarantee the row order. If we wan to make it, we need extra huge overhead.

ueshin:
As for the `head` behavior, we added a config to perform natural ordering when head (#1231).
If you turn `compute.ordered_head` to `True`, you will see the same result as pandas'. Please be careful that it will cause a performance overhead.

Issue #1064, opened 2019-11-22
Does spark DataFrame.limit guarantees any order? 


guyao:
This is just a question, not an issue.  
Out of curiosity, does spark.DataFrame.limit guarantees the order ? 

That method is used to implement many methods like `.head`, `.iloc`  
In pandas the data is small and the order could be guaranteed. But in koalas, the data is distributed. Is it guaranteed those methods would keep order?

HyukjinKwon:
Yeah, this is a pretty important issue in Koalas.

Currently, it does not guarantees the natural order in general. Natural order can be kept when it only performs map operation (without shuffle) but if shuffle happens, the order cannot be kept.

So, if users want the order, they should sort by, for example, index.

HyukjinKwon:
We should still think about how we'll handle this in general.

guyao:
Thank you! 

Follow up of my last question, do you mean `spark DataFrame.limit(n)` will return `first n` result ( in natural order, if no shuffle happens) 


For example, if read a file as DataFrame, the file would be split into partitions P1, P2 ... , where P1 contains first part of file, ...  If I tried to call df.limit(n), it will return first n of P1 (if n < # P1)  

Is that correct? (sorry it more like a spark question but I asked here)

HyukjinKwon:
Yes, that's correct :-).

guyao:
I was thinking about what pandas users expected when using `iloc[slice(start, end, step)]` 
Do people really know which start index, end index he need when he is using an uncountable dataframe. 

( but I think user still expected order if the DF is small )

HyukjinKwon:
Yes, it's a bit questionable at this moment. Maybe we should do one of these: ...

1. We just disallow all such cases
2. Document the case when the order can be deterministic
3. Document that the index always should be sorted before such operations
4. Internally always sort on the index in such cases

itholic:
I think we can close this for now, since we have `__natural_order__` column internally. https://github.com/databricks/koalas/pull/1146.

Issue #411, opened 2019-05-31
Series index order question


yuwillrun:
Recently i'm creating some simple function related to  serise index.
When writing demo to see how index work, i found the series order is confusion like below. Is there something wrong?

```python
import databricks.koalas as ks
series = ks.Series([1, 2, 3])
series
2    3
1    2
0    1
Name: 0, dtype: int64

series.index
Int64Index([2, 0, 1], dtype='int64')

```

ueshin:
Hi @yuwillrun, thanks for reporting this!
Could you let us know your environment? E.g., what version of Python, pandas, Koalas?

I tried with master head in my local, but the result seems deterministic:

```py
>>> import databricks.koalas as ks
>>> series = ks.Series([1, 2, 3])
>>> series
0    1
1    2
2    3
Name: 0, dtype: int64
>>> series.index
Int64Index([0, 1, 2], dtype='int64')
```

yuwillrun:
Here is my details.
I'm in local jupyter model write python script. The script will run pyspark task via company cluster by docker(abount 500 instances). Is this the reason causing this due to some unknown distributed reasons.

python '3.6.4'
pyspark '2.3.0'
koalas '0.5.0'
pandas '0.22.0'

ueshin:
hm, Spark's `sdf.limit(n)` in Koalas' `kdf.head(n)` we use to limit the number of rows to be shown seems sometimes non-deterministic in a distributed environment.

```
>>> from pyspark.sql.session import SparkSession
>>> spark = SparkSession.builder.master('local-cluster[4, 2, 1024]').getOrCreate()

>>> sdf = spark.createDataFrame([[1],[2],[3]], ['a'])
>>> sdf.show()
+---+
|  a|
+---+
|  1|
|  2|
|  3|
+---+

>>> sdf.limit(10).show()
+---+
|  a|
+---+
|  3|
|  2|
|  1|
+---+

>>> sdf.limit(10).show()
+---+
|  a|
+---+
|  1|
|  3|
|  2|
+---+
```

PR #1231, opened 2020-01-27
Add a config to perform natural ordering when head.

ueshin:
Koalas doesn't guarantee the row ordering, so `head` could return some rows from distributed partitions and the result is not deterministic, which might confuse users. (See #411, #1064, or #1219)

```py
>>> spark = ks.utils.default_session({'spark.master': 'local-cluster[4, 2, 1024]'})
>>> spark.conf.set('spark.sql.execution.arrow.enabled', True)
>>> kdf = ks.DataFrame({'a': range(10)})
>>> pdf = kdf.to_pandas().head(3)
```

```py
>>> for _ in range(100):
...     assert kdf.head(3).to_pandas().equals(pdf), 'NOT equals.'
... else:
...     print('equals.')
...
  File "<stdin>", line 2, in <module>
AssertionError: NOT equals.
```

If `compute.ordered_head` is set to True, Koalas performs natural ordering beforehand, but it will cause a performance overhead.

```py
>>> ks.options.compute.ordered_head = True
>>>
>>> for _ in range(100):
...     assert kdf.head(3).to_pandas().equals(pdf), 'NOT equals.'
... else:
...     print('equals.')
...
equals.
```

I don't think we should enable it by default due to the overhead, but it will help users debug.

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -174,7 +174,7 @@ _options = [
         key='compute.ordered_head',
         doc=(
             "'compute.ordered_head' sets whether or not to operate head with natural ordering. "
-            "Koalas doesn't guarantee the row ordering so `head` could return some rows from "
+            "Koalas does not guarantee the row ordering so `head` could return some rows from "
             "distributed partitions. If 'compute.ordered_head' is set to True, Koalas performs "
             "natural ordering beforehand, but it will cause a performance overhead."),
         default=False,
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -171,6 +171,16 @@ _options = [
             "Index type should be one of 'sequence', 'distributed', 'distributed-sequence'.")),
 
     Option(
+        key='compute.ordered_head',
+        doc=(
+            "'compute.ordered_head' sets whether or not to operate head with natural ordering. "
+            "Koalas doesn't guarantee the row ordering so `head` could return some rows from "
+            "distributed partitions. If 'compute.ordered_head' is set to True, Koalas performs "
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
Yeah, I think adding this configuration wouldn't hurt.

databricks/koalas/config.py
```diff
@@ -170,6 +170,16 @@ def validate(self, v: Any) -> None:
             lambda v: v in ('sequence', 'distributed', 'distributed-sequence'),
             "Index type should be one of 'sequence', 'distributed', 'distributed-sequence'.")),
 
+    Option(
+        key='compute.ordered_head',
+        doc=(
+            "'compute.ordered_head' sets whether or not to operate head with natural ordering. "
+            "Koalas doesn't guarantee the row ordering so `head` could return some rows from "
```
HyukjinKwon: No big deal but I wouldn't use abbreviation (doesn't -> does not) in the documentation.

databricks/koalas/config.py
```diff
@@ -170,6 +170,16 @@ def validate(self, v: Any) -> None:
             lambda v: v in ('sequence', 'distributed', 'distributed-sequence'),
             "Index type should be one of 'sequence', 'distributed', 'distributed-sequence'.")),
 
+    Option(
+        key='compute.ordered_head',
```
HyukjinKwon: We could fix the doc together.

https://github.com/databricks/koalas/blob/defeaffef0b0527c821c9332ba5043a4db04d653/databricks/koalas/config.py#L110-L114

ueshin:
Thanks! I'd merge this for now. Please feel free to leave comments if any.

PR #1232, opened 2020-01-28
Fix clip to use `_InternalFrame.with_new_columns`.

ueshin:
Fixes `DataFrame/Series.clip` function to use `_InternalFrame.with_new_columns` and preserve its index.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4441,22 +4441,29 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         numeric_types = (DecimalType, DoubleType, FloatType, ByteType, IntegerType, LongType,
                          ShortType)
-        numeric_columns = [(c, self._internal.scol_for(c)) for c in self.columns
-                           if isinstance(self._internal.spark_type_for(c), numeric_types)]
+        numeric_columns = [(idx, self._internal.scol_for(idx))
+                           for idx in self._internal.column_index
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -4441,22 +4441,29 @@ def clip(self, lower: Union[float, int] = None, upper: Union[float, int] = None)
...
-            numeric_columns = [(c, F.when(scol < lower, lower).otherwise(scol).alias(c))
-                               for c, scol in numeric_columns]
+            numeric_columns = [(idx, (F.when(scol < lower, lower).otherwise(scol)
+                                      .alias(name_like_string(idx))))
+                               for idx, scol in numeric_columns]
         if upper is not None:
-            numeric_columns = [(c, F.when(scol > upper, upper).otherwise(scol).alias(c))
-                               for c, scol in numeric_columns]
+            numeric_columns = [(idx, (F.when(scol > upper, upper).otherwise(scol)
+                                      .alias(name_like_string(idx))))
```
HyukjinKwon: Hm .. should we maybe not use alias at all against Spark columns whenever possible?

ueshin: Ideally, yes, but so far we can't avoid column name access so we should have proper names. Otherwise, if the name becomes something like "xx`yy" by accident, it throws an error since it breaks the naming rule of Spark.

PR #1233, opened 2020-01-28
Add a test case when series.groupby series references the same column but with different values

HyukjinKwon:
A small followup of https://github.com/databricks/koalas/pull/1224 and https://github.com/databricks/koalas/pull/1229

Now, we can cover this case


ueshin:
Thanks! merging.

PR #1237, opened 2020-01-28
Introduce `_apply_series_op` util function for DataFrame.

ueshin:
Introduce `_apply_series_op` util function to consolidate the same pattern of code to call corresponding Series functions.


ueshin:
Thanks! merging.

PR #1238, opened 2020-01-29
Throw a better exception in DataFrame.sort_values when multi-index column is used

HyukjinKwon:
Previously, it threw the exception as below:

```
  File "/.../koalas/databricks/koalas/frame.py", line 5398, in sort_values
    by = [self[colname]._scol for colname in by]
  File "/.../koalas/databricks/koalas/frame.py", line 5398, in <listcomp>
    by = [self[colname]._scol for colname in by]
  File "/.../koalas/databricks/koalas/frame.py", line 8334, in __getattr__
    "'%s' object has no attribute '%s'" % (self.__class__.__name__, key))
AttributeError: 'DataFrame' object has no attribute '_scol'
```

Can be tested as below:

```python
import pandas as pd

mi = pd.MultiIndex.from_tuples([('a', 'b')])
kdf = ks.range(10)
kdf.columns = mi
kdf.sort_values("a")
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -5385,8 +5385,17 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             by = [by]  # type: ignore
         else:
             by = [b if isinstance(b, tuple) else (b,) for b in by]  # type: ignore
-        by = [self[colname]._scol for colname in by]
-        return self._sort(by=by, ascending=ascending,
+
+        new_by = []
+        for colname in by:
+            ser = self[colname]
+            if not isinstance(ser, ks.Series):
+                raise ValueError(
+                    "The column %s is not unique. For a multi-index, the label must "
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -5385,8 +5385,17 @@ def sort_values(self, by: Union[str, List[str], Tuple[str, ...], List[Tuple[str,
...
-        by = [self[colname]._scol for colname in by]
-        return self._sort(by=by, ascending=ascending,
+
+        new_by = []
+        for colname in by:
+            ser = self[colname]
+            if not isinstance(ser, ks.Series):
+                raise ValueError(
+                    "The column %s is not unique. For a multi-index, the label must "
+                    "be a tuple with elements corresponding to each level." % colname)
```
ueshin: `name_like_string(colname)` instead of `colname`?

PR #1242, opened 2020-01-29
Add `_apply_series_op` to Series as well.

ueshin:
Add `_apply_series_op` to Series as well to be able to use it in `_Frame`.

PR #1243, opened 2020-01-30
Apply the similar `_apply_series_op` to GroupBy.

ueshin:
This is a follow-up of #1237 and #1242 to apply the similar way to `GroupBy`.


ueshin:
Thanks! I'd merge this for now.

Issue #1248, opened 2020-01-31
Output of print ks.Series


beobest2:
I tried to check the __print()__ function output of koalas Series
I noticed that there is a difference in output from pandas Series

```python
>>> ks = ks.Series([1, 3, 5, np.nan, 6, 8])
>>> print(ks)
Column<b'0'>
>>> print(str(ks))
Column<b'0'>
>>> print(repr(ks))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>>
>>> ps = pd.Series([1, 3, 5, np.nan, 6, 8])
>>> print(ps)
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
>>> print(str(ps))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
>>> print(repr(ps))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
```
Output of print(ks) is  "Column<b'0'>" 
This does not represent data inside the Series.

beobest2:
I simply removed the \_\_str\_\_function from *series.py* and it works fine.
please check this out.

```python
>>> ks = ks.Series([1, 3, 5, np.nan, 6, 8])
>>> print(ks)
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>> print(str(ks))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>> print(repr(ks))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>>
>>> ps = pd.Series([1, 3, 5, np.nan, 6, 8])
>>> print(ps)
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
>>> print(str(ps))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
>>> print(repr(ps))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
```

beobest2:
I submitted PR https://github.com/databricks/koalas/pull/1250

PR #1250, opened 2020-01-31
Fix output of print() matches with pandas of Series

beobest2:
Resolve https://github.com/databricks/koalas/issues/1248

Since the Python interpreter calls \_\_repr\_\_ if \_\_str\_\_ is not implemented, 
the output will be identical to pandas using the implemented \_\_repr\_\_ of __series.py__.

```python
>>> ks = ks.Series([1, 3, 5, np.nan, 6, 8])
>>> print(ks)
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>> print(str(ks))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>> print(repr(ks))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
Name: 0, dtype: float64
>>> ps = pd.Series([1, 3, 5, np.nan, 6, 8])
>>> print(ps)
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
>>> print(str(ps))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
>>> print(repr(ps))
0    1.0
1    3.0
2    5.0
3    NaN
4    6.0
5    8.0
dtype: float64
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4292,10 +4292,6 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
                 return partial(property_or_func, self)
         return self.getField(item)
 
-    def __str__(self):
-        # TODO: figure out how to reuse the original one.
-        return 'Column<%s>' % self._scol._jc.toString().encode('utf8')
-
     def _to_internal_pandas(self):
         """
         Return a pandas Series directly from _internal to avoid overhead of copy.


itholic:
LGTM

itholic:
LGTM, too if there is no reason we should use the existing behavior.

itholic:
LGTM, too if there is no proper reason we should use the existing behavior.

itholic:
nit: we better avoid `ks` as a variable name of `Series` since there is a high probability of overwriting the koalas package name, `ks`.

beobest2:
> nit: we better avoid `ks` as a variable name of `Series` since there is a high probability of overwriting the Koalas package name, `ks`. (I recommend `kser` instead, generally)

Thanks for the good tip :)

Issue #1252, opened 2020-02-01
 Check parameters when grouping in multiple columns


beobest2:
```python
>>> pdf = pd.DataFrame({'A': ['foo', 'bar', 'foo', 'bar',
...                     'foo', 'bar', 'foo', 'foo'],
...                     'B': ['one', 'one', 'two', 'three',
...                     'two', 'two', 'one', 'three'],
...                     'C': np.random.randn(8),
...                     'D': np.random.randn(8)})
>>> kdf = ks.from_pandas(pdf)
```

Checking grouping by multiple columns forms a hierarchical index. I works fine.
```python
>>> kdf.groupby(['B', 'A']).sum().sort_index()
                  C         D
B     A
one   bar -0.193929  0.193662
      foo  1.134526 -0.062965
three bar  1.075517 -0.621216
      foo  1.297346 -0.462547
two   bar  0.509226  1.057209
      foo  0.911734  0.284388
>>> pdf.groupby(['B', 'A']).sum()
                  C         D
B     A
one   bar -0.193929  0.193662
      foo  1.134526 -0.062965
three bar  1.075517 -0.621216
      foo  1.297346 -0.462547
two   bar  0.509226  1.057209
      foo  0.911734  0.284388
```

I missed square brackets around the parameter to __groupby__.
koalas couldn't catch a mistake and did __groupby__ job with the first parameter ('B' column).

```python
>>> kdf.groupby('B', 'A').sum().sort_index()
              C         D
B
one    0.940597  0.130697
three  2.372863 -1.083763
two    1.420961  1.341598
>>> pdf.groupby('B', 'A').sum()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/site-packages/pandas/core/generic.py", line 7883, in groupby
    axis = self._get_axis_number(axis)
  File "/usr/local/lib/python3.7/site-packages/pandas/core/generic.py", line 411, in _get_axis_number
    raise ValueError("No axis named {0} for object type {1}".format(axis, cls))
ValueError: No axis named A for object type <class 'pandas.core.frame.DataFrame'>
>>>
```

itholic:
Thanks for the report.

are you interested in trying to create a Pull Request to fix this?

beobest2:
@itholic You're welcome :) I'm going to take a look at this

beobest2:
I submitted PR https://github.com/databricks/koalas/pull/1253

PR #1253, opened 2020-02-02
Add as_index check logic to groupby parameter

beobest2:
Resolve https://github.com/databricks/koalas/issues/1252

By doing type validation on item __as_index__  in groupby, 
I think it will be fine if other parameters are added later.
Now, when checking grouping of multiple columns, koalas can find mistakes.

```python
>>> pdf = pd.DataFrame({'A': ['foo', 'bar', 'foo', 'bar',
...                   'foo', 'bar', 'foo', 'foo'],
...                    'B': ['one', 'one', 'two', 'three',
...                    'two', 'two', 'one', 'three'],
...                    'C': np.random.randn(8),
...                    'D': np.random.randn(8)})
>>> kdf = ks.from_pandas(pdf)
>>> kdf.groupby('B', 'A').sum().sort_index()
  File "/Users/hwpark/Desktop/dev/git_koalas/koalas/databricks/koalas/generic.py", line 1286, in groupby
    'got [%s]' % type(as_index))
TypeError: as_index must be an boolean; however, got [<class 'str'>]
>>> pdf.groupby('B', 'A').sum()
  File "/Users/hwpark/Desktop/dev/git_koalas/venv/lib/python3.7/site-packages/pandas/core/generic.py", line 7883, in groupby
    axis = self._get_axis_number(axis)
  File "/Users/hwpark/Desktop/dev/git_koalas/venv/lib/python3.7/site-packages/pandas/core/generic.py", line 411, in _get_axis_number
    raise ValueError("No axis named {0} for object type {1}".format(axis, cls))
ValueError: No axis named A for object type <class 'pandas.core.frame.DataFrame'>
```

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -1282,7 +1282,7 @@ class _Frame(object):
         if not len(by):
             raise ValueError('No group keys passed!')
         if not isinstance(as_index, bool):
-            raise TypeError('as_index must be an boolean; however, '
+            raise TypeError('as_index must be a boolean; however, '
                             'got [%s]' % type(as_index))
         if isinstance(df_or_s, DataFrame):
             df = df_or_s  # type: DataFrame
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -1285,7 +1285,7 @@ class _Frame(object):
             raise ValueError('No group keys passed!')
         axis = validate_axis(axis)
         if axis != 0:
-            raise ValueError('axis should be either 0 or "index" currently.')
+            raise NotImplementedError('axis should be either 0 or "index" currently.')
         if isinstance(df_or_s, DataFrame):
             df = df_or_s  # type: DataFrame
             col_by = [_resolve_col(df, col_or_s) for col_or_s in by]
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -170,8 +170,7 @@ class _Frame(object):
     # TODO: add 'axis' parameter
     def cumsum(self, skipna: bool = True):
         """
-        Return cumulative sum over a DataFrame or Series  axis=0,
-   axis.
+        Return cumulative sum over a DataFrame or Series axis.
 
         Returns a DataFrame or Series of the same size containing the cumulative sum.
 
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -34,7 +34,7 @@ from pyspark.sql.types import DataType, DoubleType, FloatType
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
 from databricks.koalas.indexing import AtIndexer, iAtIndexer, iLocIndexer, LocIndexer
 from databricks.koalas.internal import _InternalFrame, NATURAL_ORDER_COLUMN_NAME
-from databricks.koalas.utils import validate_arguments_and_invoke_function, scol_for
+from databricks.koalas.utils import validate_arguments_and_invoke_function, scol_for, validate_axis
 from databricks.koalas.window import Rolling, Expanding
 
 
@@ -1209,6 +1209,8 @@ class _Frame(object):
             If Series is passed, the Series or dict VALUES
             will be used to determine the groups. A label or list of
             labels may be passed to group by the columns in ``self``.
+        axis : int, default 0 or 'index'
+            Can only be set to 0 at the moment.
         as_index : bool, default True
             For aggregated output, return object with group labels as the
             index. Only relevant for DataFrame input. as_index=False is
@@ -1281,6 +1283,7 @@ class _Frame(object):
             raise ValueError("Grouper for '{}' not 1-dimensional".format(type(by)))
         if not len(by):
             raise ValueError('No group keys passed!')
+        axis = validate_axis(axis)
         if axis != 0:
             raise ValueError('axis should be either 0 or "index" currently.')
         if isinstance(df_or_s, DataFrame):
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -170,7 +170,8 @@ class _Frame(object):
     # TODO: add 'axis' parameter
     def cumsum(self, skipna: bool = True):
         """
-        Return cumulative sum over a DataFrame or Series axis.
+        Return cumulative sum over a DataFrame or Series  axis=0,
+   axis.
 
         Returns a DataFrame or Series of the same size containing the cumulative sum.
 
@@ -1193,7 +1194,7 @@ class _Frame(object):
 
     # TODO: by argument only support the grouping name and as_index only for now. Documentation
     # should be updated when it's supported.
-    def groupby(self, by, as_index: bool = True):
+    def groupby(self, by, axis=0, as_index: bool = True):
         """
         Group DataFrame or Series using a Series of columns.
 
@@ -1281,6 +1282,8 @@ class _Frame(object):
             raise ValueError("Grouper for '{}' not 1-dimensional".format(type(by)))
         if not len(by):
             raise ValueError('No group keys passed!')
+        if axis != 0:
+            raise ValueError('axis sould be either 0 or "index" currently.')
         if not isinstance(as_index, bool):
             raise TypeError('as_index must be a boolean; however, '
                             'got [%s]' % type(as_index))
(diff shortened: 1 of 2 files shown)


databricks/koalas/generic.py
```diff
@@ -1281,6 +1281,9 @@ def groupby(self, by, as_index: bool = True):
             raise ValueError("Grouper for '{}' not 1-dimensional".format(type(by)))
         if not len(by):
             raise ValueError('No group keys passed!')
+        if not isinstance(as_index, bool):
+            raise TypeError('as_index must be an boolean; however, '
```
itholic: nit: `as_index must be an boolean` -> `as_index must be a boolean`

itholic:
Otherwise, looks fine.

databricks/koalas/generic.py
itholic: hmm, maybe this way looks fine for now, but i think maybe we should handle `by` parameter directly rather than `as_index` since now behavior looks some hacky (e.g. if other parameters will be added to second positional parameter, it will not work properly (maybe `axis`, `level` like the pandas are doing that like the below)

```python
    @Appender(_shared_docs["groupby"] % _shared_doc_kwargs)
    def groupby(
        self,
        by=None,
        axis=0,
        level=None,
        as_index: bool = True,
        sort: bool = True,
        group_keys: bool = True,
        squeeze: bool = False,
        observed: bool = False,
    ) -> "groupby_generic.DataFrameGroupBy":
```

beobest2: Oppose~ I made a mistake. I'll fix this.

itholic: i think we can handle axis parameter like pandas.

```python
ValueError: No axis named A for object type <class 'pandas.core.frame.DataFrame'>
```

beobest2: > hmm, maybe this way looks fine for now, but i think we better handle `by` parameter directly rather than `as_index` since now behavior looks some hacky (e.g. if other parameters will be added to second positional parameter, it will not work properly - maybe `axis` or `level` like the below pandas are doing-)
> 
> ```python
>     @Appender(_shared_docs["groupby"] % _shared_doc_kwargs)
>     def groupby(
>         self,
>         by=None,
>         axis=0,
>         level=None,
>         as_index: bool = True,
>         sort: bool = True,
>         group_keys: bool = True,
>         squeeze: bool = False,
>         observed: bool = False,
>     ) -> "groupby_generic.DataFrameGroupBy":
> ```

Pandas recognized the value of the second argument as an "axis". Koalas does not yet support "axis" parameters, so it is recognized as "as_index". It is expected that "as_index" will be safely verified no matter what function is added in the future.

beobest2: > i think we can handle `axis` parameter like pandas.
> 
> ```python
> ValueError: No axis named A for object type <class 'pandas.core.frame.DataFrame'>
> ```
> 
> for example, you can add `axis` parameter with default 0, and raise `NotImplementedError` for when `axis=1` for now.


Let me proceed that way. Thank you :)

itholic: yes, go for it!

FYI: although i think your approach is not bad, but the point is that we always try to mimic pandas as possible as we can 😃

itholic: And pandas doesn't raise exception when `as_index` is not boolean.

- for string

```python
>>> pdf.groupby(['A', 'B'], as_index='koalas').sum()
                  C         D
A   B
bar one   -0.517176  1.082738
    three  2.075121  0.745619
    two    1.062240 -0.289503
foo one    0.028217 -0.131834
    three -0.374145 -0.422457
    two    2.320670  1.416018
```

- for number

```python
>>> pdf.groupby(['A', 'B'], as_index=100).sum()
                  C         D
A   B
bar one   -0.517176  1.082738
    three  2.075121  0.745619
    two    1.062240 -0.289503
foo one    0.028217 -0.131834
    three -0.374145 -0.422457
    two    2.320670  1.416018
```

but we are 

```python
>>> kdf.groupby('A', as_index='koalas').sum()
...
TypeError: as_index must be an boolean; however, got [<class 'str'>]
```

beobest2: > And pandas doesn't raise exception when `as_index` is not boolean.
> (They treat 0 as False otherwise True)
> 
> * `True` for string or number which is not 0.
> 
> ```python
> >>> pdf.groupby('A', as_index='koalas').sum()
>             C         D
> A
> bar -0.998532  1.623860
> foo  3.844849  1.563355
> 
> >>> pdf.groupby('A', as_index=100).sum()
>             C         D
> A
> bar -0.998532  1.623860
> foo  3.844849  1.563355
> ```
> 
> * `False` for 0
> 
> ```python
> >>> pdf.groupby('A', as_index=0).sum()
>      A         C         D
> 0  bar -0.998532  1.623860
> 1  foo  3.844849  1.563355
> ```
> 
> but we're not.
> 
> ```python
> >>> kdf.groupby('A', as_index='koalas').sum()
> Traceback (most recent call last):
> ...
> TypeError: as_index must be an boolean; however, got [<class 'str'>]
> ```

I got it! I will remove the "as_index" validation to work with the example shown above.

beobest2:
@itholic 
"as_index" validation is removed
```python
>>> kdf.groupby('A', as_index='koalas').sum()
            C         D
A
bar  2.124855 -3.710326
foo -0.271959 -0.680334
```

also I add axis parameter with default 0, and raise error for when axis=1 for now.
```python
>>> kdf.groupby('B', 'A').sum().sort_index()
  File "/root/git_koalas/koalas/databricks/koalas/generic.py", line 1285, in groupby
    raise ValueError('axis should be either 0 or "index" currently.')
ValueError: axis sould be either 0 or "index" currently.
```

itholic:
LGTM if test is passed.

databricks/koalas/generic.py
```diff
@@ -1193,7 +1193,7 @@ def abs(self):
 
     # TODO: by argument only support the grouping name and as_index only for now. Documentation
     # should be updated when it's supported.
-    def groupby(self, by, as_index: bool = True):
+    def groupby(self, by, axis=0, as_index: bool = True):
         """
         Group DataFrame or Series using a Series of columns.
```
HyukjinKwon: The parameter should be fixed too. Actually, why don't you try to implement the other axis? It wouldn't be impossible to do if we use pandas UDF from a cursory look. We have enough time before the next release currently.

beobest2: @HyukjinKwon  I'll be happy to do that.

beobest2: @HyukjinKwon  For now, I modified it to only supports axis = 0. (also docstring is fix)
I've looked into Pandas, but I don't have confidence in Koalas yet, 
so the implementation of other axis seems to take some time. 
Hmm.. Is it better I implement another axis as the next step after this PR?
 or do you want me to keep developing in this PR?

databricks/koalas/generic.py
```diff
@@ -1281,6 +1283,9 @@ def groupby(self, by, as_index: bool = True):
             raise ValueError("Grouper for '{}' not 1-dimensional".format(type(by)))
         if not len(by):
             raise ValueError('No group keys passed!')
+        axis = validate_axis(axis)
```
itholic: if we're not going to implement `axis=1` for now, i think we don't need this yet.

databricks/koalas/generic.py
```diff
@@ -1281,6 +1283,9 @@ def groupby(self, by, as_index: bool = True):
...
+        if axis != 0:
+            raise ValueError('axis should be either 0 or "index" currently.')
```
itholic: Let's raise `NotImplementedError` since https://github.com/databricks/koalas/pull/1256#issue-370544117

databricks/koalas/tests/test_groupby.py
```diff
@@ -80,6 +80,10 @@ def test_groupby(self):
 
         self.assertRaises(TypeError, lambda: kdf.a.groupby(kdf.b, as_index=False))
 
+        self.assertRaises(ValueError, lambda: kdf.groupby('a', axis=1))
```
itholic: so should be fixed here also

ueshin: Could you add tests to specify `axis=0` and `'index'`?

beobest2: @itholic @ueshin okay let me fix it. thank you!

databricks/koalas/generic.py
HyukjinKwon: Sure.

PR #1254, opened 2020-02-03
Migrate to GitHub Actions

itholic:
I suggest introducing [GitHub Actions](https://help.github.com/en/actions/automating-your-workflow-with-github-actions) to Koalas.

- Since it is a tool provided by GitHub itself, I think that stable support for GitHub repository is possible compared to other CI tools.

- Currently it supports 20 concurrent jobs vs Travis 3 concurrent jobs.

- It also support Mac and Windows.

- And i think it's simple to use and the [marketplace](https://github.com/marketplace) is so active that it's easy to apply a lot of additional features in the future.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -54,21 +54,26 @@ jobs:
     - name: Update Codecov
       run: bash <(curl -s https://codecov.io/bash)
 
-  python-conda:
+  python:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         python: [3.6, 3.7]
         include:
-          - python: 3.6
-            pandas: 0.24.2
-            pyarrow: 0.13.0
+          - python-version: 3.6
+            spark-version: 2.4.4
+            pandas-version: 0.24.2
+            pyarrow-version: 0.13.0
             logger: databricks.koalas.usage_logging.usage_logger
-          - python: 3.7
-            pandas: 0.25.3
-            pyarrow: 0.14.1
+          - python-version: 3.7
+            spark-version: 2.4.4
+            pandas-version: 0.25.3
+            pyarrow-version: 0.14.1
     env:
-      SPARK_VERSION: 2.4.4
+      PYTHON_VERSION: ${{ matrix.python-version }}
+      SPARK_VERSION: ${{ matrix.spark-version }}
+      PANDAS_VERSION: ${{ matrix.pandas-version }}
+      PYARROW_VERSION: ${{ matrix.pyarrow-version }}
       DISPLAY: 0.0
       QT_QPA_PLATFORM: offscreen
       KOALAS_USAGE_LOGGER: ${{ matrix.logger }}
@@ -96,13 +101,13 @@ jobs:
         # Useful for debugging any issues with conda
         conda info -a
         # Replace dep1 dep2 ... with your dependencies
-        conda create -c conda-forge -q -n test-environment python=3.6
+        conda create -c conda-forge -q -n test-environment python=$PYTHON_VERSION
         conda activate test-environment
         conda install -c conda-forge --yes codecov
-        conda config --env --add pinned_packages python=${{ matrix.python }}
-        conda config --env --add pinned_packages pandas==${{ matrix.pandas }}
-        conda config --env --add pinned_packages pyarrow==${{ matrix.pyarrow }}
-        conda install -c conda-forge --yes pandas==${{ matrix.pandas }} pyarrow==${{ matrix.pyarrow }}
+        conda config --env --add pinned_packages python=$PYTHON_VERSION
+        conda config --env --add pinned_packages pandas==$PANDAS_VERSION
+        conda config --env --add pinned_packages pyarrow==$PYARROW_VERSION
+        conda install -c conda-forge --yes pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
         conda install -c conda-forge --yes --freeze-installed --file requirements-dev.txt
         conda list
         export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -58,7 +58,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.7]
+        python-version: [3.6, 3.7]
         include:
           - python-version: 3.6
             spark-version: 2.4.4
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -80,7 +80,6 @@ jobs:
       run: ./dev/download_travis_dependencies.sh
     - name: Install dependencies & Run tests
       run: |
-        sudo apt-get update
         curl -s https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > miniconda.sh
         bash miniconda.sh -b -p $HOME/miniconda
         . $HOME/miniconda/etc/profile.d/conda.sh
@@ -127,7 +126,6 @@ jobs:
       run: ./dev/download_travis_dependencies.sh
     - name: Install dependencies & Run tests
       run: |
-        sudo apt-get update
         curl -s https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > miniconda.sh
         bash miniconda.sh -b -p $HOME/miniconda
         . $HOME/miniconda/etc/profile.d/conda.sh
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -32,6 +32,13 @@ jobs:
         sudo apt-get install tk-dev python3.5-tk python3 python3.5 python-pip python3-pip python-pytest
         sudo rm /usr/bin/python
         sudo ln -s /usr/bin/python3.5 /usr/bin/python
+    - name: Using cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
     - name: Download dependencies
       run: ./dev/download_travis_dependencies.sh
     - name: Install Python dependencies
@@ -72,6 +79,13 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.6
+    - name: Using cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
     - name: Download dependencies
       run: ./dev/download_travis_dependencies.sh
     - name: Install Python dependencies
@@ -108,6 +122,13 @@ jobs:
       uses: actions/setup-python@v1
       with:
         python-version: 3.7
+    - name: Using cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
     - name: Download dependencies
       run: ./dev/download_travis_dependencies.sh
     - name: Install Python dependencies
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -9,137 +9,119 @@ on:
       - master
 
 jobs:
-  python3_5:
+  python35:
     runs-on: ubuntu-latest
     env:
       SPARK_VERSION: 2.3.4
-      PANDAS_VERSION: 0.23.4
+      PANDAS_VERSION: 0.23.3
       PYARROW_VERSION: 0.10.0
-
+      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
+      QT_QPA_PLATFORM: offscreen
     steps:
     - uses: actions/checkout@v2
-
     - uses: actions/setup-java@v1
       with:
         java-version: 1.8
-
-    - name: setup_python
+    - name: Setup Python 3.5
       uses: actions/setup-python@v1
       with:
         python-version: 3.5
-
-    - name: before_install
+    - name: Download dependencies
       run: ./dev/download_travis_dependencies.sh
-
-    - name: install
+    - name: Install Python dependencies
       run: |
         sudo apt-get update
+        export DISPLAY=:0.0
+        sudo apt-get install xclip
         pip install --upgrade cython
+        pip install pyperclip
         pip install -r requirements-dev.txt
         pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
         pip install 'pyspark>=2.4'
         pip list
-
-    - name: before_script
+    - name: Setup environment variables
       run: |
         export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
-        export QT_QPA_PLATFORM="offscreen"
         ./dev/lint-python
-
-    - name: script
+    - name: Run tests
       run: ./dev/pytest
-
-    - name: after_success
+    - name: Update Codecov
       run: bash <(curl -s https://codecov.io/bash)
 
-  python3_6:
+  python36:
     runs-on: ubuntu-latest
     env:
       SPARK_VERSION: 2.4.4
       PANDAS_VERSION: 0.24.2
       PYARROW_VERSION: 0.13.0
-
+      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
+      QT_QPA_PLATFORM: offscreen
+      DISPLAY: localhost:0.0
     steps:
     - uses: actions/checkout@v2
-
     - uses: actions/setup-java@v1
       with:
         java-version: 1.8
-
-    - name: setup_python
+    - name: Setup Python 3.6
       uses: actions/setup-python@v1
       with:
         python-version: 3.6
-  
-    - name: before_install
+    - name: Download dependencies
       run: ./dev/download_travis_dependencies.sh
-  
-    - name: install
+    - name: Install Python dependencies
       run: |
         sudo apt-get update
+        sudo apt-get install xclip
         pip install --upgrade cython
         pip install pyperclip
         pip install -r requirements-dev.txt
         pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
         pip install 'pyspark>=2.4'
         pip list
-
-    - name: before_script
+    - name: Setup environment variables
       run: |
         export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
-        export QT_QPA_PLATFORM="offscreen"
-        python --version
-        pip list
         ./dev/lint-python
-  
-    - name: script
+    - name: Run tests
       run: ./dev/pytest
-  
-    - name: after_success
+    - name: Update Codecov
       run: bash <(curl -s https://codecov.io/bash)
 
-  python3_7:
+  python37:
     runs-on: ubuntu-latest
     env:
       SPARK_VERSION: 2.4.4
       PANDAS_VERSION: 0.25.3
       PYARROW_VERSION: 0.14.1
-
+      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
+      QT_QPA_PLATFORM: offscreen
+      DISPLAY: localhost:0.0
     steps:
     - uses: actions/checkout@v2
-
     - uses: actions/setup-java@v1
       with:
         java-version: 1.8
-
-    - name: setup_python
+    - name: Setup Python 3.7
       uses: actions/setup-python@v1
       with:
         python-version: 3.7
-  
-    - name: before_install
+    - name: Download dependencies
       run: ./dev/download_travis_dependencies.sh
-  
-    - name: install
+    - name: Install Python dependencies
       run: |
         sudo apt-get update
+        sudo apt-get install xclip
         pip install --upgrade cython
         pip install pyperclip
         pip install -r requirements-dev.txt
         pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
         pip install 'pyspark>=2.4'
         pip list
-
-    - name: before_script
+    - name: Setup environment variables
       run: |
         export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
-        export QT_QPA_PLATFORM="offscreen"
-        python --version
-        pip list
         ./dev/lint-python
-  
-    - name: script
+    - name: Run tests
       run: ./dev/pytest
-  
-    - name: after_success
+    - name: Update Codecov
       run: bash <(curl -s https://codecov.io/bash)
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -21,17 +21,6 @@ jobs:
     - uses: actions/setup-java@v1
       with:
         java-version: 1.8
-    # Setup Python 3.5 via `apt-get install` since the dafault Python3.5 from `actions/setup-python`
-    # seems to have some problem with Tkinter, so we should manually install the python3.5-tk.
-    # for this, we should use the Python we manually installed, not the default one from `actions/setup-python`
-    - name: Setup Python 3.5
-      run: |
-        sudo add-apt-repository ppa:deadsnakes/ppa
-        sudo apt-get install tk-dev python3.5-tk python3.5
-        sudo rm /usr/bin/python
-        sudo ln -s /usr/bin/python3.5 /usr/bin/python
-        # Add the path which packages are actually installed to `PATH`
-        echo "::add-path::/home/runner/.local/bin"
     - name: Using cache
       uses: actions/cache@v1
       with:
@@ -39,29 +28,50 @@ jobs:
         key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
-    - name: Download dependencies
-      run: ./dev/download_travis_dependencies.sh
+    # Setup Python3.5 via `apt-get install` since the dafault Python3.5 from `actions/setup-python`
+    # seems to have some problem with Tkinter, so we should manually install the python3.5-tk.
+    # for this, we should use the Python manually installed, not the default one from `actions/setup-python`
+    - name: Setup Python 3.5
+      run: |
+        sudo add-apt-repository ppa:deadsnakes/ppa
+        sudo apt-get install tk-dev python3.5-tk python3.5
+        sudo rm /usr/bin/python
+        sudo ln -s /usr/bin/python3.5 /usr/bin/python
+        # Below command is required for ensuring pip installed executables to be in the path.
+        echo "::add-path::/home/runner/.local/bin"
     - name: Install dependencies & run tests
       run: |
+        ./dev/download_travis_dependencies.sh
         sudo apt-get install xclip
-        python -m pip install --upgrade pip
-        python -m pip install -r requirements-dev.txt
-        python -m pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
-        python -m pip list
+        pip install setuptools
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip list
         export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
         ./dev/lint-python
         ./dev/pytest
+    # Push the results back to codecov
     - name: Update Codecov
       run: bash <(curl -s https://codecov.io/bash)
 
-  python36:
+  python-conda:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python: [3.6, 3.7]
+        include:
+          - python: 3.6
+            pandas: 0.24.2
+            pyarrow: 0.13.0
+            logger: databricks.koalas.usage_logging.usage_logger
+          - python: 3.7
+            pandas: 0.25.3
+            pyarrow: 0.14.1
     env:
       SPARK_VERSION: 2.4.4
-      PANDAS_VERSION: 0.24.2
-      PYARROW_VERSION: 0.13.0
-      PYTHON_EXECUTABLE: xvfb-run python
-      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
+      DISPLAY: 0.0
+      QT_QPA_PLATFORM: offscreen
+      KOALAS_USAGE_LOGGER: ${{ matrix.logger }}
     steps:
     - uses: actions/checkout@v2
     - uses: actions/setup-java@v1
@@ -74,10 +84,9 @@ jobs:
         key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
         restore-keys: |
           ${{ runner.os }}-pip-
-    - name: Download dependencies
-      run: ./dev/download_travis_dependencies.sh
     - name: Install dependencies & run tests
       run: |
+        ./dev/download_travis_dependencies.sh
         curl -s https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > miniconda.sh
         bash miniconda.sh -b -p $HOME/miniconda
         . $HOME/miniconda/etc/profile.d/conda.sh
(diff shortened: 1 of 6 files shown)


.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+      PYARROW_VERSION: 0.10.0
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    - name: setup_python
```
HyukjinKwon: Can you name it pretty? e.g.) `Set up Python 3.5`

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+  python3_5:
+    runs-on: ubuntu-latest
+    env:
+      SPARK_VERSION: 2.3.4
+      PANDAS_VERSION: 0.23.4
+      PYARROW_VERSION: 0.10.0
+
+    steps:
+    - uses: actions/checkout@v2
+
```
HyukjinKwon: nit, I would remove this newline.

itholic:
Since it is very first time to me using GitHub Actions, need some advice 😅 

cc @ueshin @HyukjinKwon

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
+    - name: setup_python
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.5
+
+    - name: before_install
```
HyukjinKwon: Name it pretty here as well. e.g.) `Download dependencies`

.github/workflows/master.yml
itholic: ah, yes. i'll take a look for other also 👍

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+      run: ./dev/download_travis_dependencies.sh
+
+    - name: install
```
HyukjinKwon: Name it nicer like Install Python dependencies

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+      run: |
+        sudo apt-get update
+        pip install --upgrade cython
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
+        pip list
+
+    - name: before_script
```
HyukjinKwon: before_script -> "Set up environment variables"

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+      run: |
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
+        export QT_QPA_PLATFORM="offscreen"
+        ./dev/lint-python
+
+    - name: script
+      run: ./dev/pytest
+
+    - name: after_success
```
HyukjinKwon: "Update Codecov"

.github/workflows/master.yml
itholic: yeah i just use the existing naming of travis at first,

but i also think we better have more pretty naming. thanks! :)

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+        pip install 'pyspark>=2.4'
+        pip list
+
+    - name: before_script
+      run: |
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
+        export QT_QPA_PLATFORM="offscreen"
+        ./dev/lint-python
+
+    - name: script
```
HyukjinKwon: Run tests

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+      with:
+        python-version: 3.6
+  
+    - name: before_install
+      run: ./dev/download_travis_dependencies.sh
+  
+    - name: install
+      run: |
+        sudo apt-get update
+        pip install --upgrade cython
```
HyukjinKwon: Can you test and setup env via using `conda` for Python 3.6 and 3.7? We should test Conda as well, which is the preferred installation way.

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
+        pip list
+
+    - name: before_script
+      run: |
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
+        export QT_QPA_PLATFORM="offscreen"
+        python --version
```
HyukjinKwon: Seems like a mistake.

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+        pip list
```
HyukjinKwon: `install` seems already show dependency installations. Same to Python 3.7 job

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+    branches: 
+      - master
+
+jobs:
```
HyukjinKwon: I would indent this file as below:

```
jobs:
  python35:
    ... no newlines.

  python36:
    ... no newlines.

  ...
```

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+  python3_5:
```
HyukjinKwon: python3_5 -> python35

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+        pip install pyperclip
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
+        pip list
+
+    - name: before_script
+      run: |
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
+        export QT_QPA_PLATFORM="offscreen"
```
itholic: @HyukjinKwon i'm not pretty sure, we don't need this line for now until `xclip` is solved ?

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+
+    - name: after_success
+      run: bash <(curl -s https://codecov.io/bash)
+
+  python3_6:
+    runs-on: ubuntu-latest
+    env:
+      SPARK_VERSION: 2.4.4
+      PANDAS_VERSION: 0.24.2
+      PYARROW_VERSION: 0.13.0
```
HyukjinKwon: Can you keep this env? (https://github.com/databricks/koalas/blob/master/.travis.yml#L27). It tests the Koalas logger

databricks/koalas/tests/test_dataframe_conversion.py
```diff
@@ -158,6 +159,8 @@ def test_to_json_with_path(self):
         output_path = "%s/%s" % (self.tmp_dir, output_paths[0])
         self.assertEqual("[%s]" % open(output_path).read().strip(), expected)
 
+    # TODO: `to_clipboard` raises Error since `xclip` seems not work in GitHub Actions on Linux
+    @unittest.skip("Skip for now until `xclip` issue is solved in GitHub Actions")
```
HyukjinKwon: Is this because you used `ubuntu-latest`? How does it fail with what kind of errors?

.github/workflows/master.yml
itholic: okay, you mean something like that?

```yml
jobs:
  python3_5:
    runs-on: ubuntu-latest
    env:
      SPARK_VERSION: 2.3.4
      PANDAS_VERSION: 0.23.4
      PYARROW_VERSION: 0.10.0
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-java@v1
      with:
        java-version: 1.8
    - name: Setup Python 3.5
      uses: actions/setup-python@v1
      with:
        python-version: 3.5
    - name: Download dependencies
      run: ./dev/download_travis_dependencies.sh
    - name: Install Python dependencies
      run: |
        sudo apt-get update
        pip install --upgrade cython
        pip install -r requirements-dev.txt
        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
        pip install 'pyspark>=2.4'
        pip list
    - name: Setup environment variables
      run: |
        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
        export QT_QPA_PLATFORM="offscreen"
        ./dev/lint-python
    - name: Run tests
      run: ./dev/pytest
    - name: Update Codecov
      run: bash <(curl -s https://codecov.io/bash)

  python3_6:
    runs-on: ubuntu-latest
    env:
      SPARK_VERSION: 2.4.4
      PANDAS_VERSION: 0.24.2
      PYARROW_VERSION: 0.13.0
    steps:
    - uses: actions/checkout@v2
    - uses: actions/setup-java@v1
      with:
        java-version: 1.8
    - name: setup_python
      uses: actions/setup-python@v1
      with:
        python-version: 3.6
    - name: before_install
      run: ./dev/download_travis_dependencies.sh
    - name: install
      run: |
        sudo apt-get update
        pip install --upgrade cython
        pip install pyperclip
        pip install -r requirements-dev.txt
        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
...
```

HyukjinKwon: Yup

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+        sudo apt-get update
+        pip install --upgrade cython
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
+        pip list
+
+    - name: before_script
+      run: |
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
```
HyukjinKwon: I think you can move those environment variables to `env`

.github/workflows/master.yml
```diff
@@ -0,0 +1,145 @@
...
+        pip list
+        ./dev/lint-python
+  
+    - name: script
+      run: ./dev/pytest
+  
+    - name: after_success
+      run: bash <(curl -s https://codecov.io/bash)
+
+  python3_7:
```
HyukjinKwon: Can we use `strategy` & `matrix` to deduplicate the configurations here? Seems it duplicates some settings.

.github/workflows/master.yml
HyukjinKwon: It looks not related to xclip. https://github.com/databricks/koalas/commit/66f5119c37aa7621f95ff81d9b99be38a278091e

itholic: ah, i just read that PR. thanks! :)

.github/workflows/master.yml
itholic: it looks better. :)

.github/workflows/master.yml
itholic: okay let me fix. thanks!! 😃

harupy:
Is it possible to cache Spark as the current Travis config does?

https://github.com/databricks/koalas/blob/0e56ea4644775b352faab1859e1fe69e6bbd3380/.travis.yml#L3-L5

itholic:
@harupy hey, how have you been :)

i tried like the below with using `actions/cache`,

```yml
    - uses: actions/cache@v1
      id: cache
      with:
        path: $HOME/.cache/spark-versions
```

 but they want to `key` something like the below.

```yml
    - uses: actions/cache@v1
      id: cache
      with:
        path: $HOME/.cache/spark-versions
        key: ${{ runner.os }}-primes
```

Any idea what value I should put in `key`?

FYI, they say about `key`:  An explicit key for restoring and saving the cache

HyukjinKwon:
@itholic I think you can use `hashFiles` against https://github.com/databricks/koalas/blob/master/requirements-dev.txt and or maybe https://github.com/databricks/koalas/blob/master/setup.py. See https://github.com/actions/cache/blob/master/examples.md#python---pip

itholic:
@HyukjinKwon 
ah, maybe I was confused with something else.
i can manage this, thanks! :)

harupy:
Just FYI

> A repository can have up to 2GB of caches. Once the 2GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.

https://github.com/actions/cache#cache-limits

itholic:
@harupy thanks for reminding me!! :D

itholic:
i'm working for adding `cache` and `codecov` report after building succeed

itholic:
While Here, our test result for GitHub Actions will be displayed like the below.

스크린샷 2020-02-05 오후 3 12 08

README.md
```diff
@@ -27,6 +27,7 @@ This project is currently in beta and is rapidly evolving, with a bi-weekly rele
 
 Try the Koalas 10 minutes tutorial on a live Jupyter notebook [here](https://mybinder.org/v2/gh/databricks/koalas/master?filepath=docs%2Fsource%2Fgetting_started%2F10min.ipynb). The initial launch can take up to several minutes.
 
+[![Github Actions](https://github.com/databricks/koalas/workflows/master/badge.svg)](https://github.com/databricks/koalas/actions)
 [![Build Status](https://travis-ci.com/databricks/koalas.svg?token=Rzzgd1itxsPZRuhKGnhD&branch=master)](https://travis-ci.com/databricks/koalas)
```
HyukjinKwon: Can you change `Build Status` -> `Travis CI`?

dev/pytest
```diff
@@ -38,17 +38,19 @@ if [ -n "$KOALAS_USAGE_LOGGER" ]; then
...
+# The X virtual framebuffer (Xvfb) display server is used to perform all the graphical operations in memory without showing any screen output.
 if [ "$#" = 0 ]; then
     if [[ "$SPARK_VERSION" == 2.3* ]] || [[ "$SPARK_VERSION" == 2.4.1* ]] || [[ "$SPARK_VERSION" == 2.4.2* ]]; then
         # Delta requires Spark 2.4.2+. We skip the related doctests.
         if [[ "$SPARK_VERSION" == 2.3* ]]; then
-            pytest --cov=databricks -k " -melt -to_delta -read_delta" --verbose --showlocals --doctest-modules databricks "${logopts[@]}"
+            xvfb-run python -m pytest --cov=databricks -k " -melt -to_delta -read_delta" --verbose --showlocals --doctest-modules databricks "${logopts[@]}"
         else
-            pytest --cov=databricks -k "-to_delta -read_delta" --verbose --showlocals --doctest-modules databricks "${logopts[@]}"
+            xvfb-run python -m pytest --cov=databricks -k "-to_delta -read_delta" --verbose --showlocals --doctest-modules databricks "${logopts[@]}"
```
HyukjinKwon: @itholic, if we do this, the dev env requires `xvfb-run` but `xvfb-run` is only required when the display isn't available. From the offline discussions, it matters in `to_clipboard` test right? Let's just change the test if it throws the same exception or same output.

dev/pytest
```diff
@@ -38,17 +38,19 @@ if [ -n "$KOALAS_USAGE_LOGGER" ]; then
...
 
 # Runs both doctests and unit tests by default, otherwise hands arguments over to pytest.
+# we're doing pytest with `xvfb-run` since without this some of the tests would fail for GUI library because there is no display server.
+# The X virtual framebuffer (Xvfb) display server is used to perform all the graphical operations in memory without showing any screen output.
 if [ "$#" = 0 ]; then
     if [[ "$SPARK_VERSION" == 2.3* ]] || [[ "$SPARK_VERSION" == 2.4.1* ]] || [[ "$SPARK_VERSION" == 2.4.2* ]]; then
         # Delta requires Spark 2.4.2+. We skip the related doctests.
         if [[ "$SPARK_VERSION" == 2.3* ]]; then
-            pytest --cov=databricks -k " -melt -to_delta -read_delta" --verbose --showlocals --doctest-modules databricks "${logopts[@]}"
+            xvfb-run python -m pytest --cov=databricks -k " -melt -to_delta -read_delta" --verbose --showlocals --doctest-modules databricks "${logopts[@]}"
```
HyukjinKwon: @itholic, in this way, it requires `xvfb-run` to the dev env even where the display is available. Can you factor this out as an environment variable and set it in Github Actions only?

e.g.):

In this file:

```bash
PYTHON_EXECUTABLE="${PYTHON_EXECUTABLE:-python}"
...
"$PYTHON_EXECUTABLE" -m pytest --cov=databricks -k ...
```

In Github Actions, where it's required,

```bash
PYTHON_EXECUTABLE="xvfb-run python" ./dev/pytest ...
```

.github/workflows/master.yml
```diff
@@ -0,0 +1,149 @@
...
+
+jobs:
+  python35:
+    runs-on: ubuntu-latest
+    env:
+      SPARK_VERSION: 2.3.4
+      PANDAS_VERSION: 0.23.4
+      PYARROW_VERSION: 0.10.0
+      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
+      QT_QPA_PLATFORM: offscreen
```
HyukjinKwon: Can you add a comment for his variable per 66f5119?

.github/workflows/master.yml
```diff
@@ -0,0 +1,149 @@
...
+        pip install wheel
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
+        pip list
+    - name: Setup environment variables
+      run: |
+        sudo apt-get install flake8
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
+        ./dev/lint-python
```
HyukjinKwon: I think we should involve this into `Run tests`

.github/workflows/master.yml
```diff
@@ -0,0 +1,149 @@
...
+        python3 -m pip install --upgrade pip
+        pip install setuptools
+        pip install wheel
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
+        pip list
+    - name: Setup environment variables
+      run: |
+        sudo apt-get install flake8
```
HyukjinKwon: Can we move this to `Download dependencies`

.github/workflows/master.yml
```diff
@@ -0,0 +1,149 @@
...
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
```
HyukjinKwon: I think we can just remove `Setup environment variables`, and move this and the next line to `Run tests`.

.github/workflows/master.yml
```diff
@@ -0,0 +1,149 @@
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Download dependencies
+      run: ./dev/download_travis_dependencies.sh
+    - name: Install Python dependencies
+      run: |
+        sudo apt-get update
```
HyukjinKwon: Should we apt-get upate here?

HyukjinKwon:
Once conda issue is resolved and the comments above are resolved, I think it's roughly good enough to go

itholic:
@HyukjinKwon thanks for the all you commented. i'll manage them and i'm considering about using `matrix` so that we can various combinations of tests at once.

.github/workflows/master.yml
itholic: @HyukjinKwon actually we don't need the anymore since that issue is resolved by `xvfb-run`.

FYI.

https://github.com/databricks/koalas/blob/aebfd56812dd16a41e388f7a97eb1ce7530238f7/dev/pytest#L41-L42

dev/pytest
itholic: that makes sense. thanks for the comment!

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Download dependencies
+      run: ./dev/download_travis_dependencies.sh
+    - name: Install dependencies & Run tests
+      run: |
+        sudo apt-get install xclip python-pytest flake8
```
HyukjinKwon: Hm, should we install `python-pytest` from `apt-get`? Why we didn't install it from PIP?

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+    - name: Using cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Download dependencies
+      run: ./dev/download_travis_dependencies.sh
+    - name: Install dependencies & Run tests
```
HyukjinKwon: `Install dependencies & Run tests` -> `Install dependencies and run tests`

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+      run: |
+        sudo apt-get install xclip python-pytest flake8
+        python3 -m pip install --upgrade pip
+        pip install setuptools
+        pip install wheel
+        pip install -r requirements-dev.txt
+        pip install pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        pip install 'pyspark>=2.4'
```
HyukjinKwon: We download Spark from `./dev/download_travis_dependencies.sh`, and the environment variables are set properly via `./dev/pytest` -> `dev/env_setup.sh`. So, I think we don't need to install this.

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+        conda info -a
+        conda create -c conda-forge -q -n test-environment python=3.6
+        conda activate test-environment
+        conda install -c conda-forge --yes codecov
+        conda config --env --add pinned_packages python=3.6 && \
+        conda config --env --add pinned_packages pandas==$PANDAS_VERSION && \
+        conda config --env --add pinned_packages pyarrow==$PYARROW_VERSION && \
+        conda install -c conda-forge --yes pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION && \
+        conda install -c conda-forge --yes --freeze-installed --file requirements-dev.txt && \
+        conda install 'pyspark>=2.4' && \
```
HyukjinKwon: This case too.  I don't think we should install pyspark here again.

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+        bash miniconda.sh -b -p $HOME/miniconda
+        . $HOME/miniconda/etc/profile.d/conda.sh
+        hash -r
+        conda config --set always_yes yes --set changeps1 no
+        conda update -q conda
+        conda info -a
+        conda create -c conda-forge -q -n test-environment python=3.6
+        conda activate test-environment
+        conda install -c conda-forge --yes codecov
+        conda config --env --add pinned_packages python=3.6 && \
```
HyukjinKwon: I concatenated with `&&` to use if-else in Travis. We don't need to do this anymore.

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+    - name: Install dependencies & Run tests
+      run: |
+        curl -s https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh > miniconda.sh
+        bash miniconda.sh -b -p $HOME/miniconda
+        . $HOME/miniconda/etc/profile.d/conda.sh
+        hash -r
+        conda config --set always_yes yes --set changeps1 no
+        conda update -q conda
+        conda info -a
+        conda create -c conda-forge -q -n test-environment python=3.6
```
HyukjinKwon: can you keep the comments like "# Replace dep1 dep2 ... with your dependencies" from Travis?

dev/pytest
```diff
@@ -38,17 +40,20 @@ if [ -n "$KOALAS_USAGE_LOGGER" ]; then
 fi
 
 # Runs both doctests and unit tests by default, otherwise hands arguments over to pytest.
+# We're doing pytest with `xvfb-run` for CI since without this some of the tests would fail for GUI library because there is no display server.
```
HyukjinKwon: I think you should move this comments to `.github/workflows/master.yml `

.github/workflows/master.yml
itholic: Because it was not working with 'not installed' error message when we test `Install dependencies` and `Run tests` separately.

but now we integrated them, so i'm testing again. if not necessary, i'll remove it from apt-get install list.

HyukjinKwon: `flake8` too. It's from `requirements-dev.txt`.

itholic: you're right, so i'm testing both of them

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+      with:
+        java-version: 1.8
+    # Setup Python 3.5 via `apt-get install` since the dafault Python3.5 from `actions/setup-python`
+    # seems to have some problem with Tkinter, so we should manually install the python3.5-tk.
+    # for this, we should use the Python we manually installed, not the default one from `actions/setup-python`
+    - name: Setup Python 3.5
+      run: |
+        sudo add-apt-repository ppa:deadsnakes/ppa
+        sudo apt-get update
+        sudo apt-get install tk-dev python3.5-tk python3 python3.5 python-pip python3-pip
```
HyukjinKwon: Why do we need to install `python3` and `python-pip`?  Also, does `python3-pip` guarantees to install Python 3.5's PIP? If we're not sure,  you can just try `/usr/bin/python3.5 -m ensurepip` instead.

.github/workflows/master.yml
```diff
@@ -0,0 +1,147 @@
...
+    - uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+    # Setup Python 3.5 via `apt-get install` since the dafault Python3.5 from `actions/setup-python`
+    # seems to have some problem with Tkinter, so we should manually install the python3.5-tk.
+    # for this, we should use the Python we manually installed, not the default one from `actions/setup-python`
+    - name: Setup Python 3.5
+      run: |
+        sudo add-apt-repository ppa:deadsnakes/ppa
+        sudo apt-get update
```
HyukjinKwon: Why should we apt-get update?

.github/workflows/master.yml
itholic: it was raise `no module names 'pyspark' `without this line, but i'll check again in more detail. thanks !

.github/workflows/master.yml
itholic: ah, it's my mistake. thanks :)

.github/workflows/master.yml
itholic: okay i'll add them. thanks!

.github/workflows/master.yml
HyukjinKwon: We should probably merge `./dev/download_travis_dependencies.sh` line to `Install dependencies & Run tests
` too.

.github/workflows/master.yml
```diff
@@ -0,0 +1,156 @@
...
+
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+    branches: 
+      - master
+
+# We're doing pytest with `xvfb-run` for CI since without this some of the tests would fail for GUI library because there is no display server.
```
HyukjinKwon: Can you move this comments right above `PYTHON_EXECUTABLE`?  Also, please keep these comments with the 100 char length to make it easier to line.

.github/workflows/master.yml
```diff
@@ -0,0 +1,156 @@
...
+    - name: Using cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Download dependencies
+      run: ./dev/download_travis_dependencies.sh
+    - name: Install dependencies & Run tests
```
HyukjinKwon: Here too  -> `Install dependencies and run tests`

.github/workflows/master.yml
```diff
@@ -0,0 +1,156 @@
...
+    - name: Update Codecov
+      run: bash <(curl -s https://codecov.io/bash)
+
+  python37:
+    runs-on: ubuntu-latest
+    env:
+      SPARK_VERSION: 2.4.4
+      PANDAS_VERSION: 0.25.3
+      PYARROW_VERSION: 0.14.1
+      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
```
HyukjinKwon: We dont need to test logger in python37. It's enough to test only in Python 3.6.

.github/workflows/master.yml
```diff
@@ -0,0 +1,156 @@
...
+# If you want to run your test with Xvfb, set the environment variable `PYTHON_EXECUTABLE` to `xvfb-run python`.
+jobs:
+  python35:
+    runs-on: ubuntu-latest
+    env:
+      SPARK_VERSION: 2.3.4
+      PANDAS_VERSION: 0.23.4
+      PYARROW_VERSION: 0.10.0
+      PYTHON_EXECUTABLE: xvfb-run python
+      KOALAS_USAGE_LOGGER: databricks.koalas.usage_logging.usage_logger
```
HyukjinKwon: Remove this one too.

.github/workflows/master.yml
itholic: Thanks for the comment!

it was fixed by adding the path of packages are installed by PyPI to the `PATH` using `add-path`.

dev/pytest
```diff
@@ -15,6 +15,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+
+# We're doing pytest with `xvfb-run` for CI since without this some of the tests would fail
+# for GUI library because there is no display server.
+# The X virtual framebuffer(Xvfb) display server is used to perform all the graphical operations
+# in memory without showing any screen output.
+# If you want to run test with Xvfb, set environment `PYTHON_EXECUTABLE` to `xvfb-run python`.
```
HyukjinKwon: I think you can remove these comments for now. Dev env usually has display available, and it's mostly only useful in running CI.

.github/workflows/master.yml
```diff
@@ -0,0 +1,150 @@
...
+    # Setup Python 3.5 via `apt-get install` since the dafault Python3.5 from `actions/setup-python`
+    # seems to have some problem with Tkinter, so we should manually install the python3.5-tk.
+    # for this, we should use the Python we manually installed, not the default one from `actions/setup-python`
+    - name: Setup Python 3.5
+      run: |
+        sudo add-apt-repository ppa:deadsnakes/ppa
+        sudo apt-get install tk-dev python3.5-tk python3.5
+        sudo rm /usr/bin/python
+        sudo ln -s /usr/bin/python3.5 /usr/bin/python
+        # Add the path which packages are actually installed to `PATH`
```
HyukjinKwon: Can you clarify this `echo "::add-path::/home/runner/.local/bin"` was required for pip installed executables to be in the path?

.github/workflows/master.yml
```diff
@@ -0,0 +1,150 @@
...
+    - name: Update Codecov
+      run: bash <(curl -s https://codecov.io/bash)
+
+  python36:
+    runs-on: ubuntu-latest
+    env:
+      SPARK_VERSION: 2.4.4
+      PANDAS_VERSION: 0.24.2
+      PYARROW_VERSION: 0.13.0
+      PYTHON_EXECUTABLE: xvfb-run python
```
HyukjinKwon: For Python 3.6 and 3.7, `DISPLAY=0.0` works, right? Let's don't use `xvfb-run` for Python 3.6 and 3.7. `xvfb-run` is a bit overkill.

dev/pytest
```diff
@@ -15,6 +15,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+PYTHON_EXECUTABLE="${PYTHON_EXECUTABLE:-python}"
+
+
```
HyukjinKwon: nit: let's remove one newline here.

.github/workflows/master.yml
```diff
@@ -0,0 +1,150 @@
...
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Download dependencies
+      run: ./dev/download_travis_dependencies.sh
+    - name: Install dependencies & run tests
+      run: |
+        sudo apt-get install xclip
+        python -m pip install --upgrade pip
```
HyukjinKwon: Why do we need to upgrade pip?

.github/workflows/master.yml
```diff
@@ -0,0 +1,150 @@
...
+        python -m pip install -r requirements-dev.txt
```
HyukjinKwon: now `pip` is in the path. Can we remove `python -m`?

.github/workflows/master.yml
```diff
@@ -0,0 +1,150 @@
...
+        java-version: 1.8
+    - name: Using cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements-dev.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    - name: Download dependencies
+      run: ./dev/download_travis_dependencies.sh
```
HyukjinKwon: Let's just move this into "Install dependencies & run tests" and remove "Download dependencies".

.github/workflows/master.yml
```diff
@@ -0,0 +1,150 @@
...
+        conda config --env --add pinned_packages pandas==$PANDAS_VERSION
+        conda config --env --add pinned_packages pyarrow==$PYARROW_VERSION
+        conda install -c conda-forge --yes pandas==$PANDAS_VERSION pyarrow==$PYARROW_VERSION
+        conda install -c conda-forge --yes --freeze-installed --file requirements-dev.txt
+        conda list
+        export SPARK_HOME="$HOME/.cache/spark-versions/spark-$SPARK_VERSION-bin-hadoop2.7"
+        ./dev/lint-python
+        ./dev/pytest
+    - name: Update Codecov
+      run: bash <(curl -s https://codecov.io/bash)
```
HyukjinKwon: Can you add a comment from existing travis CI "# Push the results back to codecov"

.github/workflows/master.yml
itholic: okay i'll fix it to work dynamically. thanks 👍

.github/workflows/master.yml
itholic: if then, using `/usr/bin/pip`.

maybe i think we better keep `python -m` since the PyPI official documentation is recommending that way. (https://packaging.python.org/guides/installing-using-pip-and-virtual-environments/)

itholic: yes we can! thanks 😄

itholic:
Applied matrix to integrate python 3.6 & 3.7 tests.

harupy:
> But when I open a PR from my forked repository to the upstream repository, the Github Action is not being triggered on the upstream repository.

https://github.community/t5/GitHub-Actions/Github-Workflow-not-running-from-pull-request-from-forked/td-p/32979

itholic:
@harupy Thanks for the report! :smiley: 

Maybe we should delay merging this PR until that issue solved.

I'll test on my repository and try out to find the solution. Thanks :D

PR #1256, opened 2020-02-03
Raising NotImplementedError for elements that actually are not implemented.

itholic:
we're raising `ValueError` for some 'NotImplemented' elements.

i think it's better to raise `NotImplementedError` rather than `ValueError` for those elements so that we can find more easily when we want to implement them.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -6123,7 +6123,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         if isinstance(other, ks.Series):
             raise ValueError("DataFrames.append() does not support appending Series to DataFrames")
         if sort:
-            raise ValueError("The 'sort' parameter is currently not supported")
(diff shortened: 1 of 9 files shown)


itholic:
while i'm here, i find more cases.

databricks/koalas/frame.py
```diff
@@ -5458,9 +5458,10 @@ def sort_index(self, axis: int = 0,
         b 1  0  3
         """
         if axis != 0:
-            raise ValueError("No other axes than 0 are supported at the moment")
+            raise NotImplementedError("No other axis than 0 are supported at the moment")
```
ueshin: Should we use `validate_axis` and support `'index'` as well while we are here?

itholic: yeah, we'd better do that. i'll check if another place to add, too.

thanks for the comment! :)

ueshin:
Thanks! merging.

PR #1257, opened 2020-02-06
Apply util function to `DataFrame._assign`.

ueshin:
We can apply `_InternalFrame.with_new_columns` util function in `DataFrame._assign` as well.


ueshin:
Thanks! merging.

PR #1258, opened 2020-02-07
Remove workarounds in `_InternalFrame.scol_for` and `.column_name_for`.

ueshin:
Now that we can remove workarounds added previously in `_InternalFrame.scol_for` and `.column_name_for`.
Now they only allow tuples for columns or strings as index column names to avoid misuse of these methods.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -622,25 +622,27 @@ class _InternalFrame(object):
     def _column_index_to_name(self) -> Dict[Tuple[str, ...], str]:
         return dict(zip(self.column_index, self.data_columns))
 
-    def column_name_for(self, column_name_or_index: Union[str, Tuple[str, ...]]) -> str:
+    def column_name_for(self, column_index_or_index_column: Union[str, Tuple[str, ...]]) -> str:
         """ Return the actual Spark column name for the given column name or index. """
-        if column_name_or_index in self._column_index_to_name:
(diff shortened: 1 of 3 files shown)


databricks/koalas/internal.py
```diff
@@ -622,25 +622,27 @@ def offset(id):
     def _column_index_to_name(self) -> Dict[Tuple[str, ...], str]:
         return dict(zip(self.column_index, self.data_columns))
 
-    def column_name_for(self, column_name_or_index: Union[str, Tuple[str, ...]]) -> str:
+    def column_name_for(self, column_index_or_index_column: Union[str, Tuple[str, ...]]) -> str:
```
HyukjinKwon: BTW, is it a proper term used in pandas? Column index and index column .... seems very confusing :-)..

ueshin: Agree! 😄 
hm, should we use "label" instead?

https://pandas.pydata.org/pandas-docs/stable/reference/frame.html

> DataFrame.index | The index (row labels) of the DataFrame.
> DataFrame.columns | The column labels of the DataFrame.

HyukjinKwon: sounds good!

ueshin: maybe let me change it in a separate PR?
I guess we should rename others as well, like `column_index` -> `column_labels`?

HyukjinKwon: no problem

PR #1259, opened 2020-02-07
Implement DataFrame.apply

### LINKED ISSUE #65: UDF using DataFrame.apply
These are the patterns that I have seen with the apply() function:
```
df = pd.DataFrame({'a': [1, 2, 3, 4, 5, 6, 7], 'b': [7, 6, 5, 4, 3, 2, 1]})

# Option 1: Use a series and apply on it to run a UDF (Technically a Series.apply)
df['c'] = df['a'].apply(lambda x: x*2)

# Option 2: Use the entire row from a dataframe
df['d'] = df.apply(lambda x: x.a*x.b, axis=1)

# Option 3: Use some columns in a row
df['e'] = df[['a', 'b']].apply(lambda x: x.a*x.b, axis=1)

# Option 4: Apply aggregates on every column to create a new dataframe
new_df = df.apply(lambda x: x.sum())

# Option 5: Apply aggregates but return multiple values - here it adjusts the shorted columns with NaNs
new_df = df.apply(lambda x: x[:x[0]])
```

### LINKED ISSUE #1228: DataFrame.apply(func, axis=1)
I'm missing the access of the whole row in the DataFrame.apply function, like

```python
df = pd.DataFrame({"date":pd.to_datetime(['2012-01-01','2012-01-02']),
              "hour": [5, 5], 
              })

print(df)

df['datetime'] = df.apply(lambda x: x['date'].replace(hour=x['hour']), axis=1)

print(df)
```

Prints out:
```bash
        date  hour
0 2012-01-01     5
1 2012-01-02     5

        date  hour            datetime
0 2012-01-01     5 2012-01-01 05:00:00
1 2012-01-02     5 2012-01-02 05:00:00
```

We should implement `DataFrame.apply(row_func, axis=1)` using the UDFs.

HyukjinKwon:
This PR proposes to implement `DataFrame.apply` with both `axis` 0 and 1. Note that, `DataFrame.apply(..., axis=1)` with global aggregations is impossible.

It can be tested with the examples below:

```python
import numpy as np

df = ks.DataFrame([[4, 9]] * 10, columns=['A', 'B'])

df.apply(np.sqrt, axis=0)

def sqrt(x) -> ks.Series[float]:
    return np.sqrt(x)
df.apply(sqrt, axis=0)


df.apply(np.sum, axis=1)

def summation(x) -> int:
   return np.sum(x)

df.apply(summation, axis=1)
```

Basically the approach is using group map Pandas UDF by grouping by partitions.

```python
from pyspark.sql.functions import pandas_udf, PandasUDFType
from pyspark.sql import functions as F

df = spark.createDataFrame(
    [(1, 1.0), (1, 2.0), (2, 3.0), (2, 5.0), (2, 10.0)],
    ("id", "v"))

@pandas_udf("id long, v double", PandasUDFType.GROUPED_MAP)
def func(pdf):
    return pdf.apply(...)

df.groupby(F.spark_partition_id()).apply(func).show()
```

Resolves #1228
Resolves #65

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -879,8 +879,10 @@ class GroupBy(object):
             kdf = DataFrame(pdf)
             return_schema = kdf._sdf.drop(*HIDDEN_COLUMNS).schema
 
-        sdf = self._spark_group_map_apply(
+        sdf = GroupBy._spark_group_map_apply(
+            self._kdf,
             lambda pdf: pdf.groupby(input_groupnames).apply(func),
+            self._groupkeys_scols,
             return_schema,
             retain_index=should_infer_schema)
 
@@ -938,17 +940,18 @@ class GroupBy(object):
         def pandas_filter(pdf):
             return pdf.groupby(groupby_names).filter(func)
 
-        sdf = self._spark_group_map_apply(
-            pandas_filter, data_schema, retain_index=True)
+        sdf = GroupBy._spark_group_map_apply(
+            self._kdf, pandas_filter, self._groupkeys_scols, data_schema, retain_index=True)
         return DataFrame(self._kdf._internal.copy(
             sdf=sdf,
             column_scols=[scol_for(sdf, col) for col in self._kdf._internal.data_columns]))
 
-    def _spark_group_map_apply(self, func, return_schema, retain_index):
-        index_columns = self._kdf._internal.index_columns
-        index_names = self._kdf._internal.index_names
-        data_columns = self._kdf._internal.data_columns
-        column_index = self._kdf._internal.column_index
+    @staticmethod
+    def _spark_group_map_apply(kdf, func, groupkeys_scols, return_schema, retain_index):
+        index_columns = kdf._internal.index_columns
+        index_names = kdf._internal.index_names
+        data_columns = kdf._internal.data_columns
+        column_index = kdf._internal.column_index
 
         def rename_output(pdf):
             # TODO: This logic below was borrowed from `DataFrame.pandas_df` to set the index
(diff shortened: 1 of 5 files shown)


patryk-oleniuk:
Great!
> Note that, DataFrame.apply(..., axis=1) with global aggregations is impossible.
Could you tell what do you exactly mean by that? 

Is access to each element of the row available with axis=1?
Can I do this? 

```python
df = ks.DataFrame([[4, 9]] * 10, columns=['param1', 'param2'])

def my_func( x ) -> int:
  # x should be a row here 
  return run_regression_and_log_to_mlflow(x['param1'], x['param2'])

#should return a series of ints
df.apply(summation, axis=1) 
```

Thanks!

databricks/koalas/groupby.py
```diff
@@ -938,17 +940,18 @@ def filter(self, func):
...
             sdf=sdf,
             column_scols=[scol_for(sdf, col) for col in self._kdf._internal.data_columns]))
 
-    def _spark_group_map_apply(self, func, return_schema, retain_index):
-        index_columns = self._kdf._internal.index_columns
-        index_names = self._kdf._internal.index_names
-        data_columns = self._kdf._internal.data_columns
-        column_index = self._kdf._internal.column_index
+    @staticmethod
+    def _spark_group_map_apply(kdf, func, groupkeys_scols, return_schema, retain_index):
```
ueshin: Just out of curiosity, why did you make it `@staticmethod`?

ueshin: nvm, I see the `self` is not `GroupBy` anymore.

HyukjinKwon:
@patryk-oleniuk, yup, that works:

```python
df = ks.DataFrame([[4, 9]] * 10, columns=['param1', 'param2'])

def my_func( x ) -> int:
  # x should be a row here 
  print(x)
  return 1

#should return a series of ints
df.apply(my_func, axis=1) 
```

```
param1    4
param2    9
Name: 5, dtype: int64
param1    4
param2    9
Name: 5, dtype: int64
...
```


```
0    1
1    1
2    1
3    1
4    1
5    1
6    1
7    1
8    1
9    1
Name: 0, dtype: int32
```

HyukjinKwon:
Thanks, @ueshin. Merged!

PR #1262, opened 2020-02-07
Add a util method to verify temporary column names.

ueshin:
This is a follow-up of #1221 and #1223.
Adding a util method to verify temporary column names.


databricks/koalas/utils.py
```diff
@@ -411,6 +411,50 @@ def validate_axis(axis=0, none_axis=0):
...
+    +---+-----------+
+    |  a|          0|
+    |  b|          0|
+    |  c|          0|
+    +---+-----------+
+
+    >>> temp_column_name(sdf, 'tmp')
+    '__tmp_0__'
+    >>> temp_column_name(sdf, 'dummy')
+    '__dummy_1__'
```
itholic: cool.

itholic:
LGTM

databricks/koalas/utils.py
```diff
@@ -411,6 +411,50 @@ def validate_axis(axis=0, none_axis=0):
     return {None: none_axis, 'index': 0, 'columns': 1}.get(axis, axis)
 
 
+def temp_column_name(sdf: spark.DataFrame, gen: Union[str, Callable[[int], str]]) -> str:
```
HyukjinKwon: @ueshin do you plan to use `temp_column_name(sdf, '{}'.format)` style very soon? Seems currently not being used, and might be best to just remove.

databricks/koalas/utils.py
```diff
@@ -411,6 +411,50 @@ def validate_axis(axis=0, none_axis=0):
...
+    |  b|  0|
+    |  c|  0|
+    +---+---+
+    >>> temp_column_name(sdf, '{}'.format)
+    '2'
+    """
+    if isinstance(gen, str):
+        base = gen
+        gen = lambda i: '__{}_{}__'.format(base, i)
+    for i in range(len(sdf.columns) + 1):
```
HyukjinKwon: I am still less sure about numbering based upon one DataFrame. If we're really worried about the name being conflicted (and, for instance, `F.col(...)` doesn't work), we should make the column globally unique by using a global variable or things like UUID.

databricks/koalas/utils.py
ueshin: ah, yeah, it's from the original idea and I don't use it anymore. We can remove it.

databricks/koalas/utils.py
ueshin: So far I'm not worried about a global conflicts but only breaking the existing columns.

HyukjinKwon:
> I don't think we should rely on manpower.
I think we should have some mechanism for the case where we "leak" such columns.

If it doesn't throw an exception and just generate the column names when it conflicts, we will never notice when such "leak" happens ... I think any issue by such field conflict was found yet. It wouldn't be too late to think when somebody actually faces this issue.

ueshin:
Even now Koalas doesn't throw an exception but rewrite such columns silently, and it will cause Koalas to behave strangely.
Actually this PR is trying to allow the "leak" so that we won't need to worry about it.

Or, do we want to rather throw an exception if the column name we want to add conflicts?

Anyways, I think we need something to allow or detect such a case.

HyukjinKwon:
> Or, do we want to rather throw an exception if the column name we want to add conflicts?

Yes.. probably that's what I am thinking for now, and think about doing the name generation if users actually face such errors...

HyukjinKwon:
> Even now Koalas doesn't throw an exception but rewrite such columns silently, and it will cause Koalas to behave strangely.

Yup, I agree it could cause a correctness issue if this happens. Is there an actual case found at your side? Maybe I am missing something if this is already happening.

ueshin:
>> Or, do we want to rather throw an exception if the column name we want to add conflicts?
>
> Yes.. probably that's what I am thinking for now, and think about doing the name generation if users actually face such errors...

I'm fine with throwing an exception and let's see.

> Is there an actual case found at your side?

Actually there is not an issue yet, but we can easily expect it could happen in the future, so I want to set something to make it work or detect it.

HyukjinKwon:
Maybe if we can throw an exception first .. and change it to generate names if it's actually reported multiple times ...

HyukjinKwon:
Yea.. let's throw an exception first if you don't mind.. and consider changing it to generate names later when it's actually reported.

ueshin:
Sure, I updated to throw an exception.

HyukjinKwon:
Thanks @ueshin.

PR #1267, opened 2020-02-09
Cache downloaded Spark, and don't cache Conda

HyukjinKwon:
We shouldn't install miniconda if already cached. Currently Github Actions fails as below when it's cached.

```
ERROR: File or directory already exists: '/home/runner/miniconda'
If you want to update an existing installation, use the -u option.
##[error]Process completed with exit code 1.
```

Seems it's a bit complicated to cache Conda usage. I excluded for now to make the builds pass.

While I am here, I changes to cache downloaded Spark as well.


HyukjinKwon:
I am going to merge it to unblock other PRs.

PR #1270, opened 2020-02-10
'Add not support api to docs

itholic:


itholic:
I roughly listed-up once and added them to FAQ section.

If there is a missing elements or any idea for improving this documentation, just let me know so that i can add to list while i'm here.

docs/source/user_guide/faq.rst
```diff
@@ -62,3 +62,52 @@ lot longer (in the order of days)
...
+
+Koalas doesn't support several APIs that may cause memory issues mostly due to the size of the data.
+
+For example, implementing and using `DataFrame.values` in Koalas can cause all data belonging to the
+
+DataFrame to be loaded into the driver's memory, causing memory errors like OOM.
+
+The following is a list of APIs that Koalas doesn't plan to support.
+
+- DataFrame
```
HyukjinKwon: Can we automatically generate this list and the APIs we didn't implement yet?

itholic: okay let me try,

and IMO, i think we'd better separate the list into `Not yet implemented` and `Not will be implemented`.

PR #1273, opened 2020-02-12
Implement DataFrame.query

itholic:
This PR proposes to implement `DataFrame.query`

```python
>>> df = ks.DataFrame({'A': range(1, 6),
...                    'B': range(10, 0, -2),
...                    'C C': range(10, 5, -1)})
>>> df
   A   B  C C
0  1  10   10
1  2   8    9
2  3   6    8
3  4   4    7
4  5   2    6

>>> df.query('A > B')
   A  B  C C
4  5  2    6

The previous expression is equivalent to

>>> df[df.A > df.B]
   A  B  C C
4  5  2    6

For columns with spaces in their name, you can use backtick quoting.

>>> df.query('B == `C C`')
   A   B  C C
0  1  10   10

The previous expression is equivalent to

>>> df[df.B == df['C C']]
   A   B  C C
0  1  10   10
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8345,17 +8345,17 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
               you can just workaround by using :meth:`DataFrame.map_in_pandas`
               although it is less performant. See the example below.
 
-            >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
-            >>> df
-               A  B
-            0  1  2
-            1  3  4
-            2  5  6
-            >>> num = 1
-            >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))  # doctest: +SKIP
-               A  B
-            1  3  4
-            2  5  6
+                >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
+                >>> df
+                   A  B
+                0  1  2
+                1  3  4
+                2  5  6
+                >>> num = 1
+                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))  # doctest: +SKIP
+                   A  B
+                1  3  4
+                2  5  6
 
         Parameters
         ----------
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8390,6 +8390,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
            A   B  C C
         0  1  10   10
         """
+        if isinstance(self.columns, pd.MultiIndex):
+            raise ValueError("Doesn't support for MultiIndex columns")
         if not isinstance(expr, str):
             raise ValueError(
                 'expr must be a string to be evaluated, {} given'
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8352,7 +8352,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             1  3  4
             2  5  6
             >>> num = 1
-            >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
+            >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))  # doctest: +SKIP
                A  B
             1  3  4
             2  5  6
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8335,6 +8335,14 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         Query the columns of a DataFrame with a boolean expression.
 
+        .. note::
+            * MultiIndex columns are not supported.
+            * If you want to use MultiIndex columns, you can just workaround
+              by using `DataFrame.apply` although it is less performance.
+            * Internal columns that starting with __.. are able to access,
+              however, they are not supposed to be accessed.
+            * This delegates to Spark SQL so the syntax follows Spark SQL
+
         Parameters
         ----------
         expr : str
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8336,14 +8336,12 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         Query the columns of a DataFrame with a boolean expression.
 
         .. note::
-            * MultiIndex columns are not supported.
             * Using variables in the environment with `@` is not supported.
             * Internal columns that starting with __. are able to access,
               however, they are not supposed to be accessed.
             * This delegates to Spark SQL so the syntax follows Spark SQL
-            * If you want to use MultiIndex columns, or variables in the environment,
-              you can just workaround by using :meth:`DataFrame.map_in_pandas`
-              although it is less performant. See the example below.
+            * If you want the exactly same syntax with pandas' you can work around
+              by using :meth:`DataFrame.map_in_pandas`. See the example below.
 
                 >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
                 >>> df
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8338,8 +8338,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         .. note::
             * MultiIndex columns are not supported.
             * If you want to use MultiIndex columns, you can just workaround
-              by using `DataFrame.apply` although it is less performance.
-            * Internal columns that starting with __.. are able to access,
+              by using :meth:`DataFrame.apply` although it is less performant.
+            * Internal columns that starting with __. are able to access,
               however, they are not supposed to be accessed.
             * This delegates to Spark SQL so the syntax follows Spark SQL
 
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,82 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
...
+        if not isinstance(expr, str):
+            raise ValueError(
+                'expr must be a string to be evaluated, {} given'
+                .format(type(expr)))
+        if not isinstance(inplace, bool):
+            raise ValueError(
+                'For argument "inplace" expected type bool, received type {}.'
+                .format(type(inplace).__name__))
+
+        sdf = self._sdf.filter(expr)
```
HyukjinKwon: @itholic how does pandas handle multi-index columns?

itholic: let me test!

itholic: i just tested, they seems not support for multi-index columns.

i added the code for catching that exception.

HyukjinKwon: @itholic can you add the limitation of this API in the docstring? You could use `.. note::`.

- multiindex columns are not supported.
- internal columns that starting with `__`.. are able to access; however, they are not supposed to be accessed.
- this delegates to Spark SQL so the syntax follows Spark SQL

HyukjinKwon: Also, adds another note that, if you want to use multidex columns, people can easily work around by using `DataFrame.apply`.

itholic: yes, i got it. thanks!! 👍

itholic: > internal columns that starting with __.. are able to access; however, they are not supposed to be accessed.

I'm curious about this; if we add note about this, then should we add this to all APIs since all APIs are using internal columns??

databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,92 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
...
         return DataFrame(internal) if not result_as_series else DataFrame(internal).T[key]
 
+    def query(self, expr, inplace=False):
+        """
+        Query the columns of a DataFrame with a boolean expression.
+
+        .. note::
+            * MultiIndex columns are not supported.
+            * If you want to use MultiIndex columns, you can just workaround
+              by using `DataFrame.apply` although it is less performance.
```
HyukjinKwon: `` `DataFrame.apply` `` -> `` :meth:`DataFrame.apply` ``. Can you also check the generated doc?

databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,92 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
...
+            * Internal columns that starting with __.. are able to access,
```
HyukjinKwon: There's two dots.

databricks/koalas/frame.py
HyukjinKwon: `less performance` -> `less performant`

databricks/koalas/tests/test_dataframe.py
```diff
@@ -2397,3 +2397,50 @@ def test_mask(self):
...
+                    ValueError,
+                    'For argument "inplace" expected type bool, received type {}.'
+                    .format(type(inplace).__name__)):
+                kdf.query('a < b', inplace=inplace)
+
+        # doesn't support for MultiIndex columns
+        columns = pd.MultiIndex.from_tuples([('A', 'Z'), ('B', 'X'), ('C', 'C')])
+        kdf.columns = columns
+        with self.assertRaisesRegex(ValueError, "Doesn't support for MultiIndex columns"):
+            kdf.query("('A', 'Z') > ('B', 'X')")
```
HyukjinKwon: Wait, is it right syntax in pandas? Seems pandas [doesn't support multi-index column yet] (https://github.com/pandas-dev/pandas/issues/13348)

databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,92 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
 
         return DataFrame(internal) if not result_as_series else DataFrame(internal).T[key]
 
+    def query(self, expr, inplace=False):
+        """
+        Query the columns of a DataFrame with a boolean expression.
+
```
HyukjinKwon: @itholic, can you elaborate what works and what doesn't work based on https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#the-query-method ?

Also, actually sorry I think `DataFrame.apply` doesn't work as a workaround as it iterates each series.

I will make a API like [`map_partitions`](https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.map_partitions) soon which can actually work around. I think you should include the workaround example as well.

databricks/koalas/tests/test_dataframe.py
itholic: they're supporting both

```python
>>> d
  L1
   A  B
0  0  0
1  4  1
2  4  1
3  0  3

>>> d.L1.A
0    0
1    4
2    4
3    0
Name: A, dtype: int64

>>> d[('L1', 'A')]
0    0
1    4
2    4
3    0
Name: (L1, A), dtype: int64
```

databricks/koalas/frame.py
itholic: > I think you should include the workaround example as well.

Totally agree. we'd better this for all docstring of other APIs which can work around.

databricks/koalas/tests/test_dataframe.py
itholic: > Seems pandas doesn't support multi-index column yet

and yes, pandas doesn't support MultiIndex columns at least for `query`.

databricks/koalas/frame.py
HyukjinKwon: @itholic, here's the PR https://github.com/databricks/koalas/pull/1276. This will be able to workaround (see the example in the docstring).

itholic: @HyukjinKwon 

Thanks for the comment!! i just added :)

databricks/koalas/frame.py
```diff
@@ -8337,11 +8337,25 @@ def query(self, expr, inplace=False):
...
-              by using :meth:`DataFrame.apply` although it is less performant.
+            * Using variables in the environment with `@` is not supported.
             * Internal columns that starting with __. are able to access,
               however, they are not supposed to be accessed.
             * This delegates to Spark SQL so the syntax follows Spark SQL
+            * If you want to use MultiIndex columns, or variables in the environment,
+              you can just workaround by using :meth:`DataFrame.map_in_pandas`
+              although it is less performant. See the example below.
+
+            >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
```
HyukjinKwon: I think it should have four spaces more. Does it render okay in the doc?

itholic: yeah i checked, and it was looks fine to me.

스크린샷 2020-02-14 오후 12 55 15

but i just tested with 4 more indent as you mentioned, it looks better.

스크린샷 2020-02-14 오후 12 58 57

itholic: fixed! thanks for the comment :)

databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,106 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
...
+        """
+        Query the columns of a DataFrame with a boolean expression.
+
+        .. note::
+            * MultiIndex columns are not supported.
+            * Using variables in the environment with `@` is not supported.
+            * Internal columns that starting with __. are able to access,
+              however, they are not supposed to be accessed.
+            * This delegates to Spark SQL so the syntax follows Spark SQL
+            * If you want to use MultiIndex columns, or variables in the environment,
```
HyukjinKwon: > If you want to use MultiIndex columns

I don't think pandas supports this, right? You can just describe "if you want the exactly same syntax with pandas' you can work around by using ...`

databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,106 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
 
         return DataFrame(internal) if not result_as_series else DataFrame(internal).T[key]
 
+    def query(self, expr, inplace=False):
+        """
+        Query the columns of a DataFrame with a boolean expression.
+
+        .. note::
+            * MultiIndex columns are not supported.
```
HyukjinKwon: Since pandas doesn't support this case, let's remove it.

itholic:
@HyukjinKwon Thanks for the review!

databricks/koalas/frame.py
```diff
@@ -8331,6 +8331,104 @@ def quantile(self, q=0.5, axis=0, numeric_only=True, accuracy=10000):
...
+              by using :meth:`DataFrame.map_in_pandas`. See the example below.
+
+                >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
+                >>> df
+                   A  B
+                0  1  2
+                1  3  4
+                2  5  6
+                >>> num = 1
+                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))  # doctest: +SKIP
```
HyukjinKwon: @itholic can you enable this since `map_in_pandas` is merged?

itholic: @HyukjinKwon sure, i'm waiting for!

PR #1276, opened 2020-02-13
Implement DataFrame.map_in_pandas

HyukjinKwon:
This PR implements a Koalas specific API in order to directly use pandas DataFrame APIs.

For pandas Series APIs, we have a workaround for instance:

```python
>>> ks.range(10)[['id']].apply(lambda series: series)['id']
0    0
1    1
2    2
3    3
4    4
5    5
6    6
7    7
8    8
9    9
Name: id, dtype: int64
```

however, there is no way to use the APIs in the pandas DataFrame.

There is a similar API called [`map_partitions`](https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.map_partitions) in Dask and this API is inspired by it.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1801,6 +1801,137 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
     T = property(transpose)
 
+    def map_in_pandas(self, func):
+        """
+        Map a function that takes pandas DataFrame and outputs pandas DataFrame. The pandas
+        DataFrame given to the function is of a batch used internally.
+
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
@ueshin, let me merge this for now. The implementation is similar with `DataFrame.apply` so there won't be too many things to closely review.

Also, this PR has a reference from Dask, it allows many workarounds for example:

```python
        >>> def take_func(pdf) -> ks.DataFrame[int, int]:
        ...     return pdf.query('A == 1')
        >>> df.map_in_pandas(take_func)
```

Let me know if you have any concern!

PR #1278, opened 2020-02-13
Move the way to create Series with a proper label into DataFrame.

ueshin:
We can skip `loc` indexer when the label is apparently resolvable.
Note that we should use the method only if the label is related to its `_InternalFrame`.


databricks/koalas/frame.py
```diff
@@ -484,10 +484,16 @@ def calculate_columns_axis(*cols):
         else:
             raise ValueError("No axis named %s for object type %s." % (axis, type(axis)))
 
+    def _kser_for(self, label):
```
HyukjinKwon: @ueshin, I manually debugged the code path being executed. Which code path is the concern for this shortcut? and, can we move this shortcut into `DataFrame.__getitem__` or `LocIndexer`?

HyukjinKwon: @ueshin, out of curiosity, which code path was the concern this shortcut avoids specifically? and ..  are we just able to move this shortcut into `DataFrame.__getitem__` or `LocIndexer`?

ueshin: `LocIndexer._get_from_multiindex_column` could be an overhead, especially for deep level labeled columns.
As for the movement, we need to look up the column if the input is from users, so I don't think we should move this into `DataFrame.__getitem__` or `LocIndexer`.

HyukjinKwon: Hm .. maybe I missed something but the number of columns are expected to be small. Wouldn't it the performance improvement is trivial? I was wondering if it's worthwhile having a shortcut which makes a bit difficult to read. I myself would be confused when to use `self._kser_for(label)` vs `self[label]`.

ueshin: Probably the PR title is confusing you?
This PR is just moving the way to create Series with a proper label into DataFrame.
Basically we can just use `self._kser_for(label)`, and if the label is from user input, we should use `self[label]`.

PR #1281, opened 2020-02-14
Add util function validate_bool_kwarg

itholic:
There is some mismatch between pandas and Koalas when using `inplace` parameter like the below.

- pandas (only allows `True` or `False`)
```python
>>> pdf
   A    B    C  D
0  1  2.0  NaN  0
1  1  4.0  NaN  1
2  2  NaN  NaN  5
3  2  3.0  1.0  4

>>> pdf.fillna('A', inplace=1)
...
ValueError: For argument "inplace" expected type bool, received type int.
```

- Koalas (allows any type)
```python
>>> kdf
   A    B    C  D
0  1  2.0  NaN  0
1  1  4.0  NaN  1
2  2  NaN  NaN  5
3  2  3.0  1.0  4

>>> kdf.fillna('A', inplace=1)
>>> kdf
   A    B    C  D
0  1  2.0    A  0
1  1  4.0    A  1
2  2    A    A  5
3  2  3.0  1.0  4
```

i think we should fix them to make work same with pandas.

so, i think we'd better have an util function `validate_bool_kwarg` for this (and also pandas has same one)

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -43,8 +43,7 @@ from databricks.koalas.missing.groupby import _MissingPandasLikeDataFrameGroupBy
     _MissingPandasLikeSeriesGroupBy
 from databricks.koalas.series import Series, _col
 from databricks.koalas.config import get_option
-from databricks.koalas.utils import (column_labels_level, scol_for, name_like_string,
-                                     validate_bool_kwarg)
+from databricks.koalas.utils import column_labels_level, scol_for, name_like_string
 from databricks.koalas.window import RollingGroupby, ExpandingGroupby
 
 # to keep it the same as pandas
@@ -1299,7 +1298,7 @@ class GroupBy(object):
         """
         return self._fillna(value, method, axis, inplace, limit)
 
-    def bfill(self, inplace=False, limit=None):
+    def bfill(self, limit=None):
         """
         Synonym for `DataFrame.fillna()` with ``method=`bfill```.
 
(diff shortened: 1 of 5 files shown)


databricks/koalas/groupby.py
```diff
@@ -1396,7 +1397,7 @@ def ffill(self, limit=None):
         2  NaN  NaN  5
         3  3.0  1.0  4
         """
-        return self._fillna(method='ffill', limit=limit)
+        return self._fillna(method='ffill', inplace=inplace, limit=limit)
```
HyukjinKwon: Can we add a test case for this one? I think it doesn't work.

databricks/koalas/tests/test_utils.py
```diff
@@ -65,6 +66,20 @@ def test_lazy_property(self):
         self.assert_eq(obj.lazy_prop, 1)
         self.assert_eq(obj.lazy_prop, 1)
 
+    def test_validate_bool_kwarg(self):
+        # This should pass and run fine
+        koalas = True
+        self.assert_eq(validate_bool_kwarg(koalas, "koalas"), True)
+        koalas = False
+        self.assert_eq(validate_bool_kwarg(koalas, "koalas"), False)
+
```
HyukjinKwon: Can we test `None` case too

databricks/koalas/groupby.py
itholic: ah, you're right, and pandas doesn't support them.

- koalas
```python
>>> df.groupby(['A']).ffill(inplace=True).sort_index()
...
AssertionError: (3, 4)
```

- pandas
```python
>>> df.to_pandas().groupby(['A']).ffill(inplace=True).sort_index()
TypeError: pad() got an unexpected keyword argument 'inplace'
```

i'll revert those kinda change.

PR #1283, opened 2020-02-14
Rename data columns prior to filter to make sure the column names are as expected.

ueshin:
This is a follow-up of #1273.
The Spark column names are not always the same as its column label.
This PR is to rename data columns prior to filter to make sure the column names are as expected.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8469,13 +8469,13 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 'For argument "inplace" expected type bool, received type {}.'
                 .format(type(inplace).__name__))
 
-        column_names = [label[0] for label in self._internal.column_labels]
+        data_columns = [label[0] for label in self._internal.column_labels]
         sdf = self._sdf.select(self._internal.index_scols
                                + [scol.alias(col) for scol, col
-                                  in zip(self._internal.column_scols, column_names)]) \
+                                  in zip(self._internal.column_scols, data_columns)]) \
             .filter(expr)
         internal = self._internal.copy(sdf=sdf,
-                                       column_scols=[scol_for(sdf, col) for col in column_names])
+                                       column_scols=[scol_for(sdf, col) for col in data_columns])
 
         if inplace:
             self._internal = internal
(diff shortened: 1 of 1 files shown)


databricks/koalas/frame.py
```diff
@@ -8469,8 +8469,13 @@ def query(self, expr, inplace=False):
                 'For argument "inplace" expected type bool, received type {}.'
                 .format(type(inplace).__name__))
 
-        sdf = self._sdf.filter(expr)
-        internal = self._internal.copy(sdf=sdf)
+        data_columns = [label[0] for label in self._internal.column_labels]
```
ueshin: If we support multi-index column later, we need to rename to fit the pandas' requirement.

ueshin:
also cc @itholic

itholic:
@ueshin 

Thanks for cc me!

Does it mean that there are cases where `data_columns` and `column_labels` are different?

ueshin:
yes, I guess you saw such cases several times?

itholic:
@ueshin 

yeah, maybe i guess.

so i'm trying to reproduce such cases, but couldn't yet (even after column re-naming)

could you show me a simple example when you available ?

itholic:
Anyway, LGTM if the cases could be happened!

ueshin:
e.g.,:

```py
>>> kdf = ks.DataFrame({('x','a'): [1,2,3], ('x','b'): [4,5,6], ('y','c'): [7,8,9]})
>>> kdf['x']
   a  b
0  1  4
1  2  5
2  3  6
>>> kdf['x'].query('a > 1')
...
pyspark.sql.utils.AnalysisException: "cannot resolve '`a`' given input columns: [__index_level_0__, (x, a), (x, b), __natural_order__]; ...
```

ueshin:
Thanks! I'd merge this now. Please feel free to leave comments if any.

itholic:
> e.g.,:
> 
> ```python
> >>> kdf = ks.DataFrame({('x','a'): [1,2,3], ('x','b'): [4,5,6], ('y','c'): [7,8,9]})
> >>> kdf['x']
>    a  b
> 0  1  4
> 1  2  5
> 2  3  6
> >>> kdf['x'].query('a > 1')
> Traceback (most recent call last):
> ...
> pyspark.sql.utils.AnalysisException: "cannot resolve '`a`' given input columns: [__index_level_0__, (x, a), (x, b), __natural_order__]; ...
> ```

Thanks !!

PR #1284, opened 2020-02-16
Standardize code style using Black

### LINKED ISSUE #755: Add precommit in workflow to standardize code style

deepyaman:
Close #755 

TODO:

- [x] Add pre-commit config (testing build first)
- [ ] Pare down list of formatted files, if necessary
- [ ] Document pre-commit


requirements-dev.txt
```diff
@@ -18,6 +18,7 @@ pypandoc
 ipython
 
 # Linter
+black==19.10b0; python_version >= "3.6"
```
HyukjinKwon: FYI, this file is being shared with conda of the CI too, which has a slightly different syntax comparing to the regular PIP.

itholic:
i think we have discussed about using `Black` before.

can you refer to [this PR](https://github.com/databricks/koalas/pull/761) and [this comment](https://github.com/databricks/koalas/pull/761#issuecomment-535984396) ?

requirements-dev.txt
deepyaman: Yep... I was hopeful https://github.com/conda/conda-build/pull/1984 supported it, but I think it's for something different. :( I'll add it differently.

deepyaman: Actually, I was trying to figure out yesterday where the other library requiring Python (Conda) 3.6+ is and couldn't find it; figured I'd support this similarly.

https://github.com/databricks/koalas/blob/master/.travis.yml#L69

deepyaman:
> maybe i think we have same discussed about using `Black` before.
> 
> can you refer to [this PR](https://github.com/databricks/koalas/pull/761) and [this comment](https://github.com/databricks/koalas/pull/761#issuecomment-535984396) ?

I saw previous discussions (#755 and #756), but I think I missed that part of the discussion on #761. I would be feel pretty against using Black just on diffs--rather than a consistent code style, you end up with an even worse mix between old and now.

If there are certain formatting rules would like to tweak, Black supports that, as long as those options are applied consistently throughout.

HyukjinKwon:
>  I would be feel pretty against using Black just on diffs--rather than a consistent code style, you end up with an even worse mix between old and now.

The problem is that, it overwrites the commit logs and difficult to track the changes. Also, I would like to keep it as an option, not a hard requirement because some people prefers a different style.

After all, we still run multiple linters so it's fine as long as it follows. PEP 8

.pre-commit-config.yaml
```diff
@@ -0,0 +1,6 @@
+repos:
```
HyukjinKwon: Hm, I was thinking about using `.git/hooks/pre-commit` right away rather then using a dependency for now.

HyukjinKwon:
Okay, let me take a look and fix it if it's really worthwhile. I will port some changes here and make the commit co-authored.

HyukjinKwon:
Let me make https://github.com/databricks/koalas/pull/1301 co-authored. thanks @deepyaman!

PR #1290, opened 2020-02-18
Implement Series.combine_first

itholic:
Implement [Series.combine_first](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.combine_first.html#pandas.Series.combine_first)

- basic example
```python
>>> s1 = ks.Series([1, np.nan])
>>> s2 = ks.Series([3, 4])
>>> s1.combine_first(s2)
0    1.0
1    4.0
Name: 0, dtype: float64
```

- MultiIndex
```python
>>> midx1 = pd.MultiIndex([['lama', 'cow', 'falcon', 'koala'],
...                        ['speed', 'weight', 'length', 'power']],
...                       [[0, 3, 1, 1, 1, 2, 2, 2],
...                        [0, 2, 0, 3, 2, 0, 1, 3]])
>>> midx2 = pd.MultiIndex([['lama', 'cow', 'falcon'],
...                        ['speed', 'weight', 'length']],
...                       [[0, 0, 0, 1, 1, 1, 2, 2, 2],
...                        [0, 1, 2, 0, 1, 2, 0, 1, 2]])
>>> kser1 = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1], index=midx1)
>>> kser2 = ks.Series([-45, 200, -1.2, 30, -250, 1.5, 320, 1, -0.3], index=midx2)
>>> kser1
lama    speed      45.0
koala   length    200.0
cow     speed       1.2
        power      30.0
        length    250.0
falcon  speed       1.5
        weight    320.0
        power       1.0
Name: 0, dtype: float64
>>> kser2
lama    speed     -45.0
        weight    200.0
        length     -1.2
cow     speed      30.0
        weight   -250.0
        length      1.5
falcon  speed     320.0
        weight      1.0
        length     -0.3
Name: 0, dtype: float64

>>> kser1.combine_first(kser2)
cow     length    250.0
        power      30.0
        speed       1.2
        weight   -250.0
falcon  length     -0.3
        power       1.0
        speed       1.5
        weight    320.0
koala   length    200.0
lama    length     -1.2
        speed      45.0
        weight    200.0
Name: 0, dtype: float64
```


itholic:
i considered putting `Series.combine` and `Series.combine_first` in single PR,

but their implementation concept was way different than i thought, so i separated them.

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,58 @@ def pct_change(self, periods=1):
 
         return self._with_new_scol((scol - prev_row) / prev_row)
 
+    def combine_first(self, other):
+        """
+        Combine Series values, choosing the calling Series's values first.
+
+        .. note:: This API internally performs a join operation which can be pretty expensive
+            in general. if you want to use though, set `compute.ops_on_diff_frames` to True.
```
HyukjinKwon: I think it's fine to don't respect `compute.ops_on_diff_frames` when users call this API. There's no way to use joining to combine two dataframes.

HyukjinKwon: You can enable it via `option_context(...)`.

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,58 @@ def pct_change(self, periods=1):
...
+        -----
+        Result index will be the union of the two indexes.
+
+        Examples
+        --------
+        >>> from databricks.koalas.config import set_option, reset_option
+        >>> set_option("compute.ops_on_diff_frames", True)
+        >>> s1 = ks.Series([1, np.nan])
+        >>> s2 = ks.Series([3, 4])
+        >>> s1.combine_first(s2)
```
HyukjinKwon: We don't have to perform `compute.ops_on_diff_frames` when two series are from the same dataframe.

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,58 @@ def pct_change(self, periods=1):
...
+        0    1.0
+        1    4.0
+        Name: 0, dtype: float64
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        if not isinstance(other, ks.Series):
+            raise ValueError("`combine_first` only allows `Series` for parameter `other`")
+        this = "__this_0"
```
HyukjinKwon: what is this? Series can have arbitrary name. Are you assuming the series alway to have the name `0`?

databricks/koalas/series.py
itholic: It seems we still need this?? 😮

```python
>>> kdf
   A  B
0  1  3
1  2  4
>>> kdf.A.combine_first(kdf.B)
...
ValueError: Cannot combine the series or dataframe because it comes from a different dataframe. In order to allow this operation, enable 'compute.ops_on_diff_frames' option.
```

itholic: It seems we still need this. maybe i missed something??

```python
>>> kdf = ks.DataFrame({'A':[1, 2], 'B': [3, 4]})
>>> combine_frames(kdf.A, kdf.B)
...
ValueError: Cannot combine the series or dataframe because it comes from a different dataframe. In order to allow this operation, enable 'compute.ops_on_diff_frames' option.
```

itholic: ah, i got what you meant. thanks!

databricks/koalas/series.py
itholic: Ah, I should have checked once again. thanks!

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,57 @@ def pct_change(self, periods=1):
...
+
+        Notes
+        -----
+        Result index will be the union of the two indexes.
+
+        Examples
+        --------
+        >>> s1 = ks.Series([1, np.nan])
+        >>> s2 = ks.Series([3, 4])
+        >>> s1.combine_first(s2)
```
HyukjinKwon: Do you know how it works without setting `compute.ops_on_diff_frames`? Both series are from different frames, and `combine_frames` checks it too.

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,57 @@ def pct_change(self, periods=1):
...
+        if not isinstance(other, ks.Series):
+            raise ValueError("`combine_first` only allows `Series` for parameter `other`")
+        if self._kdf is other._kdf:
+            this = self.name
+            that = other.name
+            combined = self._kdf
+        else:
+            this = "__this_{}".format(self.name)
+            that = "__that_{}".format(other.name)
+            with option_context("compute.ops_on_diff_frames", True):
```
HyukjinKwon: @itholic, it should throw an exception if `compute.ops_on_diff_frames` is not set. Let users to decide if they are going to allow it or not.

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,57 @@ def pct_change(self, periods=1):
...
+                combined = combine_frames(self.to_frame(), other)
+        index_scols = combined._internal.index_scols
+        sdf = combined._sdf
+        # If `self` has missing value, use value of `other`
+        cond = F.when(sdf[this].isNull(), sdf[that]).otherwise(sdf[this])
+        sdf = sdf.select(*index_scols, cond.alias(self.name))
+        internal = _InternalFrame(sdf=sdf, index_map=self._internal.index_map)
```
HyukjinKwon: Why don't we just copy from `kdf._internal` when `self._kdf is other._kdf`? Also we should use `self._with_new_scol` to keep it as the same DataFrame.

databricks/koalas/series.py
```diff
@@ -4324,6 +4324,57 @@ def pct_change(self, periods=1):
...
+        Result index will be the union of the two indexes.
+
+        Examples
+        --------
+        >>> s1 = ks.Series([1, np.nan])
+        >>> s2 = ks.Series([3, 4])
+        >>> s1.combine_first(s2)
+        0    1.0
+        1    4.0
+        Name: 0, dtype: float64
```
HyukjinKwon: Seems like it works differently if the index is mismatched.

```python
import numpy as np
import pandas as pd

s1 = ks.Series([1,2,3], index=["a", "b", "c"])
s2 = ks.Series([3,4,6], index=["c", "d", "e"])
s1.combine_first(s2)
```

```
>>> s1.combine_first(s2)

e    6
d    4
c    3
b    2
a    1
```

HyukjinKwon: Seems like this case works differently from pandas'

```python
import numpy as np
import pandas as pd

index = pd.MultiIndex.from_tuples([("a", "b"), ("a", "b"), ("a", "c")])
# s1 = ks.Series([1,2,3], index=index)
# s2 = ks.Series([3,4,6], index=index)
s1 = pd.Series([1,2,3], index=index)
s2 = pd.Series([3,4,6], index=index)
s1.combine_first(s2)
```


```
a  c    3
   b    1
   b    1
   b    2
   b    2
```

We might need to select distict columns.

itholic: Ah, let me check. thanks for the example

HyukjinKwon:
I think you should rebase and sync to the current master, @itholic .

databricks/koalas/series.py
```diff
@@ -4335,6 +4335,57 @@ def pct_change(self, periods=1):
...
+        """
+        if not isinstance(other, ks.Series):
+            raise ValueError("`combine_first` only allows `Series` for parameter `other`")
+        if self._kdf is other._kdf:
+            this = self.name
+            that = other.name
+            combined = self._kdf
```
HyukjinKwon: @itholic, when they are same frames, we should keep anchor without creating new DataFrame. Otherwise, the next operation will require the operation from different DataFrames. e.g.: `kdf.id.combine_first(kdf.id.combine_first(kdf.id))`. It should use `self._with_new_scol` whenever possible.

databricks/koalas/series.py
```diff
@@ -4335,6 +4335,57 @@ def pct_change(self, periods=1):
...
+            this = "__this_{}".format(self.name)
+            that = "__that_{}".format(other.name)
+            with option_context("compute.ops_on_diff_frames", True):
+                combined = combine_frames(self.to_frame(), other)
+        index_scols = combined._internal.index_spark_columns
+        sdf = combined._sdf
+        # If `self` has missing value, use value of `other`
+        cond = F.when(sdf[this].isNull(), sdf[that]).otherwise(sdf[this])
+        sdf = sdf.select(*index_scols, cond.alias(self.name))
+        internal = _InternalFrame(spark_frame=sdf, index_map=self._internal.index_map)
```
HyukjinKwon: Seems like it will use column index, and column index names.

HyukjinKwon: Seems like it will lose column index and column index names.

HyukjinKwon:
Looks fine otherwise.

databricks/koalas/series.py
itholic: Thanks! I'll keep that in mind

databricks/koalas/series.py
itholic: Sorry It was my mistake. Thanks for fixing :)

PR #1291, opened 2020-02-18
[MINOR] Uncomment doctest for DataFrame.query

itholic:
As requested at https://github.com/databricks/koalas/pull/1273#discussion_r380129074 ,

and since `map_in_pandas` (#1276 ) has been merged,

just uncommented existing doctest for DataFrame.query

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8539,7 +8539,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 1  3  4
                 2  5  6
                 >>> num = 1
-                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))  # doctest: +SKIP
+                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
                    A  B
                 1  3  4
                 2  5  6


databricks/koalas/frame.py
```diff
@@ -8539,7 +8539,7 @@ def query(self, expr, inplace=False):
                 1  3  4
                 2  5  6
                 >>> num = 1
-                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))  # doctest: +SKIP
+                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
```
ueshin: I'm wondering whether this example works with Spark.
I guess the `num` can't be seen from executors.
Actually, it throws an exception:

```py
>>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
>>> num = 1
>>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
   A  B
1  3  4
2  5  6

>>> ks.options.compute.shortcut_limit = 1
>>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
org.apache.spark.api.python.PythonException: Traceback (most recent call last):

...

pandas.core.computation.ops.UndefinedVariableError: local variable 'num' is not defined
```

itholic: @ueshin 

yes, i also just found some problem with `map_in_pandas` for those kinda methods, that it only works properly when the `compute.shortcut_limit` is larger than the size of the data during i'm working at #1292 .

so i think this example for `query` is not illustrated properly, we better remove this example or adding the note that we can work around with `map_in_pandas` when only the size of the data is smaller than `compute.shortcut_limit`.

which way is more proper do you think??

Thanks for the comment :)

HyukjinKwon: @itholic can you fix the example? Also update "the exactly same syntax with pandas" in the docstring too

itholic: okay, i'll start right away, thanks for the comment !

PR #1292, opened 2020-02-18
Implement DataFrame.take

itholic:
This PR proposes Implement of [DataFrame.take](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.take.html#pandas.DataFrame.take)

```python
>>> df = ks.DataFrame([('falcon', 'bird', 389.0),
...                    ('parrot', 'bird', 24.0),
...                    ('lion', 'mammal', 80.5),
...                    ('monkey', 'mammal', np.nan)],
...                   columns=['name', 'class', 'max_speed'],
...                   index=[0, 2, 3, 1])
>>> df
     name   class  max_speed
0  falcon    bird      389.0
2  parrot    bird       24.0
3    lion  mammal       80.5
1  monkey  mammal        NaN

>>> df.take([0, 3])
     name   class  max_speed
0  falcon    bird      389.0
1  monkey  mammal        NaN

>>> df.take([1, 2], axis=1)
    class  max_speed
0    bird      389.0
2    bird       24.0
3  mammal       80.5
1  mammal        NaN

>>> df.take([-1, -2])
     name   class  max_speed
1  monkey  mammal        NaN
3    lion  mammal       80.5
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8644,6 +8644,86 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         self._internal.spark_internal_df.explain(extended)
 
+    def take(self, indices, axis=0, **kwargs):
+        """
+        Return the elements in the given *positional* indices along an axis.
+
+        This means that we are not indexing according to actual values in
+        the index attribute of the object. We are indexing according to the
+        actual position of the element in the object.
+
+        ..note:: This method has some disadvantage in terms of performance since
+            implemented by using :meth:`DataFrame.map_in_pandas`.
+            See the note of :meth:`DataFrame.map_in_pandas` if you want to know
+            about those disadvantage for more detail.
+
+        Parameters
+        ----------
+        indices : array-like
+            An array of ints indicating which positions to take.
+        axis : {0 or 'index', 1 or 'columns', None}, default 0
+            The axis on which to select elements. ``0`` means that we are
+            selecting rows, ``1`` means that we are selecting columns.
+        **kwargs
+            For compatibility with :meth:`numpy.take`. Has no effect on the
+            output.
+
+        Returns
+        -------
+        taken : same type as caller
+            An array-like containing the elements taken from the object.
+
+        See Also
+        --------
+        DataFrame.loc : Select a subset of a DataFrame by labels.
+        DataFrame.iloc : Select a subset of a DataFrame by positions.
+        numpy.take : Take elements from an array along an axis.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame([('falcon', 'bird', 389.0),
+        ...                    ('parrot', 'bird', 24.0),
+        ...                    ('lion', 'mammal', 80.5),
+        ...                    ('monkey', 'mammal', np.nan)],
+        ...                   columns=['name', 'class', 'max_speed'],
+        ...                   index=[0, 2, 3, 1])
+        >>> df
+             name   class  max_speed
+        0  falcon    bird      389.0
+        2  parrot    bird       24.0
+        3    lion  mammal       80.5
+        1  monkey  mammal        NaN
+
+        Take elements at positions 0 and 3 along the axis 0 (default).
+
+        Note how the actual indices selected (0 and 1) do not correspond to
+        our selected indices 0 and 3. That's because we are selecting the 0th
+        and 3rd rows, not rows whose indices equal 0 and 3.
+
+        >>> df.take([0, 3])
+             name   class  max_speed
+        0  falcon    bird      389.0
+        1  monkey  mammal        NaN
+
+        Take elements at indices 1 and 2 along the axis 1 (column selection).
+
+        >>> df.take([1, 2], axis=1)
+            class  max_speed
+        0    bird      389.0
+        2    bird       24.0
+        3  mammal       80.5
+        1  mammal        NaN
+
+        We may take elements using negative integers for positive indices,
+        starting from the end of the object, just like with Python lists.
+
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -8644,6 +8644,86 @@ def explain(self, extended: bool = False):
...
+
+        We may take elements using negative integers for positive indices,
+        starting from the end of the object, just like with Python lists.
+
+        >>> df.take([-1, -2])
+             name   class  max_speed
+        1  monkey  mammal        NaN
+        3    lion  mammal       80.5
+        """
+        return self.map_in_pandas(lambda pdf: pdf.take(indices, axis, **kwargs))
```
HyukjinKwon: @itholic, I don't think this will work when there are multiple partitions. This function will be executed against each batch, and it will execute `take` for each batch.

databricks/koalas/frame.py
```diff
@@ -8644,6 +8644,86 @@ def explain(self, extended: bool = False):
...
+        1  monkey  mammal        NaN
+
+        Take elements at indices 1 and 2 along the axis 1 (column selection).
+
+        >>> df.take([1, 2], axis=1)
+            class  max_speed
+        0    bird      389.0
+        2    bird       24.0
+        3  mammal       80.5
+        1  mammal        NaN
```
HyukjinKwon: For example, the results here are different.

```python
>>> df.take([1,2])
     name   class  max_speed
2  parrot    bird       24.0
3    lion  mammal       80.5
```

databricks/koalas/frame.py
HyukjinKwon: After more thoughts, seems we won't be able to work around by `map_in_pandas`. Probably you should have a distributed sequence, and filter out by its position.

databricks/koalas/frame.py
itholic: thanks for the comment!

i agree with https://github.com/databricks/koalas/pull/1292#discussion_r380462332, but this case seems missing `axis=1` .

```python
>>> kdf.take([1, 2], axis=1)
    class  max_speed
0    bird      389.0
2    bird       24.0
3  mammal       80.5
1  mammal        NaN

>>> pdf.take([1, 2], axis=1)
    class  max_speed
0    bird      389.0
2    bird       24.0
3  mammal       80.5
1  mammal        NaN
```

databricks/koalas/frame.py
itholic: i see. let me test more case with bigger dataset and find the solution,

Thanks for the comment!! 👍

itholic:
We have to find another solution since `map_in_pandas` won't work for this method.

While i'm here, i'm searching and fixing the existing implementations.

databricks/koalas/frame.py
```diff
@@ -9578,6 +9578,86 @@ def explain(self, extended: bool = False):
...
+
+        We may take elements using negative integers for positive indices,
+        starting from the end of the object, just like with Python lists.
+
+        >>> df.take([-1, -2])
+             name   class  max_speed
+        1  monkey  mammal        NaN
+        3    lion  mammal       80.5
+        """
+        return self.map_in_pandas(lambda pdf: pdf.take(indices, axis, **kwargs))
```
HyukjinKwon: How come this works? It will apply `take` for each partition; however, `indices` is an index in a global sequence.

itholic: It's not working now, we should find another way

itholic: Maybe I think I can workaround with another way!

itholic:
Implementation for `axis=0` is done here,

And while I'm here, I'm working for implementing `axis=1`.

Could someone reviews `axis=0` if when available?

databricks/koalas/tests/test_dataframe.py
```diff
@@ -2966,6 +2966,39 @@ def test_query(self):
...
+        # MultiIndex columns with axis=0 (default)
+        self.assert_eq(kdf.take([1, 2]).sort_index(), pdf.take([1, 2]).sort_index())
+        self.assert_eq(kdf.take([-1, -2]).sort_index(), pdf.take([-1, -2]).sort_index())
+
+        # self.assert_eq(
+        #     kdf.take([1, 2], axis=1).sort_index(), pdf.take([1, 2], axis=1).sort_index()
+        # )
+        # self.assert_eq(
+        #     kdf.take([-1, -2], axis=1).sort_index(), pdf.take([-1, -2], axis=1).sort_index()
+        # )
```
itholic: I'm working now for making those commented tests work.

ueshin:
Seems like we can just delegate to `iloc` indexer?

itholic:
@ueshin Ah, I should have took a look at more carefully. Thanks for the comment :)

databricks/koalas/frame.py
```diff
@@ -9578,6 +9578,85 @@ def explain(self, extended: bool = False):
...
+
+        We may take elements using negative integers for positive indices,
+        starting from the end of the object, just like with Python lists.
+
+        >>> df.take([-1, -2]).sort_index()
+             name   class  max_speed
+        1  monkey  mammal        NaN
+        3    lion  mammal       80.5
+        """
+        axis = validate_axis(axis)
```
ueshin: We should check the type of `indices`?

itholic: Thanks. I'll add :)

ueshin:
Could you update the PR description?

itholic:
@ueshin Yup, I just removed unmatched descriptions

ueshin:
Thanks! merging.

PR #1293, opened 2020-02-18
Fixing & Adding example in docstring for DataFrame.query

itholic:
As we discussed at https://github.com/databricks/koalas/pull/1291#discussion_r380478346 ,

made the docstring more properly with adding more example and explanations.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -9062,40 +9062,29 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             * Internal columns that starting with __. are able to access,
               however, they are not supposed to be accessed.
             * This delegates to Spark SQL so the syntax follows Spark SQL
-            * If you want the exactly same syntax with pandas you can work around
-              by using :meth:`DataFrame.map_in_pandas`, when only the size of the
-              data is smaller than `compute.shortcut_limit`. See the example below.
+            * If you want the pandas syntax, you can work around. See the example below.
 
-                >>> from databricks.koalas.config import set_option, reset_option, get_option
-                >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
-                >>> df
-                   A  B
-                0  1  2
-                1  3  4
-                2  5  6
-                >>> # You can check the value of `compute.shortcut_limit` like the below.
-                ... get_option('compute.shortcut_limit')
-                1000
-                >>> # Since 1000 is larger than the size of the data(in this case, 3),
-                ... # we can work around with `map_in_pandas`.
-                ... num = 1
-                >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
-                   A  B
-                1  3  4
-                2  5  6
-
-                >>> # After then, let's set `compute.shortcut_limit` to 2,
-                ... # which is smaller value than the size of the data.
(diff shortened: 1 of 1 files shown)


databricks/koalas/frame.py
```diff
@@ -8529,21 +8529,41 @@ def query(self, expr, inplace=False):
             * Internal columns that starting with __. are able to access,
               however, they are not supposed to be accessed.
             * This delegates to Spark SQL so the syntax follows Spark SQL
-            * If you want the exactly same syntax with pandas' you can work around
-              by using :meth:`DataFrame.map_in_pandas`. See the example below.
+            * If you want the exactly same syntax with pandas you can work around
```
HyukjinKwon: @itholic, let's just say as below:

> If you want the pandas syntax, you can work around.  See the example below.

databricks/koalas/frame.py
```diff
@@ -8529,21 +8529,41 @@ def query(self, expr, inplace=False):
...
+                >>> from databricks.koalas.config import set_option, reset_option, get_option
                 >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
                 >>> df
                    A  B
                 0  1  2
                 1  3  4
                 2  5  6
-                >>> num = 1
+                >>> # You can check the value of `compute.shortcut_limit` like the below.
+                ... get_option('compute.shortcut_limit')
```
HyukjinKwon: Koalas targets big data. I think there is not so much point to show how to work around with small dataset.

databricks/koalas/frame.py
```diff
@@ -8529,21 +8529,41 @@ def query(self, expr, inplace=False):
...
+                1000
+                >>> # Since 1000 is larger than the size of the data(in this case, 3),
+                ... # we can work around with `map_in_pandas`.
+                ... num = 1
                 >>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
```
HyukjinKwon: You should show an example with type hints. So it avoids short-cut, and shows a good practice example.

For the limitation by the referred variable not being serialized into Python worker sides, you can just add some descriptions such as :

> `map_in_pandas` is performed at workers in different nodes; however, currently the variable referred in pandas syntax is not serialized together. Therefore, in order to use such syntax,  you should make sure the variable is serialized by, for example, putting it within the closure.

```python
>>> def query_func(df):
...     num = 1
...     pdf.query('A > @num')
>>> df.map_in_pandas(lambda pdf: pdf.query('A > @num'))
```

databricks/koalas/frame.py
itholic: i have some tested, the index will be lost after map_in_pandas (data looks okay)

```python
>>> df = ks.DataFrame({'A': range(1000), 'B': range(1000)})
>>> def query_func(pdf) -> ks.DataFrame[int, int]:
...     num = 990
...     return pdf.query('A > @num')
...
>>> df.map_in_pandas(query_func)
    c0   c1
0  991  991
1  992  992
2  993  993
3  994  994
4  995  995
5  996  996
6  997  997
7  998  998
8  999  999
```

is it expected behaviour? or just a bug?

HyukjinKwon: Yea, that's expected at this moment. See https://github.com/databricks/koalas/issues/1307

HyukjinKwon:
Let me just merge and rephrase by myself.

PR #1294, opened 2020-02-18
Add another utility method to `_InternalFrame`.

ueshin:
Added `with_new_sdf` utility to `_InternalFrame`.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -782,16 +782,34 @@ class _InternalFrame(object):
                                for name in index_names]
         return pdf
 
+    def with_new_sdf(self, sdf: spark.DataFrame,
+                     data_columns: Optional[List[str]] = None) -> '_InternalFrame':
+        """ Copy the immutable _InternalFrame with the updates by the specified Spark DataFrame.
+
+        :param sdf: the new Spark DataFrame
+        :param data_columns: the new column names.
+            If None, the original one is used.
+        :return: the copied _InternalFrame.
+        """
+        if data_columns is None:
+            data_columns = self.data_columns
+        else:
+            assert len(data_columns) == len(self.column_labels), \
+                (len(data_columns), len(self.column_labels))
+        sdf = sdf.drop(NATURAL_ORDER_COLUMN_NAME)
+        return self.copy(sdf=sdf, column_scols=[scol_for(sdf, col) for col in data_columns])
+
     def with_new_columns(self, scols_or_ksers: List[Union[spark.Column, 'Series']],
                          column_labels: Optional[List[Tuple[str, ...]]] = None,
                          keep_order: bool = True) -> '_InternalFrame':
-        """ Copy the immutable DataFrame with the updates by the specified Spark Columns or Series.
+        """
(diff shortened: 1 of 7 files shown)


databricks/koalas/internal.py
```diff
@@ -782,16 +782,34 @@ def pandas_df(self):
...
         return pdf
 
+    def with_new_sdf(self, sdf: spark.DataFrame,
+                     data_columns: Optional[List[str]] = None) -> '_InternalFrame':
+        """ Copy the immutable _InternalFrame with the updates by the specified Spark DataFrame.
+
+        :param sdf: the new Spark DataFrame
+        :param data_columns: the new column names.
+            If None, the original one is used.
+        :return: the copied _InternalFrame.
```
itholic: Although i'm not sure if it is really necessary,

but how are you think about we have a simple example for these kinda util methods?
(since sometimes i got a little confused when i try to use such kinda of methods 😿 )

itholic:
can you give me an opinion about https://github.com/databricks/koalas/pull/1294#discussion_r381067247 ?

otherwise, LGTM 👍

databricks/koalas/internal.py
```diff
@@ -832,10 +850,10 @@ def with_new_columns(self, scols_or_ksers: List[Union[spark.Column, 'Series']],
             scol=None)
 
     def with_filter(self, pred: Union[spark.Column, 'Series']):
-        """ Copy the immutable DataFrame with the updates by the predicate.
+        """ Copy the immutable _InternalFrame with the updates by the predicate.
 
         :param pred: the predicate to filter.
-        :return: the copied immutable DataFrame.
+        :return: the copied _InternalFrame.
```
itholic: +1

databricks/koalas/internal.py
ueshin: I think these are simpler versions of `copy`.
I'm afraid I'm not quite sure where you are being confused.
Maybe you can add some examples (and description?) to make them clearer?

ueshin:
Thanks! I'd merge this now. Please feel free to leave comments if any.

databricks/koalas/internal.py
itholic: @ueshin 

Ah, sorry my comment was totally not enough.

i meant how about we have a simple example for util methods or methods those are only used under-the-hood looks like the below. (it didn't necessarily talk about `with_new_sdf`)

```python
def combine_frames(this, *args, how="full"):
    """
    ...
    Examples
    --------
    >>> from databricks.koalas.config import set_option
    >>> set_option("compute.ops_on_diff_frames", True)
    >>> kdf = ks.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]})
    >>> kser = ks.Series([10, 20, 30])
    >>> kdf
       a  b
    0  1  4
    1  2  5
    2  3  6
    >>> kser
    0    10
    1    20
    2    30
    Name: 0, dtype: int64
    
    >>> combine_frames(kdf, kser)
      this    that
         a  b    0
    0    1  4   10
    1    2  5   20
    2    3  6   30
    """
    ...
```

(if we add such examples for under-the-hood methods like the above, i think it's okay to just skip their doctest)

PR #1295, opened 2020-02-19
Implement DataFrame.unstack() but without level

### LINKED ISSUE #886: DataFrame.unstack
Implement `DataFrame.unstack`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.unstack.html

HyukjinKwon:
This PR implements `DataFrame.unstack()`. This PR intentionally does not implement `level` with multi-index. See the discussion at https://github.com/databricks/koalas/issues/886/
When multi-index is not use, it is okay to implement.

Resolves #886


databricks/koalas/frame.py
```diff
@@ -7710,9 +7710,9 @@ def melt(self, id_vars=None, value_vars=None, var_name=None, value_name="value")
 
         Examples
         --------
-        >>> df = ks.DataFrame({'A': {0: 'a', 1: 'b', 2: 'c'},
-        ...                    'B': {0: 1, 1: 3, 2: 5},
-        ...                    'C': {0: 2, 1: 4, 2: 6}},
+        >>> df = ks.DataFrame({'A': {'0': 'a', '1': 'b', '2': 'c'},
+        ...                    'B': {'0': '1', '1': '3', '2': '5'},
+        ...                    'C': {'0': '2', '1': '4', '2': '6'}},
```
ueshin: Why changed here? Seems not related?

HyukjinKwon: It was because of `array(struct(...), ...)` function type coercion in Spark 2.3. Nested structs require the exactly same type in Spark 2.3. From Spark 2.4, seems they can coerce types properly. So I had to change it to strings to make it pass in Spark 2.3 too.

HyukjinKwon: Oops, yes. Seems I mistakenly changed it.

PR #1296, opened 2020-02-20
Let extra options have higher priority to allow workarounds in read_csv

HyukjinKwon:
This PR proposes to let extra options have higher priority to allow workarounds.

For instance, after this PR, you can disable schema inference:

```python
>>> ks.read_csv("tmp.csv").dtypes
a    int32
b    int32
c    int32
dtype: object
>>> ks.read_csv("tmp.csv", inferSchema=False).dtypes
a    object
b    object
c    object
```


HyukjinKwon:
@ueshin, let me merge this. Trivial and I am pretty sure :-).

PR #1299, opened 2020-02-21
Matching the behaviour to pandas>=1.0.0 at most.

itholic:
Follow-up for #1197

Since we're following latest version of pandas, should fix several TODOs with matching pandas>=1.0.0 for now.

## For example.

the behaviour of `Expanding.count()` and `ExpandingGroupby.count()` are different depending on what pandas version has been installed.

- pandas < 1.0.0
```python
>>> s = pd.Series([2, 2, 3, 3, 3, 4, 4, 4, 4, 5, 5])
>>> s.groupby(s).expanding(3).count().sort_index()
2  0     1.0
   1     2.0
3  2     1.0
   3     2.0
   4     3.0
4  5     1.0
   6     2.0
   7     3.0
   8     4.0
5  9     1.0
   10    2.0
dtype: float64
```

- pandas >= 1.0.0
```python
>>> s = pd.Series([2, 2, 3, 3, 3, 4, 4, 4, 4, 5, 5])
>>> s.groupby(s).expanding(3).count().sort_index()
2  0     NaN
   1     NaN
3  2     NaN
   3     NaN
   4     3.0
4  5     NaN
   6     NaN
   7     3.0
   8     4.0
5  9     NaN
   10    NaN
dtype: float64
```

Since we're following latest version of pandas, need to fix this.


databricks/koalas/tests/test_expanding.py
```diff
@@ -115,7 +120,10 @@ def _test_groupby_expanding_func(self, f):
             repr(getattr(pdf.groupby([("a", "x"), ("a", "y")]).expanding(2), f)().sort_index()))
 
     def test_groupby_expanding_count(self):
-        self._test_groupby_expanding_func("count")
+        # The behaviour of ExpandingGroupby.count are different between pandas>=1.0.0 and lower,
```
HyukjinKwon: @itholic, do we handle all of TODOs here? I still see many TODOs left, for example, in https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-b2490bcc905f49fa2f41e71c72121af8R780

- If and only if the behaviour in pandas 1.0.0 is a bug, we skip or test differently. We should explain in the comments why it looks a bug. Ideally, we should link the pandas issue where the behaviour was changed.
- Otherwise, we match to latest pandas' behaviour.

itholic: @HyukjinKwon 

Thanks for the comment !

> do we handle all of TODOs here?

I think all TODO has been resolved yesterday before the #1197 merged, except two.

- one is: https://github.com/databricks/koalas/pull/1197/files#r381830438 
  - since it seems requiring lots of effort, i'd like to work this in separated PR.
- and the other is: https://github.com/databricks/koalas/pull/1197/files#r382345421 which i made here.

am i missing something?

databricks/koalas/window.py
```diff
@@ -1104,7 +1095,12 @@ def count(self):
         2  2.0
         3  3.0
         """
-        return super(Expanding, self).count()
+        def count(scol):
+            return F.when(
+                F.row_number().over(self._unbounded_window) >= self._min_periods,
+                F.count(scol).over(self._window)
+            ).otherwise(F.lit(None))
+        return self._apply_as_series_or_frame(count).astype('float64')
```
ueshin: I might be missing something, but seems like pandas did the same thing for rolling as well?

https://github.com/pandas-dev/pandas/pull/30923/files#diff-0de5c5d9abfcdd141e83701eaaec4358L1194-L1195

What happens if we move this up to `_RollingAndExpanding`?

itholic: @ueshin 

> What happens if we move this up to _RollingAndExpanding?

I tried, and it occurs test failure of `Rolling.count()` and `RollingGroupby.count()`.

and i think i found the reason like the below, which seems we might be fixing.

https://github.com/databricks/koalas/blob/f8b60d223c49da386b10e453ea2131db0d8623d2/databricks/koalas/generic.py#L1544-L1546

databricks/koalas/tests/test_expanding.py
HyukjinKwon: Let's do it all here in this PR. Does this PR handle all of them below?

https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-d3f67c2eab2a8dd3df7283426995819fR86
https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-497e17c16cdf5dca7bbc65d1b130574eR127
https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-497e17c16cdf5dca7bbc65d1b130574eR750
https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-b2490bcc905f49fa2f41e71c72121af8R780
https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-bbc5f26000bddf5eb70586e69d18f8f8R57

We should describe why it looks like a bug in pandas, ideally with some issue or PR links in pandas side.

databricks/koalas/window.py
itholic: i have tried, but it occurred the test failure for `Rolling.count()` and `RollingGroupby.count()` maybe because of below:

https://github.com/databricks/koalas/blob/f8b60d223c49da386b10e453ea2131db0d8623d2/databricks/koalas/window.py#L134-L139

i think we might be fixing this first.

itholic: in short, the problem we have now like the below.

```python
>>> pser.rolling(5).count().sort_index()
0     1.0
1     2.0
2     3.0
3     4.0
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
dtype: float64

>>> kser.rolling(5).count().sort_index()
0     1.0
1     2.0
2     3.0
3     4.0
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
Name: 0, dtype: float64
```

it's okay, but.

```python
>>> pser.rolling(5, min_periods=3).count().sort_index()
0     NaN
1     NaN
2     3.0
3     4.0
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
dtype: float64

>>> kser.rolling(5, min_periods=3).count().sort_index()
0     1.0
1     2.0
2     3.0
3     4.0
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
Name: 0, dtype: float64
```

and if we do 

> What happens if we move this up to _RollingAndExpanding?

```python
>>> pser.rolling(5).count().sort_index()
0     1.0
1     2.0
2     3.0
3     4.0
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
dtype: float64

>>> kser.rolling(5).count().sort_index()
0     NaN
1     NaN
2     NaN
3     NaN
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
Name: 0, dtype: float64
```

since we're treating the `min_periods` is same with `window` when `min_periods` is not given. 

```python
kser.rolling(5).count() == kser.rolling(5, min_periods=5).count()
0     True
1     True
2     True
3     True
4     True
5     True
6     True
7     True
8     True
9     True
10    True
Name: 0, dtype: bool
```

we should fix it.

HyukjinKwon: Which error and test failure happens? It should be best to fix it together here.

itholic: @HyukjinKwon 

Thanks for responding !

1. as i commented above, the parameter `min_periods`  is not working for `Rolling.count()` and `RollingGroupby.count()`



> ```python
> >>> pser.rolling(5, min_periods=3).count().sort_index()
> 0     NaN
> 1     NaN
> 2     3.0
> 3     4.0
> 4     5.0
> 5     5.0
> 6     5.0
> 7     5.0
> 8     5.0
> 9     5.0
> 10    5.0
> dtype: float64
> 
> >>> kser.rolling(5, min_periods=3).count().sort_index()
> 0     1.0
> 1     2.0
> 2     3.0
> 3     4.0
> 4     5.0
> 5     5.0
> 6     5.0
> 7     5.0
> 8     5.0
> 9     5.0
> 10    5.0
> Name: 0, dtype: float64
> ```

2. this is because the `count()` for `Rolling` and `RollingGroupby` is implemented like the below, without considering about `min_periods` at all

```python
    def count(self):
        def count(scol):
            return F.count(scol).over(self._window)

        return self._apply_as_series_or_frame(count).astype('float64')
```

3. so i have tried to fix this with applying same logic with `Expanding.count()` and `ExpandingGroupby.count()` like the below:

```python
        def count(scol):
            return F.when(
                F.row_number().over(self._unbounded_window) >= self._min_periods,
                F.count(scol).over(self._window)
            ).otherwise(F.lit(None))
        return self._apply_as_series_or_frame(count).astype('float64')
```

4. but it drive another bug looks like the below.

```python
>>> pser.rolling(5).count().sort_index()
0     1.0
1     2.0
2     3.0
3     4.0
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
dtype: float64

>>> kser.rolling(5).count().sort_index()
0     NaN
1     NaN
2     NaN
3     NaN
4     5.0
5     5.0
6     5.0
7     5.0
8     5.0
9     5.0
10    5.0
Name: 0, dtype: float64
```

since we're treating the `min_periods` as same as `window` if specific value is not given for `min_periods`

https://github.com/databricks/koalas/blob/f8b60d223c49da386b10e453ea2131db0d8623d2/databricks/koalas/window.py#L134-L139

databricks/koalas/tests/test_expanding.py
itholic: Ah, i thought you mean the comment in the previous PR you left yesterday.

okay i just changed the title of this PR to `Matching the behaviour to pandas>=1.0.0 at most.`, and will start cleaning up remaining TODOs right away.

Thanks! :)

databricks/koalas/window.py
HyukjinKwon: Okay, let's fix it separately.

HyukjinKwon:
@deepyaman do you mind rebasing and syncing to the master? There are many conflicts as of https://github.com/databricks/koalas/commit/30b33344cdf7c41e2a2061daed314cdd55ec3215

HyukjinKwon:
@itholic do you mind rebasing and syncing to the master? There are many conflicts as of https://github.com/databricks/koalas/commit/30b33344cdf7c41e2a2061daed314cdd55ec3215

itholic:
@HyukjinKwon okay i did it :)

databricks/koalas/tests/test_expanding.py
itholic: - [ ] [24726bd#diff-d3f67c2eab2a8dd3df7283426995819fR86](https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-d3f67c2eab2a8dd3df7283426995819fR86)
- [ ] [24726bd#diff-497e17c16cdf5dca7bbc65d1b130574eR127](https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-497e17c16cdf5dca7bbc65d1b130574eR127)
- [ ] [24726bd#diff-497e17c16cdf5dca7bbc65d1b130574eR750](https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-497e17c16cdf5dca7bbc65d1b130574eR750)
- [ ] [24726bd#diff-b2490bcc905f49fa2f41e71c72121af8R780](https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-b2490bcc905f49fa2f41e71c72121af8R780)
- [ ] [24726bd#diff-bbc5f26000bddf5eb70586e69d18f8f8R57](https://github.com/databricks/koalas/commit/24726bdf3c69f224c77152361b9641ce6a2408df#diff-bbc5f26000bddf5eb70586e69d18f8f8R57)

itholic:
okay, i think the remained TODOs are roughly solved.

See the https://github.com/databricks/koalas/pull/1299#discussion_r384467192 for detail.

databricks/koalas/tests/test_groupby.py
```diff
@@ -168,7 +168,20 @@ def test_groupby_multiindex_columns(self):
...
+        # ValueError: Can only tuple-index with a MultiIndex
+        # >>> kdf[('x', 'a')].groupby(kdf[('x', 'b')]).sum().sort_index()
+        # (x, b)
+        # 1    13
+        # 2     9
+        # 3     8
+        # 4     1
+        # 7     6
+        # Name: (x, a), dtype: int64
+        # So, skip this for now.
```
HyukjinKwon: Can you manually construct a DataFrame as same as the output, and then test it? You can keep the comments.

PR #1308, opened 2020-02-25
Add `values` back.

ueshin:
Previously we discussed and decided not to add `values` property (#146), but there are still many users using the property. We should add it back, but with a warning to encourage them to use `to_numpy`.


HyukjinKwon:
@ueshin, can we update "Guardrails to prevent users from shooting themselves in the foot" at `docs/source/development/design.rst` too?

Issue #1307, opened 2020-02-25
DataFrame.apply(func, axis=1) returns Series with unmatched index


patryk-oleniuk:
Env: koalas 0.27.0, Databricks DBR 6.2.

Consider the following code for pandas:
```python
df = pd.DataFrame({"timestamp":[0.0, 0.5, 1.0, 0.0, 0.5, 100.0, 101.0, 102.0],
              "car_id": ['A','A','A','B','B', 'C', 'C', 'C'], 
              "battery_charge": [100, 90, 80,100,90, 50, 60, 70]
              })

df2 = df.groupby("car_id").sum()

def calc_somthing(row) -> float:
  # doing some job per row
  return row['battery_charge']/(row['timestamp'] +1)

print(df2.apply(calc_somthing, axis=1))
```
That prints a series with the proper index: 
```bash
car_id
A    108.000000
B    126.666667
C      0.592105
dtype: float64
```
But if we do the same for koalas, we get a wrong index (and also wrong name, BTW):
```bash
0    126.666664
1    108.000000
2      0.592105
Name: 0, dtype: float32
```
That causes a lot of problems when merging that index with a new df, e.g. 
```python
df2["newcol"] = df2.apply(calc_somthing, axis=1)
``` 
The workaround is to always reset_index before apply.

HyukjinKwon:
The problem here is that we can't tell what's the index given the type hints. So the index is being lost currently. Yes, `reset_index` is a nice workaround.

We should find a proper way to specify index type in the type hint... Let me at least document it for now.

patryk-oleniuk:
Hmm, not sure why couldn't we just get the same index as the originating df?
Especially for `df.apply(func, axis=1)` that returns single type value per row.

HyukjinKwon:
It's because the output index can be different from the original DataFrame:

```python
>>> import pandas as pd
>>> pd.DataFrame({'a': [1,2,3]}).apply(lambda _: pd.Series([1,2,3], index=["a", "b", "c"]))
   a
a  1
b  2
c  3
```

We can't detect the index type before the function execution. That's why it can return index type properly when the type hints are not specified (hence the type is inferred after executing the function once).

HyukjinKwon:
[...]

PR #1310, opened 2020-02-26
Add a note when type hint is provided to DataFrame.apply

HyukjinKwon:
This PR adds some documentations for the fact that index is switched to the default index when type hints are specified.

See also #1307.


HyukjinKwon:
@ueshin, let me just merge this. It's just doc change.

PR #1314, opened 2020-02-27
Use `author` instead of `maintainer`.

ueshin:
According to https://packaging.python.org/tutorials/packaging-projects/#creating-setup-py, we had better use `author` instead of `maintainer`.


ueshin:
I'd merge this now.

PR #1322, opened 2020-03-03
Support MultiIndex in DataFrame.unstack.

ueshin:
We can support unstack with MultiIndex using pivot_table.
But as discussed at https://github.com/databricks/koalas/issues/886#issuecomment-585593019, we should still be careful to use it.


ueshin:
We need to revisit `pivot_table` later. I will handle it in a separate PR.

PR #1325, opened 2020-03-05
Implements difference() for Index & MultiIndex

itholic:
This PR proposes for implementing difference for Index & MultiIndex.
(https://pandas.pydata.org/docs/reference/api/pandas.Index.difference.html#pandas.Index.difference)

- Index.difference()
```python
>>> idx1 = ks.Index([2, 1, 3, 4])
>>> idx2 = ks.Index([3, 4, 5, 6])
>>> idx1.difference(idx2)
Int64Index([1, 2], dtype='int64')
>>> idx1.difference(idx2, sort=False)
Int64Index([2, 1], dtype='int64')
```

- MultiIndex.difference()
```python
>>> midx1 = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2), ('c', 'z', 3)])
>>> midx2 = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'z', 2), ('k', 'z', 3)])
>>> midx1.difference(midx2)
MultiIndex([('b', 'y', 2),
            ('c', 'z', 3)],
           )
```


databricks/koalas/indexes.py
```diff
@@ -1391,6 +1391,69 @@ def set_names(self, names, level=None, inplace=False):
...
+        if not is_list_like(other):
+            raise TypeError("Input must be Index or array-like")
+        if not isinstance(sort, (type(None), type(True))):
+            raise ValueError(
+                "The 'sort' keyword only takes the values of None or True; {} was passed.".format(
+                    sort
+                )
+            )
+        # Handling MultiIndex
+        if isinstance(self, ks.MultiIndex):
```
HyukjinKwon: This two `if`s can be combined.

itholic: fixed it. thanks for the review!

HyukjinKwon:
I am going to manually merge this to keep the author

databricks/koalas/indexes.py
```diff
@@ -1391,6 +1391,75 @@ def set_names(self, names, level=None, inplace=False):
...
+        MultiIndex
+
+        >>> midx1 = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'y', 2), ('c', 'z', 3)])
+        >>> midx2 = ks.MultiIndex.from_tuples([('a', 'x', 1), ('b', 'z', 2), ('k', 'z', 3)])
+        >>> midx1.difference(midx2)  # doctest: +SKIP
+        MultiIndex([('b', 'y', 2),
+                    ('c', 'z', 3)],
+                   )
+        """
+        if not is_list_like(other):
```
HyukjinKwon, line 1430: I think we should check if `other` is `Index` instance. `is_list_like` is pandas' and we should rely on that.

databricks/koalas/indexes.py
```diff
@@ -1391,6 +1391,75 @@ def set_names(self, names, level=None, inplace=False):
...
+            raise TypeError("Input must be Index or array-like")
+        if not isinstance(sort, (type(None), type(True))):
+            raise ValueError(
+                "The 'sort' keyword only takes the values of None or True; {} was passed.".format(
+                    sort
+                )
+            )
+        # Handling MultiIndex
+        if isinstance(self, ks.MultiIndex) and not isinstance(other, ks.MultiIndex):
+            if not all([isinstance(item, tuple) for item in other]):
```
HyukjinKwon, line 1440: Here it does a for loop which collects everything into driver side. We should probably avoid.

HyukjinKwon: Actually, is this condition tested? doing a for loop against Koalas instance will result in something like:

```
pyspark.pandas.exceptions.PandasNotImplementedError: The method `pd.Index.__iter__()` is not implemented. If you want to collect your data as an NumPy array, use 'to_numpy()' instead.
```

PR #1328, opened 2020-03-05
Implement repeat() for Series, Index, and MultiIndex.

ueshin:
Adding `repeat()` for `Series`, `Index`, and `MultiIndex`.

- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.repeat.html
- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.repeat.html


databricks/koalas/indexes.py
```diff
@@ -1494,6 +1494,53 @@ def is_all_dates(self):
...
+        Index(['a', 'b', 'c', 'a', 'b', 'c'], dtype='object')
+        """
+        if not isinstance(repeats, int):
+            raise ValueError("`repeats` argument must be integer, but got {}".format(type(repeats)))
+        elif repeats < 0:
+            raise ValueError("negative dimensions are not allowed")
+
+        sdf = self._internal.sdf.select(self._internal.scol)
+        internal = _InternalFrame(
+            sdf=sdf, index_map=[(sdf.columns[0], self._internal.index_names[0])]
```
itholic: I'm just curious about this, why just don't use `self._internal.index_map` here rather create index_map again manually? 🤔 (i guess maybe do they works in the same way ?)

itholic:
LGTM, except for just one short question. 👍

databricks/koalas/indexes.py
ueshin: because `self._internal.scol` could have a different name.

databricks/koalas/indexes.py
```diff
@@ -1494,6 +1494,53 @@ def is_all_dates(self):
...
+        Index(['a', 'b', 'c'], dtype='object')
+        >>> idx.repeat(2)
+        Index(['a', 'b', 'c', 'a', 'b', 'c'], dtype='object')
+        """
+        if not isinstance(repeats, int):
+            raise ValueError("`repeats` argument must be integer, but got {}".format(type(repeats)))
+        elif repeats < 0:
+            raise ValueError("negative dimensions are not allowed")
+
+        sdf = self._internal.sdf.select(self._internal.scol)
```
itholic: Ah, wait. we can integrate with MultiIndex if we

1. `self._internal.scol` -> `self._internal.index_scols`
2. and use `self._internal.index_map` when create `index_map`
?

like this
```python
        if not isinstance(repeats, int):
            raise ValueError("`repeats` argument must be integer, but got {}".format(type(repeats)))
        elif repeats < 0:
            raise ValueError("negative dimensions are not allowed")

        sdf = self._internal.sdf.select(self._internal.index_scols)
        internal = _InternalFrame(
            sdf=sdf, index_map=self._internal.index_map
        )
        kdf = DataFrame(internal)  # type: DataFrame
        if repeats == 0:
            return DataFrame(kdf._internal.with_filter(F.lit(False))).index
        else:
            return ks.concat([kdf] * repeats).index
```
```

Comment (ueshin) [reply databricks/koalas/indexes.py]: `scol` in `Index` could be different from `_internal.index_scols[0]`.

Comment (itholic) [reply databricks/koalas/indexes.py]: @ueshin Ah, yeah i just checked Implementation of them. seems we don't need to use `index_scols[0]` for `Index`.

File: databricks/koalas/indexes.py (hunk 1 above)
Comment (itholic) [reply databricks/koalas/indexes.py]: @ueshin Ah.. I think found the case. Thanks!! :)

```python
>>> self = ks.Series(['a', 'b', 'c'], index=[1, 2, 3])
>>> sdf = self._internal.sdf.select(self._internal.scol)
>>> self._internal.index_map
[('__index_level_0__', None)]
>>> [(sdf.columns[0], self._internal.index_names[0])]
[('0', None)]
```

itholic:
Thanks for the replies, LGTM.

databricks/koalas/indexes.py
ueshin: Actually that's a `Series` case.
The example for `Index` is like:

```py
>>> kidx = ks.Index([1,2,3])
>>> kidx1 = (kidx + 1)
>>> kidx1._internal.index_scols, kidx1._internal.scol
([Column<b'__index_level_0__'>], Column<b'(__index_level_0__ + 1)'>)
>>> kidx1._internal.sdf.select(kidx1._internal.index_scols + [kidx1._internal.scol])
DataFrame[__index_level_0__: bigint, (__index_level_0__ + 1): bigint]
```

ueshin: But actually you raised a good point.
Maybe we should revisit the infrastructure of `Index`.
Let me think.

itholic: Oops... sorry I forgot extracting `Index` from `Series` when i tested.

Thanks for the example and I totally agree that we consolidate them! 😺

itholic: Ah, the example you replied is super good. Thanks again!!

PR #1329, opened 2020-03-06
Implement stack() for DataFrame.

### LINKED ISSUE #877: DataFrame.stack
Implement `DataFrame.stack`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.stack.html

ueshin:
Adding `stack()` for `DataFrame`.

- https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.stack.html

Resolves #877.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8100,7 +8100,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
             # FIXME: for Spark 2.3
             spark_type = self._internal.spark_type_for(label)
-            if new_label not in spark_types and not isinstance(spark_type, NullType):
+            if new_label not in spark_types:
                 spark_types[new_label] = spark_type
 
         column_labels = OrderedDict(sorted(column_labels.items(), key=lambda x: x[0]))
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -8017,11 +8017,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
              weight  height
         cat       0       1
         dog       2       3
-        >>> df_single_level_cols.stack()
-        cat  weight    0
-             height    1
-        dog  weight    2
-             height    3
+        >>> df_single_level_cols.stack().sort_index()
+        cat  height    1
+             weight    0
+        dog  height    3
+             weight    2
         Name: 0, dtype: int64
 
         **Multi level columns: simple case**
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -7965,6 +7967,211 @@ def melt(self, id_vars=None, value_vars=None, var_name=None, value_name="value")
...
+        """
+        from databricks.koalas.series import _col
+
+        if len(self._internal.column_labels) == 0:
+            return DataFrame(self._internal.with_filter(F.lit(False)))
+
+        column_labels = {}
+        index_values = []
+        returns_series = False
+        spark_types = {}  # FIXME: for Spark 2.3
```
HyukjinKwon: @ueshin, can we add a short comment about the problem here?

databricks/koalas/frame.py
```diff
@@ -7965,6 +7967,211 @@ def melt(self, id_vars=None, value_vars=None, var_name=None, value_name="value")
...
+            m      4.0     NaN
+        """
+        from databricks.koalas.series import _col
+
+        if len(self._internal.column_labels) == 0:
+            return DataFrame(self._internal.with_filter(F.lit(False)))
+
+        column_labels = {}
+        index_values = []
+        returns_series = False
```
HyukjinKwon: nit: `returns_series` -> `should_returns_series`

databricks/koalas/frame.py
```diff
@@ -7965,6 +7967,211 @@ def melt(self, id_vars=None, value_vars=None, var_name=None, value_name="value")
...
+        spark_types = {}  # FIXME: for Spark 2.3
+        for label in self._internal.column_labels:
+            new_label = label[:-1]
+            if len(new_label) == 0:
+                new_label = ("0",)
+                returns_series = True
+            value = label[-1]
+
+            scol = self._internal.scol_for(label)
+            if new_label not in column_labels:
```
HyukjinKwon: no big deal but we can use `defaultdict`:


```python
from collections import defaultdict
column_labels  = defaultdict(dict)
column_labels['a']['b'] = 'column'
print(list(column_labels.items()))
```

databricks/koalas/frame.py
```diff
@@ -7965,6 +7967,211 @@ def melt(self, id_vars=None, value_vars=None, var_name=None, value_name="value")
...
+                column_labels[new_label] = {value: scol}
+            else:
+                column_labels[new_label][value] = scol
+
+            if value not in index_values:
+                index_values.append(value)
```
HyukjinKwon: No big deal but we can use a set for `index_values` I believe.

databricks/koalas/frame.py
HyukjinKwon: Oh, I see. the type coercion problem in array of structs. Yes, it behaves differently. I didn't actually bother handling this case because we don't officially support spark 2.4 (https://koalas.readthedocs.io/en/latest/getting_started/install.html#dependencies). It might be fine to just remove this workaround. We can remove it together later when we remove Spark 2.3 in our CI too.

PR #1331, opened 2020-03-06
Refine _InternalFrame by renaming and leveraging OrderedDict

HyukjinKwon:
This PR proposes to refine and refactor `_InternalFrame`.

- `index_map`: uses `OrderedDict` instead of `List[Tuple[...]]`

</br>

- `column_name_for` -> `spark_column_name_for`
    Now, it purely searches by labels (and index name) instead of internal Spark column name when it searches index.
- `scol_for` -> `spark_column_for`

</br>

- `sdf` -> `spark_frame`
- `scol` -> `spark_column`
- `scols` -> `spark_columns`
- `columns ` -> `spark_column_names`

</br>

- `data_columns` -> `data_spark_column_names`
- `column_scols` -> `data_spark_columns`

</br>

- `index_columns` -> `index_spark_column_names`
- `index_scols` -> `index_spark_columns`

</br>

- `spark_internal_df` -> `to_internal_spark_frame`
- `spark_df` -> `to_external_spark_frame`
- `pandas_df` -> `to_pandas_frame`

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -674,14 +674,7 @@ class _InternalFrame(object):
 
     def spark_column_name_for(self, labels: Union[str, Tuple[str, ...]]) -> str:
         """ Return the actual Spark column name for the given column name or index. """
-        column_labels_to_name = dict(zip(self.column_labels, self.data_spark_column_names))
-        if labels in column_labels_to_name:
-            return column_labels_to_name[labels]  # type: ignore
-        else:
-            for index_spark_column_name, index_name in self.index_map.items():
-                if index_name == labels:
-                    return index_spark_column_name
-            raise KeyError(name_like_string(labels))
+        return self._sdf.select(self.spark_column_for(labels)).columns[0]
 
     def spark_column_for(self, labels: Union[str, Tuple[str, ...]]):
         """ Return Spark Column for the given column name or index. """
@@ -689,9 +682,7 @@ class _InternalFrame(object):
         if labels in column_labels_to_scol:
             return column_labels_to_scol[labels]  # type: ignore
         else:
-            if labels not in self.index_spark_column_names:
-                raise KeyError(name_like_string(labels))
-            return scol_for(self._sdf, self.spark_column_name_for(labels))
+            raise KeyError(name_like_string(labels))
 
     def spark_type_for(self, labels: Union[str, Tuple[str, ...]]) -> DataType:
         """ Return DataType for the given column name or index. """
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -672,20 +672,20 @@ class _InternalFrame(object):
             (sdf[offset_column] + sdf[row_number_column] - 1).alias(column_name), *scols
         )
 
-    def spark_column_name_for(self, labels: Union[str, Tuple[str, ...]]) -> str:
-        """ Return the actual Spark column name for the given column name or index. """
+    def spark_column_name_for(self, labels: Tuple[str, ...]) -> str:
+        """ Return the actual Spark column name for the given column name. """
         return self._sdf.select(self.spark_column_for(labels)).columns[0]
 
-    def spark_column_for(self, labels: Union[str, Tuple[str, ...]]):
-        """ Return Spark Column for the given column name or index. """
+    def spark_column_for(self, labels: Tuple[str, ...]):
+        """ Return Spark Column for the given column name. """
         column_labels_to_scol = dict(zip(self.column_labels, self.data_spark_columns))
         if labels in column_labels_to_scol:
             return column_labels_to_scol[labels]  # type: ignore
         else:
             raise KeyError(name_like_string(labels))
 
-    def spark_type_for(self, labels: Union[str, Tuple[str, ...]]) -> DataType:
-        """ Return DataType for the given column name or index. """
+    def spark_type_for(self, labels: Tuple[str, ...]) -> DataType:
+        """ Return DataType for the given column name. """
         return self._sdf.select(self.spark_column_for(labels)).schema[0].dataType
 
     @property
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -20,6 +20,7 @@ An internal immutable DataFrame with some metadata to manage indexes.
 import re
 from typing import Dict, List, Optional, Tuple, Union, TYPE_CHECKING
 from itertools import accumulate
+from collections import OrderedDict
 
 import numpy as np
 import pandas as pd
@@ -60,9 +61,7 @@ SPARK_INDEX_NAME_PATTERN = re.compile(r"__index_level_[0-9]+__")
 
 NATURAL_ORDER_COLUMN_NAME = "__natural_order__"
 
-HIDDEN_COLUMNS = set([NATURAL_ORDER_COLUMN_NAME])
-
-IndexMap = Tuple[str, Optional[Tuple[str, ...]]]
+HIDDEN_COLUMNS = {NATURAL_ORDER_COLUMN_NAME}
 
 
 class _InternalFrame(object):
@@ -93,7 +92,7 @@ class _InternalFrame(object):
     However, all columns including index column are also stored in Spark DataFrame internally
     as below.
 
-    >>> kdf._internal.spark_internal_df.show()  # doctest: +NORMALIZE_WHITESPACE
+    >>> kdf._internal.to_internal_spark_frame.show()  # doctest: +NORMALIZE_WHITESPACE
     +-----------------+---+---+---+---+---+
     |__index_level_0__|  A|  B|  C|  D|  E|
     +-----------------+---+---+---+---+---+
@@ -106,24 +105,30 @@ class _InternalFrame(object):
     In order to fill this gap, the current metadata is used by mapping Spark's internal column
     to Koalas' index. See the method below:
 
-    * `sdf` represents the internal Spark DataFrame
+    * `spark_frame` represents the internal Spark DataFrame
 
-    * `data_columns` represents non-indexing columns
+    * `data_spark_column_names` represents non-indexing Spark column names
 
-    * `index_columns` represents internal index columns
+    * `data_spark_columns` represents non-indexing Spark columns
 
-    * `columns` represents all columns
+    * `index_spark_column_names` represents internal index Spark column names
 
-    * `index_names` represents the external index name
+    * `index_spark_columns` represents internal index Spark columns
 
-    * `index_map` is zipped pairs of `index_columns` and `index_names`
+    * `spark_column_names` represents all columns
 
-    * `spark_df` represents Spark DataFrame derived by the metadata
+    * `index_names` represents the external index name as a label
 
-    * `pandas_df` represents pandas DataFrame derived by the metadata
+    * `index_map` is zipped pairs of `index_spark_column_names` and `index_names`
+
+    * `to_external_spark_frame` represents Spark DataFrame derived by the metadata.
+
+    * `to_internal_spark_frame` represents Spark DataFrame derived by the metadata. Includes index.
+
+    * `to_pandas_frame` represents pandas DataFrame derived by the metadata
 
     >>> internal = kdf._internal
-    >>> internal.sdf.show()  # doctest: +NORMALIZE_WHITESPACE +ELLIPSIS
+    >>> internal.spark_frame.show()  # doctest: +NORMALIZE_WHITESPACE +ELLIPSIS
     +-----------------+---+---+---+---+---+-----------------+
     |__index_level_0__|  A|  B|  C|  D|  E|__natural_order__|
     +-----------------+---+---+---+---+---+-----------------+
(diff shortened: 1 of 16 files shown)


databricks/koalas/internal.py
```diff
@@ -395,28 +401,29 @@ class _InternalFrame(object):
...
     def __init__(
         self,
-        sdf: spark.DataFrame,
-        index_map: Optional[List[IndexMap]],
+        spark_frame: spark.DataFrame,
+        index_map: Optional[Dict[str, Optional[Tuple[str, ...]]]],
         column_labels: Optional[List[Tuple[str, ...]]] = None,
-        column_scols: Optional[List[spark.Column]] = None,
+        data_spark_columns: Optional[List[spark.Column]] = None,
         column_label_names: Optional[List[str]] = None,
```
HyukjinKwon: For `column_labels` and `column_label_names`, I am currently thinking about reusing Pandas' Index directly.

HyukjinKwon: If that doesn't work, I plan to rename `column_labels` -> `data_column_labels`, `data_column_label_names` -> `data_column_label_names`

databricks/koalas/internal.py
```diff
@@ -60,9 +61,7 @@
 
 NATURAL_ORDER_COLUMN_NAME = "__natural_order__"
 
-HIDDEN_COLUMNS = set([NATURAL_ORDER_COLUMN_NAME])
-
-IndexMap = Tuple[str, Optional[Tuple[str, ...]]]
+HIDDEN_COLUMNS = {NATURAL_ORDER_COLUMN_NAME}
 
 
 class _InternalFrame(object):
```
HyukjinKwon: I also currently plan to add `_InternalColumn` for Series, Index and MultiIndex, which inherits `_InternalFrame`

databricks/koalas/internal.py
HyukjinKwon: If that doesn't work, I plan to rename something like `column_index_labels` and `column_index_names`.

HyukjinKwon:
So, I currently target the constructor in `_InternalFrame` as below:

**_InternalFrame:**

```python
    def __init__(
        self,
        spark_frame,  # Spark DataFrame
        data_spark_columns,  # Spark Columns
        index_map,  # Pandas <> Spark (row) Index
        column_index,  # Pandas <> Spark column index
```

**_InternalColumn:**


```python
    def __init__(
        self,
        spark_frame,  # Spark DataFrame
        spark_column,  # Spark Column
        index_map,  # Pandas <> Spark (row) Index
        column_index,  # Pandas <> Spark column index
```

but details might be changed ...

databricks/koalas/internal.py
```diff
@@ -657,96 +672,89 @@ def offset(id):
...
-            if column_labels_or_index_column not in self.index_columns:
-                raise KeyError(name_like_string(column_labels_or_index_column))
-            return column_labels_or_index_column  # type: ignore
-
-    @lazy_property
-    def _column_labels_to_scol(self) -> Dict[Tuple[str, ...], spark.Column]:
-        return dict(zip(self.column_labels, self.column_scols))
+            for index_spark_column_name, index_name in self.index_map.items():
+                if index_name == labels:
+                    return index_spark_column_name
```
ueshin: Actually I disagree that this method searches by labels (and index name) because multiple index names could be `None`, then it will return a wrong column name.

databricks/koalas/internal.py
```diff
@@ -657,96 +672,89 @@ def offset(id):
...
+        column_labels_to_scol = dict(zip(self.column_labels, self.data_spark_columns))
+        if labels in column_labels_to_scol:
+            return column_labels_to_scol[labels]  # type: ignore
         else:
-            if column_labels_or_index_column not in self.index_columns:
-                raise KeyError(name_like_string(column_labels_or_index_column))
-            return scol_for(self._sdf, self.column_name_for(column_labels_or_index_column))
+            if labels not in self.index_spark_column_names:
+                raise KeyError(name_like_string(labels))
+            return scol_for(self._sdf, self.spark_column_name_for(labels))
```
ueshin: We can't get index spark column properly.

Is this `label` for index, string as index column name or tuple as index name?
If string, it will fail when `self.spark_column_name_for(labels)`, otherwise it will fail when `labels not in self.index_spark_column_names`.

databricks/koalas/internal.py
```diff
@@ -657,96 +672,89 @@ def offset(id):
...
 
-    def spark_type_for(
-        self, column_labels_or_index_column: Union[str, Tuple[str, ...]]
-    ) -> DataType:
+    def spark_type_for(self, labels: Union[str, Tuple[str, ...]]) -> DataType:
         """ Return DataType for the given column name or index. """
-        return self._sdf.select(self.scol_for(column_labels_or_index_column)).schema[0].dataType
+        return self._sdf.select(self.spark_column_for(labels)).schema[0].dataType
```
ueshin: ditto.

databricks/koalas/internal.py
HyukjinKwon: True.. I will better have to remove this branch.

databricks/koalas/internal.py
HyukjinKwon: I was thinking `spark_column_for` should only search based on labels or index name, not the actual Spark name. I thought the previous version mixes both external (pandas-like labels) and internal (Spark column name). For example, for regular series, it requires labels to search instead of the physical column names.

The current version, it searches only by pandas-like labels or pandas-like index name now. For Spark columns, it has to be manually searched against Spark column names via:

```diff
- kdf._internal.scol_for(index_spark_column_name)
+ scol_for(kdf.spark_frame, index_spark_column_name)
```

For pandas-like search via labels or index name, it has to always be:

```python
kdf._internal.scol_for(labels_or_index_name)
```

Does this make sense to you?

databricks/koalas/internal.py
HyukjinKwon: Let me reply on here https://github.com/databricks/koalas/pull/1331/files/13fd0d289a7396c3adc0b78433893040be7202c8#r389330455. Seems it's the similar topic.

HyukjinKwon: Maybe I should throw an exception if multiple columns are selected ...

databricks/koalas/internal.py
```diff
@@ -395,28 +401,29 @@ class _InternalFrame(object):
 
     def __init__(
         self,
-        sdf: spark.DataFrame,
-        index_map: Optional[List[IndexMap]],
+        spark_frame: spark.DataFrame,
+        index_map: Optional[Dict[str, Optional[Tuple[str, ...]]]],
```
HyukjinKwon: BTW, `OrderedDict` seems not working as a type hint. So I had to use `Dict`.

databricks/koalas/internal.py
```diff
@@ -463,22 +470,26 @@ def __init__(
...
+            index_map = OrderedDict({SPARK_DEFAULT_INDEX_NAME: None})
 
-        if NATURAL_ORDER_COLUMN_NAME not in sdf.columns:
-            sdf = sdf.withColumn(NATURAL_ORDER_COLUMN_NAME, F.monotonically_increasing_id())
+        if NATURAL_ORDER_COLUMN_NAME not in spark_frame.columns:
+            spark_frame = spark_frame.withColumn(
+                NATURAL_ORDER_COLUMN_NAME, F.monotonically_increasing_id()
+            )
 
+        assert isinstance(index_map, OrderedDict), index_map
```
HyukjinKwon: BTW, `OrderedDict` seems not working as a type hint. So I had to use `Dict`, and add an `assert` here.

databricks/koalas/internal.py
ueshin: The fact that the current change doesn't cause test failures makes me wonder whether we need to keep support index columns in these three methods. I mean, those three methods are only for looking for data columns by the column labels.

ueshin: The following change doesn't cause test failure. We don't use `spark_column_name_for`, `spark_column_for`, and `spark_type_for` for looking for index columns at all with the current changes.

```diff
$ git diff
diff --git a/databricks/koalas/internal.py b/databricks/koalas/internal.py
index 6c09033..b9a1590 100644
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -674,14 +674,7 @@ class _InternalFrame(object):

     def spark_column_name_for(self, labels: Union[str, Tuple[str, ...]]) -> str:
         """ Return the actual Spark column name for the given column name or index. """
-        column_labels_to_name = dict(zip(self.column_labels, self.data_spark_column_names))
-        if labels in column_labels_to_name:
-            return column_labels_to_name[labels]  # type: ignore
-        else:
-            for index_spark_column_name, index_name in self.index_map.items():
-                if index_name == labels:
-                    return index_spark_column_name
-            raise KeyError(name_like_string(labels))
+        return self._sdf.select(self.spark_column_for(labels)).columns[0]

     def spark_column_for(self, labels: Union[str, Tuple[str, ...]]):
         """ Return Spark Column for the given column name or index. """
@@ -689,9 +682,7 @@ class _InternalFrame(object):
         if labels in column_labels_to_scol:
             return column_labels_to_scol[labels]  # type: ignore
         else:
-            if labels not in self.index_spark_column_names:
-                raise KeyError(name_like_string(labels))
-            return scol_for(self._sdf, self.spark_column_name_for(labels))
+            raise KeyError(name_like_string(labels))

     def spark_type_for(self, labels: Union[str, Tuple[str, ...]]) -> DataType:
         """ Return DataType for the given column name or index. """
```

How about we drop support for looking for index columns with these methods?
Maybe the way to look for index columns is different from for data columns, then we should have another method for it?

HyukjinKwon: +100!

databricks/koalas/internal.py
```diff
@@ -657,96 +672,80 @@ def offset(id):
             (sdf[offset_column] + sdf[row_number_column] - 1).alias(column_name), *scols
         )
 
-    @lazy_property
-    def _column_labels_to_name(self) -> Dict[Tuple[str, ...], str]:
-        return dict(zip(self.column_labels, self.data_columns))
-
-    def column_name_for(self, column_labels_or_index_column: Union[str, Tuple[str, ...]]) -> str:
+    def spark_column_name_for(self, labels: Union[str, Tuple[str, ...]]) -> str:
```
ueshin: `label: Tuple[str, ...]`?

databricks/koalas/internal.py
```diff
@@ -657,96 +672,80 @@ def offset(id):
...
-                raise KeyError(name_like_string(column_labels_or_index_column))
-            return column_labels_or_index_column  # type: ignore
-
-    @lazy_property
-    def _column_labels_to_scol(self) -> Dict[Tuple[str, ...], spark.Column]:
-        return dict(zip(self.column_labels, self.column_scols))
+        return self._sdf.select(self.spark_column_for(labels)).columns[0]
 
-    def scol_for(self, column_labels_or_index_column: Union[str, Tuple[str, ...]]):
+    def spark_column_for(self, labels: Union[str, Tuple[str, ...]]):
```
ueshin: ditto.

databricks/koalas/internal.py
```diff
@@ -657,96 +672,80 @@ def offset(id):
...
         else:
-            if column_labels_or_index_column not in self.index_columns:
-                raise KeyError(name_like_string(column_labels_or_index_column))
-            return scol_for(self._sdf, self.column_name_for(column_labels_or_index_column))
+            raise KeyError(name_like_string(labels))
 
-    def spark_type_for(
-        self, column_labels_or_index_column: Union[str, Tuple[str, ...]]
-    ) -> DataType:
+    def spark_type_for(self, labels: Union[str, Tuple[str, ...]]) -> DataType:
```
ueshin: ditto.

databricks/koalas/internal.py
```diff
@@ -657,96 +672,80 @@ def offset(id):
             (sdf[offset_column] + sdf[row_number_column] - 1).alias(column_name), *scols
         )
 
-    @lazy_property
-    def _column_labels_to_name(self) -> Dict[Tuple[str, ...], str]:
-        return dict(zip(self.column_labels, self.data_columns))
-
-    def column_name_for(self, column_labels_or_index_column: Union[str, Tuple[str, ...]]) -> str:
+    def spark_column_name_for(self, labels: Union[str, Tuple[str, ...]]) -> str:
         """ Return the actual Spark column name for the given column name or index. """
```
ueshin: We should remove `or index`?

databricks/koalas/internal.py
```diff
@@ -657,96 +672,80 @@ def offset(id):
...
-            return column_labels_or_index_column  # type: ignore
-
-    @lazy_property
-    def _column_labels_to_scol(self) -> Dict[Tuple[str, ...], spark.Column]:
-        return dict(zip(self.column_labels, self.column_scols))
+        return self._sdf.select(self.spark_column_for(labels)).columns[0]
 
-    def scol_for(self, column_labels_or_index_column: Union[str, Tuple[str, ...]]):
+    def spark_column_for(self, labels: Union[str, Tuple[str, ...]]):
         """ Return Spark Column for the given column name or index. """
```
ueshin: ditto.

databricks/koalas/internal.py
```diff
@@ -657,96 +672,80 @@ def offset(id):
...
-            if column_labels_or_index_column not in self.index_columns:
-                raise KeyError(name_like_string(column_labels_or_index_column))
-            return scol_for(self._sdf, self.column_name_for(column_labels_or_index_column))
+            raise KeyError(name_like_string(labels))
 
-    def spark_type_for(
-        self, column_labels_or_index_column: Union[str, Tuple[str, ...]]
-    ) -> DataType:
+    def spark_type_for(self, labels: Union[str, Tuple[str, ...]]) -> DataType:
         """ Return DataType for the given column name or index. """
```
ueshin: ditto.

HyukjinKwon:
Last changes are only comments and type hints. Linter passed so I will just merge.

PR #1334, opened 2020-03-11
Fix Index._with_new_scol to update its anchor.

### LINKED ISSUE #1190: Fix the index of `Index.to_series()`.
The index of `Index.to_series()` doesn't address the change of the index itself.

E.g.,

```py
>>> pidx
Int64Index([1, 2, 3], dtype='int64')
>>> (pidx + 1).to_series()
2    2
3    3
4    4
dtype: int64
```

whereas

```py
>>> kidx
Int64Index([1, 2, 3], dtype='int64')
>>> (kidx + 1).to_series()
1    2
2    3
3    4
Name: 0, dtype: int64
```

ueshin:
As discussed at https://github.com/databricks/koalas/pull/1328#discussion_r388681347, we can consolidate the `Index.repeat()` and `MultiIndex.repeat()` by updating its anchor with a new scol.
Also resolves #1190 and other functions which were mistakenly using `index_scols` in `Index`, whose usage is valid now.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -99,26 +99,16 @@ class Index(IndexOpsMixin):
     Index(['a', 'b', 'c'], dtype='object')
     """
 
-    def __init__(
-        self,
-        data: Union[DataFrame, list],
-        dtype=None,
-        name=None,
-        scol: Optional[spark.Column] = None,
-    ) -> None:
+    def __init__(self, data: Union[DataFrame, list], dtype=None, name=None) -> None:
         if isinstance(data, DataFrame):
             assert dtype is None
             assert name is None
             kdf = data
         else:
-            assert scol is None
             kdf = DataFrame(index=pd.Index(data=data, dtype=dtype, name=name))
-        if scol is None:
-            scol = kdf._internal.index_spark_columns[0]
         internal = kdf._internal.copy(
-            spark_column=scol,
+            spark_column=kdf._internal.index_spark_columns[0],
             column_labels=kdf._internal.index_names,
-            data_spark_columns=kdf._internal.index_spark_columns,
             column_label_names=None,
         )
         IndexOpsMixin.__init__(self, internal, kdf)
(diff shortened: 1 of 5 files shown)


databricks/koalas/indexes.py
```diff
@@ -130,7 +120,11 @@ def _with_new_scol(self, scol: spark.Column) -> "Index":
         :param scol: the new Spark Column
         :return: the copied Index
         """
-        return Index(self._kdf, scol=scol)
+        sdf = self._internal.spark_frame.select(scol)
+        internal = _InternalFrame(
+            spark_frame=sdf, index_map=OrderedDict(zip(sdf.columns, self._internal.index_names))
+        )
+        return DataFrame(internal).index
```
HyukjinKwon: Hm .. I think it will trigger the operations on different DataFrames among the operations in Index .. e.g.)

```python
index = ks.range(100).index
index + index + index
```

There looks another issue in `compute.ops_on_diff_frames` ...  operations against the indexes from different frames are not supported ..

ueshin: Good point.

ueshin: Good point.

The way we manage index Spark columns is different from Series, so we need further refactoring in `_InternalFrame`, e.g., we should manage them as Spark columns instead of Spark column names. The way in Series is not good enough yet, though.

Actually creating new DataFrame as an anchor is better than now compared to pandas behavior:

```py
>>> pdf = pd.DataFrame({'a':[1,2,3]})
>>> pidx = pdf.index
>>> pidx.name = 'i'
>>> pdf
   a
i
0  1
1  2
2  3
>>> (pidx + pidx).name = 'ii'
>>> pdf
   a
i
0  1
1  2
2  3
```

When operating `pidx + pidx`, the connection to the original `pdf` seems dropped.

Whereas:

```py
>>> kdf = ks.DataFrame({'a':[1,2,3]})
>>> kidx = kdf.index
>>> kidx.name = 'i'
>>> kdf
   a
i
0  1
1  2
2  3
>>> (kidx + kidx).name = 'ii'
>>> kdf
    a
ii
0   1
1   2
2   3
```

It updates index name even after `kidx + kidx`.

PR #1343, opened 2020-03-13
Explicitly don't support to disable numeric_only in stats APIs at DataFrame

HyukjinKwon:
Spark can't contain mixed types in its output. Statistic APIs in DataFrame outputs it. This PR explicitly disallow `numeric_only` or mention the limitation by it.


HyukjinKwon:
Thanks, @ueshin

PR #1347, opened 2020-03-17
Make nunique in DataFrame to return a Koalas DataFrame instead of pandas'

HyukjinKwon:
The approach is basically similar with `DataFrame._reduce_for_stat_function`.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3504,7 +3504,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         dropna: bool = True,
         approx: bool = False,
         rsd: float = 0.05,
-    ) -> pd.Series:
+    ) -> "ks.Series":
         """
         Return number of unique elements in the object.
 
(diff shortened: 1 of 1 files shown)


databricks/koalas/frame.py
```diff
@@ -3527,7 +3527,7 @@ def nunique(
 
         Returns
         -------
-        The number of unique values per column as a pandas Series.
+        The number of unique values per column as a koalas Series.
```
itholic: nit: `koalas` -> `Koalas`

databricks/koalas/frame.py
```diff
@@ -3550,22 +3550,31 @@ def nunique(
...
-        return res.T.iloc[:, 0]
+        )
+
+        # The data is expected to be small so it's fine to transpose/use default index.
+        with ks.option_context(
+            "compute.default_index_type", "distributed", "compute.max_rows", None
+        ):
+            kdf = DataFrame(sdf)  # type: ks.DataFrame
+            internal = _InternalFrame(
+                kdf._internal.spark_frame,
```
itholic: Not sure, but maybe `spark_frame=kdf._internal.spark_frame` is better?

itholic:
LGTM except 2 nit comments 👍

ueshin:
Thanks! merging.

PR #1348, opened 2020-03-17
Implements union() for Index & MultiIndex

itholic:
This PR proposes `Index.union` & `MultiIndex.union`
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.union.html#pandas.Index.union)

- Index
```python
>>> idx1 = ks.Index([1, 2, 3, 4])
>>> idx2 = ks.Index([3, 4, 5, 6])
>>> idx1.union(idx2).sort_values()
Int64Index([1, 2, 3, 4, 5, 6], dtype='int64')
```

- MultiIndex
```python
>>> midx1 = ks.MultiIndex.from_tuples([("x", "a"), ("x", "b"), ("x", "c"), ("x", "d")])
>>> midx2 = ks.MultiIndex.from_tuples([("x", "c"), ("x", "d"), ("x", "e"), ("x", "f")])
>>> midx1.union(midx2).sort_values()
MultiIndex([('x', 'a'),
            ('x', 'b'),
            ('x', 'c'),
            ('x', 'd'),
            ('x', 'e'),
            ('x', 'f')],
           )
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1677,7 +1677,7 @@ class Index(IndexOpsMixin):
             other = MultiIndex.from_tuples(other)
         if not isinstance(other, Index) and not isinstance(self, MultiIndex):
             if isinstance(other, Series):
-                other = other.values
+                other = other.to_numpy()
             if isinstance(other, DataFrame):
                 raise ValueError("Index data must be 1-dimensional")
             other = Index(other)
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1666,6 +1666,19 @@ class Index(IndexOpsMixin):
         >>> idx2 = ks.Index([3, 4, 5, 6])
         >>> idx1.union(idx2).sort_values()
         Int64Index([1, 2, 3, 4, 5, 6], dtype='int64')
+
+        MultiIndex
+
+        >>> midx1 = ks.MultiIndex.from_tuples([("x", "a"), ("x", "b"), ("x", "c"), ("x", "d")])
+        >>> midx2 = ks.MultiIndex.from_tuples([("x", "c"), ("x", "d"), ("x", "e"), ("x", "f")])
+        >>> midx1.union(midx2).sort_values()  # doctest: +SKIP
+        MultiIndex([('x', 'a'),
+                    ('x', 'b'),
+                    ('x', 'c'),
+                    ('x', 'd'),
+                    ('x', 'e'),
+                    ('x', 'f')],
+                   )
         """
         if not isinstance(other, Index) and isinstance(self, MultiIndex):
             if isinstance(other, list) and not all([isinstance(item, tuple) for item in other]):
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1672,18 +1672,20 @@ class Index(IndexOpsMixin):
         sort = True if sort is None else sort
         sort = validate_bool_kwarg(sort, "sort")
         if not isinstance(other, Index) and isinstance(self, MultiIndex):
-            if isinstance(other, list) and not all([isinstance(item, tuple) for item in other]):
+            if not isinstance(other, list) or not all([isinstance(item, tuple) for item in other]):
                 raise TypeError("other must be a MultiIndex or a list of tuples")
             other = MultiIndex.from_tuples(other)
         if not isinstance(other, Index) and not isinstance(self, MultiIndex):
+            if isinstance(other, Series):
+                other = other.values
+            if isinstance(other, DataFrame):
+                raise ValueError("Index data must be 1-dimensional")
             other = Index(other)
         if type(self) is not type(other):
             # TODO: We can't support different type of values in a single column for now.
             raise NotImplementedError("Union between Index and MultiIndex is not yet supported")
         sdf_self = self._internal._sdf.select(self._internal.index_spark_columns)
-        sdf_other = other._internal._sdf.select(
-            other._internal.index_spark_columns
-        ).drop_duplicates()
+        sdf_other = other._internal._sdf.select(other._internal.index_spark_columns)
         sdf = sdf_self.union(sdf_other.subtract(sdf_self))
         if isinstance(self, MultiIndex):
             sdf = sdf.drop_duplicates()
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1684,7 +1684,7 @@ class Index(IndexOpsMixin):
                     raise NotImplementedError(
                         "Union between Index and MultiIndex is not yet supported"
                     )
-                if isinstance(other, Series):
+                elif isinstance(other, Series):
                     other = other.to_frame().set_index(other.name).index
                 elif isinstance(other, DataFrame):
                     raise ValueError("Index data must be 1-dimensional")
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1649,7 +1649,7 @@ class Index(IndexOpsMixin):
         Examples
         --------
 
-        Union matching dtypes
+        Index
 
         >>> idx1 = ks.Index([1, 2, 3, 4])
         >>> idx2 = ks.Index([3, 4, 5, 6])
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1672,13 +1672,13 @@ class Index(IndexOpsMixin):
         sort = True if sort is None else sort
         sort = validate_bool_kwarg(sort, "sort")
         if type(self) is not type(other):
-            if type(self) is MultiIndex:
+            if isinstance(self, MultiIndex):
                 if not isinstance(other, list) or not all(
                     [isinstance(item, tuple) for item in other]
                 ):
                     raise TypeError("other must be a MultiIndex or a list of tuples")
                 other = MultiIndex.from_tuples(other)
-            elif type(self) is Index:
+            else:
                 if isinstance(other, MultiIndex):
                     # TODO: We can't support different type of values in a single column for now.
                     raise NotImplementedError(
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1679,15 +1679,17 @@ class Index(IndexOpsMixin):
                     raise TypeError("other must be a MultiIndex or a list of tuples")
                 other = MultiIndex.from_tuples(other)
             elif type(self) is Index:
+                if isinstance(other, MultiIndex):
+                    # TODO: We can't support different type of values in a single column for now.
+                    raise NotImplementedError(
+                        "Union between Index and MultiIndex is not yet supported"
+                    )
                 if isinstance(other, Series):
                     other = other.to_frame().set_index(other.name).index
                 elif isinstance(other, DataFrame):
                     raise ValueError("Index data must be 1-dimensional")
                 else:
                     other = Index(other)
-            else:
-                # TODO: We can't support different type of values in a single column for now.
-                raise NotImplementedError("Union between Index and MultiIndex is not yet supported")
         sdf_self = self._internal._sdf.select(self._internal.index_spark_columns)
         sdf_other = other._internal._sdf.select(other._internal.index_spark_columns)
         sdf = sdf_self.union(sdf_other.subtract(sdf_self))
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1671,20 +1671,23 @@ class Index(IndexOpsMixin):
         """
         sort = True if sort is None else sort
         sort = validate_bool_kwarg(sort, "sort")
-        if type(self) is MultiIndex and not isinstance(other, Index):
-            if not isinstance(other, list) or not all([isinstance(item, tuple) for item in other]):
-                raise TypeError("other must be a MultiIndex or a list of tuples")
-            other = MultiIndex.from_tuples(other)
-        elif type(self) is Index and not isinstance(other, Index):
-            if isinstance(other, Series):
-                other = other.to_frame().set_index(other.name).index
-            elif isinstance(other, DataFrame):
-                raise ValueError("Index data must be 1-dimensional")
-            else:
-                other = Index(other)
         if type(self) is not type(other):
-            # TODO: We can't support different type of values in a single column for now.
-            raise NotImplementedError("Union between Index and MultiIndex is not yet supported")
+            if type(self) is MultiIndex:
+                if not isinstance(other, list) or not all(
+                    [isinstance(item, tuple) for item in other]
+                ):
+                    raise TypeError("other must be a MultiIndex or a list of tuples")
+                other = MultiIndex.from_tuples(other)
+            elif type(self) is Index:
+                if isinstance(other, Series):
+                    other = other.to_frame().set_index(other.name).index
+                elif isinstance(other, DataFrame):
+                    raise ValueError("Index data must be 1-dimensional")
+                else:
+                    other = Index(other)
+            else:
+                # TODO: We can't support different type of values in a single column for now.
+                raise NotImplementedError("Union between Index and MultiIndex is not yet supported")
         sdf_self = self._internal._sdf.select(self._internal.index_spark_columns)
         sdf_other = other._internal._sdf.select(other._internal.index_spark_columns)
         sdf = sdf_self.union(sdf_other.subtract(sdf_self))
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1631,6 +1631,62 @@ class Index(IndexOpsMixin):
         else:
             return ks.concat([kdf] * repeats).index
 
+    def union(self, other, sort=None):
+        """
+        Form the union of two Index objects.
+
+        If the Index objects are incompatible, both Index objects will be
+        cast to dtype('object') first.
+
+        Parameters
+        ----------
+        other : Index or array-like
+        sort : bool or None, default None
+            Whether to sort the resulting Index.
+
+            * None : Sort the result, except when
+
+              1. `self` and `other` are equal.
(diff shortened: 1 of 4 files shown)


itholic:
Tests are failing in pandas 2.4.0. Let me check more test in various versions of pandas.

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,67 @@ def repeat(self, repeats: int) -> "Index":
...
+                    ('x', 'e'),
+                    ('x', 'f')],
+                   )
+        """
+        sort = True if sort is None else sort
+        sort = validate_bool_kwarg(sort, "sort")
+        if not isinstance(other, Index) and isinstance(self, MultiIndex):
+            if isinstance(other, list) and not all([isinstance(item, tuple) for item in other]):
+                raise TypeError("other must be a MultiIndex or a list of tuples")
+            other = MultiIndex.from_tuples(other)
```
ueshin: Could you explain what you are checking here?

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,67 @@ def repeat(self, repeats: int) -> "Index":
...
+        if not isinstance(other, Index) and not isinstance(self, MultiIndex):
+            other = Index(other)
```

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,67 @@ def repeat(self, repeats: int) -> "Index":
...
+        if type(self) is not type(other):
+            # TODO: We can't support different type of values in a single column for now.
+            raise NotImplementedError("Union between Index and MultiIndex is not yet supported")
+        sdf_self = self._internal._sdf.select(self._internal.index_spark_columns)
+        sdf_other = other._internal._sdf.select(
+            other._internal.index_spark_columns
+        ).drop_duplicates()
+        sdf = sdf_self.union(sdf_other.subtract(sdf_self))
```
ueshin: Do we need such a complex plan? when?

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,67 @@ def repeat(self, repeats: int) -> "Index":
...
+        if isinstance(self, MultiIndex):
+            sdf = sdf.drop_duplicates()
```
ueshin: Why only for `MultiIndex`?

databricks/koalas/tests/test_indexes.py
```diff
@@ -1001,3 +1001,64 @@ def test_unique(self):
...
+        if LooseVersion(pd.__version__) >= LooseVersion("0.24"):
+            # The `sort` argument is added in pandas 0.24.
+            self.assert_eq(kidx1.union(kidx2, sort=None), pidx1.union(pidx2, sort=None))
+            self.assert_eq(kidx2.union(kidx1, sort=None), pidx2.union(pidx1, sort=None))
+            self.assert_eq(
+                kidx1.union([3, 4, 5, 6], sort=None), pidx1.union([3, 4, 5, 6], sort=None),
+            )
+            self.assert_eq(
+                kidx2.union([1, 2, 3, 4], sort=None), pidx2.union([1, 2, 3, 4], sort=None),
+            )
```
ueshin: Since the default value for `sort` is `None`, do we need these?
Don't we need to check with other values?

databricks/koalas/tests/test_indexes.py
```diff
@@ -1001,3 +1001,64 @@ def test_unique(self):
...
+            self.assert_eq(kmidx1.union(kmidx2, sort=None), pmidx1.union(pmidx2, sort=None))
+            self.assert_eq(kmidx2.union(kmidx1, sort=None), pmidx2.union(pmidx1, sort=None))
+            self.assert_eq(
+                kmidx1.union([("x", "a"), ("x", "b"), ("x", "c"), ("x", "d")], sort=None),
+                pmidx1.union([("x", "a"), ("x", "b"), ("x", "c"), ("x", "d")], sort=None),
+            )
+            self.assert_eq(
+                kmidx2.union([("x", "a"), ("x", "b"), ("x", "a"), ("x", "b")], sort=None),
+                pmidx2.union([("x", "a"), ("x", "b"), ("x", "a"), ("x", "b")], sort=None),
+            )
```
ueshin: ditto.

databricks/koalas/tests/test_indexes.py
```diff
@@ -1001,3 +1001,64 @@ def test_unique(self):
...
         self.assert_eq(kmidx.unique().sort_values(), pmidx.unique().sort_values())
+
+    def test_union(self):
+        # Index
+        pidx1 = pd.Index([1, 2, 3, 4])
+        pidx2 = pd.Index([3, 4, 5, 6])
+        kidx1 = ks.from_pandas(pidx1)
+        kidx2 = ks.from_pandas(pidx2)
+
+        self.assert_eq(kidx1.union(kidx2).sort_values(), pidx1.union(pidx2).sort_values())
```
ueshin: The values must be sorted by default, do we need `sort_values()`?

databricks/koalas/indexes.py
itholic: When `self` is `MultiIndex`, `other` also should be `MultiIndex` or the list of tuples.

If so, convert `other` into `MultiIndex` by using `from_tuples`

databricks/koalas/indexes.py
itholic: If type of `self` is `Index` while `other` is not any type of Index (e.g. `list`, `tuple`, `doct`), then we convert `other` into `Index` by Index constructer.

databricks/koalas/indexes.py
ueshin: then maybe the line 1675 must be:

```py
if not isinstance(other, list) or not all([isinstance(item, tuple) for item in other]):
```

?

databricks/koalas/indexes.py
itholic: Let me show you some examples of `union` in pandas.

If there are duplicated values ​​in `other` which among the common values between `self` and `other`, pandas just doesn't use those values. So I use `sdf_other.subtract(sdf_self)` first.

```python
>>> idx1 = pd.Index([1, 2, 3, 4])
>>> idx2 = pd.Index([3, 4, 3, 4, 3, 4, 5, 6])
>>> idx1.union(idx2)
Int64Index([1, 2, 3, 4, 5, 6], dtype='int64')
```

But if there are duplicated values in `self` which among the common value between `self` and `other`, pandas use them all. So I just union them. (self and subtracted other)

```python
>>> idx1 = pd.Index([1, 2, 3, 4, 3, 4, 3, 4])
>>> idx2 = pd.Index([3, 4, 5, 6])
>>> idx1.union(idx2)
Int64Index([1, 2, 3, 3, 3, 4, 4, 4, 5, 6], dtype='int64')
```

databricks/koalas/indexes.py
ueshin: then don't we need to check the actual type of `other` before we pass it to `Index()`? e.g., what if the `other` is `ks.DataFrame` or `ks.Series`?

databricks/koalas/indexes.py
itholic: pandas has different behavior between `Index` and `MultiIndex`.

For `Index`, they use all of `self` even they are duplicated.

```python
>>> idx1 = pd.Index([1, 2, 3, 4, 3, 4, 3, 4])
>>> idx2 = pd.Index([3, 4, 5, 6])
>>> idx1.union(idx2)
Int64Index([1, 2, 3, 3, 3, 4, 4, 4, 5, 6], dtype='int64')
```

But for `MultiIndex`, they get rid of the duplicated values from `self`.

```python
>>> pmidx1 = pd.MultiIndex.from_tuples([(1, 1), (1, 2), (1, 3), (1, 4), (1, 3), (1, 4)])
>>> pmidx2 = pd.MultiIndex.from_tuples([(1, 3), (1, 4), (1, 5), (1, 6)])
>>> pmidx1.union(pmidx2)
MultiIndex([(1, 1),
            (1, 2),
            (1, 3),
            (1, 4),
            (1, 5),
            (1, 6)],
           )
```

That is the reason i just use `drop_duplicates()` for `MultiIndex` only

databricks/koalas/tests/test_indexes.py
itholic: Ah, I think that's my mistake.

Maybe we should use `False` as the default value unlike pandas since the cost issue, right?

databricks/koalas/tests/test_indexes.py
itholic: This is related with https://github.com/databricks/koalas/pull/1348#discussion_r394047937 .

I think we better use `False` as default, what do you think ??

databricks/koalas/tests/test_indexes.py
itholic: yes, ditto in comment above

itholic:
@ueshin , Thanks for the review this.

databricks/koalas/indexes.py
ueshin: Thanks for the examples. Seems like it's a behavior change in pandas>=1.0.
Shall we add the example in the tests?

btw, I guess we don't need `drop_duplicates()` in line 1686:

```py
>>> idx1
Int64Index([1, 2, 3, 4, 3, 4, 3, 4], dtype='int64')
>>> idx2
Int64Index([3, 4, 3, 4, 3, 4, 3, 4, 5, 6, 5, 6, 5, 6], dtype='int64')
>>> idx1.union(idx2)
Int64Index([1, 2, 3, 3, 3, 4, 4, 4, 5, 5, 5, 6, 6, 6], dtype='int64')
```

databricks/koalas/indexes.py
ueshin: I see. Shall we add the example in the test?

databricks/koalas/tests/test_indexes.py
ueshin: If we have `sort` argument (or something the same meaning), we usually follow pandas behavior.

databricks/koalas/indexes.py
itholic: @ueshin That's right. The behavior has changed since pandas>=1.0.

Okay, then I'm gonna add the tests again and make them run when only pandas>=1.0 .
(Actually, I removed them in https://github.com/databricks/koalas/pull/1348/commits/16a0a44f6ec2c8b4765535568b8c644a7b1b19c1 to compatibility with pandas<1.0 at the first time 😹)

itholic: > btw, I guess we don't need `drop_duplicates()` in line 1686:

Oh, It looks so, I'll fix them. Thanks for the comment!

databricks/koalas/indexes.py
itholic: @ueshin Sure, I'll add.

(Anyway, sometimes It is really hard to understand of pandas' behavior 😂)

databricks/koalas/tests/test_indexes.py
itholic: @ueshin Ah, okay i got it. I can manage this.

Thanks for the comment :)

databricks/koalas/indexes.py
itholic: My mistake, Thanks for fixing

databricks/koalas/indexes.py
itholic: Ah, I missed it. Thanks! :)

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,69 @@ def repeat(self, repeats: int) -> "Index":
...
+        """
+        sort = True if sort is None else sort
+        sort = validate_bool_kwarg(sort, "sort")
+        if not isinstance(other, Index) and isinstance(self, MultiIndex):
+            if not isinstance(other, list) or not all([isinstance(item, tuple) for item in other]):
+                raise TypeError("other must be a MultiIndex or a list of tuples")
+            other = MultiIndex.from_tuples(other)
+        if not isinstance(other, Index) and not isinstance(self, MultiIndex):
+            if isinstance(other, Series):
+                other = other.values
```
HyukjinKwon: I think we should use `to_numpy()` instead.

HyukjinKwon: Actually can we convert Series to Index without collecting the data into driver side?

itholic: I think there is no workaround converting from `Index` to `Series` without collecting for now.

I think maybe we can add warning to docstring or don't support Series for now?

ueshin: How about
other.to_frame().set_index(other.name).index

ueshin: How about
other.to_frame().set_index(other.name).index

itholic: Oh, that might be work. Thanks!! 👍

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,70 @@ def repeat(self, repeats: int) -> "Index":
...
+                    ('x', 'b'),
+                    ('x', 'c'),
+                    ('x', 'd'),
+                    ('x', 'e'),
+                    ('x', 'f')],
+                   )
+        """
+        sort = True if sort is None else sort
+        sort = validate_bool_kwarg(sort, "sort")
+        if not isinstance(other, Index) and isinstance(self, MultiIndex):
```
HyukjinKwon: Can we use if-elif-else here?

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,70 @@ def repeat(self, repeats: int) -> "Index":
...
+            if not isinstance(other, list) or not all([isinstance(item, tuple) for item in other]):
+                raise TypeError("other must be a MultiIndex or a list of tuples")
+            other = MultiIndex.from_tuples(other)
+        if not isinstance(other, Index) and not isinstance(self, MultiIndex):
```
HyukjinKwon: Do you want `self` to be `Index`? If that's the case, we can explicitly compare the type `type(self) == Index`.

HyukjinKwon:
Looks good otherwise.

databricks/koalas/indexes.py
itholic: Thanks :)!!

databricks/koalas/indexes.py
itholic: Thanks for reviewing, just applied it ! :)

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,70 @@ def repeat(self, repeats: int) -> "Index":
...
+                    ('x', 'b'),
+                    ('x', 'c'),
+                    ('x', 'd'),
+                    ('x', 'e'),
+                    ('x', 'f')],
+                   )
+        """
+        sort = True if sort is None else sort
+        sort = validate_bool_kwarg(sort, "sort")
+        if type(self) is MultiIndex and not isinstance(other, Index):
```
HyukjinKwon: I think you can also change this to:

```python
if type(self) is not type(other):
    if type(self) is MultiIndex:
        ...
    elif type(self) is Index:
        ...
    else:
        raise NotImplementedError("...")

```

itholic: Thanks for example. I just fixed it, thanks 👍

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,75 @@ def repeat(self, repeats: int) -> "Index":
...
+                ):
+                    raise TypeError("other must be a MultiIndex or a list of tuples")
+                other = MultiIndex.from_tuples(other)
+            elif type(self) is Index:
+                if isinstance(other, MultiIndex):
+                    # TODO: We can't support different type of values in a single column for now.
+                    raise NotImplementedError(
+                        "Union between Index and MultiIndex is not yet supported"
+                    )
+                if isinstance(other, Series):
```
HyukjinKwon: elif

itholic: Sorry my mistake, thanks for fixing

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,75 @@ def repeat(self, repeats: int) -> "Index":
...
+                    ('x', 'c'),
+                    ('x', 'd'),
+                    ('x', 'e'),
+                    ('x', 'f')],
+                   )
+        """
+        sort = True if sort is None else sort
+        sort = validate_bool_kwarg(sort, "sort")
+        if type(self) is not type(other):
+            if type(self) is MultiIndex:
```
ueshin: `if isinstance(self, MultiIndex):`?

databricks/koalas/indexes.py
```diff
@@ -1631,6 +1632,75 @@ def repeat(self, repeats: int) -> "Index":
...
+                if not isinstance(other, list) or not all(
+                    [isinstance(item, tuple) for item in other]
+                ):
+                    raise TypeError("other must be a MultiIndex or a list of tuples")
+                other = MultiIndex.from_tuples(other)
+            elif type(self) is Index:
```
ueshin: `elif isinstance(self, Index):`? or maybe we can just `else:`?

databricks/koalas/tests/test_indexes.py
```diff
@@ -1001,3 +1001,147 @@ def test_unique(self):
...
+        kidx2 = ks.from_pandas(pidx2)
+
+        self.assert_eq(kidx1.union(kidx2), pidx1.union(pidx2))
+        self.assert_eq(kidx2.union(kidx1), pidx2.union(pidx1))
+        self.assert_eq(
+            kidx1.union([3, 4, 5, 6]), pidx1.union([3, 4, 5, 6]),
+        )
+        self.assert_eq(
+            kidx2.union([1, 2, 3, 4]), pidx2.union([1, 2, 3, 4]),
+        )
```
ueshin: Shall we add one more test using `Series`?
Seems like we don't have the test for `pandas<0.24`.

itholic: Sure, just added it.
Thanks for finding and also other comments above :D

PR #1350, opened 2020-03-17
Implements asof() for Index

itholic:
This PR proposes `Index.asof`
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.asof.html#pandas.Index.asof)

```python
>>> idx = ks.Index(['2013-12-31', '2014-01-02', '2014-01-03'])
>>> idx.asof('2014-01-01')
'2013-12-31'

>>> idx.asof('2014-01-02')
'2014-01-02'

>>> idx.asof('1999-01-02')
nan
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1640,6 +1640,9 @@ class Index(IndexOpsMixin):
         is in the index, or return the previous index label if the passed one
         is not in the index.
 
+        .. note:: This API is dependent on :meth:`Index.is_monotonic_increasing`
+            which can be expensive.
+
         Parameters
         ----------
         label : object
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1631,6 +1631,55 @@ class Index(IndexOpsMixin):
         else:
             return ks.concat([kdf] * repeats).index
 
+    def asof(self, label):
+        """
+        Return the label from the index, or, if not present, the previous one.
+
+        Assuming that the index is sorted, return the passed index label if it
+        is in the index, or return the previous index label if the passed one
+        is not in the index.
+
+        Parameters
+        ----------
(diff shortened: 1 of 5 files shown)


databricks/koalas/indexes.py
```diff
@@ -1631,6 +1631,55 @@ def repeat(self, repeats: int) -> "Index":
...
+        '2014-01-02'
+
+        If all of the labels in the index are later than the passed label,
+        NaN is returned.
+
+        >>> idx.asof('1999-01-02')
+        nan
+        """
+        sdf = self._internal._sdf
+        if self.is_monotonic_increasing:
```
HyukjinKwon: I would like to avoid the excessive usage of `is_monotonic_decreasing`. It requires to send all data into one JVM and the APIs dependent on `is_monotonic_decreasing` or `is_monotonic_increasing` won't work at all on actual large dataset.

itholic: Agree. hope #1354 can lower the overhead

databricks/koalas/indexes.py
```diff
@@ -1632,6 +1632,55 @@ def repeat(self, repeats: int) -> "Index":
...
             return ks.concat([kdf] * repeats).index
 
+    def asof(self, label):
+        """
+        Return the label from the index, or, if not present, the previous one.
+
+        Assuming that the index is sorted, return the passed index label if it
+        is in the index, or return the previous index label if the passed one
+        is not in the index.
+
```
HyukjinKwon: Can you add a note that this API is dependent on `is_monotonic_increasing` which can be expensive? Also, seems we should update the note in the docstring at `truncate`.

itholic: Yes, then it is okay that I'll modify both in this PR?

PR #1354, opened 2020-03-18
Compute Index.is_monotonic/Index.is_monotonic_decreasing in a distributed manner

HyukjinKwon:
This PR proposes to compute `is_monotonic`/`is_monotonic_decreasing` in a distributed manner.

The approach is:

1. Check if it's ordered within each partition
2. Calculate min/max from each partition and check if the min/max of partitions are ordered.

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -568,19 +568,24 @@ class LocIndexer(_LocIndexerLike):
                 if (start is None and rows_sel.start is not None) or (
                     stop is None and rows_sel.stop is not None
                 ):
+
+                    inc = index_column.is_monotonic_increasing
+                    if inc is False:
+                        dec = index_column.is_monotonic_decreasing
+
                     if start is None and rows_sel.start is not None:
                         start = rows_sel.start
-                        if index_column.is_monotonic_increasing is not False:
+                        if inc is not False:
                             cond.append(index_column._scol >= F.lit(start).cast(index_data_type))
-                        elif index_column.is_monotonic_decreasing is not False:
+                        elif dec is not False:
                             cond.append(index_column._scol <= F.lit(start).cast(index_data_type))
                         else:
                             raise KeyError(rows_sel.start)
                     if stop is None and rows_sel.stop is not None:
                         stop = rows_sel.stop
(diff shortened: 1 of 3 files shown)


databricks/koalas/indexing.py
```diff
@@ -568,30 +568,19 @@ def _select_rows(self, rows_sel):
...
                             cond.append(index_column._scol <= F.lit(start).cast(index_data_type))
                         else:
                             raise KeyError(rows_sel.start)
                     if stop is None and rows_sel.stop is not None:
                         stop = rows_sel.stop
-                        if inc is not False:
+                        if index_column.is_monotonic_increasing is not False:
                             cond.append(index_column._scol <= F.lit(stop).cast(index_data_type))
-                        elif dec is not False:
+                        elif index_column.is_monotonic_decreasings is not False:
```
HyukjinKwon: Maybe we should ideally calculate both increasing and decreasing in one job but I noticed the codes become too much complicated here.

HyukjinKwon:
Basically, it checks:

1. If the data in each partition is ordered
2. If each partition's min/max are ordered.

HyukjinKwon:
Can't reproduce the test failure in my local ..

databricks/koalas/indexing.py
```diff
@@ -568,30 +568,19 @@ def _select_rows(self, rows_sel):
...
                             cond.append(index_column._scol <= F.lit(start).cast(index_data_type))
                         else:
                             raise KeyError(rows_sel.start)
                     if stop is None and rows_sel.stop is not None:
                         stop = rows_sel.stop
-                        if inc is not False:
+                        if index_column.is_monotonic_increasing is not False:
                             cond.append(index_column._scol <= F.lit(stop).cast(index_data_type))
-                        elif dec is not False:
+                        elif index_column.is_monotonic_decreasing is not False:
```
ueshin: How about:

```py
inc = index_column.is_monotonic_increasing
if inc:
   dec = index_column.is_monotonic_decreasing

if start is None and rows_sel.start is not None:
    start = rows_sel.start
    if inc:
        ...
    elif dec:
        ...

...
```

itholic:
Nice work

itholic:
LGTM

PR #1358, opened 2020-03-20
Delegate read_csv, read_json, and read_parquet to pandas IO if Spark throws an exception.

ueshin:
If the protocol in the specified path is not supported by Spark, Koalas fails even if pandas supports the protocol, e.g., `http` or `https`.
This PR proposes to add a config to enable to fallback to pandas with warnings.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -17,11 +17,13 @@
 """
 Wrappers around spark that correspond to common pandas functions.
 """
-from typing import Optional, Union, List, Tuple
 from collections import OrderedDict
 from collections.abc import Iterable
+from contextlib import contextmanager
 from functools import reduce
 import itertools
+from typing import Optional, Union, List, Tuple
+import warnings
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


databricks/koalas/namespace.py
```diff
@@ -272,14 +273,35 @@ def read_csv(
...
+                sdf = reader.csv(path)
+                if header is None:
+                    sdf = sdf.selectExpr(
+                        *["`%s` as `%s`" % (field.name, i) for i, field in enumerate(sdf.schema)]
+                    )
+        except Py4JJavaError:
+            if quotechar is None:
+                quotechar = '"'
+            pdf = pd.read_csv(
+                path,
```
itholic: Just curious; Is there a no problem with just use `pd.read_csv` directly with for large dataset?

ueshin: Good point.
Maybe we need a config to enable this or user warning.

itholic: Totally agree.
+ And I was just thinking, How about we add the parameter like `use_pandas` for all methods for allowing users can use pandas API directly.
Because maybe sometimes(maybe for someone often) users could want to just use pandas API with small dataset without switching the library from `ks` to `pd`.
(You don't have to seriously think about this, just my opinion 😸 )

itholic: Hmm.. just forget about above comment. maybe It's not that necessary for now.

databricks/koalas/namespace.py
```diff
@@ -272,14 +293,36 @@ def read_csv(
...
+                sdf = reader.csv(path)
+                if header is None:
+                    sdf = sdf.selectExpr(
+                        *["`%s` as `%s`" % (field.name, i) for i, field in enumerate(sdf.schema)]
+                    )
+        except Py4JJavaError as e:
+            with _fallback_to_pandas(e):
+                if quotechar is None:
+                    quotechar = '"'
+                pdf = pd.read_csv(
```
HyukjinKwon, line 309: Let's don't fallback to pandas' .. I thought about it multiple times and I think there's too many differences between Koalas and pandas. Users will expect to load files in a distributed manner. E.g., users might try to load `/tmp/abc` with default theme `fs.default.name` as `hdfs`. In the fallback, it will try to search `/tmp/abc` on local file name.

ueshin:
Let me close this as per the comment https://github.com/databricks/koalas/pull/1358#discussion_r411256145.

PR #1359, opened 2020-03-20
Implements DataFrame.eval()

itholic:
This PR proposes `DataFrame.eval()`
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.eval.html#pandas.DataFrame.eval)

```python
>>> df = ks.DataFrame({'A': range(1, 6), 'B': range(10, 0, -2)})
>>> df
   A   B
0  1  10
1  2   8
2  3   6
3  4   4
4  5   2
>>> df.eval('A + B')
0    11
1    10
2     9
3     8
4     7
Name: 0, dtype: int64
```

Assignment is allowed though by default the original DataFrame is not
modified.

```python
>>> df.eval('C = A + B')
   A   B   C
0  1  10  11
1  2   8  10
2  3   6   9
3  4   4   8
4  5   2   7
>>> df
   A   B
0  1  10
1  2   8
2  3   6
3  4   4
4  5   2
```

Use ``inplace=True`` to modify the original DataFrame.

```python
>>> df.eval('C = A + B', inplace=True)
>>> df
   A   B   C
0  1  10  11
1  2   8  10
2  3   6   9
3  4   4   8
4  5   2   7
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -9734,38 +9734,41 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         3  4   4   8
         4  5   2   7
         """
+        from databricks.koalas.series import _col
+
         if isinstance(self.columns, pd.MultiIndex):
             raise ValueError("`eval` is not supported for multi-index columns")
         inplace = validate_bool_kwarg(inplace, "inplace")
-        is_series = False
-        is_scalar_ = False
+        should_return_series = False
+        should_return_scalar = False
 
         # Since `eva_func` doesn't have a type hint, inferring the schema is always preformed
         # in the `map_in_pandas`. Hence, the variables `is_seires` and `is_scalar_` can be updated.
         def eval_func(pdf):
-            nonlocal is_series
-            nonlocal is_scalar_
+            nonlocal should_return_series
+            nonlocal should_return_scalar
             result_inner = pdf.eval(expr, inplace=inplace)
             if inplace:
                 result_inner = pdf
             if isinstance(result_inner, pd.Series):
-                is_series = True
+                should_return_series = True
                 result_inner = result_inner.to_frame()
             elif is_scalar(result_inner):
-                is_scalar_ = True
+                should_return_scalar = True
                 result_inner = pd.Series(result_inner).to_frame()
             return result_inner
 
         result = self.map_in_pandas(eval_func)
         if inplace:
+            # Here, the result is always a frame because the error is thrown during schema inference
+            # from pandas.
             self._internal = result._internal
+        elif should_return_series:
+            return _col(result)
+        elif should_return_scalar:
+            return _col(result)[0]
         else:
-            if is_series:
-                series_name = result._internal.column_labels[0][0]
-                result = result[series_name]
-            elif is_scalar_:
-                series_name = result._internal.column_labels[0][0]
-                result = result[series_name][0]
+            # Returns a frame
             return result
 
     def _to_internal_pandas(self):
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -30,7 +30,7 @@ from typing import Any, Optional, List, Tuple, Union, Generic, TypeVar, Iterable
 
 import numpy as np
 import pandas as pd
-from pandas.api.types import is_list_like, is_dict_like
+from pandas.api.types import is_list_like, is_dict_like, is_scalar
 
 if LooseVersion(pd.__version__) >= LooseVersion("0.24"):
     from pandas.core.dtypes.common import infer_dtype_from_object
@@ -4324,9 +4324,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         from databricks.koalas.series import Series
 
         for k, v in kwargs.items():
-            if not (
-                isinstance(v, (Series, spark.Column)) or callable(v) or pd.api.types.is_scalar(v)
-            ):
+            if not (isinstance(v, (Series, spark.Column)) or callable(v) or is_scalar(v)):
                 raise TypeError(
                     "Column assignment doesn't support type " "{0}".format(type(v).__name__)
                 )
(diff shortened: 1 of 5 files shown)


databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,118 @@ class  max_speed
...
+        if isinstance(self.columns, pd.MultiIndex):
+            raise ValueError("`eval` is not supported for multi-index columns")
+        inplace = validate_bool_kwarg(inplace, "inplace")
+        is_series = False
+        is_scalar_ = False
+
+        # TODO: Cannot operate `inplace=True` if there is not assignment in `expr`.
+        # There is no way to check now, so leave it to pandas for now with minimum data.
+        if inplace:
+            self.head(0).to_pandas().eval(expr, inplace)
```
itholic: I think there is some controversial here.
I couldn't found the better way to check if assignment is included given `expr`.
Any better idea is always welcome!

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,118 @@ class  max_speed
...
+           A   B
+        0  1  10
+        1  2   8
+        2  3   6
+        3  4   4
+        4  5   2
+
+        Use ``inplace=True`` to modify the original DataFrame.
+
+        >>> df.eval('C = A + B', inplace=True)
```
HyukjinKwon: We should add a test with type hint to avoid schema inference.

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,118 @@ class  max_speed
...
+        is_scalar_ = False
+
+        # TODO: Cannot operate `inplace=True` if there is not assignment in `expr`.
+        # There is no way to check now, so leave it to pandas for now with minimum data.
+        if inplace:
+            self.head(0).to_pandas().eval(expr, inplace)
+
+        def eval_func(pdf):
+            nonlocal is_series
+            nonlocal is_scalar_
```
HyukjinKwon: The variable doesn't work here because the function is serielaised to another node. It's difficult to have one global status for all other nodes. We should infer this from either type hint or schema inference from executing it against small sample of the current DataFrame.

itholic: I got it, let me fix. Thanks! :)

databricks/koalas/frame.py
itholic: @HyukjinKwon , Thanks for the comment! However, how can we provide the type hint to the `DataFrame.eval()` while there is no parameter like `func` something?

itholic: Maybe should I add the parameter like `**kwargs` and make supporting type hint possible for inferring the schema?

databricks/koalas/frame.py
itholic: I just inspected about this !
These variables are needed for after running `map_in_pandas` to make proper result matched with `pd.DataFrame.eval()`, not for type inferring
The type inferring you mentioned is performed in the `map_in_pandas` automatically since we're utilizing it.

In short, I think we still need these variables here to post-managing.

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,118 @@ class  max_speed
...
+        1  2   8  10
+        2  3   6   9
+        3  4   4   8
+        4  5   2   7
+        """
+        if isinstance(self.columns, pd.MultiIndex):
+            raise ValueError("`eval` is not supported for multi-index columns")
+        inplace = validate_bool_kwarg(inplace, "inplace")
+        is_series = False
+        is_scalar_ = False
```
ueshin: nit: why underscore at the end?

databricks/koalas/frame.py
ueshin: @HyukjinKwon Interestingly now it works since the function `eval_func` here doesn't have a type-hint and it always falls into infer-schema clause in `map_in_pandas` and update the variable in pandas execution locally.

databricks/koalas/frame.py
ueshin: I guess we can delay to `map_in_pandas` since currently we always run pandas once.
then at line 9751:

```py
result_inner = pdf.eval(expr, inplace=inplace)
if inplace:
    result_inner = pdf
```

HyukjinKwon: +1

databricks/koalas/frame.py
HyukjinKwon: Hm.. I see. If we will infer the schema always, it should be fine. @itholic can you clarify this with come comments?

databricks/koalas/frame.py
itholic: +10 !

databricks/koalas/frame.py
itholic: Its name is duplicated with imported `is_scalar` at line 33:

```python
   33 from pandas.api.types import is_list_like, is_dict_like, is_scalar
```

databricks/koalas/frame.py
itholic: @ueshin @HyukjinKwon , Thanks for the comments. 

Okay, I'll add the comments

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,117 @@ class  max_speed
...
+        0  1  10  11
+        1  2   8  10
+        2  3   6   9
+        3  4   4   8
+        4  5   2   7
+        """
+        if isinstance(self.columns, pd.MultiIndex):
+            raise ValueError("`eval` is not supported for multi-index columns")
+        inplace = validate_bool_kwarg(inplace, "inplace")
+        is_series = False
```
HyukjinKwon: `is_series ` -> `should_return_series`

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,117 @@ class  max_speed
...
+        is_scalar_ = False
```
HyukjinKwon: `is_scalar_` -> `should_return_scalar`

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,117 @@ class  max_speed
...
+            if isinstance(result_inner, pd.Series):
+                is_series = True
+                result_inner = result_inner.to_frame()
+            elif is_scalar(result_inner):
+                is_scalar_ = True
+                result_inner = pd.Series(result_inner).to_frame()
+            return result_inner
+
+        result = self.map_in_pandas(eval_func)
+        if inplace:
```
HyukjinKwon: Seems pandas throws an exception when it returns series or scalar when `inplace` is `True`. Shall we match?

HyukjinKwon: How about:

```python
if inplace:
    # Here, it always returns a DataFrame because the error is thrown during schema inference
    # from pandas.
    self._internal = result._internal
elif should_return_series:
    series_name = result._internal.column_labels[0][0]
    return result[series_name]
elif should_return_scalar:
    series_name = result._internal.column_labels[0][0]
    return result[series_name][0]
else:
    return result
```

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,117 @@ class  max_speed
...
+            self._internal = result._internal
+        else:
+            if is_series:
+                series_name = result._internal.column_labels[0][0]
+                result = result[series_name]
```
ueshin: `result = _col(result)`?

databricks/koalas/frame.py
```diff
@@ -9659,6 +9657,117 @@ class  max_speed
...
+            elif is_scalar_:
+                series_name = result._internal.column_labels[0][0]
+                result = result[series_name][0]
```
ueshin: `result = _col(result)[0]`?

itholic: Thanks! :)

databricks/koalas/frame.py
itholic: Seems better. thanks! :)

databricks/koalas/frame.py
itholic: ditto! 👍

databricks/koalas/frame.py
itholic: Looks clearer. thanks for the comment 👍

databricks/koalas/series.py
```diff
@@ -4686,6 +4686,9 @@ def _unpack_scalar(sdf):
 
 
 def _col(df):
+    """
+    Takes a DataFrame and returns the first column of the DataFrame as a Series
+    """
```
itholic: nit: Added docstring for `_col` here.

itholic:
Seems that Travis server has some problem now.

HyukjinKwon:
Let me just merge since Github Action builds passed.

PR #1364, opened 2020-03-24
[MINOR] Add copy param testcase to_numpy in indexes

beobest2:
image

It's a small part, but I've added it to improve code coverage.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -386,6 +386,8 @@ class Index(IndexOpsMixin):
         array([0, 1, 2, 3])
         >>> ks.DataFrame({'a': ['a', 'b', 'c']}, index=[[1, 2, 3], [4, 5, 6]]).index.to_numpy()
         array([(1, 4), (2, 5), (3, 6)], dtype=object)
+        >>> ks.DataFrame({'a': ['a', 'b', 'c']}, index=[[1, 2, 3], [4, 5, 6]]).index.to_numpy(copy=True)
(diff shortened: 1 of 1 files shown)


databricks/koalas/indexes.py
```diff
@@ -386,6 +386,8 @@ def to_numpy(self, dtype=None, copy=False):
         array([0, 1, 2, 3])
         >>> ks.DataFrame({'a': ['a', 'b', 'c']}, index=[[1, 2, 3], [4, 5, 6]]).index.to_numpy()
         array([(1, 4), (2, 5), (3, 6)], dtype=object)
+        >>> ks.DataFrame({'a': ['a', 'b', 'c']}, index=[[1, 2, 3], [4, 5, 6]]).index.to_numpy(copy=True)
```
itholic: too long here. (104 > 100)

itholic:
Otherwise, seems fine.

databricks/koalas/indexes.py
itholic: FYI: You can run a linter-check before push the commit via `/dev/lint-python`

HyukjinKwon: This example doesn't look useful enough to show in the docstring and documentation.
Shell we move this test case into `test_indexes`?

beobest2: @HyukjinKwon okay I will move this test case into `test_indexes`

beobest2: @itholic sorry I forgot to check lint

PR #1366, opened 2020-03-25
Implement Series.asof()

itholic:
This PR proposes `Series.asof()`
(https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.asof.html#pandas.Series.asof)

```python
>>> s = ks.Series([1, 2, np.nan, 4], index=[10, 20, 30, 40])
>>> s
10    1.0
20    2.0
30    NaN
40    4.0
Name: 0, dtype: float64
```

A scalar `where`.

```python
>>> s.asof(20)
2.0
```

For a sequence `where`, a Series is returned. The first value is
NaN, because the first element of `where` is before the first
index value.

```python
>>> s.asof([5, 20]).sort_index()
5     NaN
20    2.0
Name: 0, dtype: float64
```

Missing values are not considered. The following is ``2.0``, not
NaN, even though NaN is at the index location for ``30``.

```python
>>> s.asof(30)
2.0
```


databricks/koalas/series.py
```diff
@@ -4518,6 +4519,87 @@ def repeat(self, repeats: int) -> "Series":
         else:
             return _col(ks.concat([kdf] * repeats))
 
+    def asof(self, where, subset=None):
```
HyukjinKwon: @itholic can we deduplicate the implementation this with `Index.asof`? We could do it in `IndexOpsMixin`

itholic: Yes we can duplicate them, but I think there are more confusions than advantages because of the several reasons like the below.

- the name of parameter between `Index.asof` and `Series.asof` is different (`where` for `Series.asof`, `label` for `Index.asof`)
- And the acceptable types of those parameters are also different. (`labels` can only accept scalar type whereas `where` can accept scalar, pd.Index, pd.Series, list, tuples, dict... )

So I think the type validation logic could be too complicated.

What do you think ??

HyukjinKwon: Okay, let's leave it separate.

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,86 @@ def repeat(self, repeats: int) -> "Series":
...
+        if isinstance(self.index, ks.MultiIndex):
+            raise ValueError("asof is not supported for a MultiIndex")
+        if isinstance(where, ks.DataFrame):
+            raise ValueError("where cannot be a DataFrame")
+        if not self.index.is_monotonic_increasing:
+            raise ValueError("asof requires a sorted index")
+        if not is_list_like(where):
+            should_return_series = False
+            where = [where]
+        sdf = self._internal._sdf
```
HyukjinKwon: @itholic can you support operations on different frames?

itholic: @HyukjinKwon Thanks for the comment, but sorry I don't get exactly what you mean. 

Maybe I might be missed something, but I think this method doesn't include the operation which requires between two frames??

HyukjinKwon: I meant this case:

```python
>>> import pandas as pd
>>> pd.Series([1,2,3]).asof(pd.Series([0]))
0    1.0
dtype: float64
```

If that's difficult to implement for now, we should throw a proper exception

itholic: Thanks for the example. I understand. :)

Yes we can easily implement them if we assume that the given `Series` as a parameter could be a small enough, since It requires collecting all data of given `Series` into the single machine.

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,86 @@ def repeat(self, repeats: int) -> "Series":
...
+        index_scol = self._internal.index_spark_columns[0]
+        results = [
+            sdf.where(index_scol <= index).select(F.max(self._scol)).head()[0] for index in where
```
HyukjinKwon, line 4593: Wait, this requires to execute the job for every element which is crazy expensive. Can we convert this into single job?

HyukjinKwon: I think we did a similar trick before by collecting all maximum values from all columns, convert to pandas. In this way I think you can also pass `where` as series directly.

itholic: Thanks for the comment.

I just researched about this, Sadly I think we cannot avoid but should run a job for every each items in `where`, because this method requires get the items for every items in `where` from just a single column(`self._scol`) unlike the trick we did before.

In short, in the every step we should get the maximum values less than each items in the single column, and then we a need new single column in the next step.

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,86 @@ def repeat(self, repeats: int) -> "Series":
...
+
+        >>> s.asof(30)
+        2.0
+        """
+        should_return_series = True
+        if isinstance(self.index, ks.MultiIndex):
+            raise ValueError("asof is not supported for a MultiIndex")
+        if isinstance(where, ks.DataFrame):
+            raise ValueError("where cannot be a DataFrame")
+        if not self.index.is_monotonic_increasing:
```
HyukjinKwon, line 4585: Index.asof checks `is_monotonic_decreasing` too. should we check here too?

databricks/koalas/series.py
HyukjinKwon: Can you try something like this?

```python
from pyspark.sql import functions as F
spark.range(10).select(
    F.max(F.when(F.col(index_scol) <= index[0], F.col(index_scol)).otherwise(None)),
    F.max(F.when(F.col(index_scol) <= index[1], F.col(index_scol)).otherwise(None))).show()
```
```
+----------------------------------------------+----------------------------------------------+
|max(CASE WHEN (id <= 3) THEN id ELSE NULL END)|max(CASE WHEN (id <= 5) THEN id ELSE NULL END)|
+----------------------------------------------+----------------------------------------------+
|                                             3|                                             5|
+----------------------------------------------+----------------------------------------------+
```

databricks/koalas/series.py
itholic: We don't need to check because pandas only allows the Series with monotonically increasing Index for `Series.asof()` unlike `Index.asof()`.

```python
>>> pd.Index([4, 3, 2]).asof(1)
2

>>> pd.Series([1, 2, 3], index=[4, 3, 2]).asof(1)
...
ValueError: asof requires a sorted index
```

databricks/koalas/series.py
itholic: > Can you try something like this?
> 
> ```python
> from pyspark.sql import functions as F
> index_scol = "id"
> index = [3, 5]
> spark.range(10).select(
>     F.max(F.when(F.col(index_scol) <= index[0], F.col(index_scol)).otherwise(None)),
>     F.max(F.when(F.col(index_scol) <= index[1], F.col(index_scol)).otherwise(None))).show()
> ```
> 
> ```
> +----------------------------------------------+----------------------------------------------+
> |max(CASE WHEN (id <= 3) THEN id ELSE NULL END)|max(CASE WHEN (id <= 5) THEN id ELSE NULL END)|
> +----------------------------------------------+----------------------------------------------+
> |                                             3|                                             5|
> +----------------------------------------------+----------------------------------------------+
> ```

It works, Thanks!! 👍

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,89 @@ def repeat(self, repeats: int) -> "Series":
...
+            should_return_series = False
+            where = [where]
+        elif isinstance(where, (ks.Index, ks.Series)):
+            # Expecting the given Index or Series is small enough.
+            where = where.to_pandas()
+        sdf = self._internal._sdf
+        index_scol = self._internal.index_spark_columns[0]
+        cond = [F.max(F.when(index_scol <= index, self._scol)) for index in where]
+        result_row = sdf.select(cond).head(1)[0]
+        result_values = list(result_row.asDict().values())
```
itholic, line 4597: The test for Python 3.5 is failing since the order is not kept here. (the default dictionary type is not `OrderedDict` in Python 3.5)

I'll find the way to handle this.

databricks/koalas/series.py
```diff
@@ -4518,6 +4520,94 @@ def repeat(self, repeats: int) -> "Series":
...
+            where = [where]
+        elif isinstance(where, (ks.Index, ks.Series)):
+            # Expecting the given Index or Series is small enough.
+            where = where.to_pandas()
+        sdf = self._internal._sdf
+        index_scol = self._internal.index_spark_columns[0]
+        cond = [F.max(F.when(index_scol <= index, self._scol)) for index in where]
+        if LooseVersion(pd.__version__) >= LooseVersion("3.6"):
+            result_row = sdf.select(cond).head(1)[0]
+            result_values = list(result_row.asDict().values())
```
HyukjinKwon, line 4599: Can you `transpose()` instead so it does not trigger the job? See also `DataFrame.nunique`.

databricks/koalas/series.py
```diff
@@ -4518,6 +4520,94 @@ def repeat(self, repeats: int) -> "Series":
...
+        if isinstance(where, ks.DataFrame):
+            raise ValueError("where cannot be a DataFrame")
+        if not self.index.is_monotonic_increasing:
+            raise ValueError("asof requires a sorted index")
+        if not is_list_like(where):
+            should_return_series = False
+            where = [where]
+        elif isinstance(where, (ks.Index, ks.Series)):
+            # Expecting the given Index or Series is small enough.
+            where = where.to_pandas()
```
HyukjinKwon, line 4593: I don't think collecting here is ideal. I would just reject or compute somehow as a Spark column. We MUST avoid job execution in the middle wherever possible.

databricks/koalas/series.py
```diff
@@ -4518,6 +4520,94 @@ def repeat(self, repeats: int) -> "Series":
...
 
+    def asof(self, where):
+        """
+        Return the last row(s) without any NaNs before `where`.
+
+        The last row (for each element in `where`, if list) without any
+        NaN is taken.
+
+        If there is no good value, NaN is returned.
+
```
HyukjinKwon, line 4531: Can you also add a note like:

```
        .. note:: This API is dependent on :meth:`Index.is_monotonic_increasing`
            which can be expensive.
```

to be consistent with `Index.asof`?

databricks/koalas/series.py
itholic: I understand. just fixed this to not using to_pandas.
Thanks!!

databricks/koalas/series.py
itholic: Thanks for the comment! fixed it!

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,95 @@ def repeat(self, repeats: int) -> "Series":
...
+            raise ValueError("asof is not supported for a MultiIndex")
+        if isinstance(where, ks.DataFrame):
+            raise ValueError("where cannot be a DataFrame")
+        if not self.index.is_monotonic_increasing:
+            raise ValueError("asof requires a sorted index")
+        if not is_list_like(where):
+            should_return_series = False
+            where = [where]
+        elif isinstance(where, (ks.Index, ks.Series)):
+            where = list(map(lambda x: x.item(), where.values))
```
HyukjinKwon, line 4594: Can we just throw an exception in this case? `values` will still execute the jobs and collect everything into driver side.

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,95 @@ def repeat(self, repeats: int) -> "Series":
...
+        sdf = self._internal._sdf
+        index_scol = self._internal.index_spark_columns[0]
+        cond = [F.max(F.when(index_scol <= index, self._scol)) for index in where]
+        sdf = sdf.select(cond)
+        # The data is expected to be small so it's fine to transpose.
+        kdf = ks.DataFrame(sdf)
+        kdf.columns = pd.Index(where)
+        result_series = _col(kdf.transpose())
```
HyukjinKwon, line 4602: `transpose()` will execute the shortcut without setting `compute.max_rows`. Can you check again what configurations are set in `nunique`?

databricks/koalas/series.py
itholic: Okay, I think It will better.

databricks/koalas/series.py
itholic: Thanks!! :)

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,96 @@ def repeat(self, repeats: int) -> "Series":
...
+        index_scol = self._internal.index_spark_columns[0]
+        cond = [F.max(F.when(index_scol <= index, self._scol)) for index in where]
+        sdf = sdf.select(cond)
+        # The data is expected to be small so it's fine to transpose/use default index.
+        with ks.option_context(
+            "compute.default_index_type", "distributed", "compute.max_rows", None
+        ):
+            kdf = ks.DataFrame(sdf)
+        kdf.columns = pd.Index(where)
+        result_series = _col(kdf.transpose())
```
HyukjinKwon, line 4603: You should do `transpose` in `option_context` as well to avoid the shortcut execution.

itholic: ah, I missed it. Thanks!

HyukjinKwon:
Seems fine. Can you rebase @itholic?

databricks/koalas/series.py
```diff
@@ -4518,6 +4519,96 @@ def repeat(self, repeats: int) -> "Series":
...
+            raise ValueError("where cannot be an Index, Series or a DataFrame")
+        if not self.index.is_monotonic_increasing:
+            raise ValueError("asof requires a sorted index")
+        if not is_list_like(where):
+            should_return_series = False
+            where = [where]
+        sdf = self._internal._sdf
+        index_scol = self._internal.index_spark_columns[0]
+        cond = [F.max(F.when(index_scol <= index, self._scol)) for index in where]
+        sdf = sdf.select(cond)
```
HyukjinKwon, line 4596: Can you check `should_return_series` first and return it without executing `transpose()`?

itholic: @HyukjinKwon Fixed it. Thanks for the comment :)

PR #1371, opened 2020-03-26
Fix nan tests for Index.asof() and remove unnecessary code

itholic:
Some refactoring is done for `Index.asof()`

### DIFF
--- a/databricks/koalas/tests/test_indexes.py
+++ b/databricks/koalas/tests/test_indexes.py
@@ -1019,8 +1019,7 @@ class IndexesTest(ReusedSQLTestCase, TestUtils):
 
         self.assert_eq(kidx.asof("2014-01-01"), pidx.asof("2014-01-01"))
         self.assert_eq(kidx.asof("2014-01-02"), pidx.asof("2014-01-02"))
-        self.assert_eq(np.isnan(kidx.asof("1999-01-02")), True)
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_indexes.py
```diff
@@ -1019,8 +1019,7 @@ def test_asof(self):
 
         self.assert_eq(kidx.asof("2014-01-01"), pidx.asof("2014-01-01"))
         self.assert_eq(kidx.asof("2014-01-02"), pidx.asof("2014-01-02"))
-        self.assert_eq(np.isnan(kidx.asof("1999-01-02")), True)
-        self.assert_eq(np.isnan(pidx.asof("1999-01-02")), True)
+        self.assert_eq(repr(kidx.asof("1999-01-02")), repr(pidx.asof("1999-01-02")))
```
HyukjinKwon: Why should we use `repr` for this case only?

itholic: Since `np.nan` cannot comparable by using equality check which is used in `assery_eq` like the below.

```python
>>> np.nan == np.nan
False
```

Do we better fix other tests also to use repr for test consistency?

HyukjinKwon: you can use `isnan`. However, okay, it's fine.

PR #1374, opened 2020-03-27
Drop Travis CI and completely migrate to Github Actions

itholic:
This PR proposes of drop Travis CI from now on.

IMO, I don't think there is a need to keep two management points for CI.

The following changes are includes in this PR.

- Removed the `.travis.yml`
- Added the existing tests in the `.travis.yml` to `master.yml` which is used for GitHub Actions.
- Removed badge for the Travis CI from `README.md`
- Renamed the filename from `download_travis_dependencies.sh` to `download_spark.sh` since this script is downloading the Spark only.
- Uncomment the test which was failing on a Travis CI.
- Renamed job names


HyukjinKwon:
@ueshin WDYT?

itholic:
@HyukjinKwon okay I'll check!

itholic:
> Can you try to enable this test and see if passes?
> 
> ```
> databricks/koalas/tests/test_series.py:        # TODO: due to unknown reason, this test passes in Travis CI. Unable to reproduce in local.
> ```

Seems that It's working well 👍

.github/workflows/master.yml
```diff
@@ -12,13 +12,23 @@ jobs:
   pip_python35:
     name: PIP, Python (3.5)
```
HyukjinKwon: Can we name it as `PIP (Python, Spark, pandas, PyArrow)`?

.github/workflows/master.yml
```diff
@@ -76,9 +86,17 @@ jobs:
         include:
```
HyukjinKwon: Can we also name it `Conda (Python, Spark, pandas, PyArrow)`?

.github/workflows/master.yml
```diff
@@ -12,13 +12,23 @@ jobs:
   pip_python35:
```
HyukjinKwon: Shall we also name it `pip_pytho35` -> `pip_build` or `pip`?

.github/workflows/master.yml
HyukjinKwon: conda_python -> conda_build or conda

itholic:
@HyukjinKwon Thanks! I just renamed them.

Issue #1361, opened 2020-03-23
Duplicated key when groupby -> apply is used


syonekura:
Hello, I opened this thread in [SO](https://stackoverflow.com/questions/60790960/koalas-groupby-apply-returns-cannot-insert-key-already-exists), but viewing other issues reported here ([[1]](https://github.com/databricks/koalas/issues/1300), [[2]](https://github.com/databricks/koalas/issues/1007)), I'm suspecting that this could also be a bug.

I'm not sure if I'm missing something about the groupby - apply when moving from pandas to koalas, but the only requirement that we have is that the function to be applied cannot be type annotated since it will be used in different contexts, with different dataframes, so we don't know all the columns in advance. We're aware that koalas must run the function to infer the schema before applying the function, and we're ok with that overhead

syonekura:
Hi again, I encountered this [closed issue in pandas](https://github.com/pandas-dev/pandas/issues/32384), it seems that the grouping column is kept when one performs apply, and it is explicitly tested for (so it's a feature of pandas)

The problem arises then koala's apply has to infer the schema, runs the pandas groupby - apply and then initiates a koalas.DataFrame with the result (that after groupby-apply has index and a column with the same name):

```python
def apply(self, func):
    ...
    if should_infer_schema:
        # Here we execute with the first 1000 to get the return type.
        limit = get_option("compute.shortcut_limit")
        pdf = self._kdf.head(limit)._to_internal_pandas()
        pdf = pdf.groupby(input_groupnames).apply(func)  # <- pandas apply receive the key as column and index
        kdf = DataFrame(pdf) #  <- ValueError: cannot insert "key", already exists
        return_schema = kdf._sdf.drop(*HIDDEN_COLUMNS).schema
```

for now I can drop the key column inside `func`, I'm not sure if koalas should handle the logic for deduplicating index/column names before the initialization of `ks.DataFrame`

itholic:
Thanks for reporting the issue & detailed investigation 👍 

I think we should handle this logic and add the test explicitly like pandas.

@ueshin , @HyukjinKwon , WDYT about this?

itholic:
@ueshin , @HyukjinKwon 

For saving your time, the summary of this issue is like the below.

- in pandas.
```python
>>> df.groupby('d').sum()
        v
d
1.0   6.0
2.0  15.0

>>> df.groupby('d').apply(sum)
       d     v
d
1.0  3.0   6.0
2.0  6.0  15.0
```

- but in koalas
[...]

PR #1375, opened 2020-03-29
Solves issue with key handling during groupby - apply

syonekura:
Solves #1361 

When the function to be applied after a `groupby` operation doesn't have type annotations, koalas has to infer the schema of the result. In this context koalas raises an exception due to the duplication between index and column names in `pandas.groupby.apply` [[1]](https://github.com/pandas-dev/pandas/issues/32384).

One option to deal with this issue is get rid of the duplicated columns, getting this results:

```python
>>> df = pd.DataFrame({'d': [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], 'v': [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
>>> df
     d    v
0  1.0  1.0
1  1.0  2.0
2  1.0  3.0
3  2.0  4.0
4  2.0  5.0
5  2.0  6.0

>>> df.groupby('d').sum()
        v
d
1.0   6.0
2.0  15.0

>>> df.groupby('d').apply(sum)
       d     v
d
1.0  3.0   6.0
2.0  6.0  15.0

# Before PR
>>> ks.from_pandas(df).groupby('d').apply(sum)
...
ValueError: cannot insert d, already exists

# After PR
>>> ks.from_pandas(df).groupby('d').apply(sum)
       v
d
1.0  6.0
2.0  15.0
```

Another option that come to mind would be to rename the columns used as groupby keys (see below), but either way its still 'inconsistent' with pandas behavior.

```python
>>> ks.from_pandas(df).groupby('d').apply(sum)
       d_dup     v
d
1.0  3.0   6.0
2.0  6.0  15.0
```

how should koalas handle this?

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -928,6 +928,15 @@ class GroupBy(object):
                 pser_or_pdf = pdf.groupby(input_groupnames)[name].apply(func)
             else:
                 pser_or_pdf = pdf.groupby(input_groupnames).apply(func)
+                keys = set(input_groupnames)
+                should_drop_columns = (
+                    isinstance(pser_or_pdf, pd.DataFrame)
+                    and keys.issubset(set(pser_or_pdf.index.names))
+                    and keys.issubset(set(pser_or_pdf.columns))
+                )
+                if should_drop_columns:
+                    pser_or_pdf = pser_or_pdf.drop(input_groupnames, axis=1)
+
             kser_or_kdf = ks.from_pandas(pser_or_pdf)
             if len(pdf) <= limit:
                 return kser_or_kdf
--- a/databricks/koalas/tests/test_groupby.py
+++ b/databricks/koalas/tests/test_groupby.py
@@ -1162,15 +1162,13 @@ class GroupByTest(ReusedSQLTestCase, TestUtils):
             self.test_apply_with_new_dataframe()
 
     def test_apply_key_handling(self):
-        pdf = pd.DataFrame({'d': [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], 'v': [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
-
-        self.assert_eq(
-            pdf.groupby('d').sum(),
-            pdf.groupby('d').apply(sum).drop(['d'], axis=1)
+        pdf = pd.DataFrame(
+            {"d": [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], "v": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]}
         )
+
+        self.assert_eq(pdf.groupby("d").sum(), pdf.groupby("d").apply(sum).drop(["d"], axis=1))
         self.assert_eq(
-            ks.from_pandas(pdf).groupby('d').sum(),
-            ks.from_pandas(pdf).groupby('d').apply(sum)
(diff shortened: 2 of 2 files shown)


databricks/koalas/tests/test_groupby.py
```diff
@@ -1161,6 +1161,16 @@ def test_apply_with_new_dataframe_without_shortcut(self):
...
 
+    def test_apply_key_handling(self):
+        pdf = pd.DataFrame(
+            {"d": [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], "v": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]}
+        )
+
+        self.assert_eq(pdf.groupby("d").sum(), pdf.groupby("d").apply(sum).drop(["d"], axis=1))
+        self.assert_eq(
+            ks.from_pandas(pdf).groupby("d").sum(), ks.from_pandas(pdf).groupby("d").apply(sum)
+        )
```
itholic: Maybe

```python
        pdf = pd.DataFrame(
            {"d": [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], "v": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]}
        )
        kdf = ks.from_pandas(pdf)

        self.assert_eq(pdf.groupby("d").apply(sum), kdf.groupby("d").apply(sum))
```

seems more clearer test for this bug fixing??

syonekura: Yes, that seems simpler. I could do

```python
        pdf = pd.DataFrame(
            {"d": [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], "v": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]}
        )
        kdf = ks.from_pandas(pdf)

        self.assert_eq(pdf.groupby("d").apply(sum).drop(['d'], axis=1), kdf.groupby("d").apply(sum))
```

the `drop` is required since pandas will return d as a column and as key

databricks/koalas/groupby.py
```diff
@@ -928,6 +928,15 @@ def apply(self, func):
...
             else:
                 pser_or_pdf = pdf.groupby(input_groupnames).apply(func)
+                keys = set(input_groupnames)
+                should_drop_columns = (
+                    isinstance(pser_or_pdf, pd.DataFrame)
+                    and keys.issubset(set(pser_or_pdf.index.names))
+                    and keys.issubset(set(pser_or_pdf.columns))
+                )
+                if should_drop_columns:
+                    pser_or_pdf = pser_or_pdf.drop(input_groupnames, axis=1)
```
HyukjinKwon: Can you show before/after output in the PR description? From a cursory look, seems like it will just drop the duplicated column which isn't consistent with pandas' results.

itholic: I agree with @HyukjinKwon 's opinion since I think we should match the Koalas behavior to pandas' exactly same without any additional post-processing.

syonekura: I updated the PR description, looking forward to your comments

ueshin:
@syonekura Thanks for working on this!
I think it's a wider issue, so I submitted another PR to fix it and this as well.
Could you please take a look at #1419? Thanks!

syonekura:
@ueshin great! thanks for your work
closing this one then

PR #1379, opened 2020-03-30
Add a way to specify index column in I/O APIs

### LINKED ISSUE #1376: to_delta and to_parquet are dropping the index
Hi, I don't know if I made a mistake but `ks.DataFrame.to_delta` and `ks.read_delta` are dropping my index columns:
grafik

Is there a way to keep the index with the dataset?
Also, does this index actually give a performance boost compared to default Spark joins when I merge two large (terabyte range) datasets?

HyukjinKwon:
This PR adds a way to specify index column when reading and writing. For example,

```python
ks.range(1).to_parquet("/tmp/foo", index_col="index")
ks.read_parquet("/tmp/foo", index_col="index")
```

Resolves #1376


ueshin:
Thanks! merging.

PR #1380, opened 2020-03-31
Use the latest Sphinx in requirements-dev.txt

HyukjinKwon:
Seems we don't face the warnings from Sphinx (see https://github.com/databricks/koalas/pull/417). It seems fixed in the upstream Sphinx. This PR use the latest Sphinx in requirements-dev.txt


HyukjinKwon:
Let me merge this one - seems fine.

PR #1382, opened 2020-03-31
Reuse previously downloaded pandoc; do not download if pandoc is found in the path

HyukjinKwon:
This PR proposes to reuse previously downloaded pandoc; do not download if pandoc is found in the path. 

This targets to reduce the build time taken in the readthedocs. Currently, the documentation build in Read the Docs is being failed due to timeout.


HyukjinKwon:
Okay, doc builds passed. I will see if this makes reathedocs build pass.

PR #1383, opened 2020-03-31
Use pydata_sphinx_theme with Koalas own style

HyukjinKwon:
This PR proposes to use `pydata_sphinx_theme` theme which pandas uses but with Koalas own CSS. Demo: https://koalas.readthedocs.io/en/stable/index.html

This PR also piggy-packs two changes:

- Make Github Actions can pass with another version of Python in PIP build
- Move image to `/icons` location.


HyukjinKwon:
I am going to merge this as discussed in offline, @ueshin. I will make some followups if we should change.

PR #1387, opened 2020-04-01
Use PIP instead of Conda in Read the Docs build to speed up

HyukjinKwon:
For a bit of background:
Currently documentation generation is being failed in Read the Docs due to timeout.
The largest portion of time is being consumed for Conda installation, and seems like Read the Docs maintainers also encourage people to just use PIP instead (see also `https://github.com/readthedocs/readthedocs.org/issues/6254#issuecomment-542288316`).
We use Conda installation in order to install JDK 8 because Read the Docs has JDK 11 by default, and Spark 2.x does not support JDK 11.

For what this PR proposes:
This PR switches Conda to PIP to install Java 8 via using `install-jdk` package. Seems like they use AdoptOpenJDK instead but the jobs looks passing see https://readthedocs.org/projects/koalas/builds/10741596/.
Once Spark 3.0 is released, we can remove all related codes as Spark 3.0 supports JDK 11, and we don't need to install JDK 8 manually.

This PR decreases build time 1000 seconds (timeout) to around 900 seconds.


HyukjinKwon:
Merging .. let me see if the builds in readthedocs pass or not.

PR #1388, opened 2020-04-01
Fix iloc.__setitem__ with the other Series from the same DataFrame.

ueshin:
Fixing a bug of `iloc.__setitem__` which throws an `AnalysisException` even when the value is the other Series from the same DataFrame.

```py
>>> kdf = ks.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
>>> kser = kdf.a
>>> kser.iloc[[1]] = -kdf.b
...
pyspark.sql.utils.AnalysisException: 'Resolved attribute(s) b#2L missing from __distributed_sequence_column__#44L,__index_level_0__#0L,a#1L,__natural_order__#48L in ...
```

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -982,16 +982,10 @@ class iLocIndexer(_LocIndexerLike):
     @lazy_property
     def _internal(self):
         internal = super(iLocIndexer, self)._internal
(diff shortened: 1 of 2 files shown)


itholic:
LGTM

PR #1389, opened 2020-04-02
Implements squeeze() for Series & DataFrame

itholic:
This PR proposes [DataFrame.squeeze()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.squeeze.html#pandas.DataFrame.squeeze) and [Series.squeeze()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.squeeze.html#pandas.Series.squeeze)

```python
>>> primes = ks.Series([2, 3, 5, 7])
```

Slicing might produce a Series with a single value:

```python
>>> even_primes = primes[primes % 2 == 0]
>>> even_primes
0    2
Name: 0, dtype: int64

>>> even_primes.squeeze()
2
```

Squeezing objects with more than one value in every axis does nothing:

```python
>>> odd_primes = primes[primes % 2 == 1]
>>> odd_primes
1    3
2    5
3    7
Name: 0, dtype: int64

>>> odd_primes.squeeze()
1    3
2    5
3    7
Name: 0, dtype: int64
```
Squeezing is even more effective when used with DataFrames.

```python
>>> df = ks.DataFrame([[1, 2], [3, 4]], columns=['a', 'b'])
>>> df
   a  b
0  1  2
1  3  4
```

Slicing a single column will produce a DataFrame with the columns
having only one value:

```python
>>> df_a = df[['a']]
>>> df_a
   a
0  1
1  3
```

So the columns can be squeezed down, resulting in a Series:

```python
>>> df_a.squeeze('columns')
0    1
1    3
Name: a, dtype: int64
```

Slicing a single row from a single column will produce a single
scalar DataFrame:

```python
>>> df_0a = df.loc[[0], ['a']]
>>> df_0a
   a
0  1
```

Squeezing the rows produces a single scalar Series:

```python
>>> df_0a.squeeze('rows')
a    1
Name: 0, dtype: int64
```

Squeezing all axes will project directly into a scalar:

```python
>>> df_0a.squeeze()
1
```


databricks/koalas/generic.py
```diff
@@ -1889,6 +1889,135 @@ def get(self, key, default=None):
...
+            axis = "index" if axis == "rows" else axis
+            axis = validate_axis(axis)
+
+        if isinstance(self, ks.DataFrame):
+            is_squeezable = len(self.columns[:2]) == 1
+            # If DataFrame has multiple columns, there is no change.
+            if not is_squeezable:
+                return self
+            series_from_column = self[self.columns[0]]
+            has_single_value = len(series_from_column) == 1
```
HyukjinKwon, line 2005: Can you avoid counting? It requires to read the whole data once. It won't be beneficial when the data is big.

HyukjinKwon: If you need to know if it's a single value or not, it's better to collect single value so Spark run the job only against the first partition. See `__repr__` and `head()` to refer.

databricks/koalas/generic.py
```diff
@@ -1889,6 +1889,135 @@ def get(self, key, default=None):
...
+            # If DataFrame has only a single value, use pandas API directly.
+            if has_single_value:
+                result = self._to_internal_pandas().squeeze(axis)
+                return ks.Series(result) if isinstance(result, pd.Series) else result
+            elif axis == 0:
```
HyukjinKwon, line 2010: If should check `axis` first to avoid the count operation.

databricks/koalas/generic.py
itholic: Sorry It's my mistake! Thanks for finding!

databricks/koalas/generic.py
itholic: It returns `self` when the column has multiple value and the `axis` is 0.

So we should know whether the column has a single value or not before here.

HyukjinKwon: okay..

databricks/koalas/generic.py
```diff
@@ -1889,6 +1889,135 @@ def get(self, key, default=None):
...
+                return self
+            else:
+                return series_from_column
+        elif isinstance(self, ks.Series):
+            # The case of Series is simple.
+            # If Series has only a single value, just return it as a scalar.
+            # Otherwise, there is no change.
+            has_single_value = len(self.head(2)) == 1
+            return self[0] if has_single_value else self
```
HyukjinKwon, line 2019: Can you explicitly call `head(1)` instead of `self[0]`?

databricks/koalas/generic.py
```diff
@@ -1889,6 +1889,135 @@ def get(self, key, default=None):
...
+        if axis is not None:
+            axis = "index" if axis == "rows" else axis
+            axis = validate_axis(axis)
+
+        if isinstance(self, ks.DataFrame):
+            is_squeezable = len(self.columns[:2]) == 1
+            # If DataFrame has multiple columns, there is no change.
+            if not is_squeezable:
+                return self
+            series_from_column = self[self.columns[0]]
```
HyukjinKwon, line 2004: Why do we need to do this? seems like here we'll always have the dataframe having a single column.

HyukjinKwon: Ah, it's too have a series? you can call `_col(...)` on the frame.

itholic: Thanks for the advice. I often forgot using `_col(...)` 😿

databricks/koalas/generic.py
itholic: if then, it will returns `Series` rather scalar

e.g.

```python
>>> self = ks.Series([1, 2, 3])
>>> self[0]
1
>>> self.head(1)
0    1
Name: 0, dtype: int64
```

Maybe, did you mean `self.head(1)[0]` ???

databricks/koalas/generic.py
```diff
@@ -1889,6 +1889,137 @@ def get(self, key, default=None):
...
+            elif axis == 0:
+                return self
+            else:
+                return series_from_column
+        elif isinstance(self, ks.Series):
+            # The case of Series is simple.
+            # If Series has only a single value, just return it as a scalar.
+            # Otherwise, there is no change.
+            has_single_value = len(self.head(2)) == 1
+            return self[0] if has_single_value else self
```
HyukjinKwon, line 2021: Can we directly use the value from `self.head(2)` instead of launching another job to select one value from `self`?

itholic: Understand. Thanks!! 👍

PR #1390, opened 2020-04-02
Separate a requirements file for Read the Docs.

ueshin:
This is a follow-up of #1387.
If there is a dependency which doesn't exist in conda in a requirements file, we can't use it for conda and it takes some steps for contributors to set up their development environment.
We should separate the requirements file for Read the Docs.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -69,9 +69,6 @@ jobs:
         # as Black only works with Python 3.6+. This is hacky but we will drop
         # Python 3.5 soon so it's fine.
         sed -i '/black/d' requirements-dev.txt
-        # 'install-jdk' does not support Python 3.5. This is hacky but we will remove
-        # once Spark 3.0 is released. See also https://github.com/databricks/koalas/pull/1387
-        sed -i '/install-jdk/d' requirements-dev.txt
         pip install -r requirements-dev.txt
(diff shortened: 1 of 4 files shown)


itholic:
LGTM

HyukjinKwon:
I think this is fine as a temp fix for now. We will have to fix `docs/source/development/contributing.rst` as below:

```diff
- You can use `pip` alternatively if your Python is 3.5+.
+ You can use `pip` alternatively if your Python is 3.6+.
```

because users can't install `black` too because it requires Python 3.6+.

Another thing is that, we remove `install-jdk` dependency very soon when Spark 3.0 is released.

Last thing to consider is that, we would have another dependency that requires pip soon for `pydata_sphinx_theme` (https://github.com/databricks/koalas/pull/1383/files#diff-f4c265073ea4274a80eca85440875e56R20), which isn't Read the Docs dependency.
Once this PR is merged, it will break the contribution guide again as it requires `pip` installation.
I was simply thinking about installing the dependency via `pip` in conda env - it will still separate the package installation.

One downside of this is it won't test pure conda installation in the local environment but our CI in Koalas and koalas-feedstock will test it.

HyukjinKwon:
I am going to merge this for now as it recovers contribution guide

HyukjinKwon:
Quickly discussed offline with @ueshin. We'll stick to `conda` for now as that's encouraged by conda community.

PR #1393, opened 2020-04-03
Replace Spark download script to pip/conda installation

HyukjinKwon:
This PR proposes to remove manual Spark download script but uses `pip` and `conda` to simplify.


HyukjinKwon:
There seems not so much difference in testing time.

itholic:
It looks better than the manual install we did before

HyukjinKwon:
Let me try this out as there's no diff and it simplifies.

Merged.

PR #1394, opened 2020-04-03
Refine __setitem__ for loc/iloc with DataFrame.

ueshin:
Moving `__setitem__` for loc/iloc with DataFrame to use the similar logic as with Series.

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -329,7 +329,8 @@ class _LocIndexerLike(_IndexerLike):
 
     # TODO: support key and value as Series from different DataFrames.
     def __setitem__(self, key, value):
-        from databricks.koalas.series import Series
+        from databricks.koalas.frame import DataFrame
+        from databricks.koalas.series import Series, _col
 
         if self._is_series:
             if (isinstance(key, Series) and key._kdf is not self._kdf_or_kser._kdf) or (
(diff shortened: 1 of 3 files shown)


databricks/koalas/indexing.py
```diff
@@ -377,12 +380,88 @@ def __setitem__(self, key, value):
...
+                if len(value.columns) == 1:
+                    value = _col(value)
+                else:
+                    raise ValueError("Only a dataframe with one column can be assigned")
+
+            if (isinstance(rows_sel, Series) and rows_sel._kdf is not self._kdf_or_kser) or (
+                isinstance(value, Series) and value._kdf is not self._kdf_or_kser
+            ):
+                kdf = self._kdf_or_kser.copy()
+                kdf["__temp_natural_order__"] = F.monotonically_increasing_id()
```
itholic, line 401: I'm just curious about, don't we need to check the temporal name of index with `verify_temp_column_name` ?
(And also line 403, 405, 409, 411 which are using temporal columns)

HyukjinKwon, line 401: Should we maybe reuse `__natural_order__` we defined in Koalas DataFrames?

ueshin: The `verify_temp_column_name` is for a different layer, for Spark DataFrame.
Maybe we should add another one for Koalas DataFrame in a separate PR since there are other places to assign temporary columns to Koalas DataFrame.

ueshin: Unfortunately we can't since it will change after the assignment of key or value.

ueshin:
Thanks! Let me merge this for now. I'll address https://github.com/databricks/koalas/pull/1394#discussion_r403775995 later.

databricks/koalas/indexing.py
itholic: Thanks for the explanation! Makes sense :+1:

PR #1395, opened 2020-04-05
Update the version of pandas up-to-date for CI

itholic:
As pandas 1.0.3 released, I think It would be better match the version of pandas in our CI.


HyukjinKwon:
It doesn't affect pip build. Merging.

PR #1396, opened 2020-04-06
Install bdist_wheel explicitly and pin Sphinx version in PIP build at Github Actions

HyukjinKwon:
Dependency installation in pip build is being failed due to missing `error: invalid command 'bdist_wheel'` for unknown reason. This PR explicitly build `bdist_wheel`

Also, Sphinx documentation generation is being failed with Python 3.5. This is because:
- Sphinx change the signature of `add_directive`.
https://github.com/sphinx-doc/sphinx/commit/61098a0ae2e696a804459d36bd74ca57db76eda5#diff-9f615309782f7bbdd425e0aff31a0090L604

- In `matplotlib` 3.0.3, it calls with additional arguments: https://github.com/matplotlib/matplotlib/blob/v3.0.3/lib/matplotlib/sphinxext/plot_directive.py#L249-L262

This is fixed in the latest `matplotlib` but the problem is that `matplotlib` 3.1.x looks it dropped Python 3.5 support. So Python 3.5 + PIP build in Github Actions uses `matplotlib` 3.0.3 which is incompatible with Sphinx 3.x.


HyukjinKwon:
I am going to merge this to fix CI test failures.

PR #1399, opened 2020-04-06
Fix mod & rmod for matching with pandas.

### LINKED ISSUE #1398: Different behaviour for mod between pandas and Koalas
pandas and Koalas are using different way for calculating the `mod` for negative values like the below.

```python
>>> pser
0    10
dtype: int64

>>> kser
0    10
Name: 0, dtype: int64

>>> pser.mod(-3)
0   -2
dtype: int64

>>> kser.mod(-3)
0    1
Name: 0, dtype: int64
```

I think It would be better to match this behaviour to the pandas.

itholic:
Resolves #1398 

modulo calculation for negative numbers in Koalas is different from pandas', so fixed it.

# For example.

### 1. the Koalas'(and the PySpark') way: (only wanna get the result as a positive number)
- just switch the both numbers to positive and calculate.
```python
10 % 3  # the result will be 1
```

### 2. the pandas' way:(the result can be a negative number)
- first, switch the both numbers to positive and calculate just same as Koalas.
```python
10 % 3  # the result will be 1
```
- and then, plus the original negative number to the result
```python
1 + (-3)  # the final result will be -2
```

You can refer the https://stackoverflow.com/questions/1082917/mod-of-negative-number-is-melting-my-brain for more detail about the difference of such a modulo.


itholic:
I have no idea how can I handling the 'negative zero' like the below.

Any good idea?

```python
>>> pser
0    100.0
1      NaN
2   -300.0
3      NaN
4    500.0
5   -700.0
Name: Koalas, dtype: float64
>>> kser
0    100.0
1      NaN
2   -300.0
3      NaN
4    500.0
5   -700.0
Name: Koalas, dtype: float64

>>> pser.mod(150)
0    100.0
1      NaN
2      0.0
3      NaN
4     50.0
5     50.0
Name: Koalas, dtype: float64
>>> kser.mod(150)
0    100.0
1      NaN
2     -0.0  # << Here is the matter. how can we handle this negative zero?
3      NaN
4     50.0
5     50.0
Name: Koalas, dtype: float64
```

HyukjinKwon:
There have been ongoing issues about `0.0` vs `-0.0` in Spark itself. I think you can avoid to address it here.

databricks/koalas/base.py
```diff
@@ -86,6 +86,16 @@ def wrapper(self, *args):
...
+            # PySpark and pandas have a different way to calculate modulo operation.
+            # Below lines are needed for closing the gap.
+            if f is spark.Column.__mod__:
+                scol = F.when((self._scol * args[0] < 0) & (scol != 0), scol + args[0]).otherwise(
+                    scol
+                )
+            elif f is spark.Column.__rmod__:
+                scol = F.when(
+                    (self._scol * args[0] < 0) & (scol != 0), scol + self._scol
+                ).otherwise(scol)
```
HyukjinKwon, line 98: Can you handle these differences in the implementation side? See `__sub__` or `__add__` for example.

databricks/koalas/base.py
```diff
@@ -86,6 +86,16 @@ def wrapper(self, *args):
             args = [arg._scol if isinstance(arg, IndexOpsMixin) else arg for arg in args]
             scol = f(self._scol, *args)
             scol = booleanize_null(self._scol, scol, f)
+            # PySpark and pandas have a different way to calculate modulo operation.
+            # Below lines are needed for closing the gap.
+            if f is spark.Column.__mod__:
+                scol = F.when((self._scol * args[0] < 0) & (scol != 0), scol + args[0]).otherwise(
```
HyukjinKwon, line 92: Can you explain how it matches the differences between PySpark and pandas? Preferably in PR description.

itholic:
> There have been ongoing issues about `0.0` vs `-0.0` in Spark itself. I think you can avoid to address it here.

I made it works properly for every cases by fixing the existing calculation properly.

databricks/koalas/base.py
itholic: Okay I'll add!

ueshin:
Thanks. merging.
Since I have a concern, I'll submit a follow-up PR soon.

PR #1402, opened 2020-04-07
Help misuse of options argument.

ueshin:
I have sometimes seen misuse of a keyword argument `options`  in read_xxx/to_xxx functions. E.g.,

```py
kdf = ks.read_csv(..., options={ ... })
```

In this case, the argument `options` is actually `{'options': { ... }}`, which is not what the user wants to do.

We can help those cases by getting the `'options'` value.


itholic:
This way is not currently supported in pandas, right?
But It's LGTM though.

HyukjinKwon:
It's not supported in pandas because `options` here means PySpark specific options.

PR #1405, opened 2020-04-08
Fix the conda download url.

ueshin:
Fixing the conda build.


ueshin:
Seems like it's working, at least it's trying to install libraries.

PR #1410, opened 2020-04-09
Use TYPE_CHECKING to avoid circular reference.

ueshin:


databricks/koalas/utils.py
```diff
@@ -455,7 +459,7 @@ def validate_bool_kwarg(value, arg_name):
     return value
 
 
-def verify_temp_column_name(df: Union["ks.DataFrame", spark.DataFrame], column_name: str) -> str:
+def verify_temp_column_name(df: Union["DataFrame", spark.DataFrame], column_name: str) -> str:
```
HyukjinKwon, line 462: It's fine if it fails in some environments. But just to clarify, does it fail in your local specifically? Seems the builds are fine in CI.

ueshin:
Thanks! merging.

PR #1412, opened 2020-04-09
Fix Series.div when divide by zero

### LINKED ISSUE #1411: Incompatible behavior with pandas for Series.div() when divide by zero.
The existing implementation of `Series.div()` doesn't same as pandas.

```python
>>> pser
0    100.0
1      NaN
2   -300.0
3      NaN
4    500.0
5   -700.0
Name: Koalas, dtype: float64

>>> pser.div(0)
0    inf
1    NaN
2   -inf
3    NaN
4    inf
5   -inf
Name: Koalas, dtype: float64

>>> ks.from_pandas(pser).div(0)
0    None
1    None
2    None
3    None
4    None
5    None
Name: Koalas, dtype: object
```

itholic:
Resolves #1411 

```python
>>> pser
0    100.0
1      NaN
2   -300.0
3      NaN
4    500.0
5   -700.0
Name: Koalas, dtype: float64

>>> pser.div(0)
0    inf
1    NaN
2   -inf
3    NaN
4    inf
5   -inf
Name: Koalas, dtype: float64

>>> ks.from_pandas(pser).div(0)
0    inf
1    NaN
2   -inf
3    NaN
4    inf
5   -inf
Name: Koalas, dtype: float64
```


databricks/koalas/base.py
```diff
@@ -184,8 +184,14 @@ def __sub__(self, other):
             return _column_op(spark.Column.__sub__)(self, other)
 
     __mul__ = _column_op(spark.Column.__mul__)
-    __div__ = _numpy_column_op(spark.Column.__div__)
```
itholic, line 187: FYI: In Python 3.x, `__div__` and `__rdiv__` no more supported. Using `__truediv__` and `__rtruediv__`instead.

databricks/koalas/base.py
```diff
@@ -184,8 +184,14 @@ def __sub__(self, other):
...
-    __truediv__ = _numpy_column_op(spark.Column.__truediv__)
+
+    def __truediv__(self, other):
+        def truediv(left, right):
```
ueshin, line 189: `left` and `right` must be both Spark Column or a scalar value.

databricks/koalas/base.py
```diff
@@ -184,8 +184,14 @@ def __sub__(self, other):
...
+            return F.when(F.lit(right) == 0, F.lit("Infinity").__div__(F.lit(left))).otherwise(
+                F.lit(left).__truediv__(F.lit(right))
+            )
+
+        return self._with_new_scol(_numpy_column_op(truediv)(self, other)._scol)
```
ueshin, line 194: Why do we need to reassign?

itholic: I don't think so. Thanks for fixing!

databricks/koalas/frame.py
```diff
@@ -228,21 +228,21 @@
...
+rectangle     0.0     36.0
 
 >>> df.floordiv(10)
            angles  degrees
-circle          0       36
-triangle        0       18
-rectangle       0       36
+circle        0.0     36.0
+triangle      0.0     18.0
+rectangle     0.0     36.0
```
itholic, line 239: For `floordiv` and `rfloordiv`, the result will always float since we cannot predict the result type of each column which is already determined before Spark executes the job.

itholic: In short, since there is a possibility that the result can be a `Infinity` which is `float`, the result always be a `float`.

Someone can give me an opinion about this?

databricks/koalas/base.py
ueshin: I mean, we don't need to wrap with `F.lit()` for most cases. how about:

```diff
     def __truediv__(self, other):
         def truediv(left, right):
-            return F.when(F.lit(right) == 0, F.lit(np.inf).__div__(F.lit(left))).otherwise(
-                F.lit(left).__truediv__(F.lit(right))
+            return F.when(F.lit(right == 0), F.lit(np.inf).__div__(left)).otherwise(
+                left.__truediv__(right)
             )

         return _numpy_column_op(truediv)(self, other)
```

itholic: @ueshin Oh I got it! Thanks for the example :D

itholic:
I think I ran into some problem while I'm trying to reproduce the failure. 

Our test is failing in `Python 3.5` with `PyArrow 0.16.0` like the below,

스크린샷 2020-04-16 오전 7 40 10

After setting my test environment to equivalent to the GitHub Actions (`Python 3.5`, `PySpark 2.3.4`, `pandas 0.23.4`, `PyArrow 0.16.0`), It raises exception like the below when I import the Koalas package.

```python
WARNING:root:Found pyspark version "2.3.4" installed. pyspark>=2.4.0 is recommended.
ImportError: No module named 'numpy.core._multiarray_umath'
  File "/Users/haejoonlee/Desktop/git_store/koalas/databricks/koalas/__init__.py", line 46, in <module>
    import pyarrow
  File "/usr/local/anaconda3/envs/koalas3.5/lib/python3.5/site-packages/pyarrow/__init__.py", line 49, in <module>
    from pyarrow.lib import cpu_count, set_cpu_count
  File "pyarrow/lib.pyx", line 40, in init pyarrow.lib
ImportError: numpy.core.multiarray failed to import
```

Could somebody help to solve this??

itholic:
Oh, It was problem with version of `numpy`.

now It works after changing numpy version from `1.15.2` to `1.17.0`

databricks/koalas/frame.py
```diff
@@ -225,21 +225,21 @@
...
 
->>> df.rfloordiv(10)
+>>> df.rfloordiv(10)  # doctest: +SKIP
```
itholic, line 238: This test is moved to `tests/test_dataframe.py::DataFrameTest::test_rfloordiv` because this behaviour can be different depends on the version of pandas.

In pandas < 0.24.0
```python
In pandas < 0.24.0

```python
>>> df.rfloordiv(10)
             angles  degrees
circle          inf      0.0
triangle   3.000000      0.0
rectangle  2.000000      0.0
```

databricks/koalas/tests/test_dataframe.py
```diff
@@ -3205,6 +3205,27 @@ def test_squeeze(self):
...
+            {"angles": [0, 3, 4], "degrees": [360, 180, 360]},
+            index=["circle", "triangle", "rectangle"],
+            columns=["angles", "degrees"],
+        )
+        kdf = ks.from_pandas(pdf)
+
+        if LooseVersion(pd.__version__) < LooseVersion("1.0.0") and LooseVersion(
+            pd.__version__
+        ) >= LooseVersion("0.24.0"):
+            expected_result = ks.DataFrame(
```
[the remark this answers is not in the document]
itholic: thanks! :)

ueshin:
Thanks! merging

PR #1418, opened 2020-04-10
Add a note for Databricks Runtime for Machine Learning 6.0 and above.

ueshin:
Seems like just using a `conda` command may break MLR environment.
We should add a note to install Koalas on MLR properly.


ueshin:
Thanks! merging.

Issue #1414, opened 2020-04-09
Document that we don't support the compatibility with non-Koalas APIs yet.


HyukjinKwon:
Seems like people want to convert their codes directly from pandas to Koalas. One case I often observe is, they want to convert the codes that works together with other Python standard functions such as `max`, `min`, or list/generator comprehensions, e.g.)

```python
import pandas as pd
data = []
for a in pd.Series([1, 2, 3]):
    data.append(a)

pd.DataFrame(data)
```

In Koalas, such example does not work. We should preemptively document and guide users to stick to Koalas APIs only.

HyukjinKwon:
@beobest2 would you be interested in this?

beobest2:
@HyukjinKwon Sure, let me try. thank you

beobest2:
@HyukjinKwon 
I have written the document below, does this match your intentions? 
If not, please understand my lack of understanding.

===
FAQ

[Can I use the code used for Pandas as is?]

Replacing the Pandas module with Koalas works fine for most.
However, some compatibility with non-Kalas APIs is not yet supported.
Koalas cannot be used with other Python standard functions like `max`, `min` and `sum` or `list / generator comprehensions`.

Therefore, for the same operation, it should be changed as below:

- If you want to get `max`, `min`, and `sum` values, 
you can use the functions provided by Koalas instead of the Python standard functions.

```python
>>> max(pd.Series([1, 2, 3]))
3
>>> min(pd.Series([1, 2, 3]))
1
>>> sum(pd.Series([1, 2, 3]))
6
>>>
>>> ks.Series([1, 2, 3]).max()
3
>>> ks.Series([1, 2, 3]).min()
1
>>> ks.Series([1, 2, 3]).sum()
6
```

- If you want to use `list / generator comprehensions`, you can use Numpy object with `to_numpy()`.

```python
>>> import pandas as pd
>>> import databricks.koalas as ks
>>>
>>> data = []
>>> for a in pd.Series([1, 2, 3]):
...     data.append(a)
...
>>> for a in ks.Series([4, 5, 6]).to_numpy():
...     data.append(a)
...
>>> data
[1, 2, 3, 4, 5, 6]
```

===

PS. I am not sure where to include the guide documentation. (Whether to add to the "FAQ" or separately to the "User Guide")

HyukjinKwon:
I think we should better move them to Best Practice. I think we could rephrase, for example as below. Feel free to reword or rephrase.

[...]

PR #1420, opened 2020-04-14
Add documentation about non-Koalas APIs

beobest2:
Resolves https://github.com/databricks/koalas/issues/1414
I have added a preemptive guide to non-Koalas APIs to `best practeces`.
Please check.

### DIFF
--- a/docs/source/user_guide/best_practices.rst
+++ b/docs/source/user_guide/best_practices.rst
@@ -214,16 +214,16 @@ Therefore, it works seamlessly in pandas as below:
 
    >>> import pandas as pd
    >>> data = []
-   >>> pser = pd.Series([20, 21, 12],
-   ...                  index=['London', 'New York', 'Helsinki'])
+   >>> countries = ['London', 'New York', 'Helsinki']
+   >>> pser = pd.Series([20., 21., 12.], index=countries)
    >>> for temperature in pser:
    ...     data.append(temperature ** 2)
    ...
-   >>> pd.Series(data)
-   0    400
-   1    441
-   2    144
-   dtype: int64
+   >>> pd.Series(data, index=countries)
+   London      400.0
+   New York    441.0
+   Helsinki    144.0
+   dtype: float64
 
 However, for Koalas it does not work as the same reason above.
 The example above can be also changed to directly using Koalas APIs as below:
@@ -231,11 +231,15 @@ The example above can be also changed to directly using Koalas APIs as below:
 .. code-block:: python
 
    >>> import databricks.koalas as ks
-   >>> kser = ks.Series([20, 21, 12],
-   ...                  index=['London', 'New York', 'Helsinki'])
-   >>> kser.apply(lambda x: x ** 2)
-   0    400
-   1    441
-   2    144
-   dtype: int64
+   >>> import numpy as np
+   >>> countries = ['London', 'New York', 'Helsinki']
+   >>> kser = ks.Series([20., 21., 12.], index=countries)
+   >>> def square(x) -> np.float64:
+   ...     return x ** 2
+   ...
+   >>> kser.apply(square)
+   London      400.0
+   New York    441.0
+   Helsinki    144.0
+   Name: 0, dtype: float64
 
--- a/docs/source/user_guide/best_practices.rst
+++ b/docs/source/user_guide/best_practices.rst
@@ -217,6 +217,9 @@ Therefore, it works seamlessly in pandas as below:
    >>> countries = ['London', 'New York', 'Helsinki']
    >>> pser = pd.Series([20., 21., 12.], index=countries)
    >>> for temperature in pser:
+   ...     assert temperature > 0
+   ...     if temperature > 1000:
+   ...             temperature = None
    ...     data.append(temperature ** 2)
    ...
    >>> pd.Series(data, index=countries)
--- a/docs/source/user_guide/best_practices.rst
+++ b/docs/source/user_guide/best_practices.rst
@@ -172,14 +172,13 @@ See `Operations on different DataFrames <options.rst#operations-on-different-dat
 
 
 Use Koalas APIs directly whenever possible
----------------------------------------------------
+------------------------------------------
 
-While Koalas has similar APIs with pandas, some APIs are not explicitly supported. 
-For example, Python built-in functions such as ``min``, ``max``, ``sum``, etc.
-require the given argument to be iterable. 
-Koalas does not implement ``__iter__()`` yet to prevent users to collect
-all data into the client (driver) side from the cluster. 
-See the example below:
+Although Koalas has most of the pandas-equivalent APIs with pandas, there are several APIs not implemented yet or explicitly unsupported.
+
+As an example, Koalas does not implement ``__iter__()`` to prevent users to collect all data into the client (driver) side from the whole cluster.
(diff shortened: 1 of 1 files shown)


docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,69 @@ this operation should be avoided.
...
+   >>> ks.Series([1, 2, 3]).max()
+   3
+   >>> ks.Series([1, 2, 3]).min()
+   1
+   >>> ks.Series([1, 2, 3]).sum()
+   6
+
+Another common pattern from pandas users is to rely on list or generator comprehensions.
+In Python, any value that can follow the ``in`` keyword in a ``for`` statement is iterable.
+For instance, see below:
```
HyukjinKwon, line 211: Another common pattern from pandas users might be to rely on list comprehension or generator expression. However, it also assumes the dataset is locally iterable under the hood. Therefore, it works seamlessly in pandas as below: 

*The official name seems generator expression instead of generator comprehension per [PEP 289](https://www.python.org/dev/peps/pep-0289/).

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,69 @@ this operation should be avoided.
...
+   >>> for v in pser:
+   ...     data.append(v + 1)
+   ...
+   >>> pd.Series(data)
+   0    2
+   1    3
+   2    4
+   dtype: int64
+
+In Koalas, you can do it via:
```
HyukjinKwon, line 227: However, it does not work as the same reason above. The example above can be also changed to directly using Koalas APIs as below.

beobest2: Thanks for the detailed guidance. I will correct it with feedback.

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,69 @@ this operation should be avoided.
...
+   >>> ks.Series([1, 2, 3]).sum()
+   6
+
+Another common pattern from pandas users is to rely on list or generator comprehensions.
+In Python, any value that can follow the ``in`` keyword in a ``for`` statement is iterable.
+For instance, see below:
+
+.. code-block:: python
+
+   >>> import pandas as pd
```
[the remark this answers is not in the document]
beobest2: I have changed the example to typical summer temperatures for each city.

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,72 @@ this operation should be avoided.
...
+
+Another common pattern from pandas users might be to rely on list comprehension or generator expression.
+However, it also assumes the dataset is locally iterable under the hood.
+Therefore, it works seamlessly in pandas as below:
+
+.. code-block:: python
+
+   >>> import pandas as pd
+   >>> data = []
+   >>> pser = pd.Series([20, 21, 12],
```
HyukjinKwon, line 217: I think you can switch `pser = pd.Series([20, 21, 12],` to `pser = pd.Series([20., 21., 12.],`, and then change `kser.apply(lambda x: x ** 2)` in line 236. `apply` uses pandas UDF under the hood, and it requires to launch one job to infer schema. Avoiding these will be much more performant.

HyukjinKwon: If you intend to show that `temperature ** 2` can be an arbitrary logic, using `apply` is fine but probably we should replace `temperature ** 2` to something pandas and Koalas series APIs can't do. Also, you might want to specify the type hints for the function given to `apply` to avoid type inference.

beobest2: That's a good point. I learned a lot.

beobest2: I have fixed the example (specify the type hints for the function given to apply to avoid type inference).  also,  I intend to show that `temperature ** 2` can be an arbitrary logic. Is this okay?

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,76 @@ this operation should be avoided.
...
+Therefore, it works seamlessly in pandas as below:
+
+.. code-block:: python
+
+   >>> import pandas as pd
+   >>> data = []
+   >>> countries = ['London', 'New York', 'Helsinki']
+   >>> pser = pd.Series([20., 21., 12.], index=countries)
+   >>> for temperature in pser:
+   ...     data.append(temperature ** 2)
```
HyukjinKwon, line 220: Let's make this case a little bit more complicated intentionally :-).
```diff
    >>> for temperature in pser:
+   ...     assert temperature > 0
+   ...     if temperature > 1000:
+   ...         temperature = None
    ...     data.append(temperature ** 2)
```

beobest2: you're welcome : )  I have fixed it as more complicated.

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,79 @@ this operation should be avoided.
 
 See `Operations on different DataFrames <options.rst#operations-on-different-dataframes>`_ for more details.
 
+
+Use Koalas APIs directly whenever possible
+------------------------------------------
+
+Although Koalas has most of the pandas-equivalent APIs with pandas, there are several APIs not implemented yet or explicitly unsupported.
```
ueshin, line 177: nit: `... most of the pandas-equivalent APIs, there are ...` or `... most of the equivalent APIs to pandas, there are ...`?

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,79 @@ this operation should be avoided.
...
+
+As an example, Koalas does not implement ``__iter__()`` to prevent users to collect all data into the client (driver) side from the whole cluster.
```
ueshin, line 179: nit: maybe `... prevent users from collecting ...`?

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,79 @@ this operation should be avoided.
...
+.. code-block:: python
+
+   >>> import pandas as pd
+   >>> data = []
+   >>> countries = ['London', 'New York', 'Helsinki']
+   >>> pser = pd.Series([20., 21., 12.], index=countries)
+   >>> for temperature in pser:
+   ...     assert temperature > 0
+   ...     if temperature > 1000:
+   ...             temperature = None
```
ueshin, line 222: nit: indent?

docs/source/user_guide/best_practices.rst
```diff
@@ -170,3 +170,79 @@ this operation should be avoided.
...
+The example above can be also changed to directly using Koalas APIs as below:
+
+.. code-block:: python
+
+   >>> import databricks.koalas as ks
+   >>> import numpy as np
+   >>> countries = ['London', 'New York', 'Helsinki']
+   >>> kser = ks.Series([20., 21., 12.], index=countries)
+   >>> def square(x) -> np.float64:
+   ...     return x ** 2
```
HyukjinKwon, line 241: We should also add:

```
   ...     assert temperature > 0
   ...     if temperature > 1000:
   ...             temperature = None
```

in this function too. That was the main reason why I asked because it will leave no choice but only `kser.apply`. Otherwise, Koalas API can use `kser ** 2` directly instead of `kser.apply`.

beobest2: Aha now I understand ! I'll fix it too

HyukjinKwon:
Merged. Thanks!

PR #1424, opened 2020-04-15
Separate _select_rows and _select_cols into small methods.

ueshin:
Refactoring indexers to separate `_select_rows` and `_select_cols` into small methods.

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -220,6 +220,21 @@ class _LocIndexerLike(_IndexerLike, metaclass=ABCMeta):
     def _select_rows(
         self, rows_sel: Any
     ) -> Tuple[Optional[spark.Column], Optional[int], Optional[int]]:
+        """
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Let me take a look tomorrow ..

databricks/koalas/indexing.py
```diff
@@ -210,7 +216,121 @@ def __getitem__(self, key):
             return self._kdf_or_kser.iloc[key]
 
 
-class _LocIndexerLike(_IndexerLike):
+class _LocIndexerLike(_IndexerLike, metaclass=ABCMeta):
+    def _select_rows(
```
HyukjinKwon, line 220: @ueshin, is it all moving around the codes? If so looks fine. Can we add some docs for each method while we're here?

ueshin: sure, I'll add them soon. thanks!

HyukjinKwon:
Looks good but I will leave it to you @ueshin

ueshin:
Thanks! merging.

PR #1445, opened 2020-04-23
Specify upperbound for Python version in setup.py

HyukjinKwon:
Koalas currently can't support because Spark 2.x does not support Python 3.8. Python 3.8 is supported from Spark 3.


ueshin:
Thanks! merging.

PR #1448, opened 2020-04-24
Add links for 'Transform and apply a function' in APIs with minor corrections

HyukjinKwon:


databricks/koalas/frame.py
```diff
@@ -2017,6 +2017,9 @@ def apply_batch(self, func):
         Apply a function that takes pandas DataFrame and outputs pandas DataFrame. The pandas
         DataFrame given to the function is of a batch used internally.
 
+        See also `Transform and apply a function
+        <https://koalas.readthedocs.io/en/stable/user_guide/transform_apply.html>`_.
```
ueshin, line 2021: I'm wondering which we should use, stable or latest?

HyukjinKwon: Let me just change it to latest, thanks!

PR #1455, opened 2020-04-30
Make reset_index disallow the same name but allow it when drop=True.

ueshin:
pandas' `DataFrame.reset_index()` raises an error if the index name is the same as one of columns but allow it when `drop=True`.

```py
>>> import pandas as pd
>>> import numpy as np
>>> pdf = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]}, index=np.random.rand(3))
>>> pdf.index.name = "a"
>>> pdf.reset_index()
...
ValueError: cannot insert a, already exists
>>> pdf.reset_index(drop=True)
   a  b
0  1  4
1  2  5
2  3  6
```

whereas Koalas raises another error for both cases:

```py
>>> ks.from_pandas(pdf).reset_index()
...
pyspark.sql.utils.AnalysisException: "Reference 'a' is ambiguous, could be: a, a.;"

>>> ks.from_pandas(pdf).reset_index(drop=True)
...
pyspark.sql.utils.AnalysisException: "Reference 'a' is ambiguous, could be: a, a.;"
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3598,6 +3598,13 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
             index_map = OrderedDict(new_index_map_items)
 
+        if drop:
+            new_index_map = []
+
+        for _, name in new_index_map:
+            if name in self._internal.column_labels:
+                raise ValueError("cannot insert {}, already exists".format(name_like_string(name)))
+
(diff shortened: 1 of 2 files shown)


itholic:
LGTM.

ueshin:
Thanks! merging.

PR #1460, opened 2020-05-04
Fix Series.groupby on the Series from different DataFrames.

ueshin:
Fixing `Series.groupby` on the Series from different DataFrames.

```py
>>> kdf1 = ks.DataFrame({"C": [0.362, 0.227, 1.267, -0.562], "B": [1, 2, 3, 4]})
>>> kdf2 = ks.DataFrame({"A": [1, 1, 2, 2]})
>>> kdf1.B.groupby(kdf2.A).sum()
...
pyspark.sql.utils.AnalysisException: 'Resolved attribute(s) __tmp_groupkey_0__#118L missing from ...\n'
```


ueshin:
This is based on #1459 and the first commit here is the same as it.

PR #1462, opened 2020-05-06
Make groupby.apply in pandas<0.25 run the function only once per group.

ueshin:
`groupby.apply` in pandas<0.25 runs the functions twice for the first group, so if the function has a side effect, it could cause an unexpected behavior.

```py
>>> import pandas as pd
>>> pd.__version__
'0.24.2'

>>> kdf = ks.DataFrame({"d": [1.0, 1.0, 1.0, 2.0, 2.0, 2.0], "v": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})
>>> acc = ks.utils.default_session().sparkContext.accumulator(0)
>>> def sum_with_acc_frame(x) -> ks.DataFrame[np.float64, np.float64]:
...   global acc
...   acc += 1
...   return np.sum(x)
...
>>> kdf.groupby("d").apply(sum_with_acc_frame)
    c0    c1
0  3.0   6.0
1  6.0  15.0
>>> acc.value
4
```

whereas:

```py
>>> import pandas as pd
>>> pd.__version__
'0.25.3'

...

>>> acc.value
2
```

The `acc.value` should be `2` since there are only 2 groups and the function should be called twice.

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -977,7 +977,7 @@ class GroupBy(object):
         def pandas_groupby_apply(pdf):
 
             if not is_series_groupby and LooseVersion(pd.__version__) < LooseVersion("0.25"):
-                # `groupby.apply` in pandas<2.5 runs the functions twice for the first group.
+                # `groupby.apply` in pandas<0.25 runs the functions twice for the first group.
                 # https://github.com/pandas-dev/pandas/pull/24748
                 from pandas.core.base import SelectionMixin
 
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -983,26 +983,26 @@ class GroupBy(object):
 
                 f = SelectionMixin._builtin_table.get(func, func)
 
-                dummy_value = None
+                should_skip_first_call = True
 
-                def _func(df):
-                    nonlocal dummy_value
-                    if dummy_value is None:
+                def wrapped_func(df):
+                    nonlocal should_skip_first_call
+                    if should_skip_first_call:
+                        should_skip_first_call = False
                         if should_return_series:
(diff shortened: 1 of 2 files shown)


databricks/koalas/groupby.py
```diff
@@ -974,10 +975,34 @@ def apply(self, func):
                     return_schema = StructType([StructField("0", return_schema)])
 
         def pandas_groupby_apply(pdf):
+
+            if not is_series_groupby and LooseVersion(pd.__version__) < LooseVersion("0.25"):
+                # `groupby.apply` in pandas<2.5 runs the functions twice for the first group.
```
HyukjinKwon, line 980: nit: pandas 0.25

ueshin: wow! 🤣

databricks/koalas/groupby.py
```diff
@@ -974,10 +975,34 @@ def apply(self, func):
...
+                # https://github.com/pandas-dev/pandas/pull/24748
+                from pandas.core.base import SelectionMixin
+
+                f = SelectionMixin._builtin_table.get(func, func)
+
+                dummy_value = None
```
HyukjinKwon, line 986: How about naming it like `should_skip_first_call`, and assigning a bool? e.g.:

```python
should_skip_first_call = True

...

def wrapped_func(df):
    nonlocal should_skip_first_call
    if should_skip_first_call:
        should_skip_first_call = False
        if should_return_series:
            return pd.Series()
        else:
            return pd.DataFrame()
    else:
        return f(df)
```

ueshin: sounds great.

itholic:
LGTM. Nice fixing.

PR #1397, opened 2020-04-06
Add divmod & rdivmod for Series

itholic:
This PR proposes add `Series.divmod()` and `Series.rdivmod()`

It simply put `Series.div()` and `Series.mod()` together into the single tuple.

```python
>>> kser
0    100.0
1      NaN
2    300.0
3      NaN
4    500.0
Name: Koalas, dtype: float64

>>> kser.divmod(150)
(0    0.666667
1         NaN
2    2.000000
3         NaN
4    3.333333
Name: Koalas, dtype: float64, 0    100.0
1      NaN
2      0.0
3      NaN
4     50.0
Name: Koalas, dtype: float64)

>>> kser.rdivmod(150)
(0    1.5
1    NaN
2    0.5
3    NaN
4    0.3
Name: Koalas, dtype: float64, 0     50.0
1      NaN
2    150.0
3      NaN
4    150.0
Name: Koalas, dtype: float64)
```

PR #1463, opened 2020-05-06
Fix Series.div when div/floordiv np.inf by zero

itholic:
There is an incompatible hehaviour with pandas when divide `np.inf` by 0

```python
>>> pser
0    0.0
1    NaN
2    inf
Name: 0, dtype: float64
>>> kser
0    0.0
1    NaN
2    inf
Name: 0, dtype: float64

>>> pser.div(0)
0    NaN
1    NaN
2    inf
Name: 0, dtype: float64
>>> kser.div(0)
0   NaN
1   NaN
2   NaN  # Have to be a `inf`
Name: 0, dtype: float64
```

This PR fixed it.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -180,7 +180,7 @@ class IndexOpsMixin(object):
 
     def __truediv__(self, other):
         def truediv(left, right):
-            return F.when(F.lit(right != 0), left.__div__(right)).otherwise(
+            return F.when(F.lit(right != 0) | F.lit(right).isNull(), left.__div__(right)).otherwise(
                 F.when(F.lit(left == np.inf) | F.lit(left == -np.inf), left).otherwise(
                     F.lit(np.inf).__div__(left)
                 )
@@ -215,7 +215,9 @@ class IndexOpsMixin(object):
     def __floordiv__(self, other):
         def floordiv(left, right):
             return F.when(F.lit(right is np.nan), np.nan).otherwise(
-                F.when(F.lit(right != 0), F.floor(left.__div__(right))).otherwise(
+                F.when(
+                    F.lit(right != 0) | F.lit(right).isNull(), F.floor(left.__div__(right))
+                ).otherwise(
                     F.when(F.lit(left == np.inf) | F.lit(left == -np.inf), left).otherwise(
                         F.lit(np.inf).__div__(left)
                     )
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -185,6 +185,15 @@ class IndexOpsMixin(object):
         2. When divide positive number by zero, PySpark returns null whereas pandas returns np.inf
         3. When divide -np.inf by zero, PySpark returns null whereas pandas returns -np.inf
         4. When divide negative number by zero, PySpark returns null whereas pandas returns -np.inf
+
+        +-------------------------------------------+
+        | dividend (divisor: 0) | PySpark | pandas  |
+        |-----------------------|---------|---------|
+        |         np.inf        |   null  |  np.inf |
+        |        -np.inf        |   null  | -np.inf |
+        |           10          |   null  |  np.inf |
+        |          -10          |   null  | -np.inf |
+        +-----------------------|---------|---------+
         """
 
         def truediv(left, right):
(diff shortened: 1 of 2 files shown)


itholic:
Could someone review this PR when available to finish #1397 ?

databricks/koalas/base.py
```diff
@@ -180,8 +180,13 @@ def __sub__(self, other):
 
     def __truediv__(self, other):
         def truediv(left, right):
-            return F.when(F.lit(right == 0), F.lit(np.inf).__div__(left)).otherwise(
-                left.__truediv__(right)
+            return F.when(
+                F.lit(right == 0) & (F.lit(left != np.inf) & F.lit(left != -np.inf)),
```
HyukjinKwon, line 184: Why do we need ` & (F.lit(left != np.inf) & F.lit(left != -np.inf))` condition here? Seems if `right` is not `0`, then we will try the next branch.

databricks/koalas/tests/test_series.py
```diff
@@ -1480,7 +1480,9 @@ def test_div_zero(self):
             self.assert_eq(repr(pser.floordiv(0)), repr(kser.floordiv(0)))
             self.assert_eq(repr(pser // 0), repr(kser // 0))
         else:
-            result = pd.Series([np.inf, np.nan, -np.inf, np.nan, np.inf, -np.inf], name="Koalas")
+            result = pd.Series(
+                [np.inf, np.nan, -np.inf, np.nan, np.inf, -np.inf, np.inf, -np.inf], name="Koalas"
+            )
             self.assert_eq(repr(kser.floordiv(0)), repr(result))
             self.assert_eq(repr(kser // 0), repr(result))
```
HyukjinKwon, line 1487: Can we test the opposite case as well, for example, `kser.div(np.nan)`.

itholic: Sure, I just added and fixed related bug.

Thanks for the review!

databricks/koalas/base.py
```diff
@@ -180,8 +180,14 @@ def __sub__(self, other):
 
     def __truediv__(self, other):
         def truediv(left, right):
-            return F.when(F.lit(right == 0), F.lit(np.inf).__div__(left)).otherwise(
-                left.__truediv__(right)
+            return F.when(
```
HyukjinKwon, line 183: @itholic while we're here, can we document what behaviour this logic is in the docstring? Seems it's getting more difficult to read and follow.

itholic: @HyukjinKwon 
Yeah, I agree that.
I just simplify the code by considering every case.
could you check the simplified code once more when you available?
If you think It's still confused though, I'll add the proper docstring or comments

itholic:
@HyukjinKwon 
I simplified the existing confused `when` phrase,
Could you check this once more when you available?
If it's still not enough easy for reading, i'm gonna add an additional description as well.

databricks/koalas/base.py
```diff
@@ -212,8 +214,14 @@ def rtruediv(left, right):
 
     def __floordiv__(self, other):
         def floordiv(left, right):
-            return F.when(F.lit(right == 0), F.lit(np.inf).__div__(left)).otherwise(
-                F.when(F.lit(right) == np.nan, np.nan).otherwise(F.floor(left.__div__(right)))
+            return F.when(F.lit(right is np.nan), np.nan).otherwise(
```
HyukjinKwon, line 217: Let's just add a docstring to document the symantic differences between PySpark and pandas in the docstring at `__floordiv__` and `__truediv__`.

itholic: @HyukjinKwon Sounds good. I'll add it, thanks!

databricks/koalas/base.py
```diff
@@ -179,9 +179,19 @@ def __sub__(self, other):
     __mul__ = _column_op(spark.Column.__mul__)
 
     def __truediv__(self, other):
+        """
+        __truediv__ has different behaviour between pandas and PySpark for several cases.
+        1. When divide np.inf by zero, PySpark returns null whereas pandas returns np.inf
+        2. When divide positive number by zero, PySpark returns null whereas pandas returns np.inf
+        3. When divide -np.inf by zero, PySpark returns null whereas pandas returns -np.inf
+        4. When divide negative number by zero, PySpark returns null whereas pandas returns -np.inf
+        """
```
HyukjinKwon, line 188: Can we have an ascii table to show the diff?

itholic: Sure, I just added a table

PR #1466, opened 2020-05-07
Fix functions using GroupBy._apply_series_op.

ueshin:
Fixing groupby functions using `GroupBy._apply_series_op`, such as `cumxxx`, `diff`, `shift`, etc to use the original group keys since `Series` doesn't know the modified group keys in `SeriesGroupBy`, otherwise it doesn't follow the actual group keys.

```py
>>> kdf = ks.DataFrame({"a": [1, 2, 3, 4, 5, 6], "b": [1, 1, 2, 3, 5, 8], "c": [1, 4, 9, 16, 25, 36]})
>>> kdf
   a  b   c
0  1  1   1
1  2  1   4
2  3  2   9
3  4  3  16
4  5  5  25
5  6  8  36
>>> kdf.groupby((kdf.b // 5).rename('b')).cumsum().sort_index()
   a  b   c
0  1  1   1
1  3  2   5
2  3  2   9
3  4  3  16
4  5  5  25
5  6  8  36
```

This should be:

```py
>>> pdf.groupby((pdf.b // 5).rename('b')).cumsum().sort_index()
    a   b   c
0   1   1   1
1   3   2   5
2   6   4  14
3  10   7  30
4   5   5  25
5  11  13  61
```

Or even raises an error:

```py
>>> kdf.groupby(kdf.b // 5).cumsum().sort_index()

...

pyspark.sql.utils.AnalysisException: "cannot resolve '`CASE WHEN false THEN (Infinity / b) ELSE CASE WHEN (5 = NaN) THEN NaN ELSE FLOOR((b / 5)) END END`' given input columns: ...
```

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -626,7 +626,7 @@ class GroupBy(object):
         Name: a, dtype: float64
         """
         return self._apply_series_op(
-            lambda sg: sg._kser._diff(periods, part_cols=sg._groupkeys_scols)
+            lambda sg: sg._kser._diff(periods, part_cols=[s.spark_column for s in sg._groupkeys])
         )
 
     def cummax(self):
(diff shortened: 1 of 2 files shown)


itholic:
LGTM !

PR #1467, opened 2020-05-07
Fix GroupBy.nunique not to update the internal state.

ueshin:
`GroupBy.nunique` should not update the internal state.


ueshin:
Actually I suspect the current behavior is a pandas' bug..

PR #1478, opened 2020-05-11
Fix get dummies when uses the prefix parameter whose type is dict

### LINKED ISSUE #1469: get_dummies uses the prefix parameter whose type is dict return KeyError
Use get_dummies demo with prefix  and columns
image
pandas result like this:
image

itholic:
Resolves #1469 

```python
>>> df = ks.DataFrame({'A': ['a', 'b', 'a'], 'B': ['b', 'a', 'c'], 'C': [1, 2, 3]}, columns=['A', 'B', 'C'])

>>> ks.get_dummies(df, prefix={'B': 'ohe_B', 'C': 'ohe_C'}, columns=['B', 'C'])
   A  ohe_B_a  ohe_B_b  ohe_B_c  ohe_C_1  ohe_C_2  ohe_C_3
0  a        0        1        0        1        0        0
1  b        1        0        0        0        1        0
2  a        0        0        1        0        0        1

>>> pd.get_dummies(df.to_pandas(), prefix={'B': 'ohe_B', 'C': 'ohe_C'}, columns=['B', 'C'])
   A  ohe_B_a  ohe_B_b  ohe_B_c  ohe_C_1  ohe_C_2  ohe_C_3
0  a        0        1        0        1        0        0
1  b        1        0        0        0        1        0
2  a        0        0        1        0        0        1
```

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -1505,6 +1505,8 @@ def get_dummies(
             raise NotImplementedError(
                 "get_dummies currently does not support prefix as string types"
             )
+        elif isinstance(prefix, dict):
+            prefix = list(prefix.values())
         kdf = data.copy()
 
         if columns is None:
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -1505,8 +1505,6 @@ def get_dummies(
             raise NotImplementedError(
                 "get_dummies currently does not support prefix as string types"
             )
-        elif isinstance(prefix, dict):
-            prefix = list(prefix.values())
         kdf = data.copy()
 
         if columns is None:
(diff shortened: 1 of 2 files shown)


databricks/koalas/namespace.py
```diff
@@ -1576,6 +1576,8 @@ def get_dummies(
             "Length of 'prefix' ({}) did not match the length of "
             "the columns being encoded ({}).".format(len(prefix), len(column_labels))
         )
+    elif isinstance(prefix, dict):
+        prefix = [prefix[column_label[0]] for column_label in column_labels]
```
HyukjinKwon, line 1580: Can you handle error cases such as `pd.get_dummies(pdf, prefix={"A": "foo"})`?

HyukjinKwon:
Looks pretty good otherwise

databricks/koalas/namespace.py
itholic: Thank for the comment!

I think they are already handled in the existing code just same as pandas.

```python
>>> pd.get_dummies(pdf, prefix={"A": "bar"})
...
ValueError: Length of 'prefix' (1) did not match the length of the columns being encoded (2).

>>> ks.get_dummies(kdf, prefix={"A": "bar"})
...
ValueError: Length of 'prefix' (1) did not match the length of the columns being encoded (2).
```

HyukjinKwon: Oh, it shares the length check!

itholic: Yup! :)

PR #1479, opened 2020-05-11
Add test for Series.to_frame() with  parameter

beobest2:
Adding missing test for Series.to_frame() with parameter `name`

스크린샷 2020-05-12 오전 12 56 46


itholic:
LGTM if tests pass

PR #1482, opened 2020-05-12
Add test for Series.shape

beobest2:
Adding missing test for Series.shape

스크린샷 2020-05-12 오전 9 47 50

### DIFF
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -1494,11 +1494,24 @@ class SeriesTest(ReusedSQLTestCase, SQLTestUtils):
         kser = ks.Series(["a", "b", "c"])
         pser = kser.to_pandas()
 
-        self.assert_eq(repr(pser.to_frame(name="a")), repr(kser.to_frame(name="a")))
+        self.assert_eq(pser.to_frame(name="a"), kser.to_frame(name="a"))
 
         # for MultiIndex
         midx = pd.MultiIndex.from_tuples([("a", "x"), ("b", "y"), ("c", "z")])
         kser = ks.Series(["a", "b", "c"], index=midx)
         pser = kser.to_pandas()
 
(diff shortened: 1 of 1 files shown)


databricks/koalas/tests/test_series.py
```diff
@@ -1489,3 +1489,16 @@ def test_floordiv_nan(self):
         kser = ks.from_pandas(pser)
 
         self.assert_eq(repr(pser.floordiv(np.nan)), repr(kser.floordiv(np.nan)))
+
+    def test_shape(self):
+        kser = ks.Series(["a", "b", "c"])
+        pser = kser.to_pandas()
+
+        self.assert_eq(repr(pser.shape), repr(kser.shape))
```
ueshin, line 1497: Do we need to use `repr`?

HyukjinKwon:
@beobest2 can you rebase this to resolve the conflicts here? Looks good otherwise.

databricks/koalas/tests/test_series.py
HyukjinKwon: Yeah, everything here seems a string. So looks we won't need. Can you double check @beobest2?  Let's fix the one in https://github.com/databricks/koalas/commit/2c615ad076a262b4a3efaddca4eaa62415d6edb2 too if this is to fix.

beobest2:
> @beobest2 can you rebase this to resolve the conflicts here? Looks good otherwise.

I have fixed the confilcts and remove `repr` on both `test_shape` and `test_to_frame` on test_series.py.
Please check

HyukjinKwon:
@beobest2, seems the conflict was tried to resolve by manually picking the commits. It should either be `git merge` or `git rebase` so the PR doesn't include duplicated changes picked from the commits.

beobest2:
@HyukjinKwon 
PR has been modified to not include duplicate changes selected in the commit. I will be careful next time. Thank you

ueshin:
Thanks! merging.

PR #1484, opened 2020-05-12
Add extra arguments support in DataFrame.apply and DataFrame.apply_batch

HyukjinKwon:
This PR adds extra argument support in Koalas `DataFrame.apply` and `DataFrame.apply_batch` to match with pandas' [`DataFrame.apply`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.apply.html).

```python
ks.range(10).apply(lambda a, b, c: a + b + c, args=(1,), c=3)
```
```
   id
0   4
1   5
2   6
3   7
4   8
5   9
6  10
7  11
8  12
9  13
```

```python

def func(a, b, c, d=123) -> ks.Series[int]:
    return a + b + c + d

ks.range(10).apply(func, args=(1,), c=3)
```
```
    id
0  127
1  128
2  129
3  130
4  131
5  132
6  133
7  134
8  135
9  136
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2012,7 +2012,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
     T = property(transpose)
 
-    def apply_batch(self, func):
+    def apply_batch(self, func, args=(), **kwds):
         """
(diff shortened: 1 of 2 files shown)


ueshin:
Thanks! merging.

PR #1485, opened 2020-05-12
Add extra arguments support in DataFrame.transform, DataFrame.transform_batch and Series.transform_batch

HyukjinKwon:
This PR adds extra arguments support in DataFrame.transform, DataFrame.transform_batch and Series.transform_batch.

For example, 

```python
>>> from databricks import koalas as ks
>>> ks.range(10).transform_batch(lambda pdf, a, b, c: pdf.id + a + b + c, 1, 2, c=3)
0     6
1     7
2     8
3     9
4    10
5    11
6    12
7    13
8    14
9    15
Name: id, dtype: int64
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2404,7 +2404,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         else:
             return result
 
-    def transform(self, func):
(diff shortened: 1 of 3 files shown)


ueshin:
Thanks! merging.

PR #1488, opened 2020-05-12
Fix GroupBy apply, filter, and head to ignore temp columns when ops from different DataFrames.

ueshin:
Fixing `GroupBy.apply`, `filter`, and `head` to ignore temp columns when ops from different DataFrames, otherwise the result includes the temp columns.

```py
>>> kdf = ks.DataFrame({"a": [1, 2, 3, 4, 5, 6], "b": [1, 1, 2, 3, 5, 8], "c": [1, 4, 9, 16, 25, 36]})
>>> kkey = ks.Series([1, 1, 2, 3, 5, 8])
>>> kdf.groupby(kkey).apply(lambda x: x + x.min())
    a   b   c  __tmp_groupkey_0__
0   2   2   2                   2
5  12  16  72                  16
1   3   2   5                   2
3   8   6  32                   6
2   6   4  18                   4
4  10  10  50                  10
```

PR #1490, opened 2020-05-13
Fix GroupBy functions which need natural orderings to follow the order when opts from different DataFrames.

ueshin:
Fixing `GroupBy` functions which need natural orderings, such as `head`, `cumxxx`, etc., to follow the order when opts from different DataFrames, otherwise the result will be non-deterministic because of the shuffle needed to join the group keys.

```py
>>> kdf = ks.DataFrame({"a": [1, 1, 1, 1, 2, 2, 2, 3, 3, 3] * 3, "b": [2, 3, 1, 4, 6, 9, 8, 10, 7, 5] * 3, "c": [3, 5, 2, 5, 1, 2, 6, 4, 3, 6] * 3})
>>> kkey = ks.Series([1, 1, 1, 1, 2, 2, 2, 3, 3, 3] * 3)
>>> kdf.groupby(kkey).head(2)
    a  b  c
0   1  2  3
22  1  1  2
29  3  5  6
19  3  5  6
26  2  8  6
25  2  9  2
```

This should be:

```py
>>> pdf.groupby(pkey).head(2)
   a   b  c
0  1   2  3
1  1   3  5
4  2   6  1
5  2   9  2
7  3  10  4
8  3   7  3
```

### DIFF
--- a/databricks/koalas/utils.py
+++ b/databricks/koalas/utils.py
@@ -37,7 +37,7 @@ if TYPE_CHECKING:
     from databricks.koalas.frame import DataFrame
 
 
-def combine_frames(this, *args, how="full"):
+def combine_frames(this, *args, how="full", preserve_order_column=False):
     """
     This method combines `this` DataFrame with a different `that` DataFrame or
     Series from a different DataFrame.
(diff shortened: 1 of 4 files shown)


ueshin:
This is based on #1488. Please review it first.

databricks/koalas/utils.py
```diff
@@ -113,10 +119,15 @@ def combine_frames(this, *args, how="full"):
                 )
                 for label in that._internal.column_labels
             ]
+            + order_column
         )
 
         index_columns = set(this._internal.index_spark_column_names)
-        new_data_columns = [c for c in joined_df.columns if c not in index_columns]
+        new_data_columns = [
+            c
```
itholic, line 127: nit, I'm just curious, does `c` is generally used for indicating the `column` ?
I think `column` is more clearer than `c` if there is no special reason to should use `c`.
What do you think about? 🤔

ueshin: I just followed the previous commit, but `col` or `column` might be better. I'll update it.

itholic: Oh, I see. Thanks :)

PR #1494, opened 2020-05-13
Add doctest for Series.transform_batch with np.ufunc parameter

beobest2:
image

Adding a missing test for using the ``np.ufunc`` parameter of ``Series.transfor_batch``.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -2883,12 +2883,12 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
         Examples
         --------
-        >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
+        >>> df = ks.DataFrame([(1, -2), (3, -4), (5, -6)], columns=['A', 'B'])
         >>> df
            A  B
-        0  1  2
-        1  3  4
-        2  5  6
+        0  1 -2
+        1  3 -4
+        2  5 -6
 
         >>> def plus_one_func(pser) -> ks.Series[np.int64]:
         ...     return pser + 1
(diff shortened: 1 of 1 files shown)


beobest2:
Unexpectedly, the coverage of the targeted line is not filled. I guess I should add unittest.

databricks/koalas/series.py
```diff
@@ -2883,12 +2883,12 @@ def transform_batch(self, func, *args, **kwargs) -> "ks.Series":
 
         Examples
         --------
-        >>> df = ks.DataFrame([(1, 2), (3, 4), (5, 6)], columns=['A', 'B'])
+        >>> df = ks.DataFrame([(1, -2), (3, -4), (5, -6)], columns=['A', 'B'])
```
itholic, line 2886: Let's don't try to fix the existing tests as possible.

I think maybe you can create another `Series` for test `np.abs`, or simply just test with other `np.ufunc` just like the below if there is no special reason to should test the `np.abs`.

```python
>>> df.A.transform_batch(np.add, 10)
0    11
1    13
2    15
Name: A, dtype: int64
```

itholic:
> Unexpectedly, the coverage of the targeted line is not filled. I guess I should add unittest.

I think it is not that matter if the coverage increased or not.
I think this test more fit for doctest rather unittest because this test let users know about how to use the `transform_batch` with `np.ufunc`.

beobest2:
> > Unexpectedly, the coverage of the targeted line is not filled. I guess I should add unittest.
> 
> I think it is not that matter if the coverage increased or not.
> I think this test is more fit for doctest rather unittest because the purpose of this test is letting users know how to use the `transform_batch` with `np.ufunc`.


I agree with it. I modified it as you said.

itholic:
Looks okay to me

ueshin:
The build must be fixed.
@beobest2 Could you submit and push an empty commit to re-run the tests?

PR #1496, opened 2020-05-14
[MINOR] Fix typos in frame.py

beobest2:
Fixed typo in `frame.py`

PR #1497, opened 2020-05-14
 Added missing doctest and test cases in frame.py

beobest2:
Adding missing doctest and test cases in frame.py

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -5562,6 +5562,22 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         2     Thor  Stormbuster
         3     Hulk        Smash
 
+        You can use `tuple` or `list` types as parameters of a `subset`
+
+        >>> df.replace('Mjolnir', 'Stormbuster', subset=('weapon',))
+              name       weapon
+        0   Rescue      Mark-45
+        1  Hawkeye       Shield
+        2     Thor  Stormbuster
+        3     Hulk        Smash
+
+        >>> df.replace('Mjolnir', 'Stormbuster', subset=['weapon'])
+              name       weapon
+        0   Rescue      Mark-45
+        1  Hawkeye       Shield
+        2     Thor  Stormbuster
+        3     Hulk        Smash
+
         Dict like `to_replace`
 
         >>> df = ks.DataFrame({'A': [0, 1, 2, 3, 4],
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3114,22 +3114,6 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         3  3  400
         4  4  500
 
-        >>> df1.where(df1 > 1, df2.A).sort_index()
-           A    B
-        0  0  100
-        1 -1  200
-        2  2  300
-        3  3  400
-        4  4  500
-
-        >>> df1.where(df1 > 1, df2.A + 100).sort_index()
-             A    B
-        0  100  100
-        1   99  200
-        2    2  300
-        3    3  400
-        4    4  500
-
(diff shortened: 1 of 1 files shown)


itholic:
As @HyukjinKwon commented at https://github.com/databricks/koalas/pull/1496#issuecomment-628796555, 

Could you find & add more missing doctests while we here?

beobest2:
> missing doctests

sure!  Let me find more missing doctests.

HyukjinKwon:
Yup, let's do it in batch when possible.

beobest2:
From now on, I will give you a PR in batch so that you can review at once :)

beobest2:
image


I found that the `Series` was missing as the `other` parameter in the `DataFrame.where` function  and added it to the doctest.

beobest2:
image

I have also added doctest when `tuple` or `list` types are entered as a `subset` parameter in `DataFrame.replace` function.

databricks/koalas/frame.py
```diff
@@ -5538,6 +5562,23 @@ def replace(
...
+        such as `tuple` or `list`, as input to the `subset` parameter.
+
+        >>> df.replace('Mjolnir', 'Stormbuster', subset=('weapon',))
+              name       weapon
+        0   Rescue      Mark-45
+        1  Hawkeye       Shield
+        2     Thor  Stormbuster
+        3     Hulk        Smash
+
+        >>> df.replace('Mjolnir', 'Stormbuster', subset=['weapon'])
```
HyukjinKwon, line 5575: Let's move one of both to the test cases instead of doctest. Seems a bit verbose to have two examples to explain the iterable object as an argument.

databricks/koalas/frame.py
```diff
@@ -3106,6 +3114,22 @@ def where(self, cond, other=np.nan):
...
 
+        >>> df1.where(df1 > 1, df2.A).sort_index()
+           A    B
+        0  0  100
+        1 -1  200
+        2  2  300
+        3  3  400
+        4  4  500
+
+        >>> df1.where(df1 > 1, df2.A + 100).sort_index()
```
HyukjinKwon, line 3125: Let's remove this doctest.

HyukjinKwon:
Looks good. Mind updating PR title and description to reflect the proposed changes?

beobest2:
> Looks good. Mind updating PR title and description to reflect the proposed changes?


Of course, I will revise and update the title according to the comments you suggested. Thank you :)

databricks/koalas/frame.py
beobest2:
```python
>>> kdf.replace('Mjolnir', 'Stormbuster', subset=('weapon',))
                     name       weapon
0.342778          Ironman      Mark-45
0.087444  Captain America       Shield
0.179212             Thor  Stormbuster
0.522174             Hulk        Smash

>>> pdf.replace('Mjolnir', 'Stormbuster', subset=('weapon',))
TypeError: replace() got an unexpected keyword argument 'subset'
```

@HyukjinKwon 
When I tried adding the above test to the test case, I found that pandas does not support the `subset` parameter. So when I looked into the old version, pandas didn't support the `subset` parameter from the beginning. I found that the current `replace` parameter matches the spark `replace`. So, what are your thoughts on deleting a `subset` from Koalas for pandas?

ref>

[pandas 0.9.0 replace()](https://pandas.pydata.org/pandas-docs/version/0.9.0/generated/pandas.DataFrame.replace.html?highlight=replace#pandas.DataFrame.replace)

[pandas 1.0.1 replace()](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.replace.html)

[pyspark 2.1.3 replace()](https://spark.apache.org/docs/2.1.3/api/python/pyspark.sql.html#pyspark.sql.DataFrame.replace)

Issue #1498, opened 2020-05-14
Support 'max_rows' parameter in DataFrame.to_markdown()


beobest2:
https://github.com/databricks/koalas/blob/master/databricks/koalas/frame.py (line: 1767)
![image](https://user-images.githubusercontent.com/7010554/81952614-a85e1480-9641-11ea-9718-14de6c18fc25.png)

According to the code above, the `to_markdown` function in Koalas DataFrame seems to support the `max_rows` parameter. However, `to_markdown`, which can be supported by pandas version from 1.0.0, did not support 'max_rows' from the beginning. (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_markdown.html)

Therefore, the following error occurs. 
```python
>>> kdf = ks.DataFrame(data={"animal_1": ["elk", "pig"], "animal_2": ["dog", "quetzal"]})
>>> kdf.to_markdown(max_rows=1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/hwpark/Desktop/git_koalas/koalas/databricks/koalas/frame.py", line 1814, in to_markdown
    kdf._to_internal_pandas(), self.to_markdown, pd.DataFrame.to_markdown, args
  File "/Users/hwpark/Desktop/git_koalas/koalas/databricks/koalas/utils.py", line 413, in validate_arguments_and_invoke_function
    % (pd.__version__, param.name, pandas_func.__name__)
TypeError: The pandas version [1.0.3] available does not support parameter 'max_rows' for function 'to_markdown'.
```

In general, I'm curious whether to remove `max_rows` parameter support to match pandas in this case, or bypass `validate_arguments_and_invoke_function` to only use Koalas.

itholic:
I think we can just remove the `max_rows` parameter with related note in docstring and code since pandas doesn't support it.

beobest2:
Okay I will just remove the max_rows parameter.

itholic:
Yup, go for it :)

PR #1499, opened 2020-05-15
Remove max_rows parameter from DataFrame.to_markdown function

beobest2:
Resolves https://github.com/databricks/koalas/issues/1498

I removed the `max_rows` parameter from the `DataFrame.to_markdown` function.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1764,13 +1764,9 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             kdf._to_internal_pandas(), self.to_latex, pd.DataFrame.to_latex, args
         )
 
-    def to_markdown(self, buf=None, mode=None, max_rows=None):
+    def to_markdown(self, buf=None, mode=None):
         """
         Print DataFrame in Markdown-friendly format.
-
-        .. note:: This method should only be used if the resulting Pandas object is expected
-                  to be small, as all the data is loaded into the driver's memory. If the input
-                  is large, set max_rows parameter.
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


databricks/koalas/frame.py
```diff
@@ -1764,14 +1764,10 @@ def to_latex(
...
         )
 
-    def to_markdown(self, buf=None, mode=None, max_rows=None):
+    def to_markdown(self, buf=None, mode=None):
         """
         Print DataFrame in Markdown-friendly format.
 
-        .. note:: This method should only be used if the resulting Pandas object is expected
-                  to be small, as all the data is loaded into the driver's memory. If the input
-                  is large, set max_rows parameter.
```
itholic, line 1773: We should notice users that this API loads the all data into the driver's memory.

So, you can remove the only `max_rows` part:

```python
        .. note:: This method should only be used if the resulting Pandas object is expected
```

beobest2: okay. I will recover that part.

itholic:
Seems good

HyukjinKwon:
Looks good but let's hold off for now. Breaking changes should ideally happen in Koalas 1.0.0 and we're discussing about the timing offline.

databricks/koalas/frame.py
```diff
@@ -1764,13 +1764,12 @@ def to_latex(
             kdf._to_internal_pandas(), self.to_latex, pd.DataFrame.to_latex, args
         )
 
-    def to_markdown(self, buf=None, mode=None, max_rows=None):
+    def to_markdown(self, buf=None, mode=None):
```
HyukjinKwon, line 1767: Wait .. `to_markdown` is only supported with pandas 1.0.0. @itholic, was `max_rows` supported in pandas 1.0? Seems that parameter was never supported from the first place (https://github.com/pandas-dev/pandas/commit/7ab73a93c94e6557e5d77b2966549955e5c51cbc). Why was this added at https://github.com/databricks/koalas/commit/f44022927f3336ce666f82faa07d7276ed28de05?

itholic: @HyukjinKwon No, It was just my mistake and I didn't notice that.

HyukjinKwon: That's alright. Wanted to double check. Let's merge this as a bug fix.

PR #1501, opened 2020-05-16
Implement Series.unstack

itholic:
This PR proposes a `Series.unstack`.

```python
>>> s = ks.Series([1, 2, 3, 4],
...               index=pd.MultiIndex.from_product([['one', 'two'],
...                                                 ['a', 'b']]))
>>> s
one  a    1
     b    2
two  a    3
     b    4
Name: 0, dtype: int64

>>> s.unstack(level=-1).sort_index()
     a  b
one  1  2
two  3  4

>>> s.unstack(level=0).sort_index()
   one  two
a    1    3
b    2    4
```


databricks/koalas/series.py
```diff
@@ -4730,6 +4730,71 @@ def asof(self, where):
...
+                "Too many levels: Index has only {} levels, not {}".format(index_nlevels, level + 1)
+            )
+        elif level < 0 and (level < -index_nlevels):
+            raise IndexError(
+                "Too many levels: Index has only {} levels, {} is not a valid level number".format(
+                    index_nlevels, level
+                )
+            )
+        if self.index.has_duplicates:
+            raise ValueError("Index contains duplicate entries, cannot reshape")
```
HyukjinKwon, line 4784: Let's remove this restriction and add a note in the docstring. I would like to avoid to scan whole data just to check duplication.

databricks/koalas/series.py
```diff
@@ -4730,6 +4730,71 @@ def asof(self, where):
...
+
+        sdf = self._internal.spark_frame
+        index_scol_names = self._internal.index_spark_column_names.copy()
+        pivot_col = index_scol_names.pop(level)
+        data_scol_name = self._internal.data_spark_column_names[0]
+
+        sdf = sdf.groupby(index_scol_names).pivot(pivot_col).sum(data_scol_name)
```
HyukjinKwon, line 4791: I think you can check the duplication in the column names of pivot results.

databricks/koalas/series.py
```diff
@@ -4730,6 +4730,71 @@ def asof(self, where):
...
+                    index_nlevels, level
+                )
+            )
+        if self.index.has_duplicates:
+            raise ValueError("Index contains duplicate entries, cannot reshape")
+
+        sdf = self._internal.spark_frame
+        index_scol_names = self._internal.index_spark_column_names.copy()
+        pivot_col = index_scol_names.pop(level)
+        data_scol_name = self._internal.data_spark_column_names[0]
```
HyukjinKwon, line 4789: `self._internal.spark_column`?

itholic: @HyukjinKwon 

It seems that we need name of column rather column itself when perform groupby.sum().

otherwise, It occurs `TypeError: Column is not iterable`.

Thanks for the comment!

databricks/koalas/series.py
itholic: @HyukjinKwon Agree with that. I'll address it.

databricks/koalas/series.py
itholic: I think there is always no duplicated column in pivot result because the duplicated columns are integrated single column when performing groupby.

I just remove the duplicated check and add a proper note to docstring !

HyukjinKwon:
@itholic can you rebase please?

databricks/koalas/series.py
```diff
@@ -4759,6 +4759,75 @@ def asof(self, where):
...
+
+        sdf = sdf.groupby(index_scol_names).pivot(pivot_col).sum(data_scol_name)
+        internal = _InternalFrame(
+            spark_frame=sdf,
+            index_map=OrderedDict((index_scol_name, None) for index_scol_name in index_scol_names),
```
HyukjinKwon, line 4827: I think it wouldn't but can you check if the name of the index is kept in pandas?

itholic: Ah, It's kept in pandas. I'll fix it soon.

Thanks for the catch!

ueshin:
We should support non-numeric datatypes as well?

```py
>>> pd.Series(list('abcd'), index=pd.MultiIndex.from_product([['one', 'two'], ['a', 'b']])).unstack()
     a  b
one  a  b
two  c  d
```

itholic:
@ueshin I'll adress it, too. Thanks! :+1:

PR #1502, opened 2020-05-16
Implement Series.item

itholic:
This PR proposes `Series.item`

```python
>>> kser = ks.Series([10])
>>> kser.item()
10
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4732,7 +4732,7 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
 
     def item(self):
         """
-        Return the first element of the underlying data as a python scalar.
+        Return the first element of the underlying data as a Python scalar.
 
         Returns
         -------
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4750,9 +4750,10 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         >>> kser.item()
         10
         """
-        if len(self[:2]) != 1:
+        item_top_two = self[:2]
+        if len(item_top_two) != 1:
             raise ValueError("can only convert an array of size 1 to a Python scalar")
-        return self[0]
+        return item_top_two[0]
 
     def _cum(self, func, skipna, part_cols=()):
         # This is used to cummin, cummax, cumsum, etc.
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4730,6 +4730,30 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
         result_series.name = self.name
         return result_series
 
+    def item(self):
+        """
+        Return the first element of the underlying data as a python scalar.
+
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -4730,6 +4730,30 @@ def asof(self, where):
...
+        ValueError
+            If the data is not length-1.
+
+        Examples
+        --------
+        >>> kser = ks.Series([10])
+        >>> kser.item()
+        10
+        """
+        if len(self) != 1:
```
HyukjinKwon, line 4753: Let's avoid scanning whole data here. You can do this via collecting 2 rows.

databricks/koalas/series.py
```diff
@@ -4730,6 +4730,30 @@ def asof(self, where):
         result_series.name = self.name
         return result_series
 
+    def item(self):
+        """
+        Return the first element of the underlying data as a python scalar.
```
HyukjinKwon, line 4735: python -> Python

itholic: @HyukjinKwon Ah, I just followed the docstring of pandas, but It seems better to fix. Thanks!

databricks/koalas/series.py
itholic: Thanks!

databricks/koalas/series.py
```diff
@@ -4730,6 +4730,30 @@ def asof(self, where):
...
+        ValueError
+            If the data is not length-1.
+
+        Examples
+        --------
+        >>> kser = ks.Series([10])
+        >>> kser.item()
+        10
+        """
+        if len(self[:2]) != 1:
```
HyukjinKwon, line 4753: Can you reuse the output from `self[:2]` rather then executing the jobs twice?

itholic: Ah, I missed it. Thanks. !

databricks/koalas/series.py
```diff
@@ -4730,6 +4730,31 @@ def asof(self, where):
...
+        Examples
+        --------
+        >>> kser = ks.Series([10])
+        >>> kser.item()
+        10
+        """
+        item_top_two = self[:2]
+        if len(item_top_two) != 1:
+            raise ValueError("can only convert an array of size 1 to a Python scalar")
+        return item_top_two[0]
```
ueshin, line 4756: Seems like this still runs Spark jobs twice? We should explicitly call `to_pandas()` or `collect()`?

itholic: Ah, I'll check and fix it. Thanks !

HyukjinKwon: Let's be very careful on this @itholic next time. This one line mistake makes the performance two times slower.

itholic: @HyukjinKwon 
Yeah, I should have considered much more carefully.
Thanks for reminding me once again, @HyukjinKwon ! I must keep that in mind.

PR #1503, opened 2020-05-16
Implement Series.mad

90jam:
implement mad in Series

```python
>>> s = ks.Series([1, 2, 3, 4])
>>> s.mad()
1.0
```


itholic:
Seems good to me.

databricks/koalas/tests/test_series.py
```diff
@@ -1495,6 +1495,17 @@ def test_floordiv_nan(self):
...
+        pser = pd.Series([1, 2, 3, 4], name="Koalas")
+        kser = ks.from_pandas(pser)
+
+        self.assert_eq(pser.mad(), kser.mad())
+
+        pser = pd.Series([None, -2, 5, 10, 50, np.nan, -20], name="Koalas")
+        kser = ks.from_pandas(pser)
+
+        self.assert_eq(pser.mad(), kser.mad())
+
```
itholic, line 1508: Can we have a test with `MultiIndex` ?

you can

```python
pmidx = pd.MultiIndex.from_tuples([('x', 'a'), ('x', 'b'), ('y', 'c'), ('y', 'd')])
pser.index = pmidx
kser = ks.from_pandas(pser)

self.assert_eq(pser.mad(), kser.mad())
```

databricks/koalas/series.py
```diff
@@ -4730,6 +4730,31 @@ def asof(self, where):
...
+        Name: 0, dtype: int64
+
+        >>> s.mad()
+        1.0
+        """
+
+        sdf = self._internal.spark_frame
+        column_name = self._internal.data_spark_column_names[0]
+        avg = _unpack_scalar(sdf.select(F.avg(sdf[column_name])))
+        mad = _unpack_scalar(sdf.select(F.avg(F.abs(sdf[column_name] - avg))))
```
HyukjinKwon, line 4754: You don't have to run the jobs twice here. Can we do this in single pass? e.g.,  `return _unpack_scalar(sdf.select(F.avg(F.abs(sdf[column_name] - F.avg(sdf[column_name])))))`

90jam: I think the proposed method may not work properly because the aggregate function `avg` is used  as an argument of another aggregate function `avg`.  
I can't find any other way except running the jobs twice at this time. 
Are there any other ideas can handle it by running a job once? 
I would appreciate any comments :)

databricks/koalas/tests/test_series.py
90jam: Thanks, I added new tests with `MultiIndex`

databricks/koalas/series.py
HyukjinKwon: Okay.

HyukjinKwon, line 4754: nit:
```diff
-        column_name = self._internal.data_spark_column_names[0]
-        avg = _unpack_scalar(sdf.select(F.avg(sdf[column_name])))
-        mad = _unpack_scalar(sdf.select(F.avg(F.abs(sdf[column_name] - avg))))
-
+        spark_column = self._internal.spark_column
+        avg = _unpack_scalar(sdf.select(F.avg(spark_column)))
+        mad = _unpack_scalar(sdf.select(F.avg(F.abs(spark_column - avg))))
```

LGTM otherwise.

PR #1504, opened 2020-05-17
Add missing docs in series.rst & frame.rst

itholic:
Since several APIs are missing in the docs, just added them.

스크린샷 2020-05-17 오후 1 23 08


docs/source/reference/frame.rst
```diff
@@ -282,3 +288,4 @@ specific plotting methods of the form ``DataFrame.plot.<kind>``.
    DataFrame.plot.kde
    DataFrame.plot.density
    DataFrame.hist
+   DataFrame.kde
```
HyukjinKwon, line 291: This doesn't exist in pandas' doc. Let's remove.

itholic: @HyukjinKwon Thanks! I removed it

PR #1505, opened 2020-05-17
Fix message for AttributeError properly for Series.

itholic:
When users try to call the wrong name of property or function with Series,

The existing error message seems not proper like the below.

```python
>>> kser.itholic()  # 'kser' is Series of Koalas
...
AttributeError: Not a struct: LongType
```

Fixed it to:

```python
>>> kser.itholic()  # 'kser' is Series of Koalas
...
AttributeError: 'Series' object has no attribute 'itholic'
```

This way is same as pandas and other APIs (DataFrame, Index ...)

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4898,6 +4898,8 @@ class Series(_Frame, IndexOpsMixin, Generic[T]):
                 return property_or_func.fget(self)  # type: ignore
             else:
                 return partial(property_or_func, self)
+        elif not hasattr(ks.Series, item):
+            raise AttributeError("'Series' object has no attribute '{}'".format(item))
         return self.getField(item)
 
     def _to_internal_pandas(self):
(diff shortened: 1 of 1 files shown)


databricks/koalas/series.py
```diff
@@ -4898,6 +4898,8 @@ def __getattr__(self, item: str_type) -> Any:
                 return property_or_func.fget(self)  # type: ignore
             else:
                 return partial(property_or_func, self)
+        elif not hasattr(ks.Series, item):
+            raise AttributeError("'Series' object has no attribute '{}'".format(item))
         return self.getField(item)
```
HyukjinKwon, line 4903: I think you can remove this. We don't currently support nested structure in Series. Just directly raise `raise AttributeError("'Series' object has no attribute '{}'".format(item))`.

HyukjinKwon:
LGTM otherwise.

databricks/koalas/series.py
itholic: @HyukjinKwon Thanks! 👍

PR #1507, opened 2020-05-18
Implement DataFrame.explode

itholic:
This PR proposes `DataFrame.explode`.

```python
>>> df = ks.DataFrame({'A': [[1, 2, 3], [], [3, 4]], 'B': 1})
>>> df
           A  B
0  [1, 2, 3]  1
1         []  1
2     [3, 4]  1

>>> df.explode('A')
     A  B
0  1.0  1
0  2.0  1
0  3.0  1
1  NaN  1
2  3.0  1
2  4.0  1
```


itholic:
This method was not in our missing list, I found this at [stackoverflow question](https://stackoverflow.com/questions/60599853/what-is-the-koalas-equivalent-of-the-pandas-explode-function).

I think we'd better to update the missing list to match with pandas.

I'll check and add if there is additional missing list.

databricks/koalas/frame.py
```diff
@@ -10425,6 +10426,62 @@ def eval_func(pdf):
...
+            column = name_like_string(column)
+        else:
+            raise ValueError("column must be a scalar")
+
+        sdf = self._sdf
+        sdf = sdf.withColumn(column, F.explode_outer(sdf[column]))
+        index_scol_names = self._internal.index_spark_column_names
+        internal = _InternalFrame(
+            spark_frame=sdf,
+            index_map=OrderedDict((index_scol_name, None) for index_scol_name in index_scol_names),
```
HyukjinKwon, line 10481: Does it lose the names of index? Seems like it shouldn't be `None` to keep the names of the index.

itholic: Fixed, Thanks! 👍

databricks/koalas/frame.py
```diff
@@ -9992,6 +9993,63 @@ def eval_func(pdf):
...
+             A  B
+        0  1.0  1
+        0  2.0  1
+        0  3.0  1
+        1  NaN  1
+        2  3.0  1
+        2  4.0  1
+        """
+        if isinstance(column, tuple) or is_scalar(column):
+            spark_type = self[column].spark_type
```
HyukjinKwon, line 10036: `self[column].spark_type` -> `self[column].spark.data_type`

databricks/koalas/frame.py
```diff
@@ -9992,6 +9993,63 @@ def eval_func(pdf):
...
+        >>> df.explode('A')
+             A  B
+        0  1.0  1
+        0  2.0  1
+        0  3.0  1
+        1  NaN  1
+        2  3.0  1
+        2  4.0  1
+        """
+        if isinstance(column, tuple) or is_scalar(column):
```
HyukjinKwon, line 10035: Let's check only if `str` or `tuple` instead of `is_scalar`.

databricks/koalas/frame.py
```diff
@@ -9992,6 +9993,63 @@ def eval_func(pdf):
...
+            spark_type = self[column].spark_type
+            if not isinstance(spark_type, ArrayType):
+                return self
+            column = name_like_string(column)
```
HyukjinKwon, line 10039: Let's select Spark column directly here. `self[column].spark.column`

databricks/koalas/frame.py
```diff
@@ -9992,6 +9993,63 @@ def eval_func(pdf):
...
+            raise ValueError("column must be a scalar")
+
+        sdf = self._sdf
+        sdf = sdf.withColumn(column, F.explode_outer(sdf[column]))
+        internal = InternalFrame(
+            spark_frame=sdf,
+            index_map=self._internal.index_map,
+            column_labels=self._internal.column_labels,
+            column_label_names=self._internal.column_label_names,
+        )
```
HyukjinKwon, line 10050: Can you try to copy instead of creating new one? `self._internal.copy(...)`

HyukjinKwon:
Looks good otherwise.

databricks/koalas/frame.py
itholic: @HyukjinKwon Fixed it, thanks! :)

PR #1509, opened 2020-05-18
Fix docstring styles for consistent style and warning removal

HyukjinKwon:


databricks/koalas/frame.py
```diff
@@ -2099,6 +2099,7 @@ def apply_batch(self, func, args=(), **kwds):
         0  1  2
 
         You can also specify extra arguments.
+
```
HyukjinKwon, line 2102: For some reasons, this style makes the doc build failed in my local:

```
Warning, treated as error:
/.../koalas/databricks/koalas/frame.py:docstring of databricks.koalas.DataFrame.apply_batch:109:Unexpected indentation.
make: *** [html] Error 2
```

:confused:

itholic:
LGTM.

but curious about https://github.com/databricks/koalas/pull/1509#discussion_r426552987 since It's not reproduced in my local 😅

HyukjinKwon:
Thanks, @itholic.

PR #1510, opened 2020-05-18
Implements Series.to_markdown

itholic:
This PR proposes `Series.to_markdown()`

```python
>>> s = ks.Series(["elk", "pig", "dog", "quetzal"], name="animal")
>>> s.to_markdown()
|    | animal   |
|---:|:---------|
|  0 | elk      |
|  1 | pig      |
|  2 | dog      |
|  3 | quetzal  |
```


HyukjinKwon:
@itholic, can you rebase this please?

databricks/koalas/series.py
```diff
@@ -1395,6 +1396,52 @@ def to_latex(
 
     to_latex.__doc__ = DataFrame.to_latex.__doc__
 
+    def to_markdown(self, buf=None, mode=None):
```
HyukjinKwon, line 1399: Can we move this implementation to `generic.py`?

beobest2:
@itholic 
I'm just curious, some functions are implemented using functions of __pandas__ using `validate_arguments_and_invoke_function` (as of now), and some functions utilize __spark__ internals. 
Which criteria should I choose?

itholic:
@beobest2 

I think there is no strict criteria, we should consider it as case by case.

In this case, It is hard to avoid loading all data into the driver's memory whether we use pandas or PySpark since the result will be a string.

Plus, PySpark has no similar methods, that's why we just reuse pandas' API directly with proper warning note.

beobest2:
@itholic 

Thank you for the neat explanation. It was very helpful 👍

databricks/koalas/series.py
```diff
@@ -1395,6 +1396,52 @@ def to_latex(
...
+        # `to_markdown` is supported in pandas >= 1.0.0 since it's newly added in pandas 1.0.0.
+        if LooseVersion(pd.__version__) < LooseVersion("1.0.0"):
+            raise NotImplementedError(
+                "`to_markdown()` only supported in Koalas with pandas >= 1.0.0"
+            )
+        # Make sure locals() call is at the top of the function so we don't capture local variables.
+        args = locals()
+        kser = self
+        return validate_arguments_and_invoke_function(
+            kser._to_internal_pandas(), self.to_markdown, pd.Series.to_markdown, args
```
itholic, line 1442: @HyukjinKwon 

If we move this implementation to the `generic.py`, this part should be separated since `pd.Series.to_markdown` for `Series`, and `pd.DataFrame.to_markdown` for `DataFrame`.

Is it okay to move though?

databricks/koalas/series.py
itholic: Sure. Thanks for the comment!

itholic: Moved !

PR #1511, opened 2020-05-18
Implements Series.filter

beobest2:
Implementing `Series.filter`

```python
>>> kser = ks.Series([0, 1, 2], index=['one', 'two', 'three'])

>>> kser
one      0
two      1
three    2
Name: 0, dtype: int64

>>> kser.filter(items=['one', 'three'])
one      0
three    2
Name: 0, dtype: int64

>>> kser.filter(regex='e$')
one      0
three    2
Name: 0, dtype: int64

>>> kser.filter(like='hre')
three    2
Name: 0, dtype: int64
```


databricks/koalas/series.py
```diff
@@ -1870,6 +1870,70 @@ def drop(
         else:
             raise ValueError("Need to specify at least one of 'labels' or 'index'")
 
+    def filter(self, items=None, like=None, regex=None, axis=0):
```
HyukjinKwon, line 1873: Can you move this into `_Frame` and let it share one implementation?

beobest2: @HyukjinKwon You mean, I should also move `DataFrame.filter()` currently implemented in the `DataFrame`  into `generic._Frame` and let it share one implementation?

itholic: @beobest2 Correct.

databricks/koalas/generic.py
```diff
@@ -1632,6 +1634,136 @@ def first_valid_index(self):
...
+            raise TypeError(
+                "Keyword arguments `items`, `like`, or `regex` " "are mutually exclusive"
+            )
+
+        is_series = isinstance(self, ks.Series)
+
+        kdf = self
+        if is_series:
+            kdf = self.to_frame()
+            axis = validate_axis(axis, none_axis=0)
```
itholic, line 1707: The default value of `none_axis` is `0` we don't have to specify.

databricks/koalas/generic.py
```diff
@@ -1632,6 +1634,136 @@ def first_valid_index(self):
...
+            if isinstance(kdf, ks.Series) and axis == 1:
```
itholic, line 1708: I think `kdf` always `ks.DataFrame` ?

beobest2: This commit also supports the filter function for Series. Therefore, if the input is a Series, the axis has been checked.

databricks/koalas/generic.py
beobest2: okay I will fix it :)

databricks/koalas/generic.py
```diff
@@ -1632,6 +1634,136 @@ def first_valid_index(self):
...
+                    label
+                    for label in column_labels
+                    if any(matcher.search(i) is not None for i in label)
+                ]
+                filtered_df = kdf[output_labels]
+        else:
+            raise TypeError("Must pass either `items`, `like`, or `regex`")
+
+        if is_series:
+            return filtered_df.iloc[:, 0]
```
itholic, line 1763: If you wanna get a first `Series` from the `DataFrame`, let's use `first_series`.

you can

```python
from databricks.koalas.series import first_series
first_series(filtered_df)
```

databricks/koalas/generic.py
beobest2: my mistake. I will fix it too. thank you

databricks/koalas/tests/test_series.py
```diff
@@ -1599,3 +1599,11 @@ def test_unstack(self):
...
         kser = ks.Series([10, 20])
         self.assertRaises(ValueError, lambda: kser.item())
+
+    def test_filter(self):
+        kser = ks.Series([0, 1, 2], index=["one", "two", "three"])
+        pser = kser.to_pandas()
+
+        self.assert_eq(pser.filter(items=["one", "three"]), kser.filter(items=["one", "three"]))
+        self.assert_eq(pser.filter(regex="e$"), kser.filter(regex="e$"))
+        self.assert_eq(pser.filter(like="hre"), kser.filter(like="hre"))
```
itholic, line 1609: Could you add a test with `MultiIndex` ?

databricks/koalas/generic.py
```diff
@@ -1632,6 +1634,136 @@ def first_valid_index(self):
...
+        is_series = isinstance(self, ks.Series)
+
+        kdf = self
+        if is_series:
+            kdf = self.to_frame()
+            axis = validate_axis(axis)
+            if axis == 1:
+                raise ValueError("Series does not support columns axis.")
+        else:
+            axis = validate_axis(axis, none_axis=1)
```
itholic, line 1711: Hmm, maybe now we can check axis at once?

like

```python
        axis = validate_axis(axis)
        if is_series:
            kdf = self.to_frame()
            if axis == 1:
                raise ValueError("Series does not support columns axis.")
```

databricks/koalas/tests/test_series.py
beobest2: Of course, I will also add test with `Mutiindex`

databricks/koalas/generic.py
beobest2: I also considered that, but when the `DataFrame` is an input, the default `axis` is `1`, so I configured it like this.

databricks/koalas/tests/test_series.py
itholic: Thanks, and I think we'd better to have error test when axis=1 for `Series`.

like 

```python
        with self.assertRaisesRegex(ValueError, "Series does not support columns axis."):
            kser.filter(like="hre", axis=1)
```

beobest2: @itholic 
The `filter` function of Koalas `DataFrame` did not support `Multiindex`. So, I am currently posting PR by adding `Multiindex` support. (https://github.com/databricks/koalas/pull/1512) Test cases related to `Multiindex` will be added together in this PR.

itholic: Got it !

databricks/koalas/generic.py
itholic: Ah, okay I got it

HyukjinKwon:
Ah, https://github.com/databricks/koalas/pull/1512 caused a conflict here. Can you rebase and resolve the conflicts please?

beobest2:
@HyukjinKwon please check this.

1. Resolving code conflicts
2. Remove `filter` of `frame.py`
3. Add `multi-index` support implemented at https://github.com/databricks/koalas/commit/4a18a09f63d85086d687de69cd8d5f8e33c160e5
4. Add `multi-index` test cases in `Series`

ueshin:
Sorry for the late review.
Actually I'd say it's not good to move the function to generic.
I'll submit a PR to fix it soon.

Btw, this broke the docstring, I'll submit the PR to fix it anyway.

databricks/koalas/generic.py
```diff
@@ -1634,6 +1636,151 @@ def first_valid_index(self):
 
         return first_valid_idx
 
+    def filter(self, items=None, like=None, regex=None, axis=None):
+        """
+        Subset rows or columns of dataframe according to labels in
+        the specified index.
+        Note that this routine does not filter a dataframe on its
+        contents. The filter is applied to the labels of the index.
+        Parameters
```
HyukjinKwon, line 1645: Ah .. I didn't notice the newlines disappeared here.. it breaks the doc rendering. Let's be careful next time.

HyukjinKwon:
Okay, I had a offline discussion. We have been following pandas' structure so far. However, now it became kind of a bit weird for some instances in `generic.py` as we should handle both differently with an if-else.
We can avoid an if-else on `self` in `generic.py` which actually looks a bit odds, but only share the common codes.

itholic:
@HyukjinKwon Thanks for sharing, makes sense.

ueshin:
I don't look into all of the function in `generic.py` file yet so I'm not sure we should move all of them to the subclasses, but here is a pattern especially I'd avoid in `generic.py`:

```py
def some_function(self, ...):

    ... preprocess ...

    if isinstance(self, Series):
        kdf = self.to_frame()
    else:
        kdf = self

    ... working with kdf and the result is ret ...

    if isinstance(self, Series):
        return first_series(ret)
    else:
        return ret
```

Then we should move it to `DataFrame` and make it work with only `DataFrame`, then we can add something like the following in `Series`:

```py
def some_function(self, ...):

    ... preprocess ...

    return first_series(self.to_frame().some_function(...))
```

ueshin:
Anyway I'll submit PRs when I encounter the pattern.

databricks/koalas/generic.py
beobest2: Oppse I'll be careful next time. I'm Sorry

PR #1512, opened 2020-05-18
Implements multi-index support in Dataframe.filter

beobest2:
Implementing multi-index support in `DataFrame.filter`

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -9328,34 +9328,56 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             else:
                 raise ValueError("items should be a list-like object.")
             if axis == 0:
-                # TODO: support multi-index here
-                if len(index_scols) != 1:
-                    raise ValueError("Single index must be specified.")
-                col = None
-                for item in items:
-                    if col is None:
-                        col = index_scols[0] == F.lit(item)
-                    else:
-                        col = col | (index_scols[0] == F.lit(item))
+                if len(index_scols) > 1:
+                    # for multi-index
+                    cols = []
+                    col = None
+                    for item in items:
+                        if col is None:
+                            for i, element in enumerate(item):
+                                if col is None:
+                                    col = index_scols[i] == F.lit(element)
+                                else:
+                                    col = col & (index_scols[i] == F.lit(element))
+                        cols.append(col)
+                        col = None
+                    for c in cols:
+                        if col is None:
+                            col = c
+                        else:
+                            col = col | c
+                else:
+                    col = None
+                    for item in items:
+                        if col is None:
+                            col = index_scols[0] == F.lit(item)
+                        else:
+                            col = col | (index_scols[0] == F.lit(item))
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -9328,34 +9328,58 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+                        else:
+                            col = col | (index_scols[0] == F.lit(item))
+                elif len(index_scols) > 1:
+                    # for multi-index
+                    cols = []
+                    col = None
+                    for item in items:
+                        if col is None:
+                            for i, element in enumerate(item):
+                                if col is None:
```
itholic, line 9345: Should we check this here again? I think this is already checked at line 9343.

beobest2: I've  just removed double check

databricks/koalas/frame.py
```diff
@@ -9328,34 +9328,57 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+                    cols = []
+                    col = None
+                    for item in items:
+                        for i, element in enumerate(item):
+                            if col is None:
+                                col = index_scols[i] == F.lit(element)
+                            else:
+                                col = col & (index_scols[i] == F.lit(element))
+                        cols.append(col)
+                        col = None
```
HyukjinKwon, line 9349: I think you don't need to nullify the col here.

databricks/koalas/frame.py
```diff
@@ -9328,34 +9328,57 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+                    for c in cols:
+                        if col is None:
```
HyukjinKwon, line 9351: Is it correct? Seems `col` is always going to be `None`.

databricks/koalas/frame.py
```diff
@@ -9328,34 +9328,57 @@ def filter(self, items=None, like=None, regex=None, axis=None):
...
+                    for item in items:
+                        if col is None:
+                            col = index_scols[0] == F.lit(item)
+                        else:
+                            col = col | (index_scols[0] == F.lit(item))
+                elif len(index_scols) > 1:
+                    # for multi-index
+                    cols = []
+                    col = None
+                    for item in items:
```
HyukjinKwon, line 9342: Let's check if `item` is a tuple and throw the same error like pandas's

HyukjinKwon, line 9342: Let's do something like:

```python
col = None
for item in items:
    # Error if item is not a tuple
    # Error if item is empty
    midx_col = None
    for i, element in enumerate(item):
        if midx_col is None:
            midx_col = index_scols[i] == F.lit(element)
        else:
            midx_col = midx_col & (index_scols[i] == F.lit(element))
    if col is None:
        col = midx_col
    else:
        col = col | midx_col
```

databricks/koalas/frame.py
beobest2: Now, it is checked only when the first value is entered in `col`.

PR #1514, opened 2020-05-19
Standardize leading underscore in namespaces

HyukjinKwon:
Per PEP8, seems the leading underscore in names are simply to indicate to say they are private or internal - nothing special is done with the name itself. Due to this ambiguity, each project or person has a different style. I would like to standardize it here:

Use the leading underscore when:
- clearly it's private and it looks not going to be used across other internal modules. For example, a method or variables only referred within the module or class.
- the private variables or methods are defined under the public API classes documented. For example, `namespace.py`, `Series`, `DataFrame`, `IndexOpsMixin` which is inherited from `Series` and `DataFrame`, etc.

Don't use the leading underscore when:
- when the private method or variables are referred across different modules.
- the module or class itself is undocumented at all. We should treat them as private modules.
- you write a nested function as it is forced as an internal purpose.

In general, I propose to treat everything not documented as internal, and we don't add a leading underscore there when you're uncertain in this case. When it's documented, we should explicitly do the leading understore whenever we need to consider.


HyukjinKwon:
@ueshin @itholic WDYT?

databricks/koalas/base.py
```diff
@@ -1170,21 +1170,21 @@ def value_counts(self, normalize=False, sort=True, ascending=False, bins=None, d
...
+            internal = InternalFrame(
                 spark_frame=sdf,
                 index_map=OrderedDict({index_name: None}),
                 column_labels=column_labels,
                 data_spark_columns=[scol_for(sdf, "count")],
                 column_label_names=self._internal.column_label_names,
             )
 
-        return _col(DataFrame(internal))
+        return first_series(DataFrame(internal))
```
itholic, line 1187: I like the rename !

itholic:
Yes, I strongly agree that we have to set an our own rule for the leading underscore.

This suggestion seems reasonable to me.

HyukjinKwon:
Thanks, @ueshin and @itholic.

PR #1515, opened 2020-05-19
Fix documentations style in DataFrame.pop

HyukjinKwon:
```
Warning, treated as error:
/.../koalas/databricks/koalas/frame.py:docstring of databricks.koalas.DataFrame.pop:6:Unexpected indentation.
```

Similar with https://github.com/databricks/koalas/pull/1509. For some reasons, it didn't fail before but after new more changes in the current master, it started to fail as above. We should fix the style mistakes anyway ..


itholic:
LGTM

PR #972, opened 2019-10-29
Add get_level_values for Index and MultiIndex

charlesdong1991:
This PR is blocked by #973 where `levels` property is implemented, and by #971 where seems @itholic is into fixing it soon ^^

^ above is done!!

```python
mi = pd.MultiIndex.from_arrays((list('abc'), list('def')))
mi.names = ['level_1', 'level_2']
kdf = ks.DataFrame({'a': [1, 2, 3]}, index=mi)

# get values by level number
kdf.index.get_level_values(0)

# get values by level number
kdf.index.get_level_values('level_1')
```

PR #1517, opened 2020-05-20
Add get_level_values for Index and MultiIndex 

itholic:
Take over #972 

```python
mi = pd.MultiIndex.from_arrays((list('abc'), list('def')))
mi.names = ['level_1', 'level_2']
kdf = ks.DataFrame({'a': [1, 2, 3]}, index=mi)

# get values by level number
kdf.index.get_level_values(0)

# get values by level number
kdf.index.get_level_values('level_1')
```


databricks/koalas/indexes.py
```diff
@@ -2498,6 +2517,82 @@ def __getattr__(self, item: str) -> Any:
...
+        lev = self.levels[level]
+        name = self.names[level]
+
+        return ks.Index(lev, name=name)
+
+    def get_level_values(self, level):
+        """
+        Return vector of label values for requested level,
+        equal to the length of the index.
+        Parameters
```
HyukjinKwon, line 2570: Let's add spaces for each section

itholic: Thanks :)

databricks/koalas/indexes.py
```diff
@@ -2498,6 +2518,78 @@ def __getattr__(self, item: str) -> Any:
                 return partial(property_or_func, self)
         raise AttributeError("'MultiIndex' object has no attribute '{}'".format(item))
 
+    def _get_level_number(self, level):
+        """
+        Return the level number if a valid level is given.
+        **this is an internal method**
```
HyukjinKwon, line 2524: No need to document this because leading `_` already means that.

databricks/koalas/indexes.py
```diff
@@ -2498,6 +2518,78 @@ def __getattr__(self, item: str) -> Any:
...
+        """
+        count = self.names.count(level)
+        if (count > 1) and not isinstance(level, int):
+            raise ValueError("The name %s occurs multiple times, use a level number" % level)
+        try:
+            level = self.names.index(level)
```
HyukjinKwon, line 2530: I think you can use if-else instead of relying on try-catch.

HyukjinKwon:
Looks good otherwise.

databricks/koalas/indexes.py
itholic: Agree, I'm gonna remove this

databricks/koalas/indexes.py
itholic: Thanks @HyukjinKwon , Fixed!

PR #1518, opened 2020-05-20
Fix Series.item to run a Spark job once

itholic:
According to the comment https://github.com/databricks/koalas/pull/1502#discussion_r427560507, fixed `Series.item` to run a Spark job not twice, once.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4873,10 +4873,11 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         >>> kser.item()
         10
         """
-        item_top_two = self[:2]
+        scol = self.spark_column
+        item_top_two = self._internal._sdf.select(scol).head(2)
         if len(item_top_two) != 1:
             raise ValueError("can only convert an array of size 1 to a Python scalar")
-        return item_top_two[0]
+        return item_top_two[0][0]
 
     def _cum(self, func, skipna, part_cols=()):
         # This is used to cummin, cummax, cumsum, etc.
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -4873,10 +4873,11 @@ def item(self):
         >>> kser.item()
         10
         """
-        item_top_two = self[:2]
```
HyukjinKwon, line 4876: I think you can just simply fix this line to `self[:2].to_pandas()`

HyukjinKwon: Actually, why don't you just remove all here and call `self[:2].to_pandas().item()`?

ueshin: Also maybe we should use head(2) instead of [:2]

itholic: Thank you guys!

databricks/koalas/tests/test_series.py
```diff
@@ -1595,7 +1595,3 @@ def test_unstack(self):
         # Only support for MultiIndex
         kser = ks.Series([10, -2, 4, 7])
         self.assertRaises(ValueError, lambda: kser.unstack())
-
-    def test_item(self):
-        kser = ks.Series([10, 20])
-        self.assertRaises(ValueError, lambda: kser.item())
```
ueshin, line 1601: Shall we keep this test?

itholic: Thanks, @ueshin !

PR #1521, opened 2020-05-20
Add same_anchor util.

ueshin:
Adding `same_anchor` util to check if the anchors of the given DataFrame or Series are the same or not.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -60,9 +60,10 @@ from databricks.koalas.plot import KoalasSeriesPlotMethods
 from databricks.koalas.ml import corr
 from databricks.koalas.utils import (
     validate_arguments_and_invoke_function,
-    scol_for,
     combine_frames,
     name_like_string,
+    same_anchor,
+    scol_for,
     validate_axis,
     validate_bool_kwarg,
     verify_temp_column_name,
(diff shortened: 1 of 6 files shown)


databricks/koalas/series.py
```diff
@@ -60,9 +60,10 @@
 from databricks.koalas.ml import corr
 from databricks.koalas.utils import (
     validate_arguments_and_invoke_function,
-    scol_for,
     combine_frames,
     name_like_string,
+    same_anchor,
+    scol_for,
```
itholic, line 66: It's really not a big deal, but I'm just curious that why the position of `scol_for` has been moved?

itholic:
LGTM

databricks/koalas/utils.py
```diff
@@ -35,25 +35,27 @@
...
         this_kdf = this
     else:
-        assert isinstance(this, Series), type(this)
+        assert isinstance(this, IndexOpsMixin), type(this)
         this_kdf = this._kdf
     if isinstance(that, DataFrame):
         that_kdf = that
     else:
-        assert isinstance(that, Series), type(that)
+        assert isinstance(that, IndexOpsMixin), type(that)
```
itholic, line 58: Looks better!

databricks/koalas/series.py
ueshin: The order should basically be alphabetical order.
I'd like black or something to manage the order..

itholic: Ah, I got it! Sorry for annoying and thanks for the reply !

ueshin:
Thanks! merging.

PR #1522, opened 2020-05-21
Loosen anchor restriction.

ueshin:
Trying to loosen anchor restriction.

When the underlying Spark DataFrame is the same and index metadata is not modified, we can work without joins.

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -2175,12 +2175,13 @@ class GroupBy(object):
     @staticmethod
     def _resolve_grouping(kdf: DataFrame, by: List[Union[Series, Tuple[str, ...]]]) -> List[Series]:
         new_by_series = []
-        for col_or_s in by:
+        for i, col_or_s in enumerate(by):
             if isinstance(col_or_s, Series):
                 if col_or_s._kdf is kdf:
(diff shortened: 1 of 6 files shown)


databricks/koalas/groupby.py
```diff
@@ -2177,7 +2177,10 @@ def _resolve_grouping(kdf: DataFrame, by: List[Union[Series, Tuple[str, ...]]])
         new_by_series = []
         for col_or_s in by:
             if isinstance(col_or_s, Series):
-                new_by_series.append(col_or_s)
+                if col_or_s._kdf is kdf:
+                    new_by_series.append(col_or_s)
+                else:
+                    new_by_series.append(col_or_s.rename(col_or_s.name))
```
HyukjinKwon, line 2183: Shall we add a comment that the rename is being added because two different DataFrames are joined and the original labels in the name can be lost (is it correct?)

HyukjinKwon, line 2183: Ah, okay. So now it can contain the same Spark column but Koalas' column labels could be different .. shell we leave a short comment here?

ueshin: maybe we should use more explicit name. let me update with a comment.

PR #1528, opened 2020-05-22
Remove InternalFrame.to_external_spark_frame.

ueshin:
Seems like current `InternalFrame.to_external_spark_frame` is only used in `DataFraem.to_spark()` without `index_col` argument.
I think we can just remove it for now and consolidate the logic between with and without `index_col` argument.


itholic:
LGTM

PR #1529, opened 2020-05-22
Remove the deprecated pandas_wraps

HyukjinKwon:
This PR proposes to remote `pandas_wraps` as it's deprecated.

There are two more deprecated functions, `map_in_pandas`, and `get_dtype_counts`. The former is referred more publicity for example in https://databricks.com/blog/2020/03/31/10-minutes-from-pandas-to-koalas-on-apache-spark.html. I would like to delay the removal.

The latter is used in `DataFrame.info` internally when pandas < 1.0.0 is used. We can remove it when we use minimum pandas version as 1.0.0.


databricks/koalas/typedef/typehints.py
```diff
@@ -75,29 +75,6 @@ def __repr__(self):
         return "UnknownType[{}]".format(self.tpe)
 
 
-def _to_stype(tpe) -> typing.Union[SeriesType, DataFrameType, ScalarType, UnknownType]:
```
HyukjinKwon, line 78: Moved to `infer_return_type` as is.

ueshin:
Thanks! merging.

PR #1533, opened 2020-05-23
Refine median / quantile.

ueshin:
Introducing `databricks.koalas.functions` module and `approx_percentile` function in it, and use the function for `median` and `quantile`.


databricks/koalas/__init__.py
```diff
@@ -106,6 +106,13 @@ def _auto_patch():
     import os
     import logging
 
+    from pyspark.sql import dataframe as df, functions as F
+    from databricks.koalas import functions
+
+    for name in functions.__all__:
+        if not hasattr(F, name):
+            setattr(F, name, getattr(functions, name))
```
HyukjinKwon, line 114: Hm.. should we auto-patch? I am less sure on this yet as `percentile_approx` seems a Spark function and auto-patching it in Koalas seems a bit counterintuitive.

databricks/koalas/functions.py
```diff
@@ -0,0 +1,70 @@
+#
```
HyukjinKwon, line 1: Maybe it's best to add module docs that this file is supposed to have PySpark functions. We could name it `functions.py` to something like `spark_functions.py` too ...

ueshin:
@HyukjinKwon I updated this PR, could you take another look?

PR #1535, opened 2020-05-24
Add Series.spark.apply

### LINKED ISSUE #1523: Apply Spark Column operations directly on a Series
On issue #1492, I noticed it was noted the following operation was possible on a column of a Koalas `DataFrame`:
```python
>>> df = ks.DataFrame(["example"], columns=["column"])
>>> from pyspark.sql import functions as F
>>> df["column"] = F.trim(F.upper(F.col("column")))
>>> df
    column
0  EXAMPLE
```

By any chance, is there currently a way to apply Spark SQL/Column functions to a Koalas `Series`? I can imagine it looking something like so:
```python
>>> import pyspark.sql.functions as F
>>> kss = ks.Series(["example"])
>>> kss.apply(F.trim(F.upper(kss.spark_column)))
0    EXAMPLE
Name: 0, dtype: object
```

HyukjinKwon:
This PR adds `Series.spark.apply` to allow arbitrary length such as aggregation:

```python
import pyspark.sql.functions as F
kss = ks.Series([1, 2, 3, 4])
kss.spark.apply(lambda s: F.collect_list(s))
```

```
0    [1, 2, 3, 4]
Name: 0, dtype: object
```

Note that It will lose index which can result in performance degradation due to using the default index. I don't think we should encourage this API.

### DIFF
--- a/databricks/koalas/tests/test_indexops_spark.py
+++ b/databricks/koalas/tests/test_indexops_spark.py
@@ -50,3 +50,16 @@ class SparkIndexOpsMethodsTest(ReusedSQLTestCase, SQLTestUtils):
             )
             s = ks.Series([45, 200, 1.2, 30, 250, 1.5, 320, 1, 0.3], index=midx)
             s.index.spark.transform(lambda scol: scol)
+
+    def test_series_apply_negative(self):
+        with self.assertRaisesRegex(
+            ValueError, "The output of the function.* pyspark.sql.Column.*int"
+        ):
+            self.kser.spark.apply(lambda scol: 1)
+
+        with self.assertRaisesRegex(AnalysisException, "cannot resolve.*non-existent.*"):
+            self.kser.spark.transform(lambda scol: F.col("non-existent"))
+
+    def test_index_apply_negative(self):
+        with self.assertRaisesRegex(NotImplementedError, "Index does not support spark.apply yet"):
+            ks.range(10).index.spark.apply(lambda scol: scol)
--- a/databricks/koalas/spark/accessors.py
+++ b/databricks/koalas/spark/accessors.py
@@ -56,9 +56,9 @@ class SparkIndexOpsMethods(object):
 
     def transform(self, func):
         """
-        Applies a function that takes and returns a Spark column. It allows natively
+        Applies a function that takes and returns a Spark column. It allows to natively
         apply a Spark function and column APIs with the Spark column internally used
-        in Series or Index.
+        in Series or Index. The output length of the Spark column should be same as input's.
 
         .. note:: It requires to have the same input and output length; therefore,
             the aggregate Spark functions such as count does not work.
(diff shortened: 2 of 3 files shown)


databricks/koalas/spark.py
```diff
@@ -117,6 +117,75 @@ def transform(self, func):
...
         return new_ser
 
+    def apply(self, func):
+        """
+        Applies a function that takes and returns a Spark column. It allows to natively
+        apply a Spark function and column APIs with the Spark column internally used
+        in Series or Index.
+
+        .. note:: It forces to lose the index and end up with using default index. It is
+            preferred to use :meth:`Series.spark.transform` or `:meth:`DataFrame.spark.apply`
```
HyukjinKwon, line 127: `DataFrame.spark.apply` is at https://github.com/databricks/koalas/pull/1536

databricks/koalas/tests/test_indexops_spark.py
```diff
@@ -50,3 +50,16 @@ def test_multiindex_transform_negative(self):
...
+    def test_series_apply_negative(self):
+        with self.assertRaisesRegex(
+            ValueError, "The output of the function.* pyspark.sql.Column.*int"
+        ):
+            self.kser.spark.apply(lambda scol: 1)
+
+        with self.assertRaisesRegex(AnalysisException, "cannot resolve.*non-existent.*"):
+            self.kser.spark.transform(lambda scol: F.col("non-existent"))
+
+    def test_index_apply_negative(self):
```
itholic, line 63: Just curious, what is the reason we don't have positive tests for now ?
(Maybe because pandas doesn't have these kind of functions ?)

itholic:
LGTM except a nit question.

databricks/koalas/tests/test_indexops_spark.py
HyukjinKwon: Because I have doctests and these look handling the cases enough :-).

itholic: @HyukjinKwon I got it! Makes sense, thanks :)

databricks/koalas/spark.py
```diff
@@ -117,6 +117,75 @@ def transform(self, func):
...
+            raise NotImplementedError("Index does not support spark.apply yet.")
+        output = func(self._data.spark.column)
+        if not isinstance(output, Column):
+            raise ValueError(
+                "The output of the function [%s] should be of a "
+                "pyspark.sql.Column; however, got [%s]." % (func, type(output))
+            )
+        assert isinstance(self._data, Series)
+
+        sdf = self._data._kdf._internal.to_internal_spark_frame.select(output)
```
ueshin, line 183: `self._data._internal.spark_frame.select(output)`?

databricks/koalas/spark.py
```diff
@@ -117,6 +117,75 @@ def transform(self, func):
...
+            raise NotImplementedError("Index does not support spark.apply yet.")
+        output = func(self._data.spark.column)
+        if not isinstance(output, Column):
+            raise ValueError(
+                "The output of the function [%s] should be of a "
+                "pyspark.sql.Column; however, got [%s]." % (func, type(output))
+            )
+        assert isinstance(self._data, Series)
+
+        sdf = self._data._kdf._internal.spark_frame.select(output)
```
ueshin, line 183: I don't think we need to use `_kdf`.
Should be `self._data._internal.spark_frame.select(output)`.

HyukjinKwon: It actually need when you need to do `lambda s: s + another_column_in_kdf`

HyukjinKwon:
rebased to resolve conflicts

databricks/koalas/spark.py
HyukjinKwon: Oh, yes, I can switch because we're using `spark_frame` now.

PR #1537, opened 2020-05-25
Implement DataFrame.mad

90jam:
Implement `DataFrame.mad`

- Basic Examples
```python
>>> df = ks.DataFrame({'a': [1, 2, 3, np.nan], 'b': [0.1, 0.2, 0.3, np.nan]}, columns=['a', 'b'])
>>> df.mad()
a    0.666667
b    0.066667
Name: 0, dtype: float64

>>> df.mad(axis=1)
0    0.45
1    0.90
2    1.35
3     NaN
Name: 0, dtype: float64
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10049,6 +10049,64 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         internal = self._internal.copy(spark_frame=sdf, data_spark_columns=data_scols)
         return DataFrame(internal)
 
+    def mad(self, axis=0):
+        """
+        Return the mean absolute deviation of values.
+
+        Parameters
+        ----------
+        axis : {index (0), columns (1)}
+            Axis for the function to be applied on.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({'a': [1, 2, 3, np.nan], 'b': [0.1, 0.2, 0.3, np.nan]},
+        ...                   columns=['a', 'b'])
+
+        >>> df.mad()
+        a    0.666667
+        b    0.066667
+        Name: 0, dtype: float64
+
+        >>> df.mad(axis=1)
+        0    0.45
+        1    0.90
+        2    1.35
+        3     NaN
+        Name: 0, dtype: float64
+        """
+        from databricks.koalas import Series
+
+        axis = validate_axis(axis)
+
+        # Here we execute with the first 1000 to get the return type.
+        # If the records were less than 1000, it uses pandas API directly for a shortcut.
+        limit = get_option("compute.shortcut_limit")
+        pdf = self.head(limit + 1)._to_internal_pandas()
+        pser = pdf.mad(axis=axis)
+        if len(pdf) <= limit:
+            return Series(pser)
+
+        if axis == 0:
+            mean_kdf = self._reduce_for_stat_function(F.mean, name="mean", axis=0)
+
+            return self._reduce_for_stat_function(
+                lambda col_sdf: F.avg(F.abs(col_sdf - mean_kdf[col_sdf._jc.toString()])),
+                name="mad",
+                axis=0,
+            )
+
+        elif axis == 1:
+
+            @pandas_udf(returnType=as_spark_type(pser.dtype.type))
+            def calculate_columns_axis(*cols):
+                return pd.concat(cols, axis=1).mad(axis=1)
+
+            df = self._sdf.select(
+                calculate_columns_axis(*self._internal.data_spark_columns).alias("0")
+            )
+            return DataFrame(df)["0"]
+
     def _to_internal_pandas(self):
         """
         Return a pandas DataFrame directly from _internal to avoid overhead of copy.
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10118,10 +10118,12 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             def calculate_columns_axis(*cols):
                 return pd.concat(cols, axis=1).mad(axis=1)
 
-            df = self._sdf.select(
-                calculate_columns_axis(*self._internal.data_spark_columns).alias("0")
+            internal = self._internal.copy(
+                spark_column=calculate_columns_axis(*self._internal.data_spark_columns).alias("0"),
+                column_labels=[("0",)],
+                column_label_names=None,
             )
-            return DataFrame(df)["0"]
+            return Series(internal, anchor=self)
 
     def _to_internal_pandas(self):
         """
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10076,17 +10076,46 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         Name: 0, dtype: float64
         """
         from databricks.koalas import Series
+        from databricks.koalas.series import first_series
 
         axis = validate_axis(axis)
 
         if axis == 0:
-            mean_kdf = self._reduce_for_stat_function(F.mean, name="mean", axis=0)
+            exprs = []
+            new_column_labels = []
+            for label in self._internal.column_labels:
+                col_sdf = self._internal.spark_column_for(label)
+                col_type = self._internal.spark_type_for(label)
 
-            return self._reduce_for_stat_function(
-                lambda col_sdf: F.avg(F.abs(col_sdf - mean_kdf[col_sdf._jc.toString()])),
-                name="mad",
-                axis=0,
-            )
+                if isinstance(col_type, BooleanType):
+                    col_sdf = col_sdf.cast("integer")
+
+                label_str = name_like_string(label)
+
+                col_sdf = F.avg(
+                    F.abs(
+                        col_sdf
+                        - self._sdf.select(F.avg(col_sdf).alias(label_str)).first()[label_str]
+                    )
+                )
+
+                exprs.append(col_sdf.alias(name_like_string(label)))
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,64 @@ def explode(self, column):
...
+
+        # Here we execute with the first 1000 to get the return type.
+        # If the records were less than 1000, it uses pandas API directly for a shortcut.
+        limit = get_option("compute.shortcut_limit")
+        pdf = self.head(limit + 1)._to_internal_pandas()
+        pser = pdf.mad(axis=axis)
+        if len(pdf) <= limit:
+            return Series(pser)
+
+        if axis == 0:
```
itholic, line 10090: I think we don't need to infer the return type when `axis=0`.

Maybe we should run this lines when only `axis=1`.

itholic: Anyway, I think the line 10082 to 10108 is duplicated with `_reduce_for_stat_function`, we should duplicate it or implement this method without using `_reduce_for_stat_function` like you did in `Series.mad`

90jam: Thanks, I applied it.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -3607,3 +3607,17 @@ def test_explain_hint(self):
...
+
+        self.assert_eq(kdf.mad(), pdf.mad())
+        self.assert_eq(kdf.mad(axis=1), pdf.mad(axis=1))
+
+        with self.assertRaises(ValueError):
+            kdf.mad(axis=2)
+
+        with option_context("compute.shortcut_limit", 1):
+            self.assert_eq(kdf.mad(), pdf.mad())
+            self.assert_eq(kdf.mad(axis=1), pdf.mad(axis=1))
```
itholic, line 3623: Can you add tests for MultiIndex columns ?

you can like:

```python
columns = pd.MultiIndex.from_tuples([('A', 'X'), ('A', 'Y')])
pdf.columns = columns
kdf.columns = columns
self.assert_eq(kdf.mad(), pdf.mad())
```

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,90 @@ def explode(self, column):
...
+                return first_series(DataFrame(internal).transpose())
+
+        elif axis == 1:
+            limit = get_option("compute.shortcut_limit")
+            pdf = self.head(limit + 1)._to_internal_pandas()
+            pser = pdf.mad(axis=axis)
+            if len(pdf) <= limit:
+                return Series(pser)
+
+            @pandas_udf(returnType=as_spark_type(pser.dtype.type))
```
HyukjinKwon, line 10127: I took a quick look for `DataFrame.mad` in pandas. Seems it inherits the numpy behaviour from `numpy.mean` which will likely always return a float. I think we can just set the return type as `DoubleType()` and forget about the type infernece.

HyukjinKwon: In this case, I think you can also remove the shortcut handling as well.

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,90 @@ def explode(self, column):
...
+            def calculate_columns_axis(*cols):
+                return pd.concat(cols, axis=1).mad(axis=1)
+
+            df = self._sdf.select(
+                calculate_columns_axis(*self._internal.data_spark_columns).alias("0")
+            )
+            return DataFrame(df)["0"]
```
HyukjinKwon, line 10134: I think you can keep the same anchor in this case, for example, as below:

```python
    spark_column= calculate_columns_axis(*self._internal.data_spark_columns).alias("0"),
    column_labels=[("0",)],
    column_label_names=None,
)
return Series(internal, anchor=self)
```

So, the same DataFrame does not require to turn on `compute.ops_on_diff_frames` when it performs a computation against the series from here. e.g.) `kdf.col + kdf.mad(axis=1)`

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,90 @@ def explode(self, column):
...
+
+                col_sdf = F.avg(
+                    F.abs(
+                        col_sdf
+                        - self._sdf.select(F.avg(col_sdf).alias(label_str)).first()[label_str]
+                    )
+                )
+
+                exprs.append(col_sdf.alias(label_str))
+                new_column_labels.append(label)
```
HyukjinKwon, line 10103: Seems `new_column_labels` is always same as `self._internal.column_labels`.

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,90 @@ def explode(self, column):
...
+        from databricks.koalas import Series
+        from databricks.koalas.series import first_series
+
+        axis = validate_axis(axis)
+
+        if axis == 0:
+            exprs = []
+            new_column_labels = []
+            for label in self._internal.column_labels:
+                col_sdf = self._internal.spark_column_for(label)
```
HyukjinKwon, line 10087: I would name it like `scol` or `new_scol`

databricks/koalas/tests/test_dataframe.py
90jam: Thanks :) I added an additional test for the MultiIndex columns.

itholic: great!

databricks/koalas/frame.py
90jam: Oh, I agree that type inference is unnecessary! I forgot about it :)

databricks/koalas/frame.py
90jam: Really. I removed it.

databricks/koalas/frame.py
90jam: I Agree. 
Renamed it to scol.

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,80 @@ def explode(self, column):
...
+                scol = self._internal.spark_column_for(label)
+                col_type = self._internal.spark_type_for(label)
+
+                if isinstance(col_type, BooleanType):
+                    scol = scol.cast("integer")
+
+                label_str = name_like_string(label)
+
+                scol = F.avg(
+                    F.abs(scol - self._sdf.select(F.avg(scol).alias(label_str)).first()[label_str])
```
itholic, line 10095: Hmm... I think we'd better to avoid using `first()` for every single loop since It occurs a Spark job.

databricks/koalas/frame.py
90jam: Thanks to the review, I could understand the internal copy. I modified the code as in the example.

databricks/koalas/frame.py
90jam: I think you are right. I refactored to call first() only once.

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,88 @@ def explode(self, column):
...
+                for label in self._internal.column_labels
+            ]
+            sdf = self._sdf.select(new_columns)
+
+            with ks.option_context(
+                "compute.default_index_type", "distributed", "compute.max_rows", None
+            ):
+                kdf = DataFrame(sdf)
+                internal = InternalFrame(
+                    kdf._internal.spark_frame,
```
itholic, line 10113: How about `spark_frame=kdf._internal.spark_frame` ?

90jam: Sounds good.

HyukjinKwon:
Looks pretty good to me

itholic:
Looks fine to me

HyukjinKwon:
@ueshin, can you take a look when you're available?

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,88 @@ def explode(self, column):
...
+                if isinstance(col_type, BooleanType):
+                    scol = scol.cast("integer")
+
+                return scol
+
+            new_columns = [
+                F.avg(get_spark_column(self, label)).alias(name_like_string(label))
+                for label in self._internal.column_labels
+            ]
+            sdf_mean = self._sdf.select(new_columns).first()
```
ueshin, line 10098: Could you use `self._internal.spark_frame` instead of `self._sdf`?

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,88 @@ def explode(self, column):
...
+
+            new_columns = [
+                F.avg(
+                    F.abs(get_spark_column(self, label) - sdf_mean[name_like_string(label)])
+                ).alias(name_like_string(label))
+                for label in self._internal.column_labels
+            ]
+            sdf = self._sdf.select(new_columns)
```
ueshin, line 10106: ditto.

databricks/koalas/frame.py
```diff
@@ -10049,6 +10049,88 @@ def explode(self, column):
...
+
+        >>> df.mad(axis=1)
+        0    0.45
+        1    0.90
+        2    1.35
+        3     NaN
+        Name: 0, dtype: float64
+        """
+        from databricks.koalas import Series
+        from databricks.koalas.series import first_series
```
ueshin, line 10079: nit: the following should work?

```py
from databricks.koalas.series import Series, first_series
```

databricks/koalas/frame.py
ueshin, line 10098: Shall we use a different name for `sdf_mean` since it is actually not a Spark DataFrame.

90jam: Thanks for the reviews. I reflected the above.

PR #1541, opened 2020-05-26
Use collections.abc to suppress warnings.

ueshin:
Use `collections.abc` instead of `collection` to suppress the following warnings:

```
/path/to/databricks/koalas/groupby.py:23: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
  from collections import Callable, OrderedDict, namedtuple
```


ueshin:
I'd merge this now. Please feel free to leave comments if any.

itholic:
LGTM!

PR #1547, opened 2020-05-28
Add Index.holds_integer

itholic:
This PR proposes Index.holds_integer.

```python
>>> kidx = ks.Index([1, 2, 3, 4])
>>> kidx.holds_integer()
True
```

Returns False for string type.

```python
>>> kidx = ks.Index(["A", "B", "C", "D"])
>>> kidx.holds_integer()
False
```

Returns False for float type.

```python
>>> kidx = ks.Index([1.1, 2.2, 3.3, 4.4])
>>> kidx.holds_integer()
False
```


databricks/koalas/indexes.py
```diff
@@ -2692,6 +2716,13 @@ def get_level_values(self, level):
         )
         return ks.DataFrame(internal).index
 
+    def holds_integer(self):
+        """
+        Whether the type is an integer type.
+        Always return False for MultiIndex.
+        """
+        return False
```
ueshin, line 2724: Do we need this?

databricks/koalas/indexes.py
```diff
@@ -1884,6 +1884,30 @@ def union(self, other, sort=None):
...
+        >>> kidx.holds_integer()
+        False
+
+        Returns False for float type.
+
+        >>> kidx = ks.Index([1.1, 2.2, 3.3, 4.4])
+        >>> kidx.holds_integer()
+        False
+        """
+        return isinstance(self.spark.data_type, LongType)
```
HyukjinKwon, line 1909: Does it work if it contains nulls?

itholic: Hmm.. good point.

If Index contains null values, the type of index will be `float64` in Koalas, so the result is gonna be different with pandas.

```python
>>> ks.Index([1, 2, None])
Float64Index([1.0, 2.0, nan], dtype='float64')

>>> pd.Index([1, 2, None])
Index([1, 2, None], dtype='object')
```

I'll add the different behaviour to the Notes since confirming whether the type of values except null are all integer might be expensive.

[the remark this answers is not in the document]
itholic: Seems better, thanks!! :)

databricks/koalas/indexes.py
itholic: Oh, I think we don't need this.
I moved the docstring to the `Index.holds_integer` and removed the `MultiIndex.holds_integer`.

databricks/koalas/indexes.py
itholic: I just added Notes !

HyukjinKwon:
@itholic can you rebase?

PR #1548, opened 2020-05-28
Use spark accessor directly in Index.spark_type.

ueshin:
Using Spark accessor directly in `Index.spark_type` should be fine, and we can show more proper warning.


itholic:
LGTM

PR #1550, opened 2020-05-28
Fix Series.fillna to avoid Spark jobs.

ueshin:
`DataFrame.fillna` reuses `Series.fillna` which always runs a Spark job to check if there are null values, then it ends up with running many jobs if the DataFrame has a lot of columns.
We should avoid it, and we can check its nullability defined by Spark instead.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -637,21 +637,25 @@ class InternalFrame(object):
             (sdf[offset_column] + sdf[row_number_column] - 1).alias(column_name), *scols
         )
 
-    def spark_column_name_for(self, labels: Tuple[str, ...]) -> str:
-        """ Return the actual Spark column name for the given column name. """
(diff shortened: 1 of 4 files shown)


databricks/koalas/internal.py
```diff
@@ -637,21 +637,25 @@ def offset(id):
             (sdf[offset_column] + sdf[row_number_column] - 1).alias(column_name), *scols
         )
 
-    def spark_column_name_for(self, labels: Tuple[str, ...]) -> str:
-        """ Return the actual Spark column name for the given column name. """
-        return self.spark_frame.select(self.spark_column_for(labels)).columns[0]
+    def spark_column_name_for(self, label: Tuple[str, ...]) -> str:
```
itholic, line 640: Love this detail :)

itholic:
LGTM if tests pass

PR #1551, opened 2020-05-28
Use spark.data_type instead of spark_type.

ueshin:
There are some places remaining using `spark_type` instead of `spark.data_type`.


itholic:
LGTM if tests pass

PR #1552, opened 2020-05-28
Add dataframe property to spark accessors

itholic:
Since we have `spark` namespace now, proposes adding `dataframe` to the spark accessors for accessing to the spark dataframe more intuitively.

I think `self.spark.dataframe` is a bit more intuitive than `self._internal.spark_frame`.

### DIFF
--- a/databricks/koalas/spark/accessors.py
+++ b/databricks/koalas/spark/accessors.py
@@ -54,6 +54,11 @@ class SparkIndexOpsMethods(object):
         """
         return self._data._internal.spark_column
 
+    @property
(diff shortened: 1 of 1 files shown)


ueshin:
I don't think we should provide it. `self._internal.spark_frame` is for internal use only. You can use `to_spark()` instead.

itholic:
@ueshin Ah, okay then, closing this now. Thanks for the opinion! :)

HyukjinKwon:
Yeah, let's keep `spark_frame` as the internal purpose for now.

PR #1554, opened 2020-05-29
Manage data_spark_columns to avoid creating very many Spark DataFrames.

ueshin:
This PR makes huge changes on the way of `InternalFrame` management to avoid creating very many Spark DataFrames.
It will make a lot of DataFrame operations without enabling "compute.ops_on_diff_frames" option possible.

```py
df + df + df
df['foo'] = df['bar']['baz']
df[['x', 'y']] = df[['x', 'y']].fillna(0)
```

The new way in functions to manage `InternalFrame` is:

- If the function only needs to handle Spark columns, e.g., functions using `InternalFrame. with_new_columns`, use the new columns without creating new Spark DataFrame. Basically we can just use the `with_new_columns` to create a new `InternalFrame`.
- If the function needs to create a new Spark DataFrame, e.g., functions need filter, order, groupby, use `_internal.spark_frame` with columns from `_internal.spark_column_for`. Working with a Spark DataFrame from `_internal.spark_frame` and column names from `_internal.spark_column_name_for` will usually NOT work.
- If the function can only access by the column name, e.g., `pivot_table` or functions with udfs, use `_internal.applied.spark_frame` instead. The `_internal.applied.spark_frame` will be applied all the changes. Note that the `_internal.applied.spark_frame` won't work with Spark columns from `_internal.spark_column_for`.
- `DataFrame._sdf` was removed to explicitly specify which `spark_frame` should be used, `_internal.spark_frame` or `_internal.applied.spark_frame`.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -143,6 +143,12 @@ circle          0      720
 triangle        6      360
 rectangle       8      720
 
+>>> df + df + df
+           angles  degrees
+circle          0     1080
+triangle        9      540
+rectangle      12     1080
+
 >>> df.radd(1)
            angles  degrees
 circle          1      361
@@ -409,10 +415,6 @@ class DataFrame(Frame, Generic[T]):
             super(DataFrame, self).__init__(InternalFrame.from_pandas(pdf))
 
     @property
-    def _sdf(self) -> spark.DataFrame:
-        return self._internal.spark_frame
-
-    @property
     def ndim(self):
         """
         Return an int representing the number of array dimensions.
@@ -505,7 +507,7 @@ class DataFrame(Frame, Generic[T]):
                     exprs.append(col_sdf.alias(name_like_string(label)))
                     new_column_labels.append(label)
 
-            sdf = self._sdf.select(*exprs)
+            sdf = self._internal.spark_frame.select(*exprs)
 
             # The data is expected to be small so it's fine to transpose/use default index.
             with ks.option_context(
@@ -534,7 +536,7 @@ class DataFrame(Frame, Generic[T]):
             def calculate_columns_axis(*cols):
                 return getattr(pd.concat(cols, axis=1), name)(axis=axis, numeric_only=numeric_only)
 
-            df = self._sdf.select(
+            df = self._internal.spark_frame.select(
                 calculate_columns_axis(*self._internal.data_spark_columns).alias("0")
             )
             return DataFrame(df)["0"]
@@ -1253,7 +1255,9 @@ class DataFrame(Frame, Generic[T]):
             v = [row[c] for c in internal_data_columns]
             return k, v
 
-        for k, v in map(extract_kv_from_spark_row, self._sdf.toLocalIterator()):
+        for k, v in map(
+            extract_kv_from_spark_row, self._internal.applied.spark_frame.toLocalIterator()
+        ):
             s = pd.Series(v, index=columns, name=k)
             yield k, s
 
@@ -1945,7 +1949,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             )
         )
 
-        exploded_df = self._sdf.withColumn("pairs", pairs).select(
+        exploded_df = self._internal.spark_frame.withColumn("pairs", pairs).select(
             [
                 F.to_json(
                     F.struct(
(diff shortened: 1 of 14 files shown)


ueshin:
also cc @itholic

databricks/koalas/internal.py
```diff
@@ -793,12 +793,26 @@ def to_pandas_frame(self) -> pd.DataFrame:
...
+    @lazy_property
+    def applied(self):
+        """ Copy the immutable InternalFrame with the updates applied. """
+        sdf = self.spark_frame.select(self.spark_columns + list(HIDDEN_COLUMNS))
+        if self.spark_column is None:
+            return self.copy(
+                spark_frame=sdf,
+                data_spark_columns=[scol_for(sdf, col) for col in self.data_spark_column_names],
+            )
+        else:
```
itholic, line 805: Just for double checking.
`if` for returning the applied sdf for the `DataFrame` and `else` for the `Series`, right ?

ueshin: Yes, right.

itholic:
> ... Working with a Spark DataFrame from `_internal.spark_frame` and column names from `InternalFrame.spark_column_name_for` will usually NOT work.

Maybe `_internal.applied.spark_frame` and column names from `_internal.spark_column_name_for` will usually NOT work ??
(+ nit: what is the difference between `_internal` and `InternalFrame` ? - might be a typo ?)

databricks/koalas/internal.py
itholic: Thanks !

itholic:
Seems good enough to me except several questions.

ueshin:
@itholic ah, right. it's a typo. updated the description.

itholic:
Oh, sorry I just noticed that I missed reviewing some files. let me check these tonight.

databricks/koalas/frame.py
```diff
@@ -408,10 +414,6 @@ def __init__(self, data=None, index=None, columns=None, dtype=None, copy=False):
                 pdf = pd.DataFrame(data=data, index=index, columns=columns, dtype=dtype, copy=copy)
             super(DataFrame, self).__init__(InternalFrame.from_pandas(pdf))
 
-    @property
-    def _sdf(self) -> spark.DataFrame:
```
HyukjinKwon, line 412: +1

databricks/koalas/internal.py
```diff
@@ -793,12 +793,26 @@ def to_pandas_frame(self) -> pd.DataFrame:
             ]
         return pdf
 
+    @lazy_property
+    def applied(self):
```
HyukjinKwon, line 797: Should we maybe call it like:
- `resolved_copy`
- `applied_copy`
- `new_sdf_copy`
- ...

?

The name `applied` doesn't look clear that it's going to have a new Spark DataFrame internally that changes anchor.

databricks/koalas/internal.py
```diff
@@ -698,7 +698,7 @@ def spark_columns(self) -> List[spark.Column]:
         index_spark_columns = self.index_spark_columns
```
HyukjinKwon, line 698: Shall we add some docstrings clearly to describe when to use which clearly?

For example, `spark_frame` now should always be used via `df._internal.applied.spark_frame` for Spark DataFrame APIs that internally creates new query execution plan.

For expressions and/or functions, `df._internal.spark_frame` should be used together with Spark column instances, in order to avoid the operations on different DataFrames.

HyukjinKwon: Maybe we could also document that `spark_frame` is just the base Spark DataFrame where the expressions and functions are not applied.

We might have to consider mentioning about:
- Spark expressions/functions, to create new Spark Columns against the same DataFrame.
- Spark DataFrame APIs that internally creates query execution plans, to create a new DataFrame.

HyukjinKwon:
Looks good. The documentation (https://github.com/databricks/koalas/pull/1554#discussion_r433032633) I can do it in a separate PR if you're busy for something else.

itholic:
Um.. honestly `resolved_copy` seems more complicated for me than `applied`.

But I think we can merge this for now since I have no better idea of naming it.

Otherwise, LGTM

ueshin:
Thanks! I'd merge this now.

PR #1557, opened 2020-05-31
[MINOR] fix output of doctest in internal.py

itholic:
While I'm taking a look & reminding the internal.py, fixed some words and output of doctests.


itholic:
While I'm here, let me find if there is also minor fixes in other related files.

PR #1560, opened 2020-06-02
Implements Index.IndexesTest

beobest2:
Implementing `Index.putmask`

```python
>>> kidx = ks.Index(['a', 'b', 'c', 'd', 'e'])
>>> mask = [True if x < 2 else False for x in range(5)]
>>> value = 100

>>> kidx
Index(['a', 'b', 'c', 'd', 'e'], dtype='object')

>>> kidx.putmask(mask, value)
Index(['100', 'd', '100', 'c', 'e'], dtype='object')
```


databricks/koalas/indexes.py
```diff
@@ -1539,6 +1540,54 @@ def argmin(self):
...
+        Example
+        -------
+        >>> kidx = ks.Index(['a', 'b', 'c', 'd', 'e'])
+        >>> kidx
+        Index(['a', 'b', 'c', 'd', 'e'], dtype='object')
+
+        >>> kidx.putmask([True if x < 2 else False for x in range(5)], "Koalas").sort_values()
+        Index(['Koalas', 'Koalas', 'c', 'd', 'e'], dtype='object')
+        """
+        origin_col = self._internal.index_spark_column_names[0]
```
itholic, line 1563: I think It's not column, how about `scol_name` ?

databricks/koalas/indexes.py
```diff
@@ -1539,6 +1540,54 @@ def argmin(self):
...
+        sdf = self._internal.spark_frame.select(self.spark.column)
+
+        sequence_col = verify_temp_column_name(sdf, "__distributed_sequence_column__")
```
itholic, line 1566: How about `dist_sequence_col` because Koalas also has a `sequence` as an index type.

databricks/koalas/indexes.py
```diff
@@ -1539,6 +1540,54 @@ def argmin(self):
...
+        sdf = InternalFrame.attach_distributed_sequence_column(sdf, column_name=sequence_col)
+
+        masking_col = verify_temp_column_name(sdf, "__masking_column__")
+        masking_udf = udf(lambda x: mask[x], BooleanType())
```
itholic, line 1570: Seems not work when `mask = kidx > 2` ?

databricks/koalas/tests/test_indexes.py
```diff
@@ -278,6 +278,17 @@ def test_dropna(self):
         self.assert_eq(kidx.dropna(), pidx.dropna())
         self.assert_eq((kidx + 1).dropna(), (pidx + 1).dropna())
 
+    def test_putmask(self):
+        pidx = pd.Index(["a", "b", "c", "d", "e"])
+        kidx = ks.from_pandas(pidx)
+
+        mask = [True if x < 2 else False for x in range(5)]
```
itholic, line 285: Let's add more test cases for `mask` and `value`.

`value` also can be an Index or a Series, and even can be a list or a tuple. 🤪

beobest2: I will modify it to work in the cases mentioned above as well :)

itholic:
Could you also delete the `put_mask` of MultiIndex like the below and implement it?

```diff
@@ -58,7 +58,6 @@ class MissingPandasLikeIndex(object):
     is_type_compatible = _unsupported_function("is_type_compatible")
     join = _unsupported_function("join")
     map = _unsupported_function("map")
-    putmask = _unsupported_function("putmask")
     ravel = _unsupported_function("ravel")
     reindex = _unsupported_function("reindex")
     searchsorted = _unsupported_function("searchsorted")
@@ -131,7 +130,6 @@ class MissingPandasLikeMultiIndex(object):
     is_type_compatible = _unsupported_function("is_type_compatible")
     join = _unsupported_function("join")
     map = _unsupported_function("map")
-    putmask = _unsupported_function("putmask")
     ravel = _unsupported_function("ravel")
     reindex = _unsupported_function("reindex")
     remove_unused_levels = _unsupported_function("remove_unused_levels")
```

databricks/koalas/indexes.py
```diff
@@ -1539,6 +1540,56 @@ def argmin(self):
...
+        -------
+        Index
+
+        Example
+        -------
+        >>> kidx = ks.Index(['a', 'b', 'c', 'd', 'e'])
+        >>> kidx
+        Index(['a', 'b', 'c', 'd', 'e'], dtype='object')
+
+        >>> kidx.putmask([True if x < 2 else False for x in range(5)], "Koalas").sort_values()
```
itholic, line 1560: Let's fix this doctest more practically like the below:

```python
>>> kidx.putmask(kidx > "b", "Koalas").sort_values()
```

databricks/koalas/tests/test_indexes.py
beobest2: @itholic I modified the `mask` and `value` parameters to support `Series` and `Index`
and also `list` and `tuple`.
However, in numeric values, there is a problem that the output of the result through the internal spark operation is converted to a string. I am stuck here, 😭 could you help me how to solve this?

```
>>> pidx = pd.Index([1, 2, 3, 4, 5])
>>> kidx = ks.from_pandas(pidx)

>>> kidx.putmask(kidx < 3, [100, 200, 300, 400, 500]).sort_values()
Index(['100', '200', '3', '4', '5'], dtype='object')                            

>>> pidx.putmask(pidx < 3, [100, 200, 300, 400, 500]).sort_values()
Int64Index([3, 4, 5, 100, 200], dtype='int64')
```

HyukjinKwon:
@beobest2 can you fix the test?

beobest2:
@HyukjinKwon okay I'll fix the test

databricks/koalas/tests/test_indexes.py
beobest2: To prevent data type changes, I modified code to use pyspark's `_infer_type` to infer the data type from `value` parameter. Please check if I can use this function.

databricks/koalas/indexes.py
```diff
@@ -1540,6 +1548,75 @@ def argmin(self):
...
+        dist_sequence_col_name = verify_temp_column_name(sdf, "__distributed_sequence_column__")
+        sdf = InternalFrame.attach_distributed_sequence_column(
+            sdf, column_name=dist_sequence_col_name
+        )
+
+        replace_col = verify_temp_column_name(sdf, "__replace_column__")
+        masking_col = verify_temp_column_name(sdf, "__masking_column__")
+
+        if isinstance(value, (list, tuple)):
+            replace_udf = udf(lambda x: value[x], _infer_type(value[0]))
```
ueshin, line 1586: Is it possible to use `pandas_udf` instead of `udf`? If possible, could you replace with it?

databricks/koalas/indexes.py
```diff
@@ -1540,6 +1548,75 @@ def argmin(self):
...
+            sdf = sdf.withColumn(replace_col, replace_udf(dist_sequence_col_name))
+        elif isinstance(value, (Index, Series)):
+            value = value.to_numpy().tolist()
+            replace_udf = udf(lambda x: value[x], _infer_type(value[0]))
```
ueshin, line 1590: ditto.

databricks/koalas/indexes.py
```diff
@@ -1540,6 +1548,75 @@ def argmin(self):
...
+            sdf = sdf.withColumn(replace_col, replace_udf(dist_sequence_col_name))
+        else:
+            sdf = sdf.withColumn(replace_col, F.lit(value))
+
+        if isinstance(mask, (Index, Series)):
+            mask = mask.to_numpy().tolist()
+        elif not isinstance(mask, list) and not isinstance(mask, tuple):
+            raise TypeError("Mask data doesn't support type " "{0}".format(type(mask).__name__))
+
+        masking_udf = udf(lambda x: mask[x], BooleanType())
```
ueshin, line 1600: ditto.

databricks/koalas/indexes.py
```diff
@@ -1540,6 +1548,75 @@ def argmin(self):
...
+            sdf = sdf.withColumn(replace_col, replace_udf(dist_sequence_col_name))
+        elif isinstance(value, (Index, Series)):
+            value = value.to_numpy().tolist()
+            replace_udf = udf(lambda x: value[x], _infer_type(value[0]))
+            sdf = sdf.withColumn(replace_col, replace_udf(dist_sequence_col_name))
+        else:
+            sdf = sdf.withColumn(replace_col, F.lit(value))
+
+        if isinstance(mask, (Index, Series)):
+            mask = mask.to_numpy().tolist()
```
ueshin, line 1596: I don't think we should do this.

databricks/koalas/indexes.py
```diff
@@ -1540,6 +1548,75 @@ def argmin(self):
...
+        # |__distributed_sequence_column__|__index_level_0__|__replace_column__|__masking_column__|
+        # +-------------------------------+-----------------+------------------+------------------+
+        # |                              0|                a|               100|              true|
+        # |                              3|                d|               400|             false|
+        # |                              1|                b|               200|              true|
+        # |                              2|                c|               300|             false|
+        # |                              4|                e|               500|             false|
+        # +-------------------------------+-----------------+------------------+------------------+
+
+        cond = F.when(sdf[masking_col], sdf[replace_col]).otherwise(sdf[scol_name])
```
ueshin, line 1613: Could you use `scol_for(sdf, scol_name)`?

databricks/koalas/tests/test_indexes.py
```diff
@@ -278,6 +278,30 @@ def test_dropna(self):
...
+            pidx.putmask(pidx < "c", ("g", "h", "i", "j", "k")).sort_values(),
+        )
+        self.assert_eq(
+            kidx.putmask(kidx < "c", ks.Index(["g", "h", "i", "j", "k"])).sort_values(),
+            pidx.putmask(pidx < "c", pd.Index(["g", "h", "i", "j", "k"])).sort_values(),
+        )
+        self.assert_eq(
+            kidx.putmask(kidx < "c", ks.Series(["g", "h", "i", "j", "k"])).sort_values(),
+            pidx.putmask(pidx < "c", pd.Series(["g", "h", "i", "j", "k"])).sort_values(),
+        )
```
ueshin, line 303: What if the length of `value` is not same as the index length? Could you add the tests?

beobest2: @ueshin  Thanks for the comment! I will address it to work as you comments.

beobest2: @ueshin 
If the length of the `mask` in the pandas is different, `ValueError` is raised.

```python
>>> pidx
Index(['a', 'b', 'c', 'd', 'e'], dtype='object')
>>> pidx.putmask([True, False], pd.Series(["g", "h", "i", "j", "k"])).sort_values()
  File "/Users/hwpark/Desktop/git_koalas/venv/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 4041, in putmask
    raise err
  File "/Users/hwpark/Desktop/git_koalas/venv/lib/python3.7/site-packages/pandas/core/indexes/base.py", line 4037, in putmask
    np.putmask(values, mask, self._convert_for_op(value))
  File "<__array_function__ internals>", line 6, in putmask
ValueError: putmask: mask and data must be the same size
```

So I fixed Koalas to get an same error as well.

```python
>>> kidx.putmask([True, False], ks.Series(["g", "h", "i", "j", "k"])).sort_values()
  File "/Users/hwpark/Desktop/git_koalas/koalas/databricks/koalas/indexes.py", line 1612, in putmask
    raise ValueError("mask and data must be the same size")
ValueError: mask and data must be the same size
```

---

If the `value` ​​have different length in pandas, it works like this:

```python
>>> pidx
Index(['a', 'b', 'c', 'd', 'e'], dtype='object')
>>> pidx.putmask(pidx > 'c', pd.Series(["g", "h"])).sort_values()
Index(['a', 'b', 'c', 'g', 'h'], dtype='object')
>>> pidx.putmask(pidx < 'c', pd.Series(["g", "h"])).sort_values()
Index(['c', 'd', 'e', 'g', 'h'], dtype='object')
>>> pidx.putmask(pidx < 'c', pd.Series(["g"])).sort_values()
Index(['c', 'd', 'e', 'g', 'g'], dtype='object')
>>> pidx.putmask([True, False, True, False, True], pd.Series(["g", "h"])).sort_values()
Index(['b', 'd', 'g', 'g', 'g'], dtype='object')
```

I thought the behavior of Pandas was ambiguous, so I left the comments at line 1593 for now.

```
# TODO: We can't support different size of value for now.
```

databricks/koalas/indexes.py
beobest2: It is possible! I modified it to use `pandas_udf`

databricks/koalas/tests/test_indexes.py
beobest2: @itholic I solved the type infernece problem using `as_spark_type(pandas_value.dtype.type)`

databricks/koalas/indexes.py
```diff
@@ -1546,6 +1554,97 @@ def argmin(self):
...
+            if isinstance(value, (list, tuple)):
+                pandas_value = pd.Series(value)
+            elif isinstance(value, (Index, Series)):
+                pandas_value = value.to_pandas()
+
+            if self.size != pandas_value.size:
+                # TODO: We can't support different size of value for now.
+                raise ValueError("value and data must be the same size")
+
+            replace_return_type = as_spark_type(pandas_value.dtype.type)
```
itholic, line 1601: Let's try to use `infer_pd_series_spark_type` rather than `as_spark_type`.

You can like as below.

```python
replace_return_type = infer_pd_series_spark_type(pandas_value)
```

beobest2: thank you! :) I fixed it by using `infer_pd_series_spark_type `

itholic:
@beobest2 could you rebase this when available ?

beobest2:
@itholic sure :)

databricks/koalas/indexes.py
```diff
@@ -1578,6 +1586,97 @@ def argmin(self):
...
+
+        if isinstance(value, (list, tuple, Index, Series)):
+            if isinstance(value, (list, tuple)):
+                pandas_value = pd.Series(value)
+            elif isinstance(value, (Index, Series)):
+                pandas_value = value.to_pandas()
+
+            if self.size != pandas_value.size:
+                # TODO: We can't support different size of value for now.
+                raise ValueError("value and data must be the same size")
```
itholic, line 1631: If we can support for only same size, I think we shouldn't support this API for non-scalar objects for now.

Since we're using `pd.Series(value)` and `value.to_pandas()` above, It looks quite dangerous.

itholic: I think we better support this API only for the `ks.Index` so that we can avoid the collect all the data into single machine.

Maybe you can refer to `Series.where` which has similar concept. (https://koalas.readthedocs.io/en/latest/_modules/databricks/koalas/series.html#Series.where)

Would you tell me what do you think about this way when you available, @ueshin @HyukjinKwon  ?

xinrong-databricks:
Hi @beobest2, since Koalas has been ported to Spark as pandas API on Spark, would you like to migrate this PR to the Spark repository? Here is the ticket https://issues.apache.org/jira/browse/SPARK-36403. Otherwise, I may do that for you next week.

beobest2:
> Hi @beobest2, since Koalas has been ported to Spark as pandas API on Spark, would you like to migrate this PR to the Spark repository? Here is the ticket https://issues.apache.org/jira/browse/SPARK-36403. Otherwise, I may do that for you next week.

Hi @xinrong-databricks I would like to migrate this PR to the Spark repository. I will try to finish it by next week.

xinrong-databricks:
Please take your time :) Thank you!

beobest2:
@xinrong-databricks I created a PR at https://github.com/apache/spark/pull/33744 . Please take a look :)

xinrong-databricks:
Certainly, let's discuss in the new PR then! Thanks for the porting.

PR #1563, opened 2020-06-03
Support DataFrame.spark.explain(extended: str) case.

ueshin:
Since Spark 3.0 will support `DataFrame.explain(extended: str)` case (apache/spark#28711), we can follow it.

```py
>>> df.spark.explain("extended")  # doctest: +ELLIPSIS
== Parsed Logical Plan ==
...
== Analyzed Logical Plan ==
...
== Optimized Logical Plan ==
...
== Physical Plan ==
 ...
```

### DIFF
--- a/databricks/koalas/spark/accessors.py
+++ b/databricks/koalas/spark/accessors.py
@@ -729,6 +729,16 @@ class SparkFrameMethods(object):
         == Physical Plan ==
         ...
 
+        >>> df.spark.explain("extended")  # doctest: +ELLIPSIS
+        == Parsed Logical Plan ==
+        ...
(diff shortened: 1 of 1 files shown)


databricks/koalas/spark/accessors.py
```diff
@@ -729,6 +729,16 @@ def explain(self, extended: Optional[bool] = None, mode: Optional[str] = None):
         == Physical Plan ==
         ...
 
+        >>> df.spark.explain("extended")  # doctest: +SKIP
```
ueshin, line 732: This is not supported in Spark 3.0.0-rc2. I'd skip this for now.

PR #1564, opened 2020-06-04
Cache an internal pandas object to avoid run twice in Jupyter.

ueshin:
IPython calls Spark jobs twice due to Notebook calls both `__repr__` and `_repr_html_`.
See https://github.com/ipython/ipython/issues/9771

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10220,12 +10220,22 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         return self._internal.to_pandas_frame
 
+    def _get_or_create_repr_pandas_cache(self, n):
+        if (
+            not hasattr(self, "_repr_pandas_cache")
+            or (id(self._internal), n) not in self._repr_pandas_cache
+        ):
+            self._repr_pandas_cache = {
+                (id(self._internal), n): self.head(n + 1)._to_internal_pandas()
+            }
+        return self._repr_pandas_cache[(id(self._internal), n)]
+
     def __repr__(self):
         max_display_count = get_option("display.max_rows")
         if max_display_count is None:
             return self._to_internal_pandas().to_string()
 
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -10220,12 +10220,22 @@ def _to_internal_pandas(self):
...
 
+    def _get_or_create_repr_pandas_cache(self, n):
+        if (
+            not hasattr(self, "_repr_pandas_cache")
+            or (id(self._internal), n) not in self._repr_pandas_cache
+        ):
+            self._repr_pandas_cache = {
+                (id(self._internal), n): self.head(n + 1)._to_internal_pandas()
+            }
+        return self._repr_pandas_cache[(id(self._internal), n)]
```
HyukjinKwon, line 10231: Hm ... @ueshin should we maybe just keep one cache just to prevent the possibility of the growing cache?

ueshin: It actually keeps one cache by the key `(id(self._internal), n)`.

itholic:
Seems good to me

databricks/koalas/frame.py
HyukjinKwon: I meant, it keeps the cache if we change `display.max_rows` or operate that change `_internal`.
For example, if somebody opens a shell, and does a lot of `inplace` operations with printing out in the notebook, the cache will keep growing. I was wondering if it's better to don't keep previous caches.

HyukjinKwon: I feel like at least we should have a logic to drop the old caches conditionally if we want to keep old ones.

ueshin: It does hold only one cache.
If `inplace` operation is done, `id(self._internal)` will change, and if `display.max_rows` is changed, `n` will change.

PR #1566, opened 2020-06-04
Add SPARK_DEFAULT_SERIES_NAME to internal.py

itholic:
When we use `"0"` as a default name for Series (In fact, almost all cases), I think It seems better to specify using specific constant.

added a constant `SPARK_DEFAULT_SERIES_NAME ` and addressed related files.
(I'm not pretty sure `internal.py` is the proper file to place this constant though)


databricks/koalas/internal.py
```diff
@@ -63,6 +63,8 @@
 
 HIDDEN_COLUMNS = {NATURAL_ORDER_COLUMN_NAME}
 
+SERIES_DEFAULT_NAME = "0"
```
ueshin, line 66: How about `SPARK_DEFAULT_SERIES_NAME`?

itholic: @ueshin 

Hmm, It looks better in terms of maintaining consistency with `SPARK_DEFAULT_INDEX_NAME`.

Thanks!

ueshin:
I'm working on fixing the build failure for `Conda (Python, Spark, pandas, PyArrow) (3.7, 2.4.6, 1.0.4, 0.15.1)` at #1576.

ueshin:
@itholic Could you push an empty commit to rerun tests?

itholic:
@ueshin Yup! I just reran them, thanks :)

ueshin:
Thanks! merging.

PR #1567, opened 2020-06-05
Make filters apply first.

ueshin:
It's fine to apply filter after select, but move filter before select to make sure it works.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -910,7 +910,7 @@ class InternalFrame(object):
             spark_type = self.spark_frame.select(pred).schema[0].dataType
             assert isinstance(spark_type, BooleanType), spark_type
 
-        sdf = self.spark_frame.select(self.spark_columns).filter(pred)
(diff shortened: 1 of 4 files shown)


databricks/koalas/internal.py
```diff
@@ -910,7 +910,7 @@ def with_filter(self, pred: Union[spark.Column, "Series"]):
             spark_type = self.spark_frame.select(pred).schema[0].dataType
             assert isinstance(spark_type, BooleanType), spark_type
 
-        sdf = self.spark_frame.select(self.spark_columns).filter(pred)
+        sdf = self.spark_frame.filter(pred).select(self.spark_columns)
```
HyukjinKwon, line 913: Hmm..  I know the filter still works because of the predicate pushdown but I thought it looks logically clear that we select the columns we're interested in first (except hidden columns), and the filter out.

Actually, `df.filter("b > 1").select("a")` and `df.select("a").filter("b > 1")` might have to be distinguished in the Spark itself ...

PR #1569, opened 2020-06-08
Set upper bound for Sphinx.

ueshin:
Setting the upper bound for Sphinx for now to unblock PRs since Sphinx 3.1.0 which was published 5 hours ago causes `lint-python` failure.


ueshin:
I'd merge this now. Fixing the failures is tracked at #1570.
cc @HyukjinKwon

PR #1573, opened 2020-06-10
Support Series as repeats in Series.repeat.

ueshin:
It's good to support `Series` as `repeats` in `Series.repeat()`.

```py
>>> kdf = ks.DataFrame({"a": ["a", "b", "c"], "rep": [1, 2, 3]})
>>> kdf
   a  rep
0  a    1
1  b    2
2  c    3
>>> kdf.a.repeat(kdf.rep)
0    a
1    b
1    b
2    c
2    c
2    c
Name: a, dtype: object
```

As this PR is using a function added since Spark 2.4, it doesn't support Spark 2.3.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -22,6 +22,7 @@ import inspect
 import warnings
 from collections import OrderedDict
 from collections.abc import Iterable
+from distutils.version import LooseVersion
 from functools import partial, wraps, reduce
 from typing import Any, Generic, List, Optional, Tuple, TypeVar, Union
 
@@ -31,6 +32,7 @@ from pandas.core.accessor import CachedAccessor
 from pandas.io.formats.printing import pprint_thing
 from pandas.api.types import is_list_like
 
+import pyspark
 from pyspark import sql as spark
 from pyspark.sql import functions as F, Column
 from pyspark.sql.functions import pandas_udf, PandasUDFType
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -4701,16 +4703,43 @@ def repeat(self, repeats: int) -> "Series":
...
+        if not isinstance(repeats, (int, Series)):
+            raise ValueError(
+                "`repeats` argument must be integer or Series, but got {}".format(type(repeats))
+            )
 
-        kdf = self.to_frame()
-        if repeats == 0:
-            return first_series(DataFrame(kdf._internal.with_filter(F.lit(False))))
+        if isinstance(repeats, Series):
+            if LooseVersion(pyspark.__version__) < LooseVersion("2.4"):
```
HyukjinKwon, line 4712: @ueshin, I think we officially support Spark 2.4+ so maybe we wouldn't have to worry about this case.

HyukjinKwon: It's okay because it's just a better error message though :-)

ueshin: Yeah, maybe we can remove it after we make Koalas throw an error instead of warning (https://github.com/databricks/koalas/blob/master/databricks/koalas/__init__.py#L35-L40) when using Spark<2.4.

PR #1576, opened 2020-06-11
Rewrite requirements-dev.txt before installing in conda CI.

ueshin:
Even if packages are listed as pinned_packages, conda still tries to upgrade/downgrade them if the packages are in `requirements-dev.txt` without the version or with a version range.

```
pinned spec pyarrow==0.15.1 conflicts with explicit specs.  Overriding pinned spec.
```

The installed version satisfies the version range, though.


ueshin:
I'd merge this now to unblock PRs since the tests started without any installation issues.

PR #1577, opened 2020-06-12
Support pd.DataFrame and pd.Series in return type hints

HyukjinKwon:
This PR proposes to support the type hints with pandas Series and pandas DataFrame directly.

```python
import pandas as pd
def transform(pdf) -> pd.DataFrame[int]:
    # pdf is actually a pandas DataFrame.
    return pdf + 1

ks.range(10).apply_batch(transform)
```

```
     c0
0     1
1     2
2     3
3     4
4     5
5     6
6     7
7     8
8     9
9    10
```

Note that:

- The type hints with Pandas only works with Python 3.7+.

- The previous way with Koalas DataFrame type hint will still work but will be deprecated once we deprecate Python 3.5 and 3.6.

    ```python
    def transform(pdf) -> ks.DataFrame[int]:
        return pdf + 1

    ks.range(10).apply_batch(transform)
    ```

### DIFF
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 import os
+import sys
 from distutils.version import LooseVersion
 
 from databricks.koalas.version import __version__
@@ -102,7 +103,7 @@ __all__ = [
 ]
 
 
-def _auto_patch():
+def _auto_patch_spark():
     import os
     import logging
 
@@ -137,7 +138,27 @@ def _auto_patch():
         df.DataFrame.to_koalas = DataFrame.to_koalas
 
 
-_auto_patch()
+def _auto_patch_pandas():
+    import pandas as pd
+
+    # In order to use it in test cases.
+    global _frame_has_class_getitem
+    global _series_has_class_getitem
+
+    _frame_has_class_getitem = hasattr(pd.DataFrame, "__class_getitem__")
+    _series_has_class_getitem = hasattr(pd.Series, "__class_getitem__")
+
+    if sys.version_info >= (3, 7):
+        # Just in case pandas implements '__class_getitem__' later.
+        if not _frame_has_class_getitem:
+            pd.DataFrame.__class_getitem__ = lambda params: DataFrame.__class_getitem__(params)
+
+        if not _series_has_class_getitem:
+            pd.Series.__class_getitem__ = lambda params: Series.__class_getitem__(params)
+
+
+_auto_patch_spark()
+_auto_patch_pandas()
 
 # Import after the usage logger is attached.
 from databricks.koalas.config import *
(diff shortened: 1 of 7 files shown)


databricks/koalas/typedef/string_typehints.py
```diff
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 import numpy as np
+import pandas
```
ueshin, line 17: nit: seems not needed?

databricks/koalas/__init__.py
```diff
@@ -137,7 +138,16 @@ def _auto_patch():
         df.DataFrame.to_koalas = DataFrame.to_koalas
 
 
-_auto_patch()
+def _auto_patch_pandas():
+    if sys.version_info >= (3, 7):
+        import pandas as pd
+
+        pd.DataFrame.__class_getitem__ = lambda params: DataFrame.__class_getitem__(params)
+        pd.Series.__class_getitem__ = lambda params: Series.__class_getitem__(params)
```
ueshin, line 146: Shall we check if the pandas DataFrame/Series have this method to early detect that pandas supports it in the future.

databricks/koalas/frame.py
```diff
@@ -26,7 +26,7 @@
 from functools import partial, reduce
 import sys
 from itertools import zip_longest
-from typing import Any, Optional, List, Tuple, Union, Generic, TypeVar, Iterable, Dict, Callable
+from typing import Any, Optional, List, Tuple, Union, Iterable, Dict, Callable, Generic, TypeVar
```
itholic, line 29: I just curious why those order have been changed? (maybe reformatter changed it automatically ?)

databricks/koalas/frame.py
```diff
@@ -2146,8 +2146,9 @@ def apply_batch(self, func, args=(), **kwds):
             # If schema is inferred, we can restore indexes too.
             internal = kdf._internal.with_new_sdf(sdf)
         else:
-            return_schema = infer_return_type(original_func).tpe
-            is_return_dataframe = getattr(return_sig, "__origin__", None) == ks.DataFrame
+            return_tpe = infer_return_type(original_func)
```
itholic, line 2149: How about `return_type` rather `return_tpe` since the method name that return this variable is `infer_return_type`?

databricks/koalas/frame.py
```diff
@@ -2146,8 +2146,9 @@ def apply_batch(self, func, args=(), **kwds):
...
+            return_schema = return_tpe.tpe
```
itholic, line 2150: Ah,,, maybe because of here

databricks/koalas/typedef/string_typehints.py
HyukjinKwon: Here the code path is very hacky.. so it is in order to support a case when the return type hint is `pandas.DataFrame` as an example..

ueshin:
Thanks! merging.

PR #1578, opened 2020-06-12
Support pandas typehints names

### LINKED ISSUE #1143: Leverage NamedTuple and TypedDict for type hinting
https://docs.python.org/3/library/typing.html#typing.NamedTuple
https://docs.python.org/3/library/typing.html#typing.TypedDict

We can use both to avoid schema inference in APIs such as `DataFrame.transform`. This looks added from Python 3.6. Koalas supports Python 3.5 yet. So we should wait for now.

### LINKED ISSUE #1307: DataFrame.apply(func, axis=1) returns Series with unmatched index

HyukjinKwon:
This PR proposes a way to specify names in the type hint in DataFrame. See the examples below:

```python
import pandas as pd
pdf = pd.DataFrame({'a': [1, 2, 3], 'b': [3, 4, 5]})
kdf = ks.from_pandas(pdf)

def transform(pdf) -> pd.DataFrame[zip(pdf.columns, pdf.dtypes)]:
    return pdf + 1

kdf.apply_batch(transform)
```

```
   a  b
0  2  4
1  3  5
2  4  6
```

```python
import pandas as pd

def transform(pdf) -> pd.DataFrame["id": int, "A": int]:
    pdf['A'] = pdf.id + 1
    return pdf

ks.range(10).apply_batch(transform)
```

```
   id   A
0   0   1
1   1   2
2   2   3
3   3   4
4   4   5
5   5   6
6   6   7
7   7   8
8   8   9
9   9  10
```

I think we should take this as an experimental feature which could change between minor releases.

This PR also implements an easy way to specify the types without names too:

```python
import pandas as pd
pdf = pd.DataFrame({'a': [1, 2, 3], 'b': [3, 4, 5]})
kdf = ks.from_pandas(pdf)

def transform(pdf) -> pd.DataFrame[pdf.dtypes]:
    return pdf + 1

kdf.apply_batch(transform)
```

```
   c0  c1
0   2   4
1   3   5
2   4   6
```

Closes #1307
Closes #1143

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2072,7 +2072,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             `c0, c1, c2 ... cn`. These names are positionally mapped to the returned
             DataFrame in ``func``.
 
-            To specify the names, you can assign them in a pandas friendly style as below:
+            To specify the column names, you can assign them in a pandas friendly style as below:
 
             >>> def plus_one(x) -> ks.DataFrame["a": float, "b": float]:
             ...     return x + 1
@@ -2291,7 +2291,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             `c0, c1, c2 ... cn`. These names are positionally mapped to the returned
             DataFrame in ``func``.
 
-            To specify the names, you can assign them in a pandas friendly style as below:
+            To specify the column names, you can assign them in a pandas friendly style as below:
 
             >>> def plus_one(x) -> ks.DataFrame["a": float, "b": float]:
             ...     return x + 1
@@ -2711,7 +2711,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             `c0, c1, c2 ... cn`. These names are positionally mapped to the returned
             DataFrame in ``func``.
 
-            To specify the names, you can assign them in a pandas friendly style as below:
+            To specify the column names, you can assign them in a pandas friendly style as below:
 
             >>> def plus_one(x) -> ks.DataFrame['a': float, 'b': float]:
             ...     return x + 1
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -292,6 +292,43 @@ rectangle    16.0  2.348543e+108
 T = TypeVar("T")
 
 
+def _create_tuple_for_frame_type(params):
+    from databricks.koalas.typedef import NameTypeHolder
+
+    if isinstance(params, zip):
+        params = [slice(name, tpe) for name, tpe in params]
+
+    if isinstance(params, slice):
+        params = (params,)
+
+    if (
+        hasattr(params, "__len__")
+        and isinstance(params, Iterable)
+        and all(isinstance(param, slice) for param in params)
+    ):
+        for param in params:
+            if isinstance(param.start, str) and param.step is not None:
+                raise TypeError(
+                    "Type hints should be specified as "
+                    "DataFrame['name': type]; however, got %s" % param
+                )
+
+        name_classes = []
+        for param in params:
+            new_class = type("NameType", (NameTypeHolder,), {})
+            new_class.name = param.start
+            # When the given argument is a numpy's dtype instance.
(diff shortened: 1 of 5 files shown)


HyukjinKwon:
This is actually conceptually a bit hacky - the current Python hints don't support this way..

HyukjinKwon:
@ueshin, I will make a separate PR for documentations.

databricks/koalas/frame.py
```diff
@@ -292,6 +292,43 @@
 T = TypeVar("T")
 
 
+def _create_tuple_for_frame_type(params):
```
itholic, line 295: Can we have a short description as a docstring about this function ?

itholic: Hmm... maybe the function name itself is enough.

databricks/koalas/frame.py
```diff
@@ -2033,7 +2070,16 @@ def apply_batch(self, func, args=(), **kwds):
 
             If the return type is specified, the output column names become
             `c0, c1, c2 ... cn`. These names are positionally mapped to the returned
-            DataFrame in ``func``. See examples below.
+            DataFrame in ``func``.
+
+            To specify the names, you can assign them in a pandas friendly style as below:
```
itholic, line 2075: nit: How about `To specify the names` -> `To specify the column names` ?

databricks/koalas/frame.py
```diff
@@ -2033,7 +2070,16 @@ def apply_batch(self, func, args=(), **kwds):
...
+
+            >>> def plus_one(x) -> ks.DataFrame["a": float, "b": float]:
+            ...     return x + 1
+
+            >>> pdf = pd.DataFrame({'a': [1, 2, 3], 'b': [3, 4, 5]})
```
itholic, line 2080: nit: `'a', 'b'` -> `"a", "b"`

databricks/koalas/frame.py
```diff
@@ -2074,6 +2120,12 @@ def apply_batch(self, func, args=(), **kwds):
            c0  c1
         0   1   2
 
+        >>> def query_func(pdf) -> ks.DataFrame["A": int, "B": int]:
+        ...     return pdf.query('A == 1')
```
itholic, line 2124: nit: ditto

databricks/koalas/typedef/typehints.py
```diff
@@ -48,12 +48,17 @@ def __repr__(self):
...
 class DataFrameType(object):
-    def __init__(self, tpe):
-        # Seems we cannot specify field names. I currently gave some default names
-        # `c0, c1, ... cn`.
-        self.tpe = types.StructType(
-            [types.StructField("c%s" % i, tpe[i]) for i in range(len(tpe))]
-        )  # type: types.StructType
+    def __init__(self, tpe, names=None):
+        if names is None:
+            # Seems we cannot specify field names. I currently gave some default names
```
itholic, line 53: maybe now we can specify the field names ?

itholic:
Seems fine to me except several nit reviews.

HyukjinKwon:
I will merge this to unblock the release

PR #1587, opened 2020-06-16
Add Python 3.8 to python_requires.

ueshin:


setup.py
```diff
@@ -46,7 +46,7 @@
         'spark': ['pyspark>=2.4.0'],
         'mlflow': ['mlflow>=1.0'],
     },
-    python_requires='>=3.5,<3.8',
```
ueshin, line 49: Do we still need the upper bound `<3.9`?

setup.py
```diff
@@ -46,7 +46,7 @@
...
+    python_requires='>=3.5',
```
ueshin, line 49: Do we still need the upper bound `<3.9`?

ueshin: Let me keep the upper bound, just in case.

PR #1590, opened 2020-06-17
Fix test_indexes to suppress warnings.

ueshin:
We should use `assertRaisesRegex` instead of `assertRaisesRegexp`.

```
DeprecationWarning: Please use assertRaisesRegex instead.
```


ueshin:
This is a simple fix. I'd merge this now. cc @HyukjinKwon

HyukjinKwon:
Nice!

PR #1591, opened 2020-06-17
Fix index.__repr__ to use cached pandas Index and reuse between Index and MultiIndex.

ueshin:
This is a follow-up of #1564 which was missing the fix for `Index`.


itholic:
Cool

PR #1597, opened 2020-06-18
Fix to_csv to use the column names directly from the Spark DataFrame internally

HyukjinKwon:
Currently, if the columns contain some expressions depending on hidden columns, `to_csv` fails:
Screen Shot 2020-06-18 at 1 10 38 PM

This is because it uses the columns in the previous Spark DataFrame before the projection which does not have the natural order column:

Screen Shot 2020-06-18 at 1 00 01 PM

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -628,14 +628,14 @@ class Frame(object):
         2012-02-29 12:00:00,US,2
         2012-03-31 12:00:00,JP,3
 
-        >>> df.to_csv(path=r'%s/to_csv/foo.csv' % path, num_files=1)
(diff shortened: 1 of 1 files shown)


HyukjinKwon:
Seems it's the flaky test:

```
_______________ [doctest] databricks.koalas.frame.DataFrame.iloc _______________
399     >>> df
400           a   b     c     d
401     0     1   2     3     4
402     1   100  50   300   400
403     2  1000  50  3000  4000
404 
405     Setting value for an entire row
406 
407     >>> df.iloc[0] = 10
408     >>> df
    @@ -1,4 +1,4 @@
           a   b     c     d
    -0    10  10    10    10
    +0     1   2     3     4
     1   100  50   300   400
     2  1000  50  3000  4000

```

cc @ueshin FYI

PR #1604, opened 2020-06-22
Avoid recommending Conda

HyukjinKwon:
This PR proposes to avoid recommending one specific package manager. Each has its own benefits and disadvantages.


HyukjinKwon:
Let me merge this - it's trivial.

PR #1610, opened 2020-06-23
Clean up groupby to use the anchor.

ueshin:
Now that Series has a different anchor from the one it is originally derived from if it is modified, we don't need to check Spark column's equality in groupby.

Also fixing a bug when it needs `compute.ops_on_diff_frames` but the first group key is from the same DataFrame but modified.

E.g.,

```py
>>> kdf1 = ks.DataFrame({"C": [0.362, 0.227, 1.267, -0.562], "B": [1, 2, 3, 4]})
>>> kdf2 = ks.DataFrame({"A": [1, 1, 2, 2]})
>>> kdf1.groupby([kdf1.C + 1, kdf2.A]).sum()
...
AttributeError: 'NoneType' object has no attribute '_internal'
```


ueshin:
Let me merge this now. Please feel free to comment on this.

HyukjinKwon:
Sorry for late review @ueshin. LGTM!

PR #1613, opened 2020-06-26
Add koalas accessor and a method to attach a unique id column

### LINKED ISSUE #1612: Provide an utility to add an increasing id column
In pandas, adding a new column with an increasing sequential id is trivial:
```
df.insert(0, "id", range(len(df)))
```
so we don't need any specific function.

Koalas itself does create an increasing id when it generates a default index (see https://github.com/databricks/koalas/blob/master/databricks/koalas/internal.py#L514) but  `attach_default_index`, `attach_distributed_column` and `attach_distributed_sequence_column` methods are internal methods and they are not supposed to be directly used (see the comment at https://github.com/databricks/koalas/blob/master/databricks/koalas/internal.py#L74).

Moreover, those internal methods work on Spark dataframes (not Koalas dataframes), so to use them we should call `to_spark()` (paying attention to indexes), add the increasing id column and then convert the Spark dataframe back to a Koalas dataframe.

Please expose utilities to add increasing ids (sequential or not) to Koalas dataframes, as we can't do it the pandas way.

ueshin:
Add `koalas` accessor and a few util methods:

- `DataFrame.koalas.attach_id_column`

Resolves #1612


databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+    """ Utility features for DataFrame. Usually, The features here are specific to Koalas. """
+
+    def __init__(self, frame: "DataFrame"):
+        self._kdf = frame
+
+    def attach_sequence_column(self, column: Union[str, Tuple[str, ...]]) -> "DataFrame":
+        """
+        Attach a column of a sequence that increases one by one.
+
+        .. note:: this uses Spark's Window without specifying partition specification.
```
itholic, line 39: nit: Maybe `this uses Spark's Window without specifying partition.` or `this uses Spark's Window without  partition apecification.` is enough?
`specifying` and `specification` seems duplicated ?

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+           x  distributed_sequence
+        0  a                     0
+        1  b                     1
+        2  c                     2
+        """
+        return self._attach_default_index_like_column(
+            column, InternalFrame.attach_distributed_sequence_column
+        )
+
+    def attach_distributed_column(self, column: Union[str, Tuple[str, ...]]) -> "DataFrame":
```
itholic, line 95: What do you think about `column_name` rather `column` to make users understand this parameter more clearly?

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        Attach a column of a monotonically increasing sequence simply by using PySpark’s
+        monotonically_increasing_id function in a fully distributed manner.
+
+        Parameters
+        ----------
+        column : string or list of string
```
itholic, line 102: How about `string or tuple containing list of string` because It seems that this function doesn't allow `list`, so might be confused for someone ? (or make this method to allow `list` ?)

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+            The column name to be attached.
+
+        Returns
+        -------
+        DataFrame
+            The DataFrame attached the column.
+
+        Examples
+        --------
```
itholic, line 111: Can we also have an example for MultiIndex columns ?

```python
>>> df = ks.DataFrame({("x", "y"): ['a', 'b', 'c']})
>>> df.utils.attach_distributed_sequence_column(("distributed-x", "distributed-y")).sort_index()
   x distributed-x
   y distributed-y
0  a           ...
1  b           ...
2  c           ...
```

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+        >>> df = ks.DataFrame({"x": ['a', 'b', 'c']})
+        >>> df.utils.attach_distributed_sequence_column("distributed_sequence").sort_index()
+           x  distributed_sequence
+        0  a                     0
+        1  b                     1
+        2  c                     2
```
itholic, line 89: Could we also have an example for MultiIndex columns ?

```python
>>> df = ks.DataFrame({("x", "y"): ['a', 'b', 'c']})
>>> df.utils.attach_sequence_column(("sequence-x", "sequence-y")).sort_index()
   x sequence-x
   y sequence-y
0  a          0
1  b          1
2  c          2
```

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({"x": ['a', 'b', 'c']})
+        >>> df.utils.attach_distributed_column("distributed")
+        ... # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
+           x  distributed
+        0  a          ...
+        1  b          ...
+        2  c          ...
```
itholic, line 118: ditto ?

```python
>>> df = ks.DataFrame({("x", "y"): ['a', 'b', 'c']})
>>> df.utils.attach_distributed_sequence_column(("distributed-x", "distributed-y")).sort_index()
   x distributed-x
   y distributed-y
0  a           ...
1  b           ...
2  c           ...
```

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,177 @@
...
+            The DataFrame attached the column.
+
+        Examples
+        --------
+        >>> df = ks.DataFrame({"x": ['a', 'b', 'c']})
+        >>> df.utils.attach_sequence_column("sequence")
+           x  sequence
+        0  a         0
+        1  b         1
+        2  c         2
```
itholic, line 61: Could we also have an example for MultiIndex columns ?

```python
>>> df = ks.DataFrame({("x", "y"): ['a', 'b', 'c']})
>>> df.utils.attach_sequence_column(("sequence-x", "sequence-y")).sort_index()
   x sequence-x
   y sequence-y
0  a          0
1  b          1
2  c          2
```

databricks/koalas/utils/accessors.py
itholic, line 89: ditto ?

```python
>>> df = ks.DataFrame({("x", "y"): ['a', 'b', 'c']})
>>> df.utils.attach_distributed_sequence_column(("distributed-sequence-x", "distributed-sequence-y")).sort_index()
   x distributed-sequence-x
   y distributed-sequence-y
0  a                    ...
1  b                    ...
2  c                    ...
```

databricks/koalas/utils/accessors.py
itholic, line 118: ditto ?

```python
>>> df = ks.DataFrame({("x", "y"): ['a', 'b', 'c']})
>>> df.utils.attach_distributed_column(("distributed-x", "distributed-y")).sort_index()
   x distributed-x
   y distributed-y
0  a           ...
1  b           ...
2  c           ...
```

itholic:
LGTM except several nit comments.

candalfigomoro:
Maybe

- attach_sequence_column
- attach_distributed_sequence_column
- attach_distributed_column

could be renamed to something like

- attach_sequence_id_column
- attach_distributed_sequence_id_column
- attach_distributed_id_column

(notice the additional "id" in names)

to make it clear what it is, otherwise it's not really clear what "attach_distributed_column" does until you read the documentation.

databricks/koalas/utils/accessors.py
ueshin: Actually I copied the note from another place. Maybe we should correct all of them in a separate PR if needed.

databricks/koalas/utils/accessors.py
ueshin: Usually pandas doesn't use `column_name` or `column_label`, but `column` or `columns`. I think we should follow the convention.

databricks/koalas/utils/accessors.py
ueshin: ah, that's my bad. We should use `tuple of string`. thanks.

databricks/koalas/utils/accessors.py
itholic: @ueshin Oh I got it, I think It's not a big deal.Let's fix them in a separate PR later.

databricks/koalas/utils/accessors.py
itholic: Ah, I agree If pandas is doing so :D

databricks/koalas/utils/accessors.py
```diff
@@ -0,0 +1,209 @@
...
+
+from databricks.koalas.internal import InternalFrame, SPARK_INDEX_NAME_FORMAT
+from databricks.koalas.utils import name_like_string, scol_for
+
+if TYPE_CHECKING:
+    from databricks.koalas.frame import DataFrame
+
+
+class UtilsFrameMethods(object):
+    """ Utility features for DataFrame. Usually, The features here are specific to Koalas. """
```
HyukjinKwon, line 30: Hm, I am less sure if we should add another accessor for this. For example, `apply_batch`, `transform_batch`, `spark` are Koalas specific APIs too.

How about we just expose one method at `DataFrame`? Maybe we could name it like:
- `append_id(type=..., labels=...)`
- `append_id_series(type=..., labels=...)`
- `attach_id_series(type=..., labels=...)`
- `add_id_series(type=..., labels=...)`
- `generate_ids(type=..., labels=...)`
- ...

ueshin: I'd not prefer to have the method at `DataFrame` directly but to separate the namespace since this doesn't exist in pandas DataFrame. For `spark` accessor, as this is less related to Spark, under the accessor might not suitable.

As for the naming and the signature, I'm fine with one of them.

HyukjinKwon: I agree with the idea of having a separate accessor. Should we maybe name it `koalas` and put other Koalas specific APIs into it?

PR #1620, opened 2020-06-30
Implement Series.cov

lopez-:
This PR proposes Series.cov 

```python
>>> s1 = ks.Series([1, 2, 3, 4])
>>> s2 = ks.Series([5, 6, 7, 8])
>>> s1
0    1
1    2
2    3
3    4
Name: 0, dtype: int64

>>> s2
0    5
1    6
2    7
3    8
Name: 0, dtype: int64

>>> s1.cov(s2)
1.666666...
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4858,6 +4858,54 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
 
         return mad
 
+    def cov(self, other: "Series", min_periods: Optional[int] = None) -> float:
+        """
+        Return the covariance between two series.
+
+        Parameters
+        ----------
+        other : Series
+        min_periods : int
+
+        Examples
+        --------
+        >>> s1 = ks.Series([1, 2, 3, 4])
+        >>> s2 = ks.Series([5, 6, 7, 8])
+        >>> s1
+        0    1
+        1    2
+        2    3
+        3    4
+        Name: 0, dtype: int64
+
+        >>> s2
+        0    5
+        1    6
+        2    7
+        3    8
+        Name: 0, dtype: int64
+
+        >>> s1.cov(s2)
+        1.666666...
+        """
+
+        if not isinstance(other, Series):
+            raise ValueError("'other' must be a Series")
+
+        if len(self.index) != len(other.index):
+            raise ValueError("series are not aligned")
+
+        min_periods = 0 if min_periods is None else min_periods
+        if len(self.index) < min_periods or len(self.index) <= 1:
+            return np.nan
+
+        if same_anchor(self, other):
+            # if the have the same anchor use the more performant Spark native `cov`
+            return self._internal.spark_frame.cov(self.name, other.name)
+        else:
+            # if not on the same anchor calculate covariance manually
+            return (self - self.mean()).dot(other - other.mean()) / (len(self.index) - 1)
+
     def unstack(self, level=-1):
         """
         Unstack, a.k.a. pivot, Series with MultiIndex to produce DataFrame.
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4860,33 +4860,30 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
 
     def cov(self, other: "Series", min_periods: Optional[int] = None) -> float:
         """
-        Return the covariance between two series.
+        Compute covariance with Series, excluding missing values.
 
         Parameters
         ----------
         other : Series
-        min_periods : int
+            Series with which to compute the covariance.
+        min_periods : int, optional
+            Minimum number of observations needed to have a valid result.
+
+        Returns
+        -------
+        float
+            Covariance between Series and other normalized by N-1
+            (unbiased estimator).
 
         Examples
         --------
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -4858,6 +4858,54 @@ def mad(self):
...
         return mad
 
+    def cov(self, other: "Series", min_periods: Optional[int] = None) -> float:
+        """
+        Return the covariance between two series.
+
+        Parameters
+        ----------
+        other : Series
+        min_periods : int
```
itholic, line 4868: Maybe `min_periods` also can be an `optional` ?

databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -948,6 +948,32 @@ def test_series_repeat(self):
...
+
+        kser = ks.Series([90, 91, 85])
+        pser = kser.to_pandas()
+        kser_other = ks.Series([90, 91, 85])
+        pser_other = kser_other.to_pandas()
+
+        k_isnan = np.isnan(kser.cov(kser_other, 4))
+        p_isnan = np.isnan(pser.cov(pser_other, 4))
+        self.assert_eq(k_isnan, p_isnan)
+
```
itholic, line 976: Can we have a test when each Series has a different index?

For example,

```python
        kser = ks.Series([90, 91, 85], index=[1, 2, 3])
        pser = kser.to_pandas()
        kser_other = ks.Series([90, 91, 85], index=[-1, -2, -3])
        pser_other = kser_other.to_pandas()

        k_isnan = np.isnan(kser.cov(kser_other, 4))
        p_isnan = np.isnan(pser.cov(pser_other, 4))
        self.assert_eq(k_isnan, p_isnan)
```

databricks/koalas/series.py
```diff
@@ -4858,6 +4858,54 @@ def mad(self):
...
+        min_periods = 0 if min_periods is None else min_periods
+        if len(self.index) < min_periods or len(self.index) <= 1:
+            return np.nan
+
+        if same_anchor(self, other):
+            # if the have the same anchor use the more performant Spark native `cov`
+            return self._internal.spark_frame.cov(self.name, other.name)
+        else:
+            # if not on the same anchor calculate covariance manually
+            return (self - self.mean()).dot(other - other.mean()) / (len(self.index) - 1)
```
itholic, line 4907: `len(self.index)` is performed three times in this code.

Can we make a proper variable to store this and reuse it?
(ex. `len_index`)

itholic:
Could you add this to the docs also ??

It is placed at `docs/source/reference/series.rst` :)

databricks/koalas/series.py
```diff
@@ -4858,6 +4858,54 @@ def mad(self):
...
+        """
+        Return the covariance between two series.
+
+        Parameters
+        ----------
+        other : Series
+        min_periods : int
+
+        Examples
+        --------
```
itholic, line 4871: Let's use `option_context` to avoid test failure.

You can

```python
from databricks.koalas.config import option_context
with option_context("compute.ops_on_diff_frames", True):
    s1.dov(s2)
```

itholic:
Could you add this to the docs also ??

It is placed at docs/source/reference/series.rst :)

itholic:
Otherwise, looks fine to me.

databricks/koalas/series.py
```diff
@@ -4858,6 +4858,54 @@ def mad(self):
...
+
+        >>> s1.cov(s2)
+        1.666666...
+        """
+
+        if not isinstance(other, Series):
+            raise ValueError("'other' must be a Series")
+
+        if len(self.index) != len(other.index):
+            raise ValueError("series are not aligned")
```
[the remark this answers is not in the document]
itholic: Oops, I missed it. Thanks, @ueshin .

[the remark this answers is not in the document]
lopez-: Mmm this is interesting. Seems like pandas performs an [alignment](https://github.com/pandas-dev/pandas/blob/v1.0.5/pandas/core/series.py#L2287) between the series before computing the covariance. So, this:
```python
>>> pd.Series([1, 2, 3, 4]).cov(pd.Series([5, 6]))
0.5
```
And this:
```python
>>> pd.Series([1, 2]).cov(pd.Series([5, 6]))
0.5
```
are equivalent... I believe this `align` is not supported in Koalas today. If this is a blocker I could open an issue and wait until somebody implements this. Another option I can think of is to go ahead and have a slightly different behavior for this edge case while we wait for the `align` implementation. Do you have any thoughts/preference on how to go about this @itholic @ueshin ?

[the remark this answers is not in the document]
ueshin: @lopez- Could you file the issue for `align`?
Also, is it possible for you to implement it?

itholic:
Any updates here ? Just confirming :)

xinrong-databricks:
Hi @lopez- , since Koalas has been ported to Spark as pandas API on Spark, would you like to migrate this PR to [the Spark repository](https://github.com/apache/spark)? Here is the ticket https://issues.apache.org/jira/browse/SPARK-36401. Otherwise, I may do that for you next week.

PR #1621, opened 2020-06-30
Fix DataFrame.iterrows to return a Koalas Series.

itholic:
I'm not sure why we should return `pandas.Series` for `DataFrame.iterrows()` in the existing implementation.

```python
>>> df = ks.DataFrame([[1, 1.5]], columns=['int', 'float'])
>>> type(next(df.iterrows())[1])
<class 'pandas.core.series.Series'>
```

Fixed it to return Koalas Series.

```python
>>> df = ks.DataFrame([[1, 1.5]], columns=['int', 'float'])
>>> type(next(df.iterrows())[1])
<class 'databricks.koalas.series.Series'>
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -1319,9 +1319,8 @@ class DataFrame(Frame, Generic[T]):
         ------
         index : label or tuple of label
             The index of the row. A tuple for a `MultiIndex`.
-        data : pandas.Series
+        data : Series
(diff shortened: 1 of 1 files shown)


ueshin:
The Series object is supposed to work in Driver, so returning pandas Series should be fine.

itholic:
Thanks, @ueshin ! closing.

PR #1623, opened 2020-07-01
Make Window functions create a new DataFrame.

ueshin:
Make Window functions create a new Spark DataFrame; otherwise, `AnalysisException` will be raised when aggregate function is applied after the window function.

```py
>>> kdf = ks.DataFrame([[1, 2, 3, 4], [5, 6, 7, 8]], columns=["A", "B", "C", "D"])
>>> kdf.cumsum().sum()
...
pyspark.sql.utils.AnalysisException: It is not allowed to use a window function inside an aggregate function. Please use the inner window function in a sub-query.;
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -688,11 +688,13 @@ class DataFrame(Frame, Generic[T]):
         """
         return self._ksers[label]
 
-    def _apply_series_op(self, op):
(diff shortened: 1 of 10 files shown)


HyukjinKwon:
this is nice, sure.

itholic:
Cool!

PR #1631, opened 2020-07-06
__iter__() for Index & MultiIndex

itholic:
This PR proposes `__iter__()` for Index & MultiIndex by using `toLocalIterator` like we did at `DataFrame.iterrows` and `Series.iteritems`.

```python
>>> for value in kidx:
...     value
...
1
2
3
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1953,7 +1953,15 @@ class Index(IndexOpsMixin):
         return repr_string
 
     def __iter__(self):
-        return MissingPandasLikeIndex.__iter__(self)
+        internal_data_column = self._internal.data_spark_column_names[0]
+
+        def extract_value_from_spark_row(row):
+            return row[internal_data_column]
+
+        for value in map(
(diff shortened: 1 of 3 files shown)


HyukjinKwon:
I think these were explicitly unsupported per https://github.com/databricks/koalas/pull/836

itholic:
@HyukjinKwon , Thanks for the comment.

I just read the issue, but maybe I think the topic of that issue is more closer to **"Koalas should returning `ks.Series` rather than `np.array` for pandas APIs that returning `np.array`"** rather than **"Koalas shouldn't support the `__iter__` for `ks.Series`/`ks.Index`"** ?

Because I think it is not that danger to just iterate the items in `ks.Series` or `ks.Index` one by one, unlike returning the whole item set in a `Series` or `Index` at once like `pd.Series.unique()`.

What do you think about ?

ueshin:
If we support it, users can unconsciously do `list(kser)` or `[ do something for row in kser ]` which cause OOM so easily.

itholic:
@ueshin Thanks! I'd close this now :)

PR #1635, opened 2020-07-07
[MINOR] typo in Series.unique()

itholic:
Fixed typo


itholic:
Oh... I understand why the existing doctest ignore the index.

while I'm here I'm going to fix the another typo in `series.py` if exists.

PR #1627, opened 2020-07-03
Refine concat to handle the same anchor DataFrames properly.

ueshin:
Refine `concat` to handle the same anchor DataFrames properly and reduce unnecessary internal joins.

PR #1636, opened 2020-07-08
Add sort parameter to concat.

ueshin:
Adding `sort` parameter to `ks.concat()`.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -1786,8 +1786,7 @@ def concat(objs, axis=0, join="outer", ignore_index=False):
 
         kdfs_not_same_anchor = []
         for kdf in kdfs[1:]:
-            labels = kdf._internal.column_labels
-            duplicated = [label for label in labels if label in column_labels]
+            duplicated = [label for label in kdf._internal.column_labels if label in column_labels]
             if len(duplicated) > 0:
                 pretty_names = [name_like_string(label) for label in duplicated]
                 raise ValueError(
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -1834,6 +1834,9 @@ def concat(objs, axis=0, join="outer", ignore_index=False, sort=False):
         if ignore_index:
             concat_kdf.columns = list(map(str, _range(len(concat_kdf.columns))))
 
+        if sort:
+            concat_kdf = concat_kdf.sort_index()
+
         return concat_kdf
 
     # Series, Series ...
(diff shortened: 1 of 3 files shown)


ueshin:
This is based on #1627. Please review it first.

databricks/koalas/namespace.py
```diff
@@ -1834,9 +1843,11 @@ def resolve_func(kdf, this_column_labels, that_column_labels):
     # DataFrame, Series ... & Series, Series ...
```
HyukjinKwon, line 1843: Maybe we should throw an exception when `axis=1` and `sort` is set:

```python
>>> import pandas as pd
>>> pd.concat([pd.DataFrame({'b': [4, 5, 6]}, index=[4,3,2]), pd.DataFrame({'a': [3, 4, 5]}, index=[4,3,2])], sort=False, axis=1)
   b  a
4  4  3
3  5  4
2  6  5
>>> pd.concat([pd.DataFrame({'b': [4, 5, 6]}, index=[4,3,2]), pd.DataFrame({'a': [3, 4, 5]}, index=[4,3,2])], sort=True, axis=1)
   b  a
2  6  5
3  5  4
4  4  3
```

ueshin: or do `sort_index()`?

ueshin: Anyway pandas' behavior is not deterministic..

databricks/koalas/namespace.py
```diff
@@ -1859,35 +1870,46 @@ def resolve_func(kdf, this_column_labels, that_column_labels):
...
+            if (len(merged_columns) > 0 and len(merged_columns[0]) > 1) or sort:
+                merged_columns = sorted(merged_columns)
 
             kdfs = [kdf[merged_columns] for kdf in objs]
         elif join == "outer":
             # If there are columns unmatched, just sort the column names.
-            merged_columns = sorted(
-                list(set(itertools.chain.from_iterable(column_labelses_of_kdfs)))
-            )
+            merged_columns = column_labels_of_kdfs[0].copy()
```
HyukjinKwon, line 1899: no big deal since it's already guarded above but we could add an `assert`  for the length of `column_labels_of_kdfs` to be non-empty.

HyukjinKwon: We could also add an assert on `merged_columns` too.

databricks/koalas/namespace.py
HyukjinKwon: 😭 .. anyway is fine. We could just `sort_index()` or just throw an exception

ueshin:
Thanks! merging.

PR #1639, opened 2020-07-08
Introduce plotting.backend configuration with Plotly support

### LINKED ISSUE #1626: Plotly on koalas dataframes
I can easily use plotly's interactive charts by below 2 lines of code:

import pandas as pd
pd.options.plotting.backend = "plotly"

However, unable to use Plotly with Koalas Dataframes. Is there a workaround or can there be an enhancement to include this feature?

DumbMachine:
Aims to fix #1626
Each backend returns the `figure` in their own format, allowing for further editing or customization if required.

How to use?:
```python

kdf = ks.DataFrame([[1, 2, 3, 4], [5, 6, 7, 8]], columns=["A", "B", "C", "D"])
kdf.plot(title="Example Figure") # defaults to backend="matplotlib"
```
image

```python
kdf.plot(backend="pandas_bokeh", title="Example Figure")
## same as:
# ks.options.plotting.backend = "pandas_bokeh"
# kdf.plot(title="Example Figure")
```
image

```python
fig = kdf.plot(backend="plotly", title="Example Figure", height=500, width=500)
fig.show()
```
image

```python
# further edits can be made to the figure
fig.update_layout(template="plotly_dark")
fig.show()
```
image


ueshin:
@DumbMachine What version of black are you using? The format seems something wrong.

DumbMachine:
I have the version required by `requirements-dev.txt`
```bash
(base) ➜  koalas git:(plotly-support) ✗ black --version                        
black, version 19.10b0
```

databricks/koalas/plot.py
```diff
@@ -1118,6 +1118,73 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+            if hasattr(module, "plot"):
+                # Validate that the interface is implemented when the option
+                # is set, rather than at plot time.
+                _backends[backend] = module
+                return module
+
+    raise ValueError(
+        f"Could not find plotting backend '{backend}'. Ensure that you've installed ",
+        f"the package providing the '{backend}' entrypoint, or that the package has a ",
+        "top-level `.plot` method.",
```
ueshin, line 1158: The f-strings is not supported in Python 3.5. Could you use `"...".format(...)` instead?

databricks/koalas/plot.py
```diff
@@ -1155,8 +1243,14 @@ def __call__(
         xerr=None,
         label=None,
         secondary_y=False,
-        **kwds
+        **kwds,
     ):
+        plot_backend = _get_plot_backend(kwds.pop("backend", None))
+        # when using another backend, let the backend take the charge
+        if plot_backend.__name__ == "plotly":
+            self.data, kwds = self._get_args_map(plot_backend.__name__, self.data, kwds)
```
ueshin, line 1255: Maybe we should not update `self.data`.

databricks/koalas/plot.py
```diff
@@ -1541,8 +1666,14 @@ def __call__(
         xerr=None,
         secondary_y=False,
         sort_columns=False,
-        **kwds
+        **kwds,
     ):
+        plot_backend = _get_plot_backend(kwds.pop("backend", None))
+        # when using another backend, let the backend take the charge
+        if plot_backend.__name__ == "plotly":
+            self.data, kwds = self._get_args_map(plot_backend.__name__, self.data, kind, kwds)
```
ueshin, line 1678: ditto.

databricks/koalas/plot.py
```diff
@@ -1127,6 +1198,27 @@ class KoalasSeriesPlotMethods(PandasObject):
...
+                ("ylim", "range_y"),
+                ("yerr", "error_y"),
+                ("xerr", "error_x"),
+            ]
+        }
+        for arg_name_mpl, arg_name_ply in args_map[backend_name]:
+            if arg_name_mpl in args_map[backend_name]:
+                kwargs[arg_name_ply] = kwargs.pop(arg_name_mpl)
+
+        return data.to_pandas(), kwargs
```
ueshin, line 1220: Don't we need `data_preprocessor` here?

databricks/koalas/plot.py
```diff
@@ -382,7 +382,7 @@ def rc_defaults(
         showcaps=None,
         showbox=None,
         showfliers=None,
-        **kwargs
+        **kwargs,
```
ueshin, line 385: Hmm, why did we come to need the `,` at the end?

databricks/koalas/config.py
```diff
@@ -220,6 +220,12 @@ def validate(self, v: Any) -> None:
             "'plotting.sample_ratio' should be 1.0 >= value >= 0.0.",
         ),
     ),
+    Option(
+        key="plotting.backend",
+        doc=("Backend to use for plotting. Default is matplotlib."),
```
ueshin, line 225: Could you add a list of currently supported backends?

ueshin: Also `check_func` to avoid setting invalid backend?

databricks/koalas/plot.py
DumbMachine: When I run  `PYTHON_EXECUTABLE=python ./dev/reformat `, that automatically gets added.

DumbMachine:
I have some trouble understanding the error here. Some tests have problems with [trailing `,` ](https://github.com/databricks/koalas/blob/9466324ec3c5bedd8c2ebe8efe6f83229a35a52c/databricks/koalas/plot.py#L1260)
and at the same time, some tests recommend to run `dev/reformat` which has the following effect (adds the trailing `,`):
image
(the image is the result of :`git diff plot.py`, after running `dev/reformat`)

ueshin:
Let me try to fix it.

databricks/koalas/plot.py
```diff
@@ -430,7 +430,7 @@ def _compute_stats(data, colname, whis, precision):
                 ).alias("{}_{}%".format(colname, int(q * 100)))
                 for q in [0.25, 0.50, 0.75]
             ],
-            F.mean(colname).alias("{}_mean".format(colname))
+            F.mean(colname).alias("{}_mean".format(colname)),
```
ueshin, line 433: Could you remove `,' at the end?

databricks/koalas/plot.py
```diff
@@ -913,7 +913,7 @@ def plot_series(
         xerr=xerr,
         label=label,
         secondary_y=secondary_y,
-        **kwds
+        **kwds,
```
ueshin, line 916: ditto.

databricks/koalas/plot.py
```diff
@@ -1082,7 +1082,7 @@ def plot_frame(
         secondary_y=secondary_y,
         layout=layout,
         sort_columns=sort_columns,
-        **kwds
+        **kwds,
```
ueshin, line 1085: ditto.

databricks/koalas/plot.py
```diff
@@ -1155,8 +1257,14 @@ def __call__(
         xerr=None,
         label=None,
         secondary_y=False,
-        **kwds
+        **kwds,
```
ueshin, line 1260: ditto.

databricks/koalas/plot.py
```diff
@@ -1182,7 +1290,7 @@ def __call__(
             xerr=xerr,
             label=label,
             secondary_y=secondary_y,
-            **kwds
+            **kwds,
```
ueshin, line 1293: ditto.

databricks/koalas/plot.py
```diff
@@ -1541,8 +1680,14 @@ def __call__(
         xerr=None,
         secondary_y=False,
         sort_columns=False,
-        **kwds
+        **kwds,
```
ueshin, line 1683: ditto.

databricks/koalas/plot.py
```diff
@@ -1574,7 +1719,7 @@ def __call__(
             xerr=xerr,
             secondary_y=secondary_y,
             sort_columns=sort_columns,
-            **kwds
+            **kwds,
```
ueshin, line 1722: ditto.

ueshin:
Btw, I'm wondering how the other backends, `pandas_bokeh` and `pandas_altair`?
I tried `pandas_bokeh` locally, but it shows the same figure as `matplotlib`. Am I missing something?

HyukjinKwon:
Awesome! @DumbMachine

databricks/koalas/plot.py
```diff
@@ -1118,6 +1118,77 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+import importlib
+
+
+def _find_backend(backend: str):
+    """
+    Find a pandas plotting backend
+    """
+    # function copied from pandas.plotting._core
+
+    import pkg_resources  # Delay import for performance.
```
itholic, line 1131: Could we move this comment to the above of `import pkg_resources` to keep the consistency with other comments if there is no special reason ?

databricks/koalas/plot.py
```diff
@@ -1118,6 +1118,77 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+
+    for entry_point in pkg_resources.iter_entry_points("koalas_plotting_backends"):
+        if entry_point.name == "matplotlib":
+            # matplotlib is an optional dependency. When
+            # missing, this would raise.
+            continue
+        _backends[entry_point.name] = entry_point.load()
+
+    try:
+        return _backends[backend]
```
itholic, line 1141: Can we check if "backend" is in "_backends" and raise an `Exception` manually rather than using `try` ??

databricks/koalas/plot.py
```diff
@@ -1118,6 +1118,77 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
     return plot_obj.result
 
 
+_backends = {}
+import importlib
+
+
+def _find_backend(backend: str):
+    """
+    Find a pandas plotting backend
```
itholic, line 1127: Maybe `pandas` -> `Koalas` ?

databricks/koalas/plot.py
```diff
@@ -1118,6 +1118,77 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+
+    if backend == "matplotlib":
+        # Because matplotlib is an optional dependency and first-party backend,
+        # we need to attempt an import here to raise an ImportError if needed.
+        try:
+            import databricks.koalas.plot as module
+        except ImportError:
+            raise ImportError(
+                "matplotlib is required for plotting when the "
+                'default backend "matplotlib" is selected.'
```
itholic, line 1179: How about 

```python
"default backend 'matplotlib' is selected."
```

to use double quoter for both line to improve readability ?

databricks/koalas/plot.py
```diff
@@ -1507,6 +1615,37 @@ class KoalasFramePlotMethods(PandasObject):
     ``df.plot(kind='hist')`` is equivalent to ``df.plot.hist()``
     """
 
+    @staticmethod
+    def _get_args_map(backend_name, data, kind, kwargs):
+        """Appropriate call args and data preprocessing mapping for the backend
```
itholic, line 1620: Can we have a simple example for this methods ?

itholic:
@DumbMachine Thanks for the work for this !! 👍

databricks/koalas/plot.py
DumbMachine: We use the `try` because the `_backends` initially does not have any modules. `_backends` gets modules assigned to it when the modules are loaded, upon the user's request of the plotting module.  The use of:
```python
        return _backends[backend]
``` 

is that if the use has the module already saved in the `_backends` dict, there will be no need for reimporting the package. The image below will example better:
image

DumbMachine:
> Btw, I'm wondering how the other backends, pandas_bokeh and pandas_altair?
I tried pandas_bokeh locally, but it shows the same figure as matplotlib. Am I missing something?

Could you try again?

DumbMachine:
Is there a way of running these tests locally? When I try to run `dev/pytest`:
```bash
(base) ➜  koalas git:(plotly-support) ✗ ./dev/pytest
ERROR: usage: __main__.py [options] [file_or_dir] [file_or_dir] [...]
__main__.py: error: unrecognized arguments: --cov=databricks --cov-report databricks
  inifile: None
[diff.txt](https://github.com/databricks/koalas/files/4895806/diff.txt)

  rootdir: /home/weakstation/work/koalas

```
Also when I run `dev/reformat`, `black` adds back all those `,`s.

[diff.txt](https://github.com/databricks/koalas/files/4895810/diff.txt)

databricks/koalas/tests/test_series_plot.py
```diff
@@ -68,6 +69,30 @@ def plot_to_base64(ax):
...
+            self.assertEqual(module.__name__, backend)
+
+    def test_plot_backends_incorrect(self):
+        default_plot_backend = "matplotlib"
+        fake_plot_backend = "none_plotting_module"
+
+        ks.set_option("plotting.backend", fake_plot_backend)
+        self.assertEqual(ks.options.plotting.backend, fake_plot_backend)
+        
+        with pytest.raises(ValueError):
```
ueshin, line 90: Could you use `self.assertRaises` instead of `pytest.raises`?

databricks/koalas/plot.py
```diff
@@ -14,8 +14,10 @@
 # limitations under the License.
 #
 
+import importlib
 from distutils.version import LooseVersion
 
+_backends = {}
```
ueshin, line 20: Could you move this to the original position?

databricks/koalas/plot.py
```diff
@@ -1118,6 +1120,103 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+        _backends["matplotlib"] = module
+
+    if backend in _backends:
+        return _backends[backend]
+
+    module = _find_backend(backend)
+    _backends[backend] = module
+    return module
+
+    def _get_args_map(backend_name, data, kind, kwargs):
```
ueshin, line 1190: indent?

databricks/koalas/plot.py
```diff
@@ -1155,8 +1254,14 @@ def __call__(
         xerr=None,
         label=None,
         secondary_y=False,
-        **kwds
+        **kwds,
     ):
+        plot_backend = _get_plot_backend(kwds.pop("backend", None))
+        # when using another backend, let the backend take the charge
+        if plot_backend.__name__ != "databricks.koalas.plot":
+            plot_data, kwds = self._get_args_map(plot_backend.__name__, self.data, kind, kwds)
```
ueshin, line 1262: `_get_args_map` without `self.`?

databricks/koalas/plot.py
```diff
@@ -1541,8 +1646,14 @@ def __call__(
         xerr=None,
         secondary_y=False,
         sort_columns=False,
-        **kwds
+        **kwds,
     ):
+        plot_backend = _get_plot_backend(kwds.pop("backend", None))
+        # when using another backend, let the backend take the charge
+        if plot_backend.__name__ != "databricks.koalas.plot":
+            plot_data, kwds = self._get_args_map(plot_backend.__name__, self.data, kind, kwds)
```
ueshin, line 1654: ditto.

docs/source/user_guide/options.rst
```diff
@@ -227,10 +227,11 @@ This is conceptually equivalent to the PySpark example as below:
 Available options
 -----------------
 
+
```
ueshin, line 230: Could you revert unrelated changes?

docs/source/user_guide/options.rst
```diff
@@ -227,10 +227,11 @@ This is conceptually equivalent to the PySpark example as below:
...
 =============================== ============== =====================================================
-Option                          Default        Description
+Option                          Default        Description
```
ueshin, line 232: ditto.

databricks/koalas/plot.py
```diff
@@ -1118,6 +1120,103 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
     return plot_obj.result
 
 
+def _find_backend(backend):
+    """
+    Find a koalas plotting backend
```
ueshin, line 1125: `Koalas` instead of `koalas`?

databricks/koalas/config.py
```diff
@@ -220,6 +220,17 @@ def validate(self, v: Any) -> None:
...
+    Option(
+        key="plotting.backend",
+        doc=(
+            "Backend to use for plotting. Default is matplotlib. "
+            "Supports any package that has a top-level `.plot` method. "
+            "Some options are: [matplotlib, plotly, pandas_bokeh, pandas_altair]."
+        ),
+        default="matplotlib",
+        types=str,
+        check_func=(lambda v: isinstance(v, str), "'plotting.backend' should b a string",),
```
ueshin, line 232: This check is done by `types=str`.

How about:

```py
        check_func=(
            lambda v: v in ("matplotlib", "plotly", "pandas_bokeh", "pandas_altair"),
            "'plotting.backend' should be one of 'matplotlib', 'plotly', 'pandas_bokeh', 'pandas_altair'",
        ),
```

ueshin:
@DumbMachine For the `dev/pytest`, it should work in our local environment.
I guess you forgot to install `pytest-cov`?

Otherwise, you can just explicitly specify the arguments:

```sh
./dev/pytest --verbose --showlocals --color=yes --doctest-modules databricks
```

databricks/koalas/config.py
DumbMachine: This will force the user to choose from the given options. Any library or their own custom library that supports a top-level `.plot` function will not be usable.

databricks/koalas/plot.py
DumbMachine: Is there a problem with indentation here?

ueshin: Now that this function is top level, so we have to not indent this function.

databricks/koalas/plot.py
```diff
@@ -1118,6 +1120,104 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+                # Validate that the interface is implemented when the option
+                # is set, rather than at plot time.
+                _backends[backend] = module
+                return module
+
+    # if failed to load plotting backend, set it back to default
+    from databricks.koalas import set_option
+
+    set_option("plotting.backend", "matplotlib")
+
```
ueshin, line 1158: I don't think we should do this automatically.

DumbMachine: `Pandas` does it, I will remove it if its better that way.

databricks/koalas/tests/test_series_plot.py
```diff
@@ -68,6 +69,30 @@ def plot_to_base64(ax):
...
+    def test_plot_backends(self):
+        default_plot_backend = "matplotlib"
+        plot_backends = ["plotly", "pandas_bokeh"]
+
+        for backend in plot_backends:
+            ks.set_option("plotting.backend", backend)
+            self.assertEqual(ks.options.plotting.backend, backend)
+
+            module = ks.plot._get_plot_backend(backend)
+            self.assertEqual(module.__name__, backend)
```
ueshin, line 81: Shall we use `ks.option_context()`?

```py
with ks.option_context("plotting.backend", backend):
    ...
```

databricks/koalas/tests/test_series_plot.py
```diff
@@ -68,6 +69,30 @@ def plot_to_base64(ax):
...
+        fake_plot_backend = "none_plotting_module"
+
+        ks.set_option("plotting.backend", fake_plot_backend)
+        self.assertEqual(ks.options.plotting.backend, fake_plot_backend)
+
+        with pytest.raises(ValueError):
+            ks.plot._get_plot_backend(fake_plot_backend)
+
+        # plotting backend returns to default after failing to load custom
+        assert ks.options.plotting.backend == default_plot_backend
```
ueshin, line 94: ditto in this test.

DumbMachine:
The tests fail I believe due to `plotly` not being installed in the testing environment. What should be done here?

databricks/koalas/plot.py
ueshin: Seems like pandas throws an error when setting the option.

ueshin:
@DumbMachine you can add the libraries you need to the `Test` category in `requirements-dev.txt` file.

databricks/koalas/config.py
ueshin: ah, so any libraries which satisfy the convention, e.g. have `plot` function, etc., can work as backend?

databricks/koalas/plot.py
ueshin: Shall we remove this for now?

DumbMachine:
The tests that are failing, fail due to lack of `pandas_bokeh`. I did add the package to test requirements.

databricks/koalas/config.py
ueshin: if yes above, shall we remove the check?

ueshin:
I'll fix the conda builds.

databricks/koalas/config.py
DumbMachine: Yes

ueshin:
ah, `pandas_bokeh` is not in conda-forge. maybe we should remove the test with it.

databricks/koalas/config.py
ueshin: Could you remove the check here?

databricks/koalas/plot.py
```diff
@@ -1543,6 +1647,12 @@ def __call__(
         sort_columns=False,
         **kwds
     ):
+        plot_backend = _get_plot_backend(kwds.pop("backend", None))
+        # when using another backend, let the backend take the charge
+        if plot_backend.__name__ != "databricks.koalas.plot":
+            plot_data, kwds = _get_args_map(plot_backend.__name__, self.data, kind, kwds)
+            return plot_backend.plot(plot_data, x=x, y=y, kind=kind, **kwds)
```
ueshin, line 1654: This doesn't read the other arguments. e.g.,

```py
kdf = ks.DataFrame([[1, 2, 3, 4], [5, 6, 7, 8]], columns=["A", "B", "C", "D"])
kdf.plot(logx=True, backend="pandas_bokeh")
```

Screen Shot 2020-07-09 at 3 05 11 PM

whereas pandas:

```py
kdf.to_pandas().plot(logx=True, backend="pandas_bokeh")
```

Screen Shot 2020-07-09 at 3 06 41 PM

databricks/koalas/plot.py
```diff
@@ -1157,6 +1255,12 @@ def __call__(
         secondary_y=False,
         **kwds
     ):
+        plot_backend = _get_plot_backend(kwds.pop("backend", None))
+        # when using another backend, let the backend take the charge
+        if plot_backend.__name__ != "databricks.koalas.plot":
+            plot_data, kwds = _get_args_map(plot_backend.__name__, self.data, kind, kwds)
+            return plot_backend.plot(plot_data, kind=kind, **kwds)
```
ueshin, line 1262: ditto.

databricks/koalas/plot.py
DumbMachine: I was trying to work this out but now I feel we should NOT try to map arguments from one backend to another (which I was trying to, using `args_map`). So this would imply that: we pass the minimal important info to the `.plot` function and other arguments are passed using kwargs? Does this sound suitable?

HyukjinKwon: @DumbMachine, so are you saying we should handle known/important arguments and pass other arguments as other keyword arguments, instead of whitelisting everything? I think that way is okay.

We can fix or think about different ways later if that way is proven that it doesn't work in many other cases. I think it will work in most of important cases.

databricks/koalas/plot.py
itholic: @DumbMachine Thanks for the explanation. let's keep this as it as

DumbMachine:
Made the changes and all types of arguments are now supported. Please test it out.
image

image

databricks/koalas/tests/test_series_plot.py
```diff
@@ -68,6 +68,29 @@ def plot_to_base64(ax):
         plt.close(ax.figure)
         return b64_data
 
+    def test_plot_backends(self):
+        default_plot_backend = "matplotlib"
```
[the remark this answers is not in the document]
ueshin: Could you remove this?

databricks/koalas/tests/test_series_plot.py
```diff
@@ -68,6 +68,26 @@ def plot_to_base64(ax):
...
+        plot_backend = "plotly"
+
+        with ks.option_context("plotting.backend", plot_backend):
+            self.assertEqual(ks.options.plotting.backend, plot_backend)
+
+            module = ks.plot._get_plot_backend(plot_backend)
+            self.assertEqual(module.__name__, plot_backend)
+
+    def test_plot_backends_incorrect(self):
+        default_plot_backend = "matplotlib"
```
ueshin, line 82: ditto as we don't need this any more.

databricks/koalas/plot.py
```diff
@@ -1118,6 +1119,173 @@ def _plot(data, x=None, y=None, subplots=False, ax=None, kind="line", **kwds):
...
+        for arg_name_mpl, arg_name_ply in args_map[backend_name]:
+            if arg_name_ply is None:
+                kwargs.pop(arg_name_mpl, None)
+            else:
+                kwargs[arg_name_ply] = kwargs.pop(arg_name_mpl)
+
+    # Values not being same as default implies user is explicitly passing the arguments
+    for arg, def_val in positional_args:
+        if arg in kwargs and kwargs[arg] == def_val:
+            kwargs.pop(arg, None)
```
ueshin, line 1284: I guess we should apply this before Line 1274, otherwise ploty specific arguments could have different values?

ueshin:
@DumbMachine Can I ask you another favor while we are here?
Could you update the PR description to have some simple example codes to use the plotting backend and the result figures?

e.g.,:

```py
kdf = ...
kdf.plot(...)
```

figure

```py
ks.option.plotting.backend = ...
kdf.plot(...)
```

figure

...

databricks/koalas/plot.py
DumbMachine: Made the changes. Now we try to map if possible and remove arguements if default.

HyukjinKwon:
I will double check by myself tomorrow and make a followup to clean up.
Thank you @DumbMachine for working on this!

itholic:
Sorry for the late.
LGTM, either.
Thanks for the work on this, @DumbMachine 👍

DumbMachine:
Always happy to help. 
Is there a reason to explicitly have two different plotter `KoalasFramePlotMethods` and `KoalasSeriesPlotMethods` ? Most of the documentation and function calls are same for both plotters. I believe this could be simplified having a single class (similar to `PlotAccessor` from `pandas`) as diff functions to plot already exist (`plot_series` and `plot_frame`).

ueshin:
@DumbMachine I'm not sure about the original motivation for having two plotter classes, but it'd be great if we could simplify the plotters. Would you like to work on it?

DumbMachine:
Yes, I'd like to work on it. I will open a PR for it, soon.

HyukjinKwon:
Sure, nice!

PR #1641, opened 2020-07-09
[WIP] Add 'distributed-sequence' type default index tests for CI.

ueshin:


databricks/koalas/namespace.py
```diff
@@ -481,13 +481,19 @@ def read_delta(
...
         options = options.get("options")  # type: ignore
 
     if version is not None:
         options["versionAsOf"] = version
     if timestamp is not None:
         options["timestampAsOf"] = timestamp
+    if version is None and timestamp is None:
+        delta_table = DeltaTable.forPath(default_session(), path)
+        latest_version = delta_table.history(1).select("version").head()[0]
+        options["versionAsOf"] = latest_version
```
ueshin, line 496: If we use cache, Delta won't load the latest version unless explicitly specify the version.

databricks/koalas/internal.py
```diff
@@ -598,6 +600,10 @@ def attach_distributed_sequence_column(sdf, column_name):
         #         ...
         #     }
         sdf = sdf.withColumn(spark_partition_column, F.spark_partition_id())
+
+        # Cache the DataFrame to fix the partition ID.
+        sdf.cache()
```
ueshin, line 605: Without this, partition ID could be changed between creating counts and attaching the counts..

databricks/koalas/namespace.py
ueshin, line 496: If we use cache, Delta won't load the latest version unless explicitly specify the version.

databricks/koalas/internal.py
ueshin: Also it needs to create cache of the whole dataset, so `head()` or `tail()` will be slow.

ueshin: We might want to add a new default index type using cache here.

ueshin:
Close this in favor of #1701.

PR #1643, opened 2020-07-09
Fixed ks.Index.to_series() to work properly with name paramter

itholic:
The existing `ks.Index.to_series()` work not properly with `name` parameter
(**Although Its test was passed !**)

```python
>>> kidx.to_series(name="Koalas")
1    1
2    2
3    3
Name: 0, dtype: int64  # It should have renamed to 'Koalas'
```

This PR addressed it.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -683,9 +683,9 @@ class Index(IndexOpsMixin):
             name = self.name
         column_labels = (
             [(SPARK_DEFAULT_SERIES_NAME,)]
-            if len(kdf._internal.index_map) > 1
-            else [(SPARK_DEFAULT_SERIES_NAME,) if name is None else (name,)]
-        )
+            if len(kdf._internal.index_map) > 1 or name is None
+            else [name if isinstance(name, tuple) else (name,)]
+        )  # type: List[Tuple[str, ...]]
         internal = kdf._internal.copy(
             column_labels=column_labels, data_spark_columns=[scol], column_label_names=None
         )
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -679,6 +679,8 @@ class Index(IndexOpsMixin):
         scol = self.spark.column
         if name is not None:
             scol = scol.alias(name_like_string(name))
+        elif len(kdf._internal.index_map) == 1:
+            name = self.name
         column_labels = (
             [(SPARK_DEFAULT_SERIES_NAME,)]
             if len(kdf._internal.index_map) > 1
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -682,10 +682,7 @@ class Index(IndexOpsMixin):
         column_labels = (
             [(SPARK_DEFAULT_SERIES_NAME,)]
             if len(kdf._internal.index_map) > 1
-            else [
-                (SPARK_DEFAULT_SERIES_NAME,) if name is None else name
-                for name in kdf._internal.index_names
-            ]
+            else [(SPARK_DEFAULT_SERIES_NAME,) if name is None else (name,)]
         )
         internal = kdf._internal.copy(
             column_labels=column_labels, data_spark_columns=[scol], column_label_names=None
--- a/databricks/koalas/tests/test_indexes.py
+++ b/databricks/koalas/tests/test_indexes.py
@@ -75,7 +75,7 @@ class IndexesTest(ReusedSQLTestCase, TestUtils):
         kidx = self.kdf.index
 
         self.assert_eq(kidx.to_series(), pidx.to_series())
-        self.assert_eq(kidx.to_series(name="a"), pidx.to_series(name="a"))
+        self.assert_eq(repr(kidx.to_series(name="a")), repr(pidx.to_series(name="a")))
 
         self.assert_eq((kidx + 1).to_series(), (pidx + 1).to_series())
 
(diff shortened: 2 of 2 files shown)


itholic:
The existing test code below has been passing even though the name was not changed properly.

```python
self.assert_eq(kidx.to_series(name="a"), pidx.to_series(name="a"))
```

For now, I addressed it with `repr`, but I think we'd better find the root cause why this test has been passing before.

ueshin:
What if ks.Index([1, 2, 3], name=‘a’).to_frame() ?

itholic:
> What if ks.Index([1, 2, 3], name=‘a’).to_series() ?

Oh, I think we should manage it, too.

Thanks! :)

itholic:
@ueshin 

Anyway, could you help me to solve this mypy failure when you available ?
Since I'm not familiar with mypy that much, couldn't find the solution yet. 😿

databricks/koalas/indexes.py
```diff
@@ -679,13 +679,12 @@ def to_series(self, name: Union[str, Tuple[str, ...]] = None) -> Series:
...
+        elif len(kdf._internal.index_map) == 1:
+            name = self.name
         column_labels = (
             [(SPARK_DEFAULT_SERIES_NAME,)]
             if len(kdf._internal.index_map) > 1
-            else [
-                (SPARK_DEFAULT_SERIES_NAME,) if name is None else name
-                for name in kdf._internal.index_names
-            ]
+            else [(SPARK_DEFAULT_SERIES_NAME,) if name is None else (name,)]
```
HyukjinKwon, line 687: None of us is very fluent in mypy and type hinting, @itholic. It might be false-positive.

```
databricks/koalas/indexes.py: note: In member "to_series" of class "Index":
databricks/koalas/indexes.py:687: error: List item 0 has incompatible type "Tuple[Union[str, Tuple[str, ...]]]"; expected "Tuple[str]"
```

ueshin: Yeah, we are always doing trial and error. 😄 

How about:

```diff
             name = self.name
         column_labels = (
             [(SPARK_DEFAULT_SERIES_NAME,)]
-            if len(kdf._internal.index_map) > 1
-            else [(SPARK_DEFAULT_SERIES_NAME,) if name is None else (name,)]
-        )
+            if len(kdf._internal.index_map) > 1 or name is None
+            else [name if isinstance(name, tuple) else (name,)]
+        )  # type: List[Tuple[str, ...]]
         internal = kdf._internal.copy(
             column_labels=column_labels, data_spark_columns=[scol], column_label_names=None
         )
```

databricks/koalas/tests/test_indexes.py
```diff
@@ -75,7 +75,12 @@ def test_to_series(self):
         kidx = self.kdf.index
 
         self.assert_eq(kidx.to_series(), pidx.to_series())
-        self.assert_eq(kidx.to_series(name="a"), pidx.to_series(name="a"))
+        self.assert_eq(repr(kidx.to_series(name="a")), repr(pidx.to_series(name="a")))
+
+        pidx.name = "Koalas"
+        kidx.name = "Koalas"
+        self.assert_eq(repr(kidx.to_series()), repr(pidx.to_series()))
+        self.assert_eq(repr(kidx.to_series(name="a")), repr(pidx.to_series(name="a")))
```
ueshin, line 83: Could you add some more tests with tupled names?

- `ks.Index([1,2,3], name=('x', 'a')).to_series()`
- `ks.Index([1,2,3]).to_series(name=('x', 'a'))`

if those are valid in pandas as well?

databricks/koalas/indexes.py
itholic: Thanks @HyukjinKwon , @ueshin .

I'll have to try to get familiar with type checking with `mypy`.

Sincerely thanks for your time for addressing this !!

databricks/koalas/tests/test_indexes.py
itholic: @ueshin Sure, I'm going to add it. Thanks!

databricks/koalas/indexes.py
```diff
@@ -679,14 +679,13 @@ def to_series(self, name: Union[str, Tuple[str, ...]] = None) -> Series:
...
             [(SPARK_DEFAULT_SERIES_NAME,)]
-            if len(kdf._internal.index_map) > 1
-            else [
-                (SPARK_DEFAULT_SERIES_NAME,) if name is None else name
-                for name in kdf._internal.index_names
-            ]
-        )
+            if len(kdf._internal.index_map) > 1 or name is None
+            else [name if isinstance(name, tuple) else (name,)]
+        )  # type: List[Tuple[str, ...]]
```
itholic, line 688: Now seems working well, but this code still occurs an error in my local like the below.

`databricks/koalas/indexes.py:687: error: List item 0 has incompatible type "object"; expected "Tuple[str, ...]"`

ueshin: It doesn't happen in my local. Maybe your mypy version is old? Mine is `0.780` and GitHub Action's is `0.782`.

databricks/koalas/tests/test_indexes.py
itholic: > What about the second one to use a tupled name as a parameter?

Sounds good. I just addressed.

Thanks for the comment !

databricks/koalas/indexes.py
itholic: > It doesn't happen in my local. Maybe your mypy version is old? Mine is `0.780` and GitHub Action's is `0.782`.

You're right, It was my mypy version problem. now It's working, thanks :)

Issue #1647, opened 2020-07-13
Index.to_frame() seems not working properly.


itholic:
```python
>>> kdf = ks.DataFrame({"Koalas": [1, 2, 3]}, index=pd.Index([1, 2, 3]))
>>> kdf["NEW"] = ks.Series([100, 200, 300])
>>> kdf
   0    NEW
1  1  200.0
3  3    NaN
2  2  300.0
```

The above code is working well.

But the same shape of DataFrame which is made from Index.to_frame() seems not work properly.

```python
>>> kdf = ks.Index([1, 2, 3]).to_frame(name="Koalas")
>>> kdf["NEW"] = ks.Series([100, 200, 300])
>>> kdf
Traceback (most recent call last):
...
pyspark.sql.utils.AnalysisException: Reference 'Koalas' is ambiguous, could be: Koalas, Koalas.;
```

itholic:
Let me fix this for completing #1630

ueshin:
What's the problem?

itholic:
I thought we should always manage the `data_spark_columns` and `index_spark_columns` separately because the internal Spark columns should be changed when the external Koalas columns are changed.

For example,

```python
>>> kdf = ks.DataFrame({"Koalas": [1, 2, 3]}, index=pd.Index([1, 2, 3]))
>>> kdf["NEW"] = ks.Series([100, 200, 300])
>>> kdf
   0    NEW
1  1  200.0
3  3    NaN
2  2  300.0
```

The above code is working well.

But the same shape of `DataFrame` which is made from `Index.to_frame()` seems not work properly.

```python
>>> kdf = ks.Index([1, 2, 3]).to_frame(name="Koalas")
>>> kdf["NEW"] = ks.Series([100, 200, 300])
>>> kdf
Traceback (most recent call last):
...
pyspark.sql.utils.AnalysisException: Reference 'Koalas' is ambiguous, could be: Koalas, Koalas.;
```

Hmm.. but probably It seems not not problem with internal Spark frame... let me investigate this more carefully

itholic:
I just changed the title and description of this issue.

itholic:
Anyway, could I have some additional question about `InternalFrame` that I often feel confusing?

Don't we need to address the values of internal Spark columns after modifying the external values ?

For example, let's say we have a `Series` like the below.

```python
>>> kser
0    1
1    2
2    3
Name: 0, dtype: int64

>>> kser._internal.spark_frame.show()
[...]

PR #1649, opened 2020-07-14
Rename spark columns only when index=False.

ueshin:
Consolidates the logic for `Index.to_frame()` and `MultiIndex.to_frame()` and rename Spark columns only when `index=False`.
Related to https://github.com/databricks/koalas/issues/1647#issuecomment-657405518, but not fully fixes it.


ueshin:
Thanks! I'd merge this now.

itholic:
+1 1

PR #1650, opened 2020-07-14
Use SPARK_INDEX_NAME_FORMAT in combine_frames to avoid ambiguity.

ueshin:
Use `SPARK_INDEX_NAME_FORMAT` in `utils.combine_frames` to avoid ambiguity.

```py
>>> kdf = ks.DataFrame({"a": [1, 2, 3], "Koalas": [0, 1, 2]}).set_index("Koalas", drop=False)
>>> kdf.index.name = None
>>> kdf["NEW"] = ks.Series([100, 200, 300])
>>> kdf
...
pyspark.sql.utils.AnalysisException: Reference 'Koalas' is ambiguous, could be: Koalas, Koalas.;
```

Related to #1647 as well.


itholic:
LGTM

PR #1654, opened 2020-07-15
Fix fillna to handle "ffill" and "bfill" properly.

### LINKED ISSUE #1651: Fillna doesn't get the correct result when arg "method" is "ffill".
Fillna the method of Koalas Series  doesn't get the correct result when method "ffill" not work.

After I execute fillna with method "ffill", print the whole result, it is right. But when I print the value which is filled, it is None.

example:
```python
>>> from databricks.koalas import Series
>>> import numpy as np
>>> a = ["Bob", "John", None]
>>> b = ["Bob", "John", np.nan]
>>> s1 = Series(a)
>>> s2 = Series(b)
>>> r1 = s1.fillna(method="ffill")
>>> r1
20/07/15 15:18:44 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
0     Bob
1    John
2    John
Name: 0, dtype: object
>>> print(r1[2])
20/07/15 15:19:06 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
None
>>> r2 = s2.fillna(method="ffill")
>>> r2
20/07/15 15:19:19 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
0     Bob
1    John
2    John
Name: 0, dtype: object
>>> print(r2[2])
20/07/15 15:19:27 WARN WindowExec: No Partition Defined for Window operation! Moving all data to a single partition, this can cause serious performance degradation.
None
```

I think r1[2] and r2[2] should be "John", but I got "None".

ueshin:
`fillna` with `method="ffill"` or `method="bfill"` uses Spark's Window functions internally, so we should create a new Spark DataFrame to apply the function properly.

Fixes #1651:

```py
>>> from databricks.koalas import Series
>>> import numpy as np
>>> a = ["Bob", "John", None]
>>> b = ["Bob", "John", np.nan]
>>> s1 = Series(a)
>>> s2 = Series(b)
>>> r1 = s1.fillna(method="ffill")
>>> r1
0     Bob
1    John
2    John
Name: 0, dtype: object
>>> r1[2]
'John'
>>> r2 = s2.fillna(method="ffill")
>>> r2
0     Bob
1    John
2    John
Name: 0, dtype: object
>>> r2[2]
'John'
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4708,11 +4708,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         2  0.0  1.0  2.0  5
         3  0.0  3.0  1.0  4
         """
+        axis = validate_axis(axis)
+        if axis != 0:
+            raise NotImplementedError("fillna currently only works for axis=0 or axis='index'")
+
         if value is not None:
-            axis = validate_axis(axis)
-            inplace = validate_bool_kwarg(inplace, "inplace")
-            if axis != 0:
-                raise NotImplementedError("fillna currently only works for axis=0 or axis='index'")
             if not isinstance(value, (float, int, str, bool, dict, pd.Series)):
                 raise TypeError("Unsupported type %s" % type(value))
             if limit is not None:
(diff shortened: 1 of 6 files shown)


ueshin:
Thanks! merging.

PR #1656, opened 2020-07-15
read_csv: Implement reading of number of rows

tomspur:
Implement reading of number of rows (nrows) in read_csv by using spark's limit.

On the first read, this does not seem to help much with regards to reading speed, because `inferSchema` is True and spark seems to scan over the full data anyway (see e.g. [here](https://stackoverflow.com/questions/44277019/how-to-read-only-n-rows-of-large-csv-file-on-hdfs-using-spark-csv-package)). It would help to use it as a similar parameter to `read_csv` than in pandas to make the API more compatible.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -183,6 +183,7 @@ def read_csv(
     squeeze=False,
     mangle_dupe_cols=True,
     dtype=None,
+    nrows=None,
     parse_dates=False,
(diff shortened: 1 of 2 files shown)


ueshin:
Thanks! merging.

itholic:
Nice work. Thanks! :D

tomspur:
Thank you for the quick response and merge! :)

PR #1659, opened 2020-07-16
Fix test failures caused by numpy upgrade.

ueshin:
Since numpy 1.19, `np.delete` came to fail on out-of-bounds indices (https://github.com/numpy/numpy/pull/15804) and pandas will as well since it depends on numpy.


ueshin:
We might want to fix the behavior to follow the new numpy's behavior.

ueshin:
I'd merge this for now.

PR #1660, opened 2020-07-17
Set the lower bound for Plotly.

ueshin:
Sets the lower bound for Plotly, just in case, since Plotly works as a plotting background since `4.8`.

> Since version 0.25, Pandas has provided a mechanism to use different backends, and as of version 4.8 of plotly, you can now use a Plotly Express-powered backend for Pandas plotting.

https://plotly.com/python/pandas-backend/

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -28,4 +28,4 @@ pytest-cov
 scikit-learn
 openpyxl
 xlrd
-plotly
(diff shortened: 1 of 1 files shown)


ueshin:
This is trivial. I'd merge this now.

HyukjinKwon:
Nice!

Issue #1670, opened 2020-07-21
Remove the upper bounds for numpy and matplotlib.


ueshin:
So far pandas sets upper bounds for `numpy` and `matplotlib`.
Let's see how pandas handles the upper bounds.
https://github.com/pandas-dev/pandas/blob/master/requirements-dev.txt

ueshin:
The upper bounds will be removed in pandas 1.2. https://github.com/pandas-dev/pandas/pull/35780

ueshin:
I'd close this as they were removed at #1845 and #1839.

HyukjinKwon:
Nice!

PR #1671, opened 2020-07-21
Set the upper bound for numpy.

ueshin:
So far pandas sets the upper bound for numpy: https://github.com/pandas-dev/pandas/blob/master/requirements-dev.txt#L4
I'd revert the previous change to fix build and set the upper bound here as well.
Tracking to remove the upper bounds for `numpy` and `matplotlib` at #1670.


ueshin:
I'd merge this now.

PR #1672, opened 2020-07-21
Fix extension documents.

ueshin:
Fixes extension documents since the examples for each extension decorator don't work as-is.


ueshin:
Since this is only a document fix, I'd merge this now.

HyukjinKwon:
👍

PR #1675, opened 2020-07-25
Matching an error message same as pandas for Index.to_series

itholic:
To match an error message same as pandas.

```python
>>> pdf
   dogs  cats
a   0.2   0.3
b   0.0   0.6
c   0.6   0.0
d   0.2   0.1
>>> kdf
   dogs  cats
a   0.2   0.3
b   0.0   0.6
c   0.6   0.0
d   0.2   0.1

>>> pdf['dogs'].index.to_series(name=["x", "a"])
...
TypeError: Series.name must be a hashable type

>>> kdf['dogs'].index.to_series(name=["x", "a"])
...
TypeError: Series.name must be a hashable type
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -17,7 +17,7 @@
 """
 Wrappers for Indexes to behave similar to pandas Index, MultiIndex.
 """
-from collections import OrderedDict
+from collections import OrderedDict, abc
 from distutils.version import LooseVersion
 from functools import partial
 from typing import Any, List, Tuple, Union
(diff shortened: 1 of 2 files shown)


databricks/koalas/indexes.py
```diff
@@ -675,6 +675,8 @@ def to_series(self, name: Union[str, Tuple[str, ...]] = None) -> Series:
         d    d
         Name: 0, dtype: object
         """
+        if not isinstance(name, abc.Hashable):
```
HyukjinKwon, line 678: Should we use pandas' `is_hashable` instead?

itholic: Seems better. Thanks! 👍

PR #1678, opened 2020-07-28
Make read_excel read from DFS if the underlying Spark is 3.0.0 or above.

ueshin:
Currently `ks.read_excel` is trying to read a file in the Driver local file system, but it should read from DFS the same as the other `read_xxx` functions.
This PR uses `binaryFile` file format supported in Spark 3.0, so it only works with Spark 3.0 or above.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -25,7 +25,8 @@ from functools import reduce
 
 import numpy as np
 import pandas as pd
-from pandas.api.types import is_list_like
+from pandas.api.types import is_datetime64_dtype, is_datetime64tz_dtype, is_list_like
+import pyspark
 from pyspark import sql as spark
 from pyspark.sql import functions as F
 from pyspark.sql.types import (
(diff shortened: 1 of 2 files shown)


databricks/koalas/namespace.py
```diff
@@ -738,9 +743,10 @@ def read_excel(
     Parameters
     ----------
     io : str, file descriptor, pathlib.Path, ExcelFile or xlrd.Book
-        The string could be a URL. Valid URL schemes include http, ftp, s3,
-        gcs, and file. For file URLs, a host is expected. For instance, a local
-        file could be /path/to/workbook.xlsx.
+        The string could be a URL. If the underlying Spark is above 3.0.0, the value URL must
+        be available by the Spark; otherwise the valid URL schemes include http, ftp, s3, gcs,
```
HyukjinKwon, line 747: @ueshin, what about we just throw an exception if we have to delegate it to pandas? We could say like "use ks.from_pandas(pd.read_excel(...)) as a workaround".

ueshin: I pushed a commit. I hope it's addressing what you meant.

HyukjinKwon:
Thanks @ueshin!

PR #1681, opened 2020-07-29
Bug fixing for hasnans when non-DoubleType.

itholic:
`ks.Series.hasnans` and `ks.Index.hasnans` seems not work properly for non-DoubleType.

```python
>>> ks.Series([True, True, np.nan]).hasnans
...
pyspark.sql.utils.AnalysisException: cannot resolve 'isnan(`0`)' due to data type mismatch: argument 1 requires (double or float) type, however, '`0`' is of boolean type.;;
'Aggregate [max((isnull(0#12) OR isnan(0#12))) AS max(((0 IS NULL) OR isnan(0)))#21]
+- Project [__index_level_0__#11L, 0#12, monotonically_increasing_id() AS __natural_order__#15L]
   +- LogicalRDD [__index_level_0__#11L, 0#12], false
```

This PR fixed it.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -399,11 +399,11 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         sdf = self._internal.spark_frame
         scol = self.spark.column
 
-        if isinstance(self.spark.data_type, BooleanType):
-            # `BooleanType` cannot contain `np.nan`.
-            return sdf.select(F.max(scol.isNull())).collect()[0][0]
-        else:
+        if isinstance(self.spark.data_type, DoubleType):
+            # Check `isnan` for the DoubleType since the type of np.nan is DoubleType in PySpark.
             return sdf.select(F.max(scol.isNull() | F.isnan(scol))).collect()[0][0]
+        else:
+            return sdf.select(F.max(scol.isNull())).collect()[0][0]
 
     @property
     def is_monotonic(self):
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -398,8 +398,7 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         sdf = self._internal.spark_frame
         scol = self.spark.column
 
-        if isinstance(self.spark.data_type, DoubleType):
-            # Check `isnan` for the DoubleType since the type of np.nan is DoubleType in PySpark.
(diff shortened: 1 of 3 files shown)


databricks/koalas/base.py
```diff
@@ -391,7 +391,7 @@ def hasnans(self):
         sdf = self._internal.spark_frame
         scol = self.spark.column
 
-        return sdf.select(F.max(scol.isNull() | F.isnan(scol))).collect()[0][0]
```
itholic, line 394: Maybe I think we don't need `F.isnan(scol)` because `np.nan` is casted to `null` when the Koalas DataFrame is casted to PySpark'.

HyukjinKwon: @itholic there can be `NaN` from Spark DataFrame directly. For example,

```python
import pyspark
spark = pyspark.sql.SparkSession.builder.getOrCreate()
df = spark.createDataFrame([[1.0, float('nan')]])
df.show()
kdf = df.to_koalas()
```

Does it work with the `kdf` too?

itholic: @HyukjinKwon Ah, I got it. We should separate the condition.

`ks.DataFrame` has no attribute `hasnans` just same as `pd.DataFrame`

databricks/koalas/base.py
```diff
@@ -391,7 +399,11 @@ def hasnans(self):
...
+        if isinstance(self.spark.data_type, BooleanType):
```
ueshin, line 402: What if the other data types, such as `StringType`, `DateType`, or `TimestampType`?

itholic: Ah, I see. `np.nan` is available only for the `DoubleType` column. Thanks! :)

databricks/koalas/base.py
```diff
@@ -391,7 +398,11 @@ def hasnans(self):
         sdf = self._internal.spark_frame
         scol = self.spark.column
 
-        return sdf.select(F.max(scol.isNull() | F.isnan(scol))).collect()[0][0]
+        if isinstance(self.spark.data_type, DoubleType):
```
ueshin, line 401: Shall we add `FloatType` as Spark can have `Float.NaN` as well?

itholic: Thanks, @ueshin

PR #1682, opened 2020-07-29
Avoid using undocumented -k syntaxes.

ueshin:
As `pytest 6.0.0` which was released on Jul 28, 2020 dropped undocumented -k syntaxes (https://github.com/pytest-dev/pytest/pull/7210, https://github.com/pytest-dev/pytest/pull/7122), we should use a full sentence query for the parameter:

```
ERROR: Wrong expression passed to '-k': not melt -to_delta -read_delta -to_clipboard: at column 10: expected end of input; got identifier
```


ueshin:
I'd merge this now. cc @HyukjinKwon

HyukjinKwon:
Nice!

PR #1692, opened 2020-07-31
Add support multiple sheets in multiple files for read_excel.

ueshin:
Enables to read multiple sheets from multiple Excel files.


itholic:
LGTM

PR #1693, opened 2020-07-31
Fix the wrong usage of "binaryFile" format in read_excel.

ueshin:
Fixes the wrong usage of "binaryFile" format.

The first column is not the content, but actually `path`.

```
root
 |-- path: string (nullable = true)
 |-- modificationTime: timestamp (nullable = true)
 |-- length: long (nullable = true)
 |-- content: binary (nullable = true)
```


ueshin:
Oops, this was merged before tests passed.
I'll submit a PR to fix it soon.

PR #1695, opened 2020-08-01
Try to read pandas metadata in read_parquet if index_col is None.

### LINKED ISSUE #1645: Cannot read parquet file from Pandas
I computed some data using pandas which I wrote to a file. The df looks like this https://i.imgur.com/pxCLrUh.png 
I wrote it to parquet using `plot_df.to_parquet("test.parquet", flavor='spark')`

Now when I read it in using spark, there is no issue, yet note there being one additional column which apparently pandas adds:
```
df = spark.read.parquet("test.parquet")
df.printSchema()
root
 |-- time: timestamp (nullable = true)
 |-- node: string (nullable = true)
 |-- power_usage: double (nullable = true)
 |-- temperature: double (nullable = true)
 |-- processes_running: double (nullable = true)
 |-- memory_used: double (nullable = true)
 |-- load1: double (nullable = true)
 |-- __index_level_0__: long (nullable = true)
```

Now if I run `kdf = df.to_koalas()` I get
```
AssertionError                            Traceback (most recent call last)
<ipython-input-29-4c46820eb041> in <module>
      1 df = spark.read.parquet("test.parquet")
      2 df.printSchema()
----> 3 kdf = df.to_koalas()

/var/scratch/lvs215/miniconda3/lib/python3.8/site-packages/databricks/koalas/frame.py in to_koalas(self, index_col)
   4449 
   4450             index_map = _get_index_map(self, index_col)
-> 4451             internal = InternalFrame(spark_frame=self, index_map=index_map)
   4452             return DataFrame(internal)
   4453 

/var/scratch/lvs215/miniconda3/lib/python3.8/site-packages/databricks/koalas/internal.py in __init__(self, spark_frame, index_map, column_labels, data_spark_columns, column_label_names)
    436 
    437         if index_map is None:
--> 438             assert not any(SPARK_INDEX_NAME_PATTERN.match(name) for name in spark_frame.columns), (
    439                 "Index columns should not appear in columns of the Spark DataFrame. Avoid "
    440                 "index column names [%s]." % SPARK_INDEX_NAME_PATTERN

AssertionError: Index columns should not appear in columns of the Spark DataFrame. Avoid index column names [re.compile('__index_level_[0-9]+__')].
```

The issue looks similar to https://github.com/xitongsys/parquet-go/issues/233

How to deal with this? Should I follow the suggestion in https://github.com/apache/arrow/issues/2244 and don't preserve the index?

ueshin:
If a parquet file is stored by pandas, there is a metadata to describe index columns.
We can read it and use as `index_col` if it's not specified.
Resolves #1645.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -33,6 +33,7 @@ import pyarrow.parquet as pq
 import pyspark
 from pyspark import sql as spark
 from pyspark.sql import functions as F
+from pyspark.sql.functions import pandas_udf, PandasUDFType
 from pyspark.sql.types import (
     ByteType,
     ShortType,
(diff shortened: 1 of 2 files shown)


databricks/koalas/namespace.py
```diff
@@ -672,6 +675,26 @@ def read_parquet(path, columns=None, index_col=None, **options) -> DataFrame:
     if columns is not None:
         columns = list(columns)
 
+    index_names = None
+
+    if index_col is None and LooseVersion(pyspark.__version__) >= LooseVersion("3.0.0"):
+        # Try to read pandas metadata
+        binary = default_session().read.format("binaryFile").load(path).select("content").head()[0]
```
HyukjinKwon, line 682: My only concern is if it's worthwhile enough to have the overhead.  It adds some extra overhead when `read_parquet` is from Spark or Koalas (not from pandas).

If people read it from the file written in pandas more often, it makes sense.
If people read it from Spark / Koalas more often, I think it might not be worthwhile.

If the file is big, fetching the entire file into the driver side won't be a trivial overhead.

ueshin: Good point.
Maybe shall we have a config?

databricks/koalas/namespace.py
```diff
@@ -625,7 +629,9 @@ def read_spark_io(
     return DataFrame(InternalFrame(spark_frame=sdf, index_map=index_map))
 
 
-def read_parquet(path, columns=None, index_col=None, **options) -> DataFrame:
+def read_parquet(
+    path, columns=None, index_col=None, read_pandas_metadata=False, **options
```
HyukjinKwon, line 633: Maybe we can just name `read_pandas_metadata` to `pandas_metadata` and says it tries to respect the metadata if the Parquet file is written from pandas in the doc.

PR #1697, opened 2020-08-03
expanded allowable types for kdf indexing to include pandas Index object

sllynn:
This PR attempts to address an issue I see with column selection of Koalas DataFrames. Pandas allows use of a Pandas Index object for selecting columns, e.g.

```python
column_mask = entity.df.columns.isin([variable.id for variable in entity.variables])
index_cols = entity.df.columns[column_mask] 
dtypes = entity.df[index_cols].dtypes.astype(str).to_dict() 
```

will work fine in Pandas but fail with Koalas as the type check in the DataFrame's __getitem__ method does not include the Pandas Index type.

If I feed a Panads Index to the Koalas DataFrame.loc method like `entity.df.loc[:, index_cols]` it works. 

I propose expanding the list of allowable types when getting an item from the Frame class, e.g.
`if isinstance(key, (str, tuple, list)):` -> `if isinstance(key, (str, tuple, list, pd.Index)):` to bring it (more) in line with pandas.

I've added to the test_dataframe test to reflect but could break this out if required.

FYI, this is a blocker to running a basic featuretools demo with the proposed Koalas backend.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10094,7 +10094,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         if key is None:
             raise KeyError("none key")
-        if isinstance(key, (str, tuple, list, pd.Index)):
+
+        if isinstance(key, Series):
+            return self.loc[key.astype(bool)]
+        elif isinstance(key, str) or is_list_like(key):
+            key = list(key) if is_list_like(key) else key
             return self.loc[:, key]
         elif isinstance(key, slice):
             if any(type(n) == int or None for n in [key.start, key.stop]):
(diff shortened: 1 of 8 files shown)


databricks/koalas/frame.py
```diff
@@ -10089,7 +10089,7 @@ def __getitem__(self, key):
 
         if key is None:
             raise KeyError("none key")
-        if isinstance(key, (str, tuple, list)):
+        if isinstance(key, (str, tuple, list, pd.Index)):
```
HyukjinKwon, line 10092: Actually, I think it's okay to just change to `if isinstance(key, (str)) or is_list_like(key):` and `key = list(key) if is_list_like(key) else key` for simplicity for now.

HyukjinKwon:
Let me leave it to @ueshin

databricks/koalas/frame.py
sllynn: I think calling `list(key)` when `key` is a koalas series will generate a `databricks.koalas.exceptions.PandasNotImplementedError`. So something like `kdf[kdf["b"] > 2]` will fail.

HyukjinKwon: We can switch the if-else conditions. But okay, let's just keep it as is for now.

sllynn: Of course, makes sense. I've switched the conditions.

PR #1698, opened 2020-08-04
include pandas Index object in dataframe indexing options

sllynn:
(supersedes previous #1697)

This PR attempts to address an issue I see with column selection of Koalas DataFrames. Pandas allows use of a Pandas Index object for selecting columns, e.g.

```
column_mask = entity.df.columns.isin([variable.id for variable in entity.variables])
index_cols = entity.df.columns[column_mask] 
dtypes = entity.df[index_cols].dtypes.astype(str).to_dict() 
```
will work fine in Pandas but fail with Koalas as the type check in the DataFrame's getitem method does not include the Pandas Index type.

The changes here extend column indexing to all list like objects.

I've added to the test_dataframe test to reflect.

### DIFF
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -84,7 +84,7 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
         pdf, kdf = self.df_pair
         column_mask = pdf.columns.isin(["a", "b"])
         index_cols = pdf.columns[column_mask]
-        self.assert_eq(pdf[index_cols].shape, kdf[index_cols].shape)
+        self.assert_eq(kdf[index_cols], pdf[index_cols])
 
     def test_inplace(self):
         pdf, kdf = self.df_pair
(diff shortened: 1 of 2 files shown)


sllynn:
OK, this breaks the Koalas Dataframe groupby logic as when the tuple representing a column name for grouping is passed in to __getitem__ this now is wrapped in a list and __getitem__ returns a Dataframe, rather than a Series as expected.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -80,6 +80,12 @@ def test_dataframe(self):
         kser = ks.from_pandas(pser)
         self.assert_eq(pd.DataFrame(pser), ks.DataFrame(kser))
 
+        # check kdf[pd.Index]
+        pdf, kdf = self.df_pair
+        column_mask = pdf.columns.isin(["a", "b"])
+        index_cols = pdf.columns[column_mask]
+        self.assert_eq(pdf[index_cols].shape, kdf[index_cols].shape)
```
ueshin, line 87: Why comparing only `shape`?

ueshin:
Thanks! merging.

PR #1699, opened 2020-08-05
Use RDD.zipWithIndex for distributed-sequence.

ueshin:
Use `RDD.zipWithIndex` for `distributed-sequence`.

### DIFF
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -1314,7 +1314,7 @@ class iLocIndexer(LocIndexerLike):
         sdf = InternalFrame.attach_distributed_sequence_column(
             internal.spark_frame, column_name=self._sequence_col
         )
-        return internal.with_new_sdf(spark_frame=sdf)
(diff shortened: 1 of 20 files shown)


databricks/koalas/indexing.py
```diff
@@ -625,12 +638,16 @@ def __setitem__(self, key, value):
                     kdf[temp_key_col] = rows_sel
                 if isinstance(value, Series):
                     kdf[temp_value_col] = value
-                kdf = kdf.sort_values(temp_natural_order)
+                kdf = kdf.sort_values(temp_natural_order).drop(temp_natural_order)
 
                 if isinstance(rows_sel, Series):
-                    rows_sel = kdf[temp_key_col]
+                    rows_sel = F.col(
+                        "`{}`".format(kdf[temp_key_col]._internal.data_spark_column_names[0])
```
HyukjinKwon, line 645: I wonder if we should just document that we don't support backquotes in the column names ..

HyukjinKwon (approved):

PR #1701, opened 2020-08-07
Use checkpoint for distributed-sequence.

ueshin:
Use `checkpoint` for `distributed-sequence`.


ueshin:
Thanks! I'd merge this now.

Issue #1007, opened 2019-11-05
ks.groupby won't drop groups with one or more null keys


pandareen:
This is basically where the pandas groupby adopts its behaviour from r.
I can submit a pr if this is not on anyones roadmap

harupy:
pyspark doesn't drop null groups. I didn't know this.

```
>>> sdf.show()
+----+
|   x|
+----+
|null|
| 1.0|
| 1.0|
+----+

>>> sdf.groupBy('x').count().show()
+----+-----+
|   x|count|
+----+-----+
|null|    1|
| 1.0|    2|
+----+-----+

>>> sdf.toPandas().groupby('x').agg({'x': 'count'})
     x
x	
1.0  2
```

pandareen:
@harupy correct. That is to keep it consistent with r library

HyukjinKwon:
Yeah, I noticed this behaviour as well lately. Please go ahead for a PR!

pandareen:
Okay thanks, I'm on it

LucasG0:
Hi, pandas 1.1.0 recently added a `dropna` parameter in order to allow groupby on null keys.
By default (True), null keys are dropped, and it has pyspark behavior when set to False.
I think it would be interesting to add it to koalas dataframe `groupby` method.
May i work on it ?

ueshin:
@LucasG0 Awesome, please go ahead! Thanks!
cc @itholic as working on pandas 1.1 support

itholic:
Thanks, @LucasG0 .
And thanks for the notice, @ueshin

PR #1702, opened 2020-08-10
Implement DataFrame.groupby.cumcount

tomspur:
Here is an example of the implementation in analogy of the one from [the pandas docs](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.GroupBy.cumcount.html):

```python
>>> df = ks.DataFrame(
...     [['a'], ['a'], ['a'], ['b'], ['b'], ['a']],
...     columns=list('A'))
>>> df
   A
0  a
1  a
2  a
3  b
4  b
5  a
>>> df.groupby('A').cumcount(ascending=False).sort_index()
0    3
1    2
2    1
3    1
4    0
5    0
Name: A, dtype: int64
>>> df.groupby('A').cumcount().sort_index()
0    0
1    1
2    2
3    0
4    1
5    3
Name: A, dtype: int64
```

The tests are the same like for the other `cumxxx` tests for now.

Nevertheless, it so far does not work the case of nulls and multiple columns within the groupby (thank you for finding this @itholic):
```python
>>> df = ks.DataFrame({"a": [1, 1, 1, 4], "b": [None, 0.1, 20.0, None], "c": [4, 3, 2, 1]})
>>> df
   a     b  c
0  1   NaN  4
1  1   0.1  3
2  1  20.0  2
3  4   NaN  1
>>> df.groupby(["a", "b"]).cumcount().sort_index()
0    0
1    0
2    0
3    0
Name: a, dtype: int64
>>> df.to_pandas().groupby(["a", "b"]).cumcount().sort_index()
0    0
1    0
2    0
3    1
dtype: int64
```

This will also work, when groupby supports a default `dropna=True` argument in #1007

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -680,7 +680,7 @@ class GroupBy(object, metaclass=ABCMeta):
         3    0
         4    1
         5    3
-        Name: 0, dtype: int64
+        Name: A, dtype: int64
         >>> df.groupby('A').cumcount(ascending=False).sort_index()
         0    3
         1    2
@@ -688,18 +688,17 @@ class GroupBy(object, metaclass=ABCMeta):
         3    1
         4    0
         5    0
-        Name: 0, dtype: int64
+        Name: A, dtype: int64
 
         """
-        ret = self._apply_series_op(
-            lambda sg: sg._kser._cum(
-                F.count, True, part_cols=sg._groupkeys_scols, ascending=ascending
-            ),
-            should_resolve=True,
+        ret = (
+            self._groupkeys[0]
+            .spark.transform(lambda _: F.lit(0))
+            ._cum(F.count, True, part_cols=self._groupkeys_scols, ascending=ascending)
+            - 1
         )
-        ret -= 1
-        # Cast columns to ``"int64"`` to match `pandas.core.groupby.GroupBy.cumcount`.
-        return ret.max(axis=1).astype('int64')
+        internal = ret._internal.resolved_copy
+        return first_series(DataFrame(internal))
 
     def cummax(self):
         """
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -696,7 +696,8 @@ class GroupBy(object, metaclass=ABCMeta):
             should_resolve=True,
         )
         ret -= 1
-        return ret.max(axis=1)
+        # Cast columns to ``"int64"`` to match `pandas.core.groupby.GroupBy.cumcount`.
+        return ret.max(axis=1).astype('int64')
 
     def cummax(self):
         """
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -646,7 +646,9 @@ class GroupBy(object, metaclass=ABCMeta):
 
         Essentially this is equivalent to
 
-        >>> self.apply(lambda x: pd.Series(np.arange(len(x)), x.index))
+        .. code-block:: python
+
+            self.apply(lambda x: pd.Series(np.arange(len(x)), x.index))
 
         Parameters
         ----------
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -642,17 +642,27 @@ class GroupBy(object, metaclass=ABCMeta):
 
     def cumcount(self, ascending=True):
         """
-        Cumulative count for each group.
+        Number each item in each group from 0 to the length of that group - 1.
+
+        Essentially this is equivalent to
+
+        >>> self.apply(lambda x: pd.Series(np.arange(len(x)), x.index))
+
+        Parameters
+        ----------
+        ascending : bool, default True
+            If False, number in reverse, from length of group - 1 to 0.
 
         Returns
         -------
-        Series or DataFrame
+        Series
+            Sequence number of each element within each group.
 
         Examples
         --------
-        >>> df = ks.DataFrame(
-        ...     [['a'], ['a'], ['a'], ['b'], ['b'], ['a']],
-        ...     columns=list('A'))
+
+        >>> df = ks.DataFrame([['a'], ['a'], ['a'], ['b'], ['b'], ['a']],
+        ...                   columns=['A'])
         >>> df
            A
         0  a
@@ -669,7 +679,7 @@ class GroupBy(object, metaclass=ABCMeta):
         4    1
         5    3
         Name: 0, dtype: int64
-        >>> df.groupby('A').cumcount(ascending=False)
+        >>> df.groupby('A').cumcount(ascending=False).sort_index()
         0    3
         1    2
         2    1
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -681,7 +681,8 @@ class GroupBy(object, metaclass=ABCMeta):
         """
         ret = self._apply_series_op(
             lambda sg: sg._kser._cum(
-                F.count, True, part_cols=sg._groupkeys_scols, ascending=ascending),
+                F.count, True, part_cols=sg._groupkeys_scols, ascending=ascending
+            ),
             should_resolve=True,
         )
         ret -= 1
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -640,6 +640,53 @@ class GroupBy(object, metaclass=ABCMeta):
             lambda sg: sg._kser._diff(periods, part_cols=sg._groupkeys_scols)
         )
 
+    def cumcount(self, ascending=True):
+        """
+        Cumulative count for each group.
+
+        Returns
+        -------
+        Series or DataFrame
+
+        Examples
+        --------
+        >>> df = ks.DataFrame(
+        ...     [['a'], ['a'], ['a'], ['b'], ['b'], ['a']],
+        ...     columns=list('A'))
+        >>> df
+           A
+        0  a
+        1  a
+        2  a
+        3  b
+        4  b
+        5  a
+        >>> df.groupby('A').cumcount().sort_index()
+        0    0
+        1    1
+        2    2
+        3    0
+        4    1
+        5    3
+        Name: 0, dtype: int64
+        >>> df.groupby('A').cumcount(ascending=False)
+        0    3
+        1    2
+        2    1
+        3    1
+        4    0
+        5    0
+        Name: 0, dtype: int64
+
+        """
+        ret = self._apply_series_op(
+            lambda sg: sg._kser._cum(
+                F.count, True, part_cols=sg._groupkeys_scols, ascending=ascending),
+            should_resolve=True,
+        )
+        ret -= 1
+        return ret.max(axis=1)
+
     def cummax(self):
         """
         Cumulative max for each group.
--- a/databricks/koalas/missing/groupby.py
+++ b/databricks/koalas/missing/groupby.py
@@ -94,7 +94,6 @@ class MissingPandasLikeSeriesGroupBy(object):
     # Functions
     agg = _unsupported_function("agg")
     aggregate = _unsupported_function("aggregate")
-    cumcount = _unsupported_function("cumcount")
     describe = _unsupported_function("describe")
     get_group = _unsupported_function("get_group")
     median = _unsupported_function("median")
--- a/databricks/koalas/missing/groupby.py
+++ b/databricks/koalas/missing/groupby.py
@@ -57,7 +57,6 @@ class MissingPandasLikeDataFrameGroupBy(object):
 
     # Functions
     boxplot = _unsupported_function("boxplot")
-    cumcount = _unsupported_function("cumcount")
     get_group = _unsupported_function("get_group")
     median = _unsupported_function("median")
     ngroup = _unsupported_function("ngroup")
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -2200,26 +2200,6 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
                 repr(pdf3.transpose().sort_index()), repr(kdf3.transpose().sort_index())
             )
 
-    def _test_cumcount(self, pdf, kdf):
-        for column in pdf.columns:
-            self.assert_eq(pdf.groupby(column).cumcount(), kdf.groupby(column).cumcount())
-            self.assert_eq(
-                pdf.groupby(column).cumcount(ascending=False),
-                kdf.groupby(column).cumcount(ascending=False),
-            )
-            self.assert_eq(
-                pdf.groupby(column).cumcount().sum(), kdf.groupby(column).cumcount().sum()
-            )
-
-    def test_cumcount(self):
-        pdf = pd.DataFrame(
-            [[1, None, 4], [1, 0.1, 3], [1, 20.0, 2], [4, None, 1]],
-            columns=list("ABC"),
-            index=np.random.rand(4),
-        )
-        kdf = ks.from_pandas(pdf)
-        self._test_cumcount(pdf, kdf)
-
     def _test_cummin(self, pdf, kdf):
         self.assert_eq(pdf.cummin(), kdf.cummin())
         self.assert_eq(pdf.cummin(skipna=False), kdf.cummin(skipna=False))
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -2200,6 +2200,26 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
                 repr(pdf3.transpose().sort_index()), repr(kdf3.transpose().sort_index())
             )
 
+    def _test_cumcount(self, pdf, kdf):
+        for column in pdf.columns:
+            self.assert_eq(pdf.groupby(column).cumcount(), kdf.groupby(column).cumcount())
+            self.assert_eq(
+                pdf.groupby(column).cumcount(ascending=False),
+                kdf.groupby(column).cumcount(ascending=False),
+            )
+            self.assert_eq(
+                pdf.groupby(column).cumcount().sum(), kdf.groupby(column).cumcount().sum()
+            )
+
+    def test_cumcount(self):
+        pdf = pd.DataFrame(
+            [[1, None, 4], [1, 0.1, 3], [1, 20.0, 2], [4, None, 1]],
+            columns=list("ABC"),
+            index=np.random.rand(4),
+        )
+        kdf = ks.from_pandas(pdf)
+        self._test_cumcount(pdf, kdf)
+
     def _test_cummin(self, pdf, kdf):
         self.assert_eq(pdf.cummin(), kdf.cummin())
         self.assert_eq(pdf.cummin(skipna=False), kdf.cummin(skipna=False))
--- a/databricks/koalas/tests/test_groupby.py
+++ b/databricks/koalas/tests/test_groupby.py
@@ -852,6 +852,70 @@ class GroupByTest(ReusedSQLTestCase, TestUtils):
             pdf.groupby([("x", "a"), ("x", "b")]).rank().sort_index(),
         )
 
+    def test_cumcount(self):
+        pdf = pd.DataFrame(
+            {"a": [1, 1, 1, 4] * 3, "b": [None, 0.1, 20.0, None] * 3, "c": [4, 3, 2, 1] * 3},
+            index=np.random.rand(4 * 3),
+        )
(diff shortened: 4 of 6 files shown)


HyukjinKwon:
Thanks @tomspur for working on this.  @itholic can you review this?

itholic:
@HyukjinKwon Sure I'm going to gladly review this !

databricks/koalas/groupby.py
```diff
@@ -43,7 +43,8 @@
 )
 from pyspark.sql.functions import PandasUDFType, pandas_udf, Column
 
-from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
+# For running doctests and reference resolution in PyCharm.
+from databricks import koalas as ks
```
itholic, line 47: nit: I'd say I'm just wondering why this should be changed ?

databricks/koalas/groupby.py
```diff
@@ -640,6 +641,54 @@ def diff(self, periods=1):
             lambda sg: sg._kser._diff(periods, part_cols=sg._groupkeys_scols)
         )
 
+    def cumcount(self, ascending=True):
+        """
+        Cumulative count for each group.
+
+        Returns
+        -------
+        Series or DataFrame
```
itholic, line 650: Would you mind matching those docstring to pandas ?

you can print the whole docstring using `help(pd.core.groupby.GroupBy.cumcount)`


```python
"""
    Number each item in each group from 0 to the length of that group - 1.

    Essentially this is equivalent to

    >>> self.apply(lambda x: pd.Series(np.arange(len(x)), x.index))

    Parameters
    ----------
    ascending : bool, default True
        If False, number in reverse, from length of group - 1 to 0.

    Returns
    -------
    Series
        Sequence number of each element within each group.
    ...
"""
```

databricks/koalas/groupby.py
```diff
@@ -2034,12 +2083,16 @@ def nunique(self, dropna=True):
         Name: value1, dtype: int64
         """
         if dropna:
-            stat_function = lambda col: F.countDistinct(col)
+
+            def stat_function(col):
+                return F.countDistinct(col)
```
itholic, line 2088: nit: I think we can leave this as it was before. Is there any special reason in you mind changing this ??

databricks/koalas/groupby.py
```diff
@@ -2034,12 +2083,16 @@ def nunique(self, dropna=True):
...
         else:
-            stat_function = lambda col: (
-                F.countDistinct(col)
-                + F.when(F.count(F.when(col.isNull(), 1).otherwise(None)) >= 1, 1).otherwise(0)
-            )
+
+            def stat_function(col):
+                return F.countDistinct(col) + F.when(
+                    F.count(F.when(col.isNull(), 1).otherwise(None)) >= 1, 1
+                ).otherwise(0)
```
itholic, line 2095: ditto ?

itholic: Oh maybe It could have changed automatically when running `reformatter`, right ?

itholic:
I just submitted a question to pandas repo (https://github.com/pandas-dev/pandas/issues/35682)
Let's just keep this for now until they response to it.

itholic:
Would you mind adding this to `docs/source/reference/groupby.rst` ??

itholic:
Anyway, I'd say think we better return the result as a int64 like pandas rather than float64. ?

```python
>>> pdf
   A     B  C
0  1   NaN  4
1  1   0.1  3
2  1  20.0  2
3  4  10.0  1

>>> pdf.groupby("A").cumcount()
0    0
1    1
2    2
3    0
dtype: int64

>>> kdf.groupby("A").cumcount()
0    0.0
1    1.0
2    2.0
3    0.0
Name: 0, dtype: float64
```

databricks/koalas/groupby.py
tomspur: Yes, all of that was when running the reformatter. I can try to drop these changes, where it was to unrelated places in this change and just correct the issue in the new code, that was added here.

databricks/koalas/groupby.py
tomspur: Sure, can do so. I was more aligning with the docstring from `cummax`, but the one from pandas is also fine

databricks/koalas/groupby.py
tomspur: All those non-related changes should be gone in this PR now

tomspur:
The above changes are now all implemented. I don't quite like the explicit cast to `int64` in the code, but also couldn't find another way to ensure that the NaNs are properly skipped during the calculation.

tomspur:
Unfortunately, the order is not preserved as can be seen in this test (which I wanted to add to the testsuite now):
```

In [10]: pdf = pd.DataFrame([[1, None, 4], [1, 0.1, 3], [1, 20.0, 2], [4, None, 1]],columns=list("ABC"),index=np.random.rand(4))                                                       

In [11]: kdf.groupby(column).cumcount(ascending=False)                                                                                                                                 
Out[11]: 
0.612123    0
0.440823    1
0.718036    2
0.883475    0
Name: 0, dtype: int64

In [12]: pdf.groupby(column).cumcount(ascending=False)                                                                                                                                 
Out[12]: 
0.874863    2
0.809940    1
0.528329    0
0.655884    0
dtype: int64
```

Not sure where this could be ensured in the new `ascending=False` part in `databricks.koalas.series` ..

ueshin:
@itholic For pandas-dev/pandas#35682, `cumcount` doesn't need to care about its type, so it's natural to support with any type.

databricks/koalas/groupby.py
```diff
@@ -640,6 +640,65 @@ def diff(self, periods=1):
...
+        """
+        ret = self._apply_series_op(
+            lambda sg: sg._kser._cum(
+                F.count, True, part_cols=sg._groupkeys_scols, ascending=ascending
+            ),
+            should_resolve=True,
+        )
+        ret -= 1
+        # Cast columns to ``"int64"`` to match `pandas.core.groupby.GroupBy.cumcount`.
+        return ret.max(axis=1).astype('int64')
```
ueshin, line 700: I don't think using `_apply_series_op` is a good idea here.

How about:

```diff
@@ -689,15 +689,9 @@ class GroupBy(object, metaclass=ABCMeta):
         Name: 0, dtype: int64

         """
-        ret = self._apply_series_op(
-            lambda sg: sg._kser._cum(
-                F.count, True, part_cols=sg._groupkeys_scols, ascending=ascending
-            ),
-            should_resolve=True,
-        )
-        ret -= 1
-        # Cast columns to ``"int64"`` to match `pandas.core.groupby.GroupBy.cumcount`.
-        return ret.max(axis=1).astype('int64')
+        ret = self._groupkeys[0]._cum(F.count, True, self._groupkeys_scols, ascending=ascending) - 1
+        internal = ret._internal.resolved_copy
+        return first_series(DataFrame(internal))

     def cummax(self):
         """
```

still need to check the behavior when `self._groupkeys[0]` contains null values.

databricks/koalas/groupby.py
```diff
@@ -640,6 +640,65 @@ def diff(self, periods=1):
             lambda sg: sg._kser._diff(periods, part_cols=sg._groupkeys_scols)
         )
 
+    def cumcount(self, ascending=True):
+        """
+        Number each item in each group from 0 to the length of that group - 1.
+
+        Essentially this is equivalent to
+
+        >>> self.apply(lambda x: pd.Series(np.arange(len(x)), x.index)) # doctest: +SKIP
```
ueshin, line 649: Could you add an expected result even if we skip the test?

ueshin: Why is this skipped, btw?

ueshin: ah, maybe:

```
        .. code-block:: python
            self.apply(lambda x: pd.Series(np.arange(len(x)), x.index))
```

as per pandas doc https://github.com/pandas-dev/pandas/blob/v1.1.0/pandas/core/groupby/groupby.py#L2199-L2256

databricks/koalas/tests/test_dataframe.py
```diff
@@ -2200,6 +2200,26 @@ def test_transpose(self):
                 repr(pdf3.transpose().sort_index()), repr(kdf3.transpose().sort_index())
             )
 
+    def _test_cumcount(self, pdf, kdf):
+        for column in pdf.columns:
+            self.assert_eq(pdf.groupby(column).cumcount(), kdf.groupby(column).cumcount())
+            self.assert_eq(
+                pdf.groupby(column).cumcount(ascending=False),
+                kdf.groupby(column).cumcount(ascending=False),
+            )
```
ueshin, line 2209: Window functions won't guarantee the row order. we can just use `sort_index()`.

ueshin: Btw, these tests must be in `test_groupby` instead of `test_dataframe`.

ueshin: Could you also add tests for `SeriesGroupBy`? Please check the similar tests in `test_groupby`.

tomspur: I moved the tests and further extended it. Unfortunately, two parts are failing, where the groupby includes multiple columns and if one of those contains null values .. Couldn't find out why this is the case.

Do you have a pointer for `SeriesGroupBy`? I seem to only find DataFrame groupby tests there

databricks/koalas/groupby.py
tomspur: This did indeed not work for null values, but I found another possibility with `F.row_count` instead. It even works with the empty array example and matches the pandas implementation now!
By default the returning values seem to be `int32` now. Is that fine as well or would be the cast to `int64` be better to match pandas?

databricks/koalas/tests/test_dataframe.py
ueshin: Something like `kdf.groupby(["b"])["a"].cumcount()` is calling `SeriesGroupBy`.

databricks/koalas/groupby.py
ueshin: What happened when null values with `F.count`?

databricks/koalas/missing/groupby.py
```diff
@@ -57,7 +57,6 @@ class MissingPandasLikeDataFrameGroupBy(object):
 
     # Functions
     boxplot = _unsupported_function("boxplot")
-    cumcount = _unsupported_function("cumcount")
```
ueshin, line 60: We should also remove https://github.com/databricks/koalas/blob/50323ff4720261b1d9db5c62d6f09bb98a1ba6b6/databricks/koalas/missing/groupby.py#L97

databricks/koalas/tests/test_dataframe.py
ueshin: Could you try with `almost=True` for the case containing null values in the result?

itholic:
> @itholic For [pandas-dev/pandas#35682](https://github.com/pandas-dev/pandas/issues/35682), `cumcount` doesn't need to care about its type, so it's natural to support with any type.

Thanks, @ueshin . I'll close that QST with proper comments.

databricks/koalas/groupby.py
tomspur: Null values are explicitely set [in series._cum](https://github.com/databricks/koalas/blob/master/databricks/koalas/series.py#L5057), when avoiding that, and just executing the `.otherwise` part there, the null values are -1 instead of counting upwards. It might work to fill the nulls with a non-existing value in that column and then using count .. this seems to work fine so far:
```
keys = self._groupkeys[0].copy()
keys.fillna(keys.max() + 1, inplace=True)
ret = (
            keys._cum(
                F.count,
                True,
                nocolarg=False,
                part_cols=self._groupkeys_scols,
                ascending=ascending,
            )
            - 1
        )
internal = ret._internal.resolved_copy
return first_series(DataFrame(internal))
```
Would that be an alternative?

databricks/koalas/tests/test_dataframe.py
tomspur: The values are completely different and also `almost=True` does not work. What is wondering me, that already the groupby is yielding different results for koalas vs. pandas:
```
In [14]: kdf.groupby([("x", "a"), ("x", "b")]).first()                                                                                                                                 
Out[14]: 
               y
               c
(x, a) (x, b)   
1      NaN     4
4      NaN     1
1      20.0    2
       0.1     3

In [15]: pdf.groupby([("x", "a"), ("x", "b")]).first()                                                                                                                                 
Out[15]: 
               y
               c
(x, a) (x, b)   
1      0.1     3
       20.0    2
```

databricks/koalas/missing/groupby.py
tomspur: Agreed. `kdf.groupby(["b"])["a"].cumcount()` works also with the `_unsupported_function` above and seems to directly call DataframeGroupBy instead of the SeriesGroupBy ..

databricks/koalas/groupby.py
ueshin: `keys.max()` yields one extra Spark job, also `fillna` needs to change the value according to the data type.
How about:

```py
ret = (
            self._groupkeys[0].spark.transform(lambda _: F.lit(0))._cum(
                F.count,
                True,
                part_cols=self._groupkeys_scols,
                ascending=ascending,
            )
            - 1
        )
```

Btw, we won't need `nocolarg` parameter anymore. Could you revert the change?

databricks/koalas/tests/test_dataframe.py
ueshin: ah, the case that groupkeys contain null values is a known issue #1007.
Could you work without null values in groupkeys here?

databricks/koalas/tests/test_groupby.py
```diff
@@ -852,6 +852,70 @@ def test_rank(self):
...
+
+        self.assert_eq(
+            kdf.groupby("b").cumcount().sort_index(), pdf.groupby("b").cumcount().sort_index()
+        )
+        # TODO: Enable the following test when ks.groupby(dropna=True)
+        # is implemented (see #1007)
+        # self.assert_eq(
+        #    kdf.groupby(["a", "b"]).cumcount().sort_index(),
+        #    pdf.groupby(["a", "b"]).cumcount().sort_index(),
+        # )
```
ueshin, line 870: Shall we use the same test dataset as the other `cumxxx` tests for now? Then we can enable this and one more below. Let's see #1007 for this case.

ueshin:
@itholic Could you take another look? Thanks!

itholic:
@ueshin My pleasure. Let me check this weekend.

databricks/koalas/tests/test_groupby.py
ueshin: @tomspur Could you update the tests?

ueshin:
@tomspur Now that this is good to go, pending @itholic's another look.
Could you update the PR description to describe the changes here?

databricks/koalas/tests/test_groupby.py
tomspur: I just changed above the columns from `["a", "b"]` to `["a", "c"]`, which does not contain Nones to have some other and keep the comments to remember to enable testing Nones at least partially.

tomspur: Used the same dataset like the other tests for now. Do you want to add a None to them once #1007 is implemented?

ueshin: Yes, we should modify or add tests which include `None`s in groupkeys at #1007.

tomspur:
I changed the description and also mentioned the current limitation for the `groupby` with nulls in it

databricks/koalas/tests/test_groupby.py
```diff
@@ -852,6 +852,62 @@ def test_rank(self):
             pdf.groupby([("x", "a"), ("x", "b")]).rank().sort_index(),
         )
 
+    def test_cumcount(self):
```
itholic, line 855: Could we also have a tests for `ascending=False` ?

You could just use the below

```python
    def test_cumcount(self):
        pdf = pd.DataFrame(
            {
                "a": [1, 2, 3, 4, 5, 6] * 3,
                "b": [1, 1, 2, 3, 5, 8] * 3,
                "c": [1, 4, 9, 16, 25, 36] * 3,
            },
            index=np.random.rand(6 * 3),
        )
        kdf = ks.from_pandas(pdf)

        self.assert_eq(
            kdf.groupby("b").cumcount().sort_index(), pdf.groupby("b").cumcount().sort_index()
        )
        self.assert_eq(
            kdf.groupby(["a", "b"]).cumcount().sort_index(),
            pdf.groupby(["a", "b"]).cumcount().sort_index(),
        )
        self.assert_eq(
            kdf.groupby(["b"])["a"].cumcount().sort_index(),
            pdf.groupby(["b"])["a"].cumcount().sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby(["b"])[["a", "c"]].cumcount().sort_index(),
            pdf.groupby(["b"])[["a", "c"]].cumcount().sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby(kdf.b // 5).cumcount().sort_index(),
            pdf.groupby(pdf.b // 5).cumcount().sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby(kdf.b // 5)["a"].cumcount().sort_index(),
            pdf.groupby(pdf.b // 5)["a"].cumcount().sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby("b").cumcount().sum(), pdf.groupby("b").cumcount().sum(),
        )

        # ascending = False
        self.assert_eq(
            kdf.groupby("b").cumcount(ascending=False).sort_index(), pdf.groupby("b").cumcount(ascending=False).sort_index()
        )
        self.assert_eq(
            kdf.groupby(["a", "b"]).cumcount(ascending=False).sort_index(),
            pdf.groupby(["a", "b"]).cumcount(ascending=False).sort_index(),
        )
        self.assert_eq(
            kdf.groupby(["b"])["a"].cumcount(ascending=False).sort_index(),
            pdf.groupby(["b"])["a"].cumcount(ascending=False).sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby(["b"])[["a", "c"]].cumcount(ascending=False).sort_index(),
            pdf.groupby(["b"])[["a", "c"]].cumcount(ascending=False).sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby(kdf.b // 5).cumcount(ascending=False).sort_index(),
            pdf.groupby(pdf.b // 5).cumcount(ascending=False).sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby(kdf.b // 5)["a"].cumcount(ascending=False).sort_index(),
            pdf.groupby(pdf.b // 5)["a"].cumcount(ascending=False).sort_index(),
            almost=True,
        )
        self.assert_eq(
            kdf.groupby("b").cumcount(ascending=False).sum(), pdf.groupby("b").cumcount(ascending=False).sum(),
        )

        # multi-index columns
        columns = pd.MultiIndex.from_tuples([("x", "a"), ("x", "b"), ("y", "c")])
        pdf.columns = columns
        kdf.columns = columns

        self.assert_eq(
            kdf.groupby(("x", "b")).cumcount().sort_index(),
            pdf.groupby(("x", "b")).cumcount().sort_index(),
        )
        self.assert_eq(
            kdf.groupby([("x", "a"), ("x", "b")]).cumcount().sort_index(),
            pdf.groupby([("x", "a"), ("x", "b")]).cumcount().sort_index(),
        )

        # ascending = False
        self.assert_eq(
            kdf.groupby(("x", "b")).cumcount(ascending=False).sort_index(),
            pdf.groupby(("x", "b")).cumcount(ascending=False).sort_index(),
        )
        self.assert_eq(
            kdf.groupby([("x", "a"), ("x", "b")]).cumcount(ascending=False).sort_index(),
            pdf.groupby([("x", "a"), ("x", "b")]).cumcount(ascending=False).sort_index(),
        )
```

itholic:
Otherwise, LGTM

databricks/koalas/tests/test_groupby.py
ueshin: Sounds good, but shall we use a loop for ascending?

```py
for ascending in [True, False]:
    ...
```

itholic: Looks better 👍

itholic: And maybe we could test for `[True, False, 0, 1, None]` - Not sure maybe It's too much though -
I fixed the example code in the above comment.

ueshin:
I'd merge this now and will submit a follow-up PR to address the comment.
@tomspur Thanks for working on this!

tomspur:
Thank you for the throughout review and merging this!

PR #1704, opened 2020-08-11
Fix first_valid_index() for Empty object

itholic:
`Series.first_valid_index()` and `DataFrame.first_valid_index()` raise Exception for Empty set - but it supposed to return None, not raise Error. -

```python
>>> ks.Series([]).first_valid_index()  # It should return `None`.
...
TypeError: 'NoneType' object is not subscriptable
```

This PR fixed this and added related tests.

```python
>>> ks.Series([]).first_valid_index() is None
True
```


ueshin:
Thanks! merging.

PR #1707, opened 2020-08-11
Suggestion to add downloads badge

itholic:
Now we can easily check the download stats via "downloads" badge provided by [PePy](https://pepy.tech/project/Koalas).

스크린샷 2020-08-12 오전 8 45 19

How does it look ?


itholic:
FYI: there are more several options we also could choose. (Now we're using the top one)

스크린샷 2020-08-12 오전 8 50 56

PR #1708, opened 2020-08-12
Fix GroupBy.cum~ for matching with pandas' behavior

itholic:
In pandas, `GroupBy.cum~` raise `DataError` - A custom exception of pandas - when the `DataFrame` has not numeric type columns except group keys.

```python
>>> pdf = pd.DataFrame([[1, 'a'], [2, 'b'], [3, 'c'], [4, 'd'], [5, 'e'], [6, 'f']], columns=['A', 'B'])
>>> pdf
   A  B
0  1  a
1  2  b
2  3  c
3  4  d
4  5  e
5  6  f

>>> pdf.groupby("A").cumsum()
...
pandas.core.base.DataError: No numeric types to aggregate
```

But in Koalas, It has been returning empty `DataFrame` rather than raise any Exceptions.

```python
>>> kdf.groupby("A").cumsum()
    B
5 NaN
4 NaN
0 NaN
2 NaN
1 NaN
3 NaN
```

So, added custom exception `DataError` and managed the logic for matching the behavior with pandas'.

### DIFF
--- a/databricks/koalas/exceptions.py
+++ b/databricks/koalas/exceptions.py
@@ -19,11 +19,7 @@ Exceptions/Errors used in Koalas.
 """
 
 
-class GroupByError(Exception):
-    pass
-
-
-class DataError(GroupByError):
+class DataError(Exception):
     pass
 
 
--- a/databricks/koalas/exceptions.py
+++ b/databricks/koalas/exceptions.py
@@ -19,6 +19,14 @@ Exceptions/Errors used in Koalas.
 """
 
 
+class GroupByError(Exception):
+    pass
+
+
+class DataError(GroupByError):
+    pass
+
+
 class SparkPandasIndexingError(Exception):
     pass
 
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -68,6 +68,7 @@ from databricks.koalas.utils import (
     verify_temp_column_name,
 )
 from databricks.koalas.window import RollingGroupby, ExpandingGroupby
+from databricks.koalas.exceptions import DataError
 
 # to keep it the same as pandas
 NamedAgg = namedtuple("NamedAgg", ["column", "aggfunc"])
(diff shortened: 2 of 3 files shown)


itholic:
Seems need to more work to cover the whole case. Please give some time.

databricks/koalas/groupby.py
```diff
@@ -2328,6 +2334,8 @@ def _apply_series_op(self, op, should_resolve: bool = False):
         applied = []
         for column in self._agg_columns:
             applied.append(op(column.groupby(self._groupkeys)))
+        if not applied and (op.__name__).startswith("cum"):
+            raise DataError("No numeric types to aggregate")
```
ueshin, line 2338: Please don't rely on the function name but add an explicit argument, as `Frame._reduce_for_stat_function` has `numeric_only=True`.

itholic: @ueshin Thanks for the advice. I'll manage it 😃

databricks/koalas/exceptions.py
```diff
@@ -19,6 +19,14 @@
 """
 
 
+class GroupByError(Exception):
+    pass
+
+
+class DataError(GroupByError):
```
HyukjinKwon, line 26: Does pandas throw an exception like this?

itholic: Yep, just followed the pandas'.

https://github.com/pandas-dev/pandas/blob/master/pandas/core/base.py#L124-L125

itholic: Yes, but seems this behavior has changed. I'll address it.

HyukjinKwon (approved):

databricks/koalas/groupby.py
```diff
@@ -90,7 +91,7 @@ def _agg_columns_scols(self):
         return [s.spark.column for s in self._agg_columns]
 
     @abstractmethod
-    def _apply_series_op(self, op, should_resolve: bool = False):
+    def _apply_series_op(self, op, should_resolve: bool = False, is_cum_func: bool = False):
```
ueshin, line 94: nit: I'm feeling a little weird about the naming `is_cum_func`. We could name it `numeric_only` for more generality. I don't think we should fix it at once, but let's rename it when we have another chance.

itholic: Ah, I think I can revise right away before we forget 😂.
Will quickly fix it. Thanks for the suggestion!

PR #1718, opened 2020-08-20
Use Series._column_label instead of ._internal.column_labels[0].

ueshin:
Refactors to use `Series._column_label` instead of `._internal.column_labels[0]`.


ueshin:
Let me merge this now as this is a trivial change. Please feel free to leave comments if any.

HyukjinKwon:
Nice!

PR #1723, opened 2020-08-24
DataFrame.reindex(fill_value) does not fill existing NaN values

### LINKED ISSUE #1717: DataFrame reindex(fill_value)  fails when containing NaNs
In pandas we have :

```python
>>> pdf = pd.DataFrame({"numbers": [1.0, 2.0, None]}, index=['A', 'B', 'C'])
>>> pdf
   numbers
A      1.0
B      2.0
C      NaN
>>> pdf.reindex(index=["A", "C", "E"], fill_value=0).sort_index()
   numbers
A      1.0
C      NaN
E      0.0
```

But with koalas :

```python
>>> kdf = ks.from_pandas(pdf)
>>> kdf.reindex(index=["A", "C", "E"], fill_value=0).sort_index()
   numbers
A      1.0
C      0.0
E      0.0
```

All NaNs are replaced why only those inserted by new index values should be replaced.

Also, I think the parameter `copy` introduces an extra copy around line 7808.
One should copy here only if no copy was performed before.

I am working on this issue.

LucasG0:
This PR would close #1717.
I used float as fill value in tests in order to avoid clash bewteen int64 (from pandas) and int32 (from spark).

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7786,22 +7786,18 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         df = self
 
         if index is not None:
-            df = df._reindex_index(index)
+            df = df._reindex_index(index, fill_value)
 
         if columns is not None:
-            df = df._reindex_columns(columns)
-
-        # Process missing values.
-        if fill_value is not None:
-            df = df.fillna(fill_value)
+            df = df._reindex_columns(columns, fill_value)
 
         # Copy
-        if copy:
+        if copy and df is self:
             return df.copy()
         else:
             return df
 
-    def _reindex_index(self, index):
+    def _reindex_index(self, index, fill_value):
         # When axis is index, we can mimic pandas' by a right outer join.
         assert (
             len(self._internal.index_spark_column_names) <= 1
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -7819,15 +7815,20 @@ def _reindex_index(self, index):
...
 
-        joined_df = self._internal.resolved_copy.spark_frame.drop(NATURAL_ORDER_COLUMN_NAME).join(
-            labels, on=index_column, how="right"
-        )
-        internal = self._internal.with_new_sdf(joined_df)
+        if fill_value is not None:
+            unmatched_labels = labels.subtract(frame.select(self._internal.spark_columns[0]))
+            filled = frame.join(unmatched_labels, on=index_column, how="right").fillna(fill_value)
+            existing = frame.join(labels, on=index_column, how="inner")
+            joined_df = existing.union(filled)
```
ueshin, line 7824: I'm afraid there are too many joins here. How about:

```py
            frame_index_column = verify_temp_column_name(frame, "__frame_index_column__")
            frame = frame.withColumnRenamed(index_column, frame_index_column)

            temp_fill_value = verify_temp_column_name(frame, "__fill_value__")
            labels = labels.withColumn(temp_fill_value, F.lit(fill_value))

            frame_index_scol = scol_for(frame, frame_index_column)
            labels_index_scol = scol_for(labels, index_column)

            joined_df = frame.join(labels, on=[frame_index_scol == labels_index_scol], how="right")
            joined_df = joined_df.select(
                index_column,
                *[
                    F.when(
                        frame_index_scol.isNull() & labels_index_scol.isNotNull(),
                        scol_for(joined_df, temp_fill_value),
                    )
                    .otherwise(scol_for(joined_df, col))
                    .alias(col)
                    for col in self._internal.data_spark_column_names
                ]
            )
```

LucasG0: Yes, nice way to do it ! It will be more efficient.
I update, thanks

databricks/koalas/tests/test_dataframe.py
```diff
@@ -2375,8 +2375,8 @@ def test_drop_duplicates(self):
 
     def test_reindex(self):
         index = ["A", "B", "C", "D", "E"]
-        pdf = pd.DataFrame({"numbers": [1.0, 2.0, 3.0, 4.0, 5.0]}, index=index)
-        kdf = ks.DataFrame({"numbers": [1.0, 2.0, 3.0, 4.0, 5.0]}, index=index)
+        pdf = pd.DataFrame({"numbers": [1.0, 2.0, 3.0, 4.0, None]}, index=index)
+        kdf = ks.DataFrame({"numbers": [1.0, 2.0, 3.0, 4.0, None]}, index=index)
```
itholic, line 2379: nit: Would you mind use `ks.from_pandas(pdf)` so that we can reuse the existing `pdf` internally ?

itholic (approved):
Otherwise, LGTM.

databricks/koalas/tests/test_dataframe.py
LucasG0: Yes, done

itholic:
Thanks, @LucasG0 ! 
merging :)

PR #1728, opened 2020-08-26
Cleanup to_pandas() from tests and deprecate toPandas()

itholic:
1. Use `ks.from_pandas` instead of `to_pandas()` to avoid unnecessary copy internally.
2. Use `to_pandas()` instead of `toPandas()` since we've explicitly exposed the former in our official document.
3. Deprecate `toPandas()`

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4230,7 +4230,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         warnings.warn(
             "DataFrame.toPandas is deprecated as of DataFrame.to_pandas. "
             "Please use the API instead.",
-            DeprecationWarning,
+            FutureWarning,
         )
         return self.to_pandas()
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4226,7 +4226,13 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         return self._internal.to_pandas_frame.copy()
 
     # Alias to maintain backward compatibility with Spark
-    toPandas = to_pandas
+    def toPandas(self):
+        warnings.warn(
+            "DataFrame.toPandas is deprecated as of DataFrame.to_pandas. "
+            "Please use the API instead.",
+            FutureWarning,
+        )
+        return self.to_pandas()
 
     def assign(self, **kwargs):
         """
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4230,9 +4230,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         warnings.warn(
             "DataFrame.toPandas is deprecated as of DataFrame.to_pandas. "
             "Please use the API instead.",
-            FutureWarning,
+            DeprecationWarning,
         )
         return self.to_pandas()
+
+    toPandas.__doc__ = to_pandas.__doc__
 
     def assign(self, **kwargs):
         """
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -1017,9 +1017,9 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
             "animal": ["cat", "dog", "bat", "penguin"],
             "locomotion": ["walks", "walks", "flies", "walks"],
         }
-        kdf = ks.DataFrame(data=d)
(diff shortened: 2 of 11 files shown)


databricks/koalas/tests/test_dataframe.py
```diff
@@ -1103,7 +1103,7 @@ def test_to_numpy(self):
 
     def test_to_pandas(self):
         pdf, kdf = self.df_pair
-        self.assert_eq(kdf.toPandas(), pdf)
+        self.assert_eq(kdf.to_pandas(), pdf)
```
ueshin, line 1106: Shall we keep this one to check `toPandas` still works.

ueshin: Btw, shall we mark `toPandas()` as deprecated?

databricks/koalas/tests/test_dataframe_spark_io.py
```diff
@@ -89,8 +89,8 @@ def check(columns, expected):
             expected_idx = expected.set_index("bhello")[["f", "i32", "i64"]]
             actual_idx = ks.read_parquet(tmp, index_col="bhello")[["f", "i32", "i64"]]
             self.assert_eq(
-                actual_idx.sort_values(by="f").to_spark().toPandas(),
-                expected_idx.sort_values(by="f").to_spark().toPandas(),
+                actual_idx.sort_values(by="f").to_spark().to_pandas(),
+                expected_idx.sort_values(by="f").to_spark().to_pandas(),
```
ueshin, line 93: We can't change these since Spark DataFrame doesn't have `to_pandas()`.

ueshin: ditto for the following changes.

ueshin: Why are we using `to_spark()` here, btw ..? @HyukjinKwon do you remember the reason?

databricks/koalas/tests/test_dataframe.py
itholic: Sounds good. Let me mark it in this PR.
Thanks for the suggestion! 👍

databricks/koalas/tests/test_dataframe_spark_io.py
itholic: Oh, sorry my bad. I changed whole `toPandas()` to `to_pandas()` at once with grep. 😅

databricks/koalas/frame.py
```diff
@@ -4226,7 +4226,13 @@ def to_pandas(self):
         return self._internal.to_pandas_frame.copy()
 
     # Alias to maintain backward compatibility with Spark
-    toPandas = to_pandas
+    def toPandas(self):
+        warnings.warn(
+            "DataFrame.toPandas is deprecated as of DataFrame.to_pandas. "
+            "Please use the API instead.",
+            FutureWarning,
```
ueshin, line 4233: `DeprecationWarning`?

ueshin: Also `toPandas.__doc__ = to_pandas.__doc__` is needed?

itholic: > `DeprecationWarning`?

Oh, then maybe should we fix the other `FutureWarning`s actually related with deprecation to the `DeprecationWarning` ??

databricks/koalas/tests/test_dataframe_spark_io.py
HyukjinKwon: I don't remember .. I think we can just remove `to_spark` call.

databricks/koalas/frame.py
HyukjinKwon: Oh, I used `FutureWarning` instead of `DeprecationWarning`. The semantic of both were changed in newer Python versions. The former implies a deprecation to end users (https://docs.python.org/3/library/exceptions.html#FutureWarning). The latter is for developers (https://docs.python.org/3/library/exceptions.html#DeprecationWarning). `DeprecationWarning` is ignored by default.

ueshin: ah, I see, then `FutureWarning` would be better.

itholic: Thanks for the comments !

HyukjinKwon (approved):

databricks/koalas/tests/test_dataframe_spark_io.py
itholic: It's for dropping the `index_col` since the order of indices could be different.

PR #1730, opened 2020-08-26
Introduce `check_exact` parameter for assertions.

ueshin:
Introduces `check_exact` parameter for assertions to use pandas testing utils for more cases instead of `almost=True`.

### DIFF
--- a/databricks/koalas/testing/utils.py
+++ b/databricks/koalas/testing/utils.py
@@ -131,7 +131,7 @@ class ReusedSQLTestCase(unittest.TestCase, SQLTestUtils):
         # Please see databricks/koalas/conftest.py.
         pass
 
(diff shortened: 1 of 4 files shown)


databricks/koalas/testing/utils.py
```diff
@@ -131,15 +131,15 @@ def tearDownClass(cls):
         # Please see databricks/koalas/conftest.py.
         pass
 
-    def assertPandasEqual(self, left, right):
+    def assertPandasEqual(self, left, right, less_precise=False):
```
HyukjinKwon, line 134: What about we just name it `check_exact`?

ueshin: I thought about that first but I felt a little weird. I'm fine with the name if you prefer.

HyukjinKwon (approved):

PR #1722, opened 2020-08-22
Use pandas testing utils for exact cases.

ueshin:
Changes the testing infrastructure for exact cases to pandas' testing library.

PR #1737, opened 2020-08-30
Implement Series.reindex

### LINKED ISSUE #881: Series.reindex
Implement `Series.reindex`.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.reindex.html

LucasG0:
This PR would close  #881.
It includes moving the method `DataFrame._reindex_index` to the generic `Frame` class.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1539,6 +1539,109 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         else:
             return first_series(kdf)
 
+    def reindex(self, index: Optional[Any] = None, fill_value: Optional[Any] = None,) -> "Series":
+        """
+        Conform Series to new index with optional filling logic, placing
+        NA/NaN in locations having no value in the previous index. A new object
+        is produced.
+
+        Parameters
+        ----------
+        index: array-like, optional
+            New labels / index to conform to, should be specified using keywords.
+            Preferably an Index object to avoid duplicating data
+        fill_value : scalar, default np.NaN
+            Value to use for missing values. Defaults to NaN, but can be any
+            "compatible" value.
+
+        Returns
+        -------
+        Series with changed index.
+
+        See Also
+        --------
+        Series.reset_index : Remove row labels or move them to new columns.
+
+        Examples
+        --------
+
+        Create a series with some fictional data.
+
+        >>> index = ['Firefox', 'Chrome', 'Safari', 'IE10', 'Konqueror']
+        >>> ser = ks.Series([200, 200, 404, 404, 301],
+        ...                 index=index, name='http_status')
+        >>> ser
+        Firefox      200
+        Chrome       200
+        Safari       404
+        IE10         404
+        Konqueror    301
+        Name: http_status, dtype: int64
+
+        Create a new index and reindex the Series. By default
+        values in the new index that do not have corresponding
+        records in the Series are assigned ``NaN``.
+
+        >>> new_index= ['Safari', 'Iceweasel', 'Comodo Dragon', 'IE10',
+        ...             'Chrome']
+        >>> ser.reindex(new_index).sort_index()
+        Chrome           200.0
+        Comodo Dragon      NaN
+        IE10             404.0
+        Iceweasel          NaN
+        Safari           404.0
+        Name: http_status, dtype: float64
+
+        We can fill in the missing values by passing a value to
+        the keyword ``fill_value``.
+
+        >>> ser.reindex(new_index, fill_value=0).sort_index()
+        Chrome           200
+        Comodo Dragon      0
+        IE10             404
+        Iceweasel          0
+        Safari           404
+        Name: http_status, dtype: int64
+
+        To further illustrate the filling functionality in
+        ``reindex``, we will create a Series with a
+        monotonically increasing index (for example, a sequence
+        of dates).
+
+        >>> date_index = pd.date_range('1/1/2010', periods=6, freq='D')
+        >>> ser2 = ks.Series([100, 101, np.nan, 100, 89, 88],
+        ...                  name='prices', index=date_index)
+        >>> ser2.sort_index()
+        2010-01-01    100.0
+        2010-01-02    101.0
+        2010-01-03      NaN
+        2010-01-04    100.0
+        2010-01-05     89.0
+        2010-01-06     88.0
+        Name: prices, dtype: float64
+
+        Suppose we decide to expand the series to cover a wider
+        date range.
+
+        >>> date_index2 = pd.date_range('12/29/2009', periods=10, freq='D')
+        >>> ser2.reindex(date_index2).sort_index()
+        2009-12-29      NaN
+        2009-12-30      NaN
+        2009-12-31      NaN
+        2010-01-01    100.0
+        2010-01-02    101.0
+        2010-01-03      NaN
+        2010-01-04    100.0
+        2010-01-05     89.0
+        2010-01-06     88.0
+        2010-01-07      NaN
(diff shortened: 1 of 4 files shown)


docs/source/reference/series.rst
```diff
@@ -414,4 +415,3 @@ These can be accessed by ``Series.koalas.<function/property>``.
    :toctree: api/
 
    Series.koalas.transform_batch
-
```
itholic, line 417: Could you please revert this unnecessary fix?

databricks/koalas/tests/test_series.py
```diff
@@ -232,6 +232,33 @@ def test_drop_duplicates(self):
...
+        index = ["A", "B", "C", "D", "E"]
+        pser = pd.Series([1.0, 2.0, 3.0, 4.0, None], index=index, name="x")
+        kser = ks.from_pandas(pser)
+
+        self.assert_eq(pser, kser, almost=True)
+
+        self.assert_eq(
+            pser.reindex(["A", "B"]).sort_index(),
+            kser.reindex(["A", "B"]).sort_index(),
+            almost=True,
```
itholic, line 245: I think we can remove `almost` now since #1722   😸

databricks/koalas/tests/test_series.py
```diff
@@ -232,6 +232,33 @@ def test_drop_duplicates(self):
...
+        )
+
+        self.assert_eq(
+            pser.reindex(["A", "B", "2", "3"]).sort_index(),
+            kser.reindex(["A", "B", "2", "3"]).sort_index(),
+            almost=True,
```
itholic, line 251: ditto

databricks/koalas/tests/test_series.py
```diff
@@ -232,6 +232,33 @@ def test_drop_duplicates(self):
...
+        )
+
+        self.assert_eq(
+            pser.reindex(["A", "E", "2"], fill_value=0).sort_index(),
+            kser.reindex(["A", "E", "2"], fill_value=0).sort_index(),
+            almost=True,
```
itholic, line 257: ditto

databricks/koalas/generic.py
```diff
@@ -2459,6 +2460,57 @@ def ffill(self, axis=None, inplace=False, limit=None):
...
+        if isinstance(self, ks.DataFrame):
+            return df
+        elif isinstance(self, ks.Series):
+            from databricks.koalas.series import first_series
+
+            return first_series(df)
+        else:
+            raise TypeError(
+                "Only DataFrame or Series can be reindexed; however, " "got [%s]" % (self,)
+            )
```
itholic, line 2512: I think we don't need to check this ?

databricks/koalas/generic.py
```diff
@@ -2459,6 +2460,57 @@ def ffill(self, axis=None, inplace=False, limit=None):
...
+            joined_df = frame.join(labels, on=index_column, how="right")
+
+        internal = self._internal.with_new_sdf(joined_df)
+        df = ks.DataFrame(internal)
+        if isinstance(self, ks.DataFrame):
+            return df
+        elif isinstance(self, ks.Series):
+            from databricks.koalas.series import first_series
+
+            return first_series(df)
```
itholic, line 2508: I'd say maybe we better implement `DataFrame.reindex` and `Series.reindex` separately if we have to check the type here in the `generic.py` ?

itholic:
Basically, the logic seems fine to me.
But I think we better reposition of some codes. It's looks a bit of complicated.

itholic:
Anyway, thanks for the work on this, @LucasG0 :)

databricks/koalas/generic.py
LucasG0: Alright, actually I was not sure either. 
I saw this pattern in `Frame.groupby` e.g. and reused it.

databricks/koalas/tests/test_series.py
LucasG0: Thanks I did not notice the change !

LucasG0:
Thanks for reviewing !
Ideally we could avoid code duplicate with `_reindex_index` generic. I submit a new version without type checking in which `Series` constructor handles `InternalFrame`. I am interested in your opinion with this new way.
If it still looks complicated, I will probably just separate `DataFrame.reindex` and `Series.reindex`. :)

itholic:
Sorry for the late, @LucasG0 .
I'm not sure it's good enough reason to modify the constructor at this point.
 @ueshin @HyukjinKwon , WDYT ??

databricks/koalas/generic.py
```diff
@@ -2459,6 +2465,47 @@ def ffill(self, axis=None, inplace=False, limit=None):
...
+                    .otherwise(scol_for(joined_df, col))
+                    .alias(col)
+                    for col in self._internal.data_spark_column_names
+                ]
+            )
+        else:
+            joined_df = frame.join(labels, on=index_column, how="right")
+
+        internal = self._internal.with_new_sdf(joined_df)
+        return self._constructor(internal)
```
HyukjinKwon, line 2507: Can we just call `self.__class__(internal)`?

databricks/koalas/series.py
```diff
@@ -1540,6 +1547,129 @@ def drop_duplicates(self, keep="first", inplace=False):
...
+        2010-01-01    100.0
+        2010-01-02    101.0
+        2010-01-03      NaN
+        2010-01-04    100.0
+        2010-01-05     89.0
+        2010-01-06     88.0
+        2010-01-07      NaN
+        Name: prices, dtype: float64
+        """
+        if index is not None and not is_list_like(index):
```
HyukjinKwon, line 1656: Can we do something like:

```python
if index is not None:
    return first_series(self.to_frame().reindex(...)).rename(self.name)
```

instead? Looks like `DataFrame._reindex_index` will create a new DataFrame anyway and seems pandas doesn't support in-place update anyway so I think it's easier to do this way.

LucasG0: In deed it's easier !
I updated and then removed `copy` parameter.

databricks/koalas/generic.py
LucasG0: No need anymore as discussed below otherwise yes.

HyukjinKwon (approved):

databricks/koalas/series.py
```diff
@@ -1539,6 +1539,107 @@ def drop_duplicates(self, keep="first", inplace=False):
...
+        Firefox      200
+        Chrome       200
+        Safari       404
+        IE10         404
+        Konqueror    301
+        Name: http_status, dtype: int64
+
+        Create a new index and reindex the dataframe. By default
+        values in the new index that do not have corresponding
+        records in the dataframe are assigned ``NaN``.
```
itholic, line 1583: dataframe -> Series ?

itholic, line 1583: dataframe -> Series ?

databricks/koalas/series.py
```diff
@@ -1539,6 +1539,107 @@ def drop_duplicates(self, keep="first", inplace=False):
...
+        2010-01-01    100.0
+        2010-01-02    101.0
+        2010-01-03      NaN
+        2010-01-04    100.0
+        2010-01-05     89.0
+        2010-01-06     88.0
+        Name: prices, dtype: float64
+
+        Suppose we decide to expand the dataframe to cover a wider
+        date range.
```
itholic, line 1624: ditto ?

itholic (approved):
Otherwise, LGTM

itholic:
Thanks, @LucasG0 ! Merging :)

LucasG0:
Thanks. :)

PR #1744, opened 2020-09-02
Implemented item for Index & MultiIndex

itholic:
Since `Index.item()` is un-deprecated from pandas 1.0.0 (https://github.com/pandas-dev/pandas/issues/29250), this PR proposes `Index.item()` and `MultiIndex.item()`.

```python
>>> kidx = ks.Index([10])
>>> kidx.item()
10

>>> kmidx = ks.MultiIndex.from_tuples([('a', 'x')])
>>> kmidx.item()
('a', 'x')
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -2825,9 +2825,8 @@ class MultiIndex(Index):
         >>> kmidx.item()
         ('a', 'x')
         """
-        result = self._internal.spark_frame.head(2)
-        if len(result) == 1:
-            return tuple(result[0][nlevel] for nlevel in range(self.nlevels))
+        if len(self._kdf.head(2)) == 1:
+            return self.to_pandas().item()
         raise ValueError("can only convert an array of size 1 to a Python scalar")
 
     @property
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1980,15 +1980,10 @@ class Index(IndexOpsMixin):
         >>> kidx = ks.Index([10])
         >>> kidx.item()
         10
-
-        Support MultiIndex
-
-        >>> kmidx = ks.MultiIndex.from_tuples([('a', 'x')])
-        >>> kmidx.item()
-        ('a', 'x')
         """
-        if len(self._kdf.head(2)) == 1:
-            return self.to_pandas().item()
+        result = self._internal.spark_frame.head(2)
+        if len(result) == 1:
+            return result[0][0]
         raise ValueError("can only convert an array of size 1 to a Python scalar")
 
     def __getattr__(self, item: str) -> Any:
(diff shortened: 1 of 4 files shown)


databricks/koalas/indexes.py
```diff
@@ -1961,6 +1961,31 @@ def holds_integer(self):
...
+        ValueError
+            If the data is not length-1.
+
+        Examples
+        --------
+        >>> kidx = ks.Index([10])
+        >>> kidx.item()
+        10
+        """
+        result = self._internal.spark_frame.head(2)
```
HyukjinKwon, line 1984: I think you can call `self.head(2).to_pandas` instead of `head`.

HyukjinKwon: or `self.to_series.item()`

itholic: Okay, `self.to_series().item()` seems fine because `Index` has no `head`

Thanks!!

databricks/koalas/indexes.py
```diff
@@ -2783,6 +2805,31 @@ def get_level_values(self, level):
...
+        ValueError
+            If the data is not length-1.
+
+        Examples
+        --------
+        >>> kmidx = ks.MultiIndex.from_tuples([('a', 'x')])
+        >>> kmidx.item()
+        ('a', 'x')
+        """
+        result = self._internal.spark_frame.head(2)
```
HyukjinKwon, line 2828: I think we'll have to use `to_pandas`. For example, I think it will return a different type of instance when you use timestamp. `datetime.datetime` from PySpark vs `datetime64[ns]` from pandas IIRC.

itholic: Oh yeah, I missed it. Thanks for the catch out!

databricks/koalas/indexes.py
```diff
@@ -2783,6 +2805,30 @@ def get_level_values(self, level):
...
+        ValueError
+            If the data is not length-1.
+
+        Examples
+        --------
+        >>> kmidx = ks.MultiIndex.from_tuples([('a', 'x')])
+        >>> kmidx.item()
+        ('a', 'x')
+        """
+        if len(self._kdf.head(2)) == 1:
```
HyukjinKwon, line 2828: Can you avoid executing jobs twice? calculating the length and collecting to pandas DataFrame.

itholic: Okay! Thanks :)

databricks/koalas/indexes.py
```diff
@@ -2783,6 +2805,28 @@ def get_level_values(self, level):
...
+        ValueError
+            If the data is not length-1.
+
+        Examples
+        --------
+        >>> kmidx = ks.MultiIndex.from_tuples([('a', 'x')])
+        >>> kmidx.item()
+        ('a', 'x')
+        """
+        return self._kdf.head(2).to_pandas().index.item()
```
HyukjinKwon, line 2828:
```suggestion
        return self._kdf.head(2)._to_internal_pandas().index.item()
```

HyukjinKwon (approved):

databricks/koalas/indexes.py
ueshin: I didn't know the "suggestion" block. 😃

itholic: Convenient !

PR #1748, opened 2020-09-03
Remove most of repr comparison.

ueshin:
Removes most of `repr` comparison as now that we don't need to use it for most cases.


itholic:
Cool. Is it Thanks to #1722 ?

itholic (approved):

HyukjinKwon (approved):
Nice!

PR #1752, opened 2020-09-08
Set a flake8 rule for missing imports.

ueshin:
Let us take advantage of a flake8 rule `F405` to detect missing imports.

### DIFF
--- a/databricks/koalas/typedef/__init__.py
+++ b/databricks/koalas/typedef/__init__.py
@@ -14,6 +14,4 @@
 # limitations under the License.
 #
 
-from databricks.koalas.typedef.typehints import *
-
-__all__ = ["as_spark_type", "infer_pd_series_spark_type"]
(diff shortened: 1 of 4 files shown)


databricks/koalas/typedef/__init__.py
```diff
@@ -14,6 +14,4 @@
 # limitations under the License.
 #
 
-from databricks.koalas.typedef.typehints import *
-
-__all__ = ["as_spark_type", "infer_pd_series_spark_type"]
```
ueshin, line 19: Is this public APIs? @HyukjinKwon

HyukjinKwon (approved):

databricks/koalas/typedef/__init__.py
HyukjinKwon: I think it's not an API :).

PR #1754, opened 2020-09-09
Deprecate Series.alias.

ueshin:
`Series.alias` doesn't exist in pandas but was added by a historical reason. We should deprecate it now.


itholic (approved):
LGTM

HyukjinKwon:
Nice!

PR #1755, opened 2020-09-09
Fix tests to avoid resetting "compute.default_index_type" and related test failures.

ueshin:
`DefaultIndexTest`s reset the config `compute.default_index_type`, so the following tests are not working under the default index type given by the test environment.

Then found out some tests failed and fixed them (https://github.com/databricks/koalas/runs/1093753632):

- `GroupByTest.test_apply_with_side_effect`
- `NamespaceTest.test_concat_index_axis`

### DIFF
--- a/databricks/conftest.py
+++ b/databricks/conftest.py
@@ -96,7 +96,7 @@ def add_db(doctest_namespace):
     doctest_namespace["db"] = db_name
 
 
-@pytest.fixture(autouse=os.getenv("KOALAS_USAGE_LOGGER", None) is not None)
+@pytest.fixture(autouse=os.getenv("KOALAS_USAGE_LOGGER", "") != "")
 def add_caplog(caplog):
     with caplog.at_level(logging.INFO, logger="databricks.koalas.usage_logger"):
         yield
@@ -106,3 +106,10 @@ def add_caplog(caplog):
 def close_figs():
     yield
     plt.close("all")
+
+
+@pytest.fixture(autouse=True)
+def check_options():
+    orig_default_index_type = ks.options.compute.default_index_type
+    yield
+    assert ks.options.compute.default_index_type == orig_default_index_type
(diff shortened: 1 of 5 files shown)


ueshin:
FYI: testing against `v1.2.0` at https://github.com/ueshin/databricks-koalas/pull/3.

databricks/conftest.py
```diff
@@ -96,7 +96,7 @@ def add_db(doctest_namespace):
     doctest_namespace["db"] = db_name
 
 
-@pytest.fixture(autouse=os.getenv("KOALAS_USAGE_LOGGER", None) is not None)
+@pytest.fixture(autouse=os.getenv("KOALAS_USAGE_LOGGER", "") != "")
```
itholic, line 99: Is this because "KOALAS_USAGE_LOGGER" can be `None` ??

ueshin: It's just changing the way to check the environment variable to see the same between not-set and empty.

itholic: Oh I got it, thanks! :)

itholic (approved):
LGTM

databricks/conftest.py
HyukjinKwon, line 99: Hmm .. why do we need this though? I thought we can just `unset KOALAS_USAGE_LOGGER`.

HyukjinKwon (approved):
LGTM otherwise

databricks/conftest.py
ueshin: The variable is always set in CI with conda.
https://github.com/databricks/koalas/blob/bc51b4b06c8db7eca60a3d8f8b31aa4db98642b1/.github/workflows/master.yml#L131
I just wanted to ignore it.

ueshin: It's not a big deal. Let me just revert it.

HyukjinKwon: Ah, got it. it should be fine.

ueshin: ah, ok, let's have it.

ueshin:
Thanks! merging.

PR #1756, opened 2020-09-10
Use _to_internal_pandas to avoid copy

itholic:
To avoid overhead of copy, use `_to_internal_pandas` rather than `to_pandas()` if possible.


ueshin (approved):
ah, cool! LGTM.

ueshin:
Thanks! merging.

PR #1757, opened 2020-09-10
Remove arguments from super()

itholic:
Since the syntax for `super()` changed in [Python 3.0](https://docs.python.org/3/library/functions.html#super),

I think we can make the code more simpler and get more readability.

```diff
class OpsOnDiffFramesEnabledTest(ReusedSQLTestCase, SQLTestUtils):
    @classmethod
    def setUpClass(cls):
-       super(OpsOnDiffFramesEnabledTest, cls).setUpClass()
+       super().setUpClass()
```

### DIFF
--- a/databricks/koalas/exceptions.py
+++ b/databricks/koalas/exceptions.py
@@ -48,7 +48,7 @@ class SparkPandasNotImplementedError(NotImplementedError):
             description += " " + hint
         else:
             description = hint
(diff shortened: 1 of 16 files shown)


ueshin (approved):
I'm fine with this, pending tests.

itholic:
Thank, @ueshin !
Let me just merge now, it's trivial.

PR #1761, opened 2020-09-11
Refine Index constructor to follow pandas' behavior

### LINKED ISSUE #1758: Creating MultiIndex via Index class.
pandas can create MultiIndex via Index class as shown below.

```python
>>> pidx = pd.Index([(1, 'a'), (2, 'b'), (3, 'c')])
>>> pidx
MultiIndex([(1, 'a'),
            (2, 'b'),
            (3, 'c')],
           )

>>> type(pidx)
<class 'pandas.core.indexes.multi.MultiIndex'>
```

Koalas also seems can do, but the type is still `Index`, not `MultiIndex`

```python
>>> kidx = ks.Index([(1, 'a'), (2, 'b'), (3, 'c')])
>>> kidx
MultiIndex([(1, 'a'),
            (2, 'b'),
            (3, 'c')],
           )

>>> type(kidx) # this must be <class 'databricks.koalas.indexes.MultiIndex'>
<class 'databricks.koalas.indexes.Index'>
```

itholic:
This PR should refine the Index constructor to follow pandas' behavior

This should resolve #1758 .

```python
>>> tuples = [(1, "red"), (1, "blue"), (2, "red"), (2, "blue")]
>>> kmidx = ks.Index(tuples)

>>> kmidx
MultiIndex([(1,  'red'),
            (1, 'blue'),
            (2,  'red'),
            (2, 'blue')],
           )

>>> type(kmidx)
<class 'databricks.koalas.indexes.MultiIndex'>
```


databricks/koalas/indexes.py
```diff
@@ -108,13 +108,26 @@ class Index(IndexOpsMixin):
     Index(['a', 'b', 'c'], dtype='object')
     """
 
-    def __init__(self, data: Union[DataFrame, list], dtype=None, name=None) -> None:
+    def __new__(cls, data: Union[DataFrame, list], dtype=None, name=None, names=None):
+        if isinstance(data, list) and all([isinstance(item, tuple) for item in data]):
+            # MultiIndex
+            return MultiIndex.from_tuples(data, names=names)
```
ueshin, line 114: Is this a right usage of `__new__`?
I think it's supposed to just create an instance, and the initialization should be in `__init__`.

itholic: Hmm... actually I just referred to this kind of behavior from pandas.

```python
    def __new__(
        cls, data=None, dtype=None, copy=False, name=None, tupleize_cols=True, **kwargs
    ) -> "Index":

        ...

        # range
        if isinstance(data, RangeIndex):
            return RangeIndex(start=data, copy=copy, dtype=dtype, name=name)
        elif isinstance(data, range):
            return RangeIndex.from_range(data, dtype=dtype, name=name)
```


Let me investigate more about the usage of `__new__`.

ueshin: ah, I see. pandas rather doesn't use `__init__` at all.

itholic: It's right. what do you think about we also follow that behavior ??

ueshin: Let's follow the behavior.

databricks/koalas/indexes.py
```diff
@@ -108,13 +108,26 @@ class Index(IndexOpsMixin):
...
+        else:
+            # Index
+            return super().__new__(cls)
+
+    def __init__(self, data: Union[DataFrame, list], dtype=None, name=None, names=None) -> None:
```
HyukjinKwon, line 119: How does it work if both `name` and `names` are set? Looks like we should throw an exception properly.

itholic: Thanks, @HyukjinKwon . I'll manage this.

HyukjinKwon (approved):

databricks/koalas/indexes.py
```diff
@@ -108,14 +108,32 @@ class Index(IndexOpsMixin):
...
+        instance = None
+
         if isinstance(data, DataFrame):
             assert dtype is None
             assert name is None
         else:
-            data = DataFrame(index=pd.Index(data=data, dtype=dtype, name=name))
+            if isinstance(data, list) and all([isinstance(item, tuple) for item in data]):
+                return MultiIndex.from_tuples(data, names=names)
+            if isinstance(instance, MultiIndex):
```
ueshin, line 121: The `instance` is never `MultiIndex`..

databricks/koalas/indexes.py
```diff
@@ -108,14 +108,32 @@ class Index(IndexOpsMixin):
...
+                index = pd.Index(data=data, dtype=dtype, name=name, names=names)
+            else:
+                index = pd.Index(data=data, dtype=dtype, name=name)
+            data = DataFrame(index=index)
+
+        if instance is None:
```
ueshin, line 127: The `instance` has never been `not None`..

PR #1764, opened 2020-09-11
Add asi8 for Index & MultiIndex

itholic:
The PR proposes asi8 for Index & MultiIndex

```python
>>> ks.Index([1, 2, 3]).asi8
array([1, 2, 3])

>>> ks.Index(['a', 'b', 'c']).asi8 is None
True
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -467,8 +467,10 @@ class Index(IndexOpsMixin):
         warnings.warn("We recommend using `{}.to_numpy()` instead.".format(type(self).__name__))
         if self.dtype == "int64":
             return self.to_numpy()
+        elif self.dtype == "<M8[ns]":
+            return np.array(list(map(lambda x: x.astype(np.int64), self.to_numpy())))
         else:
-            return
+            return None
 
     @property
     def spark_type(self):
@@ -2863,8 +2865,16 @@ class MultiIndex(Index):
         """
         Return a string of the type inferred from the values.
         """
-        # It's always 'mixed' for MultiIndex
+        # Always returns "mixed" for MultiIndex
         return "mixed"
+
+    @property
+    def asi8(self):
+        """
+        Integer representation of the values.
+        """
+        # Always returns None for MultiIndex
+        return None
 
     def __iter__(self):
         return MissingPandasLikeMultiIndex.__iter__(self)
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -456,6 +456,13 @@ class Index(IndexOpsMixin):
 
         Examples
         --------
+        >>> ks.Index([1, 2, 3]).asi8
+        array([1, 2, 3])
+
+        Returns None for non-int64 dtype
+
+        >>> ks.Index(['a', 'b', 'c']).asi8 is None
+        True
         """
         warnings.warn("We recommend using `{}.to_numpy()` instead.".format(type(self).__name__))
         if self.dtype == "int64":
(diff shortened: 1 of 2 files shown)


databricks/koalas/indexes.py
```diff
@@ -439,6 +439,37 @@ def values(self):
...
+        >>> ks.Index([1, 2, 3]).asi8
+        array([1, 2, 3])
+
+        Returns None for non-int64 dtype
+
+        >>> ks.Index(['a', 'b', 'c']).asi8 is None
+        True
+        """
+        warnings.warn("We recommend using `{}.to_numpy()` instead.".format(type(self).__name__))
+        if self.dtype == "int64":
```
HyukjinKwon, line 468: @itholic can we cover datetime too?

```python
>>> pd.DataFrame({'a': [1,2,3]}, index=pd.date_range(end='1/1/2018', periods=3)).index.asi8
array([1514592000000000000, 1514678400000000000, 1514764800000000000])
```

databricks/koalas/indexes.py
```diff
@@ -439,6 +439,37 @@ def values(self):
...
+            return self.to_numpy()
+        else:
+            return
```
HyukjinKwon, line 471: I would just explicitly return `None`.

databricks/koalas/indexes.py
itholic: Sounds good. Thanks!

databricks/koalas/indexes.py
itholic: Looks better, thanks! :)

HyukjinKwon (approved):

databricks/koalas/indexes.py
```diff
@@ -439,6 +439,39 @@ def values(self):
...
+        >>> ks.Index([1, 2, 3]).asi8
+        array([1, 2, 3])
+
+        Returns None for non-int64 dtype
+
+        >>> ks.Index(['a', 'b', 'c']).asi8 is None
+        True
+        """
+        warnings.warn("We recommend using `{}.to_numpy()` instead.".format(type(self).__name__))
+        if self.dtype == "int64":
```
ueshin, line 468: What if the `dtype` is the other integral types? `int8`, `int16`, or `int32`?

```py
>>> pd.Index([1, 2, 3]).asi8
array([1, 2, 3])
>>> pd.Index([1, 2, 3]).astype('int32').asi8
array([1, 2, 3])
>>> pd.Index([1, 2, 3]).astype('int16').asi8
array([1, 2, 3])
>>> pd.Index([1, 2, 3]).astype('int8').asi8
array([1, 2, 3])
```

ueshin: I guess we should use `isinstance(self.spark.data_type, IntegralType)`? (should check the nullability as well?)

databricks/koalas/indexes.py
```diff
@@ -439,6 +439,39 @@ def values(self):
...
+            return self.to_numpy()
+        elif self.dtype == "<M8[ns]":
```
ueshin, line 470: Similarly, we should use `isinstance(self.spark.data_type, TimestampType)`?

databricks/koalas/indexes.py
itholic: Thanks fro the comment!
If the null valued included, it should returns `None`.
So I think maybe `isinstance(self.spark.data_type, IntegralType)` is enough!

databricks/koalas/indexes.py
itholic: Thanks!

ueshin (approved):

ueshin:
Thanks! merging.

PR #1765, opened 2020-09-11
Add is_type_compatible for Index & MultiIndex

itholic:
This PR proposes `is_type_compatible()` for `Index` & `MultiIndex`

```python
>>> kidx.is_type_compatible('integer')
True

>>> kidx = ks.Index([1.0, 2.0, 3.0])
>>> kidx.is_type_compatible('integer')
False
>>> kidx.is_type_compatible('floating')
True
```


databricks/koalas/indexes.py
```diff
@@ -861,6 +861,24 @@ def is_object(self):
         """
         return is_object_dtype(self.dtype)
 
+    def is_type_compatible(self, kind):
```
HyukjinKwon, line 864: Can you add it to the docs?

itholic: pandas doesn't add this function to their docs.

Shall we add this to the docs regardless of it or keep excluding ?

HyukjinKwon: They have the doc page https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.is_type_compatible.html but seems it's not just listed up. Can you add it with hidden option as below?

```rst
.. toctree::
   :hidden:
   Index.is_type_compatible
```

itholic: Oh I didn't notice that hidden option. Thanks, I'll add it !! 👍

HyukjinKwon (approved):

PR #1766, opened 2020-09-11
Add is_unique for Index & MultiIndex

itholic:
This PR proposes `is_unique` for `Index` & `MultiIndex`.

Basically it's opposite to `has_duplicates`

```python
>>> idx = ks.Index([1, 5, 7, 7])
>>> idx.is_unique
False

>>> idx = ks.Index([1, 5, 7])
>>> idx.is_unique
True

>>> idx = ks.Index(["Watermelon", "Orange", "Apple",
...                 "Watermelon"])
>>> idx.is_unique
False

>>> idx = ks.Index(["Orange", "Apple",
...                 "Watermelon"])
>>> idx.is_unique
True
```

And fixed some tests to sync with pandas'.

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -456,22 +456,57 @@ class Index(IndexOpsMixin):
 
         Examples
         --------
-        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=list('aac'))
-        >>> kdf.index.has_duplicates
+        >>> import databricks.koalas as ks
+        >>> idx = ks.Index([1, 5, 7, 7])
+        >>> idx.has_duplicates
         True
 
-        >>> kdf = ks.DataFrame({'a': [1, 2, 3]}, index=[list('abc'), list('def')])
-        >>> kdf.index.has_duplicates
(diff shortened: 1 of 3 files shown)


databricks/koalas/indexes.py
```diff
@@ -456,23 +456,56 @@ def has_duplicates(self) -> bool:
...
+        >>> idx.has_duplicates
+        False
         """
         sdf = self._internal.spark_frame.select(self.spark.column)
         scol = scol_for(sdf, sdf.columns[0])
 
         return sdf.select(F.count(scol) != F.countDistinct(scol)).first()[0]
 
+    @property
+    def is_unique(self):
```
HyukjinKwon, line 483: Can we add it to the docs?

HyukjinKwon, line 483: Let's also add the type hints `-> bool`

HyukjinKwon (approved):
Looks good otherwise.

databricks/koalas/indexes.py
itholic: Oh, my mistake. Thanks !

itholic: Sounds good

PR #1776, opened 2020-09-16
Manage column_label_names properly.

ueshin:
`InternalFrame.column_label_names` is equivalent to `index_names`. We should manage it properly.
Now it will never be `None` but a list of tuples, the same as `index_names`.


HyukjinKwon (approved):
Took a quick look and looks good

ueshin:
Thanks! Let me merge this now.

itholic:
Looks fine to me. Nice!

PR #1784, opened 2020-09-22
Basic support for non-string names.

ueshin:
Currently names in Koalas, e.g., `df.columns`, `df.colums.names`, `df.index.names`, need to be string or tuple of string, but it should allow other data types which are supported by Spark.

before:

```py
>>> kdf = ks.DataFrame([[1, 'x'], [2, 'y'], [3, 'z']])
>>> kdf.columns
Index(['0', '1'], dtype='object')
```

after:

```py
>>> kdf = ks.DataFrame([[1, 'x'], [2, 'y'], [3, 'z']])
>>> kdf.columns
Int64Index([0, 1], dtype='int64')
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -58,12 +58,9 @@ from databricks.koalas.utils import (
     compare_null_last,
     default_session,
     name_like_string,
-    name_or_tuple,
     scol_for,
     verify_temp_column_name,
     validate_bool_kwarg,
-    validate_name_like_list,
-    validate_name_like_tuple,
 )
 from databricks.koalas.internal import (
     InternalFrame,
@@ -286,7 +283,7 @@ class Index(IndexOpsMixin):
         >>> idx = ks.Index(['a', 'b', 'c'])
         >>> idx.name = "name"
         >>> midx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
-        >>> midx.names = ["nameA", "nameB"]
+        >>> midx.names = ("nameA", "nameB")
 
         For Index
 
@@ -543,22 +540,25 @@ class Index(IndexOpsMixin):
         return not self.has_duplicates
 
     @property
-    def name(self) -> Union[Any, Tuple]:
+    def name(self) -> Union[str, Tuple[str, ...]]:
         """Return name of the Index."""
         return self.names[0]
 
     @name.setter
-    def name(self, name: Union[Any, Tuple]) -> None:
+    def name(self, name: Union[str, Tuple[str, ...]]) -> None:
         self.names = [name]
 
     @property
-    def names(self) -> List[Union[Any, Tuple]]:
+    def names(self) -> List[Union[str, Tuple[str, ...]]]:
         """Return names of the Index."""
-        return [name_or_tuple(name) for name in self._internal.index_names]
+        return [
+            name if name is None or len(name) > 1 else name[0]
+            for name in self._internal.index_names  # type: ignore
+        ]
 
     @names.setter
-    def names(self, names: List[Union[Any, Tuple]]) -> None:
-        if isinstance(names, tuple) or not is_list_like(names):
+    def names(self, names: List[Union[str, Tuple[str, ...]]]) -> None:
+        if not is_list_like(names):
             raise ValueError("Names must be a list-like")
         self.rename(names, inplace=True)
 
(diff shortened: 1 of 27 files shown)


databricks/koalas/tests/test_series.py
```diff
@@ -1831,19 +1818,12 @@ def test_droplevel(self):
             self.assert_eq(pser.droplevel(-1), kser.droplevel(-1))
             self.assert_eq(pser.droplevel([0]), kser.droplevel([0]))
             self.assert_eq(pser.droplevel(["level_1"]), kser.droplevel(["level_1"]))
-            self.assert_eq(pser.droplevel((0,)), kser.droplevel((0,)))
```
itholic, line 1834: Why this is removed? seems working fine

```python
>>> pser.droplevel((0,))
level_2  level_3
a        q          1
b        w          2
c        e          3
dtype: int64

>>> kser.droplevel((0,))
level_2  level_3
a        q          1
b        w          2
c        e          3
dtype: int64
```

databricks/koalas/tests/test_series.py
```diff
@@ -1831,19 +1818,12 @@ def test_droplevel(self):
...
             self.assert_eq(
                 pser.droplevel(["level_1", "level_3"]), kser.droplevel(["level_1", "level_3"])
             )
-            self.assert_eq(pser.droplevel((1, 2)), kser.droplevel((1, 2)))
-            self.assert_eq(
-                pser.droplevel(("level_2", "level_3")), kser.droplevel(("level_2", "level_3"))
-            )
 
-            with self.assertRaisesRegex(KeyError, "Level {0, 1, 2} not found"):
-                kser.droplevel({0, 1, 2})
```
itholic, line 1846: Why we cannot these tests properly anymore ? 😲

databricks/koalas/indexes.py
```diff
@@ -1066,7 +1064,7 @@ def drop(self, labels):
         >>> index.drop([1])
         Int64Index([2, 3], dtype='int64')
         """
-        if not isinstance(labels, (tuple, list)):
+        if not is_list_like(labels):  # TODO: tuple?
```
itholic, line 1067: I think maybe we don't need to consider `tuple` here because `Index` cannot have `tuple` value in Koalas anyway ??

```python
>>> ks.Index([(1, 2), 3])  # Cannot have a tuple for Index as mixed type
...
pyarrow.lib.ArrowInvalid: Could not convert (1, 2) with type tuple: did not recognize Python value type when inferring an Arrow data type

>>> ks.Index([(1, 2), (3, 4)])  # This is not gonna be an Index but MultiIndex
MultiIndex([(1, 2),
            (3, 4)],
           )
```

Oh, but if we had any plan for supporting tuple for values, I think It's okay to remain the TODO as it is.

databricks/koalas/indexes.py
```diff
@@ -283,7 +286,7 @@ def equals(self, other):
         >>> idx = ks.Index(['a', 'b', 'c'])
         >>> idx.name = "name"
         >>> midx = ks.MultiIndex.from_tuples([('a', 'x'), ('b', 'y'), ('c', 'z')])
-        >>> midx.names = ("nameA", "nameB")
+        >>> midx.names = ["nameA", "nameB"]
```
itholic, line 289: I wonder if Is it okay that we can't support assigning names as a tuple that pandas can do?

```python
>>> pmidx.names = ("nameA", "nameB")
>>> pmidx
MultiIndex([('a', 'x'),
            ('b', 'y'),
            ('c', 'z')],
           names=['nameA', 'nameB'])
```

itholic (commented):
Overall it looks fine, but I wonder maybe if it wouldn't matter to abandon some of the features we could have supported before ??

ueshin:
Let me revert some commits. Maybe we should discuss input checks separately.

databricks/koalas/tests/test_series.py
ueshin: Let's discuss in the following PR.

databricks/koalas/tests/test_series.py
ueshin: ditto.

databricks/koalas/indexes.py
ueshin: ditto.

databricks/koalas/tests/test_series.py
itholic: Sure ! :)

itholic (approved):
Looks good enough in terms of `Basic support` !

Let's discuss the details separately.

ueshin:
Thanks! I'd merge this now. I'll submit PRs for each input check.

PR #1785, opened 2020-09-22
Implemented DataFrame.lookup

itholic:
This PR proposes `DataFrame.lookup`.

```python
>>> kdf = ks.DataFrame({'A': [3, 4, 5, 6, 7],
...                     'B': [10.0, 20.0, 30.0, 40.0, 50.0],
...                     'C': ['a', 'b', 'c', 'd', 'e']})
>>> kdf
   A     B  C
0  3  10.0  a
1  4  20.0  b
2  5  30.0  c
3  6  40.0  d
4  7  50.0  e

>>> kdf.lookup([0], ["C"])
array(['a'], dtype=object)

>>> kdf.lookup([2, 3], ["A", "B"])
array([ 5., 40.])
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10232,7 +10232,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         array of the values corresponding to each (row, col) pair.
 
         .. note:: This method should only be used when the length of `row_labels` is small enough,
-                  as all the result is loaded into the driver's memory.
+                  as all the data belongs to the `row_labels` is loaded into the driver's memory.
 
         Parameters
         ----------
@@ -10265,8 +10265,15 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         >>> kdf.lookup([2, 3], ["A", "B"])
         array([ 5., 40.])
         """
+        from databricks.koalas.series import Series
+        from databricks.koalas.indexes import Index
+
         if len(row_labels) != len(col_labels):
             raise ValueError("Row labels must have same size as column labels")
+        if isinstance(row_labels, (Series, Index)):
+            row_labels = row_labels.to_numpy().tolist()
+        if isinstance(col_labels, (Series, Index)):
+            col_labels = col_labels.to_numpy().tolist()
         lookups = [
             self.loc[row_label, col_label] for row_label, col_label in zip(row_labels, col_labels)
         ]
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10252,18 +10252,18 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         ...                     'B': [10.0, 20.0, 30.0, 40.0, 50.0],
         ...                     'C': ['a', 'b', 'c', 'd', 'e']})
         >>> kdf
-           A     B  C   D
-        0  3  10.0  a NaN
-        1  4  20.0  b NaN
-        2  5  30.0  c NaN
-        3  6  40.0  d NaN
-        4  7  50.0  e NaN
+           A     B  C
+        0  3  10.0  a
+        1  4  20.0  b
+        2  5  30.0  c
+        3  6  40.0  d
+        4  7  50.0  e
 
         >>> kdf.lookup([0], ["C"])
         array(['a'], dtype=object)
 
-        >>> kdf.lookup([2, 3], ["A", "D"])
-        array([ 5., nan])
+        >>> kdf.lookup([2, 3], ["A", "B"])
+        array([ 5., 40.])
         """
         if len(row_labels) != len(col_labels):
             raise ValueError("Row labels must have same size as column labels")
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -10224,6 +10224,54 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         return DataFrame(pd.DataFrame.from_dict(data, orient=orient, dtype=dtype, columns=columns))
 
+    def lookup(self, row_labels, col_labels) -> np.ndarray:
+        """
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -10224,6 +10224,61 @@ def from_dict(data, orient="columns", dtype=None, columns=None) -> "DataFrame":
...
 
+    def lookup(self, row_labels, col_labels) -> np.ndarray:
+        """
+        Label-based "fancy indexing" function for DataFrame.
+
+        Given equal-length arrays of row and column labels, return an
+        array of the values corresponding to each (row, col) pair.
+
+        .. note:: This method should only be used when the length of `row_labels` is small enough,
+                  as all the data belongs to the `row_labels` is loaded into the driver's memory.
```
itholic, line 10235: I think we have no choice but should iterate the whole item of input `row_labels`.

databricks/koalas/frame.py
```diff
@@ -10224,6 +10224,61 @@ def from_dict(data, orient="columns", dtype=None, columns=None) -> "DataFrame":
...
+        """
+        from databricks.koalas.series import Series
+        from databricks.koalas.indexes import Index
+
+        if len(row_labels) != len(col_labels):
+            raise ValueError("Row labels must have same size as column labels")
+        if isinstance(row_labels, (Series, Index)):
+            row_labels = row_labels.to_numpy().tolist()
+        if isinstance(col_labels, (Series, Index)):
+            col_labels = col_labels.to_numpy().tolist()
```
HyukjinKwon, line 10276: Can you do this first and then compare `if len(row_labels) != len(col_labels):`? If `Series` is given as an argument, it will execute the jobs twice.

databricks/koalas/frame.py
```diff
@@ -10224,6 +10224,61 @@ def from_dict(data, orient="columns", dtype=None, columns=None) -> "DataFrame":
...
+        array([ 5., 40.])
+        """
+        from databricks.koalas.series import Series
+        from databricks.koalas.indexes import Index
+
+        if len(row_labels) != len(col_labels):
+            raise ValueError("Row labels must have same size as column labels")
+        if isinstance(row_labels, (Series, Index)):
+            row_labels = row_labels.to_numpy().tolist()
+        if isinstance(col_labels, (Series, Index)):
```
HyukjinKwon, line 10275: Actually, why don't you just disallow this case for now like what `iloc` and `loc` do?

itholic: Sounds good! Thanks for the review 👍

itholic:
Thanks, @HyukjinKwon

databricks/koalas/tests/test_dataframe.py
```diff
@@ -4019,3 +4019,60 @@ def test_from_dict(self):
...
+
+    def test_lookup(self):
+        pdf = pd.DataFrame(
+            {
+                "A": [3, 4, 5, 6, 7],
+                "B": [10.0, 20.0, 30.0, 40.0, 50.0],
+                "C": ["a", "b", "c", "d", "e"],
+            }
+        )
+        kdf = ks.from_pandas(pdf)
```
ueshin, line 4031: What if the index or the column is duplicated?

databricks/koalas/frame.py
```diff
@@ -10224,6 +10224,65 @@ def from_dict(data, orient="columns", dtype=None, columns=None) -> "DataFrame":
...
+            raise TypeError(
+                "'col_labels' doesn't support type '{}'.".format(type(col_labels).__name__)
+            )
+        if len(row_labels) != len(col_labels):
+            raise ValueError("Row labels must have same size as column labels")
+        lookups = [
+            self.loc[row_label, col_label] for row_label, col_label in zip(row_labels, col_labels)
+        ]
+        return np.asarray(pd.Series(lookups))
```
ueshin, line 10284: How about `self.loc[list(set(row_label)), list(set(col_label))].to_pandas().lookup(row_label, col_label)`?

databricks/koalas/tests/test_dataframe.py
itholic: It should raise `ValueError: DataFrame.lookup requires unique index and columns`.
I'll address it. Thanks :)

databricks/koalas/frame.py
itholic: Oh, so we can just delegate to the pandas.
Like this idea 👍

itholic:
Thanks, @ueshin

databricks/koalas/frame.py
```diff
@@ -10246,6 +10246,74 @@ def from_dict(data, orient="columns", dtype=None, columns=None) -> "DataFrame":
...
+                self.loc[list(set(row_labels)), list(set(col_labels))]
+                .to_pandas()
+                .lookup(row_labels, col_labels)
+            )
+        else:
+            if len(row_labels) != len(col_labels):
+                raise ValueError("Row labels must have same size as column labels")
+            lookups = [
+                self.loc[row_label, col_label]
+                for row_label, col_label in zip(row_labels, col_labels)
```
HyukjinKwon, line 10313: Does this mean we launch jobs as many as labels?

databricks/koalas/frame.py
```diff
@@ -10277,7 +10277,7 @@ def lookup(self, row_labels, col_labels) -> np.ndarray:
         lookups = [
             self.loc[row_label, col_label] for row_label, col_label in zip(row_labels, col_labels)
         ]
-        return Series(lookups).to_numpy()
+        return np.asarray(pd.Series(lookups))
```
xinrong-databricks: Out of curiosity, why did we change it to `np.asarray(pd.Series(lookups))`?

xinrong-databricks:
https://issues.apache.org/jira/browse/SPARK-36434

PR #1795, opened 2020-09-24
Add numpy to Dependencies table

itholic:
Added `numpy` to the Dependencies table in the documents.
And fixed "Minimum supported version" of `pandas` properly - from `0.23` to `0.23.2` - .

스크린샷 2020-09-25 오전 12 45 55

### DIFF
--- a/docs/source/getting_started/install.rst
+++ b/docs/source/getting_started/install.rst
@@ -121,10 +121,11 @@ Dependencies
 ============= =========================
 Package       Minimum supported version
 ============= =========================
-`pandas`      0.23
+`pandas`      0.23.2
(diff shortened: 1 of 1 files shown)


ueshin:
Cool.
Btw, it might be better to put the upper bounds? cc @HyukjinKwon

HyukjinKwon:
Oh yeah. We can change "Minimal supported version" to something like "Required version" and write https://github.com/databricks/koalas/blob/master/setup.py#L51-L54

itholic:
Sounds good. I'll fix it

HyukjinKwon (approved):

PR #1796, opened 2020-09-26
Fix Series.__repr__ when Series.name is None.

ueshin:
Fix `Series.__repr__` when `Series.name` is `None` and the length is longer than a config `display.max_rows`.
Additional added some more test cases.


itholic:
Nice! LGTM.

PR #1797, opened 2020-09-26
Update missing list with new functions

itholic:
Update missing list with newly added functions in pandas > 1.0.0


itholic:
It's just sync with pandas function list. feel free to leave any comments.

databricks/koalas/missing/frame.py
```diff
@@ -39,10 +39,13 @@ class _MissingPandasLikeDataFrame(object):
     asfreq = _unsupported_function("asfreq")
     asof = _unsupported_function("asof")
     at_time = _unsupported_function("at_time")
+    backfill = _unsupported_function("_backfill")
```
ueshin, line 42: `_unsupported_function("backfill")`?

itholic: @ueshin It's removed at #1798 ! Thanks for the review :)

PR #1805, opened 2020-09-28
Cleanup of refining Index constructor.

ueshin:
This is a follow-up of #1761 to address comments on https://github.com/databricks/koalas/pull/1761#pullrequestreview-497790324.


itholic:
LGTM.

Thanks for the follow-up, @ueshin !

PR #1814, opened 2020-10-02
Fix the input check in Series.

ueshin:
Fixes the input check in `Series`.

### DIFF
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -1131,37 +1131,51 @@ class SeriesTest(ReusedSQLTestCase, SQLTestUtils):
     def test_drop(self):
         pser = pd.Series([10, 20, 15, 30, 45], name="x")
         kser = ks.Series(pser)
+
+        self.assert_eq(kser.drop(1), pser.drop(1))
+        self.assert_eq(kser.drop([1, 4]), pser.drop([1, 4]))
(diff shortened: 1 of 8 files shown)


databricks/koalas/tests/test_series.py
```diff
@@ -1131,37 +1131,51 @@ def test_aggregate(self):
...
-            kser.drop({"lama": "speed"})
+
         msg = "Cannot specify both 'labels' and 'index'"
         with self.assertRaisesRegex(ValueError, msg):
             kser.drop("lama", index="cow")
+
         msg = r"'Key length \(2\) exceeds index depth \(3\)'"
         with self.assertRaisesRegex(KeyError, msg):
             kser.drop(("lama", "speed", "x"))
-        self.assert_eq(kser.drop(("lama", "speed", "x"), level=1), kser)
```
ueshin, line 1164: @itholic Do you remember the reason of this behavior?

itholic: Hmm... I don't remember clearly but I'd say maybe there's no special reason for it - I think I just had not enough understood at the time of working on this.😅 -
Anyway, it seems that this case should be fixed to raise an error.

HyukjinKwon (approved):

ueshin:
Thanks! merging.

PR #1823, opened 2020-10-06
Fix to_list work properly in pandas==0.23

itholic:
`Koalas` doesn't support `to_list` in `pandas<0.24` because `pandas<0.24` doesn't support `to_list` as an alias for `tolist`.

```python
>>> pd.__version__
'0.23.2'
>>> pser.to_list()
  File "/Users/haejoon.lee/opt/miniconda3/envs/koalas-dev-3.6/lib/python3.6/site-packages/pandas/core/generic.py", line 4372, in __getattr__
    return object.__getattribute__(self, name)
AttributeError: 'Series' object has no attribute 'to_list'
>>> pser.tolist()
[1, 2, 3]
```

This PR fix it to make working `to_list` properly in pandas<0.24.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1478,7 +1478,11 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
             to be small, as all the data is loaded into the driver's memory.
 
         """
-        return self._to_internal_pandas().to_list()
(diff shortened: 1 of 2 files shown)


itholic:
I'm pretty sure this is minor fix. Please feel free to comment anytime !

PR #1828, opened 2020-10-06
Standardize binary operations between int and str columns

### LINKED ISSUE #1819: Operations on Columns of Differing Types
I just noticed this difference between `koalas` and `pandas` when it comes to performing operations on columns of different types e.g. `int` and `str`. Pandas will throw a `TypeError` whereas `koalas` won't and will actually return something. I'm sure this isn't a bug but I wondered if there's anywhere where this behaviour is documented that I can read about? The return types seem to be a little unpredictable. I have only looked at `str` and `int` so far but it would be great if there was more information on all the different types and why they return what they return. Also, is there a reason, we don't throw a `TypeError`? This would be a very quick check just based on the `dtype` of the column

### Koalas
```python
df = ks.DataFrame({'a':[1, 2, 3], 'b':['a', '2', 'c'], 'c':['d', 'e', 'f']})
df['add_ab'] = df['a'] + df['b']
df['sub_ab'] = df['a'] - df['b']
df['div_ab'] = df['a'] / df['b']
df['mul_ab'] = df['a'] * df['b']
df['add_bc'] = df['b'] + df['c']
df['sub_bc'] = df['b'] - df['c']
df['div_bc'] = df['b'] / df['c']
df['mul_bc'] = df['b'] * df['c']
print(df)
#    a  b  c  add_ab  sub_ab  div_ab  mul_ab add_bc  sub_bc    div_bc  mul_bc
# 0  1  a  d     NaN     NaN     inf     NaN     ad     NaN      None     NaN
# 1  2  2  e     4.0     0.0     1.0     4.0     2e     NaN  Infinity     NaN
# 2  3  c  f     NaN     NaN     inf     NaN     cf     NaN      None     NaN
```

- Interestingly, '2' as a string in column 'b' behaves as a number.
- div_ab results in `inf` whereas the other ab operations result in `NaN`
- div_bc this time results in `None` where the other operations (ignoring add) result in `NaN` again
    - except for row 2 which this time results in `Infinity`

Essentially, I'm not sure why the string version of 2 gets typecast to an integer. And why division results in behaviour that is different to the other operations. And why this is sometimes `None` and sometimes `inf`. And what the difference is between `inf` and `Infinity`.

### Pandas
```python
df = pd.DataFrame({'a':[1, 2, 3], 'b':['a', '2', 'c'], 'c':['d', 'e', 'f']})
df['mul_ab'] = df['a'] * df['b']
df['add_bc'] = df['b'] + df['c']
print(df)
#    a  b  c mul_ab add_bc
# 0  1  a  d      a     ad
# 1  2  2  e     22     2e
# 2  3  c  f    ccc     cf
```

Comparing to the `pandas` version. The two operations included above are the only two that do not result in a `TypeError`. Here the 2 remains as a string. And the allowed operations are as expected.

## Sidenote
```python
df = ks.DataFrame({'a':[1, 2, 3], 'b':['a', 2, 'c'], 'c':['d','e', 'f']})
```
throws an `ArrowTypeError` (the 2 in col 'b' is now an int rather than a string). Pandas however allows this Dataframe to be created. Furthermore, I'm not sure how to make sense of this given that the 2 gets effectively turned into an `int` anyway if given as a `str`.

## Summary

To summarise, I'm not sure what of the above are bugs (if any). But a detailed breakdown of how functions on columns of different types behave compared to their pandas counterparts would be much appreciated. 

koalas==1.2.0
pandas==1.0.5

xinrong-databricks:
## Notes
- New test cases related to **Pandas only** are added for PR reviewers to visualize Pandas' behavior.
They are commented out and will be removed before the PR merge (after PR review).
- This PR doesn't aim to cover all binary operations: only `+`, `-`, `*`, `/`, `//`, `%` are concerned.

## Proposal
Make behaviors of binary operations (`+`, `-`, `*`, `/`, `//`, `%`) between **int** and **str** columns consistent with respective pandas behaviors.

- Standardize binary operations as follows:

    `+`: raise TypeError between int column and str column (or string literal)
    `*`: act as spark SQL `repeat` between int column(or int literal) and str columns; raise TypeError if a string literal is involved
    `-`, `/`, `//`, `%(modulo)`: raise TypeError if a str column (or string literal) is involved

- Add `def repeat(col, n):` in databricks/koalas/spark/functions.py

    [`repeat` defined in scala API](https://github.com/apache/spark/blob/master/sql/core/src/main/scala/org/apache/spark/sql/functions.scala#L2580) only accepts integer literal as the method's second parameter.
    But internal [StringRepeat](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/catalyst/expressions/stringExpressions.scala#L1607) accepts `IntegerType` as the method's second parameter.

    In order to pass int columns as the second parameter, we take advantage of `callUDF`.

## Test
databricks/koalas/tests/test_dataframe.py

Resolves #1819

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -235,13 +235,15 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         return column_op(Column.__sub__)(self, other)
 
     def __mul__(self, other):
-        if isinstance(self.spark.data_type, NumericType):
-            if isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType):
-                return column_op(SF.repeat)(other, self)
-            elif isinstance(other, str):
-                raise TypeError(
-                    "multiplication can not be applied to an int series and a string literal"
-                )
+        if isinstance(self, str) or isinstance(other, str):
+            raise TypeError("multiplication can not be applied to a string literal")
+
+        if (
+            isinstance(self.spark.data_type, NumericType)
+            and isinstance(other, IndexOpsMixin)
+            and isinstance(other.spark.data_type, StringType)
+        ):
+            return column_op(SF.repeat)(other, self)
 
         if isinstance(self.spark.data_type, StringType) and (
             (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, NumericType))
@@ -291,7 +293,7 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
             or (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType))
             or isinstance(other, str)
         ):
-            raise TypeError("mod can not be applied on string series or literals.")
+            raise TypeError("modulo can not be applied on string series or literals.")
 
         def mod(left, right):
             return ((left % right) + right) % right
@@ -345,14 +347,12 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         return column_op(Column.__rsub__)(self, other)
 
     def __rmul__(self, other):
+        if isinstance(other, str):
+            raise TypeError("multiplication can not be applied to a string literal")
+
         if isinstance(self.spark.data_type, StringType):
             if isinstance(other, int):
                 return column_op(SF.repeat)(self, other)
-        if isinstance(self.spark.data_type, NumericType):
-            if isinstance(other, str):
-                raise TypeError(
-                    "multiplication can not be applied to an int series and a string literal"
-                )
 
         return column_op(Column.__rmul__)(self, other)
 
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -417,7 +417,7 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
 
     def __rmod__(self, other):
         if isinstance(self.spark.data_type, StringType) or isinstance(other, str):
-            raise TypeError("mod can not be applied on string series or literals.")
+            raise TypeError("modulo can not be applied on string series or literals.")
 
         def rmod(left, right):
             return ((right % left) + left) % left
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -262,6 +262,13 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         +-----------------------|---------|---------+
         """
 
+        if (
+            isinstance(self.spark.data_type, StringType)
+            or (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType))
+            or isinstance(other, str)
+        ):
+            raise TypeError("division can not be applied on string series or literals.")
+
         def truediv(left, right):
             return F.when(F.lit(right != 0) | F.lit(right).isNull(), left.__div__(right)).otherwise(
                 F.when(F.lit(left == np.inf) | F.lit(left == -np.inf), left).otherwise(
@@ -334,6 +341,9 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
     __rmul__ = column_op(Column.__rmul__)
 
     def __rtruediv__(self, other):
+        if isinstance(self.spark.data_type, StringType) or isinstance(other, str):
+            raise TypeError("division can not be applied on string series or literals.")
+
         def rtruediv(left, right):
             return F.when(left == 0, F.lit(np.inf).__div__(right)).otherwise(
                 F.lit(right).__truediv__(left)
@@ -358,6 +368,12 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         |          -10          |   null  | -np.inf |
         +-----------------------|---------|---------+
         """
+        if (
+            isinstance(self.spark.data_type, StringType)
+            or (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType))
+            or isinstance(other, str)
+        ):
+            raise TypeError("division can not be applied on string series or literals.")
 
         def floordiv(left, right):
             return F.when(F.lit(right is np.nan), np.nan).otherwise(
@@ -373,6 +389,9 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
         return numpy_column_op(floordiv)(self, other)
 
     def __rfloordiv__(self, other):
+        if isinstance(self.spark.data_type, StringType) or isinstance(other, str):
+            raise TypeError("division can not be applied on string series or literals.")
+
         def rfloordiv(left, right):
             return F.when(F.lit(left == 0), F.lit(np.inf).__div__(right)).otherwise(
                 F.when(F.lit(left) == np.nan, np.nan).otherwise(F.floor(F.lit(right).__div__(left)))
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -37,6 +37,7 @@ from pyspark.sql.types import (
     LongType,
     StringType,
     TimestampType,
+    NumericType
 )
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
@@ -171,7 +172,10 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
     __neg__ = column_op(Column.__neg__)
 
     def __add__(self, other):
-        if isinstance(self.spark.data_type, StringType):
+        if isinstance(self.spark.data_type, NumericType):
+            if (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType)) or isinstance(other, str):
+                raise TypeError("string addition can only be applied to string series or literals.")
+        elif isinstance(self.spark.data_type, StringType):
             # Concatenate string columns
             if isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType):
                 return column_op(F.concat)(self, other)
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -37,7 +37,7 @@ from pyspark.sql.types import (
     LongType,
     StringType,
     TimestampType,
-    NumericType
+    NumericType,
 )
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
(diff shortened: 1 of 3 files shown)


itholic:
Just FYI: You can run `formatter` and `linter` using `dev/reformat` and `dev/lint-python` to perform static analysis.

xinrong-databricks:
@itholic Thank you for reminding me!

databricks/koalas/base.py
```diff
@@ -37,6 +37,7 @@
     LongType,
     StringType,
     TimestampType,
+    NumericType,
```
ueshin, line 40: nit: shall we keep alphabetical order?

databricks/koalas/base.py
```diff
@@ -220,7 +234,24 @@ def __sub__(self, other):
                 raise TypeError("date subtraction can only be applied to date series.")
         return column_op(Column.__sub__)(self, other)
 
-    __mul__ = column_op(Column.__mul__)
+    def __mul__(self, other):
+        if isinstance(self, str) or isinstance(other, str):
```
ueshin, line 238: nit: `self` will never be `str`.

databricks/koalas/base.py
```diff
@@ -220,7 +234,24 @@ def __sub__(self, other):
...
+            raise TypeError("multiplication can not be applied to a string literal")
+
+        if (
+            isinstance(self.spark.data_type, NumericType)
```
ueshin, line 242: `IntegralType` instead of `NumericType`?

```py
>>> df = pd.DataFrame({'a':[1.5, 2.5, 3.5], 'b':['a', '2', 'c']})
>>> df
     a  b
0  1.5  a
1  2.5  2
2  3.5  c
>>> df.a * df.b
...
TypeError: can't multiply sequence by non-int of type 'float'
```

databricks/koalas/base.py
```diff
@@ -220,7 +234,24 @@ def __sub__(self, other):
...
+            and isinstance(other, IndexOpsMixin)
+            and isinstance(other.spark.data_type, StringType)
+        ):
+            return column_op(SF.repeat)(other, self)
+
+        if isinstance(self.spark.data_type, StringType) and (
+            (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, NumericType))
```
ueshin, line 249: ditto.

databricks/koalas/base.py
```diff
@@ -171,6 +173,11 @@ def spark_column(self):
     __neg__ = column_op(Column.__neg__)
 
     def __add__(self, other):
+        if isinstance(self.spark.data_type, NumericType):
```
ueshin, line 176: `not isinstance(self.spark.data_type, StringType)` might be better?

databricks/koalas/base.py
```diff
@@ -250,19 +288,36 @@ def truediv(left, right):
...
         def mod(left, right):
             return ((left % right) + right) % right
 
         return column_op(mod)(self, other)
 
     def __radd__(self, other):
         # Handle 'literal' + df['col']
-        if isinstance(self.spark.data_type, StringType) and isinstance(other, str):
-            return self._with_new_scol(F.concat(F.lit(other), self.spark.column))
+        if isinstance(self.spark.data_type, NumericType):
```
ueshin, line 305: ditto.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1750,6 +1751,115 @@ def test_binary_operators(self):
...
+        self.assertRaisesRegex(TypeError, ks_err_msg, lambda: 1 % kdf["a"])
+
+        # pdf = pd.DataFrame({"a": ["x"], "b": [1]})
+        # self.assertRaises(TypeError, lambda: pdf["a"] % pdf["b"])
+        # self.assertRaises(TypeError, lambda: pdf["b"] % pdf["a"])
+        # self.assertRaises(TypeError, lambda: 1 % pdf["a"])
+
+    def test_binary_operator_multiply(self):
+        kdf = ks.DataFrame({"a": ["x", "y"], "b": [1, 2]})
+        pdf = pd.DataFrame({"a": ["x", "y"], "b": [1, 2]})
```
ueshin, line 1841: Shall we use `ks.from_pandas()`?

```py
pdf = pd.DataFrame({"a": ["x", "y"], "b": [1, 2]})
kdf = ks.from_pandas(pdf)
```

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1750,6 +1751,115 @@ def test_binary_operators(self):
             lambda: ks.range(10).add(ks.range(10).id),
         )
 
+    def test_binary_operator_add(self):
+        kdf = ks.DataFrame({"a": ["x"], "b": [1]})
```
ueshin, line 1755: Shall we add positive tests to add string columns?

databricks/koalas/base.py
xinrong-databricks: Makes sense! I've used `IntegralType` instead and kept the alphabetical order.

databricks/koalas/base.py
xinrong-databricks: Updated!

databricks/koalas/base.py
xinrong-databricks: Good idea! This also makes changes in this PR consistent (between int and str columns).

databricks/koalas/base.py
xinrong-databricks: Updated!

databricks/koalas/base.py
xinrong-databricks: Makes sense, updated!

databricks/koalas/base.py
xinrong-databricks: Updated!

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Good idea! Updated all usage in new tests added.

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Added a few positive tests.

databricks/koalas/base.py
```diff
@@ -171,6 +173,11 @@ def spark_column(self):
     __neg__ = column_op(Column.__neg__)
 
     def __add__(self, other):
+        if not isinstance(self.spark.data_type, StringType):
+            if (
+                isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType)
+            ) or isinstance(other, str):
```
itholic, line 179: Maybe I think we can do this also with single `if` statement like other methods ?

```python
        if not isinstance(self.spark.data_type, StringType) and (
            (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, StringType))
            or isinstance(other, str)
        ):
```

databricks/koalas/tests/test_dataframe.py
```diff
@@ -1750,6 +1751,145 @@ def test_binary_operators(self):
...
+
+        # Negative
+        kdf = ks.DataFrame({"a": ["x"], "b": [1]})
+        ks_err_msg = re.escape("string addition can only be applied to string series or literals")
+
+        self.assertRaisesRegex(TypeError, ks_err_msg, lambda: kdf["a"] + kdf["b"])
+        self.assertRaisesRegex(TypeError, ks_err_msg, lambda: kdf["b"] + kdf["a"])
+        self.assertRaisesRegex(TypeError, ks_err_msg, lambda: kdf["b"] + "literal")
+        self.assertRaisesRegex(TypeError, ks_err_msg, lambda: "literal" + kdf["b"])
+        self.assertRaisesRegex(TypeError, ks_err_msg, lambda: 1 + kdf["a"])
```
itholic, line 1769: I think we're missing of test for non-string type of `IndexOpsMixin`?

For example.

```python
pdf = pd.DataFrame({"a": ["x"], "b": ["y"], "c": [1]})
kdf = ks.from_pandas(pdf)

...

self.assertRaisesRegex(TypeError, ks_err_msg, lambda: kdf["c"] + "a")
self.assertRaisesRegex(TypeError, ks_err_msg, lambda: kdf["c"] + kdf["a"])
```

or just add

```python
self.assertRaisesRegex(TypeError, ks_err_msg, lambda: ks.Series([1]) + "a")
self.assertRaisesRegex(TypeError, ks_err_msg, lambda: ks.Series([1]) + kdf["a"])
```

databricks/koalas/base.py
```diff
@@ -250,19 +288,36 @@ def truediv(left, right):
...
             return ((left % right) + right) % right
 
         return column_op(mod)(self, other)
 
     def __radd__(self, other):
         # Handle 'literal' + df['col']
-        if isinstance(self.spark.data_type, StringType) and isinstance(other, str):
-            return self._with_new_scol(F.concat(F.lit(other), self.spark.column))
+        if not isinstance(self.spark.data_type, StringType):
+            if isinstance(other, str):
```
itholic, line 306: How about 

```python
if not isinstance(self.spark.data_type, StringType) and isinstance(other, str):
```

??

databricks/koalas/tests/test_dataframe.py
```diff
@@ -16,6 +16,7 @@
 from datetime import datetime
 from distutils.version import LooseVersion
 import inspect
+import re
```
itholic, line 19: nit: Maybe I don't think we don't necessarily need `re` module because there is no string have to escape ??

```python
ks_err_msg = re.escape("multiplication can not be applied to a string literal")
ks_err_msg = "multiplication can not be applied to a string literal"  # This also could be fine.
```

itholic (approved):
Otherwise, Looks fine to me.

databricks/koalas/base.py
```diff
@@ -291,9 +346,20 @@ def __rsub__(self, other):
                 raise TypeError("date subtraction can only be applied to date series.")
         return column_op(Column.__rsub__)(self, other)
 
-    __rmul__ = column_op(Column.__rmul__)
+    def __rmul__(self, other):
+        if isinstance(other, str):
+            raise TypeError("multiplication can not be applied to a string literal")
+
+        if isinstance(self.spark.data_type, StringType):
+            if isinstance(other, int):
```
itholic, line 354: How about 

```python
if isinstance(self.spark.data_type, StringType) and isinstance(other, int):
```

?

databricks/koalas/base.py
xinrong-databricks: Makes sense! Updated.

databricks/koalas/base.py
xinrong-databricks: Updated!

databricks/koalas/base.py
xinrong-databricks: Updated.

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Good catch! Added missing test cases of non-string type of IndexOpsMixin.

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Sounds good! Removed  `re` module usage.

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

xinrong-databricks:
@ueshin Thank you for merging!

itholic:
Nice 👍

PR #1833, opened 2020-10-08
Standardize multiplication involving string column

xinrong-databricks:
## Proposal
Multiplication involves a string column can only be applied to an int column or literal; otherwise, a TypeError should be raised.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -254,7 +254,7 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
                 return column_op(SF.repeat)(self, other)
             else:
                 raise TypeError(
-                    "multiplication involves a string series can only be applied to an int series or literal"
+                    "a string series can only be multiplied to an int series or literal"
                 )
 
         return column_op(Column.__mul__)(self, other)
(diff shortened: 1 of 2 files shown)


databricks/koalas/base.py
```diff
@@ -245,11 +245,17 @@ def __mul__(self, other):
...
-            (isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, IntegralType))
-            or isinstance(other, int)
-        ):
-            return column_op(SF.repeat)(self, other)
+        if isinstance(self.spark.data_type, StringType):
+            if (
+                isinstance(other, IndexOpsMixin)
+                and isinstance(other.spark.data_type, IntegralType)
+                or isinstance(other, int)
+            ):
```
ueshin, line 253: Shall we make sure the range of `and` and `or` when they are mixed?

```suggestion
            if (
                isinstance(other, IndexOpsMixin) and isinstance(other.spark.data_type, IntegralType)
            ) or isinstance(other, int):
```

ueshin (commented):

databricks/koalas/base.py
xinrong-databricks: Makes sense, this way makes code clearer! Updated.

ueshin (approved):

xinrong-databricks:
@ueshin Thank you! Merged.

PR #1843, opened 2020-10-12
Implement DataFrame/Series rename_axis

LucasG0:
Hi, this PR implements `DataFrame.rename_axis` and `Series.rename_axis`.
I did not add `copy` parameter as a copy is performed anyway.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -9278,16 +9278,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         )
         index_map = OrderedDict(zip(self._internal.index_spark_column_names, index_names))
 
-        spark_frame = self._internal.resolved_copy.spark_frame
-        internal = InternalFrame(
-            spark_frame=spark_frame,
-            index_map=index_map,
-            column_labels=self._internal.column_labels,
-            data_spark_columns=[
-                scol_for(spark_frame, col) for col in self._internal.data_spark_column_names
-            ],
-            column_label_names=column_label_names,
-        )
+        internal = self._internal.copy(index_map=index_map, column_label_names=column_label_names)
 
         if inplace:
             self._update_internal_frame(internal)
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -9129,7 +9129,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         columns: Optional[Any] = None,
         axis: Optional[Union[int, str]] = 0,
         inplace: Optional[bool] = False,
-    ):
+    ) -> Optional["DataFrame"]:
         """
         Set the name of the axis for the index or columns.
 
@@ -9186,6 +9186,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         dog            4         0
         cat            4         0
         monkey         2         2
+
         >>> df = df.rename_axis("animal").sort_index()
         >>> df  # doctest: +NORMALIZE_WHITESPACE
                 num_legs  num_arms
@@ -9193,6 +9194,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         cat            4         0
         dog            4         0
         monkey         2         2
+
         >>> df = df.rename_axis("limbs", axis="columns").sort_index()
         >>> df # doctest: +NORMALIZE_WHITESPACE
         limbs   num_legs  num_arms
@@ -9265,8 +9267,6 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 index = mapper
             elif axis == 1:
                 columns = mapper
-            else:
-                raise ValueError("No axis named %s for object type %s." % (axis, type(axis)))
 
         column_label_names = (
             gen_names(columns, self.columns.names)
@@ -9278,19 +9278,20 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         )
         index_map = OrderedDict(zip(self._internal.index_spark_column_names, index_names))
 
+        spark_frame = self._internal.resolved_copy.spark_frame
         internal = InternalFrame(
-            self._internal.spark_frame,
+            spark_frame=spark_frame,
             index_map=index_map,
             column_labels=self._internal.column_labels,
             data_spark_columns=[
-                scol_for(self._internal.spark_frame, col)
-                for col in self._internal.data_spark_column_names
+                scol_for(spark_frame, col) for col in self._internal.data_spark_column_names
             ],
             column_label_names=column_label_names,
         )
 
         if inplace:
             self._update_internal_frame(internal)
+            return None
         else:
             return DataFrame(internal)
 
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -31,6 +31,7 @@ from typing import Any, Optional, List, Tuple, Union, Generic, TypeVar, Iterable
 
 import numpy as np
 import pandas as pd
+import pandas.core.common as com
 from pandas.api.types import is_list_like, is_dict_like, is_scalar
 
 if LooseVersion(pd.__version__) >= LooseVersion("0.24"):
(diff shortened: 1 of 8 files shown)


HyukjinKwon:
cc @itholic can you review this please?

databricks/koalas/tests/test_series.py
```diff
@@ -170,6 +170,65 @@ def test_rename_method(self):
...
 
+    def test_rename_axis(self):
+
+        index = pd.Index(["A", "B", "C"], name="index")
+        pser = pd.Series([1.0, 2.0, 3.0], index=index, name="name")
+        kser = ks.from_pandas(pser)
+
+        self.assert_eq(
+            pser.rename_axis("index2").sort_index(), kser.rename_axis("index2").sort_index(),
+        )
```
ueshin, line 181: Could you add a test like `(kser + 1).rename_axis("index2").sort_index()`?

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,178 @@ def gen_new_column_labels_entry(column_labels_entry):
...
             return DataFrame(internal)
 
+    def rename_axis(
+        self,
+        mapper: Optional[Any] = None,
+        index: Optional[Any] = None,
+        columns: Optional[Any] = None,
+        axis: Optional[Union[int, str]] = 0,
+        inplace: Optional[bool] = False,
+    ):
```
ueshin, line 9132: Shall we add the return type annotation as well?

databricks/koalas/series.py
```diff
@@ -1115,6 +1115,71 @@ def rename(self, index=None, **kwargs):
         else:
             return first_series(kdf)
 
+    def rename_axis(
+        self, mapper: Optional[Any] = None, index: Optional[Any] = None, inplace: bool = False
+    ):
```
ueshin, line 1120: ditto.

databricks/koalas/tests/test_series.py
LucasG0: Yes, `resolved_copy` is now used !

databricks/koalas/frame.py
LucasG0: Yes, but we need to use `return None` in case of `inplace=True` in order to pass mypy checks.

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,178 @@ def gen_new_column_labels_entry(column_labels_entry):
...
+            raise TypeError("Cannot specify both 'mapper' and any of 'index' or 'columns'.")
+
+        if mapper is not None:
+            axis = validate_axis(axis)
+            if axis == 0:
+                index = mapper
+            elif axis == 1:
+                columns = mapper
+            else:
+                raise ValueError("No axis named %s for object type %s." % (axis, type(axis)))
```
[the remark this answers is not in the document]
LucasG0: Yes right !

databricks/koalas/frame.py
itholic: Then maybe we should use `Union["DataFrame", None]` ?

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,179 @@ def gen_new_column_labels_entry(column_labels_entry):
...
+        inplace: Optional[bool] = False,
+    ) -> Optional["DataFrame"]:
+        """
+        Set the name of the axis for the index or columns.
+
+        Parameters
+        ----------
+        mapper : scalar, list-like, optional
+            A scalar, list-like, dict-like or functions transformations to
+            apply to the axis name attribute.
```
itholic, line 9140: Maybe this explanation for `mapper` is not correct?

In the pandas latest docs, they say:

```
Value to set the axis name attribute.
```

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,179 @@ def gen_new_column_labels_entry(column_labels_entry):
...
+            Modifies the object directly, instead of creating a new DataFrame.
+
+        Returns
+        -------
+        DataFrame, or None if `inplace` is True.
+
+        See Also
+        --------
+        DataFrame.rename : Alter DataFrame index labels or name.
+        Index.rename : Set new names on index.
```
itholic, line 9160: Maybe also `Series.rename` ?

databricks/koalas/series.py
```diff
@@ -1115,6 +1115,71 @@ def rename(self, index=None, **kwargs):
...
+        self, mapper: Optional[Any] = None, index: Optional[Any] = None, inplace: bool = False
+    ) -> Optional["Series"]:
+        """
+        Set the name of the axis for the index or columns.
+
+        Parameters
+        ----------
+        mapper, index :  scalar, list-like, dict-like or function, optional
+            A scalar, list-like, dict-like or functions transformations to
+            apply to the index values.
```
itholic, line 1128: Maybe this it's also not correct? I think you can refer to [here](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.rename_axis.html#pandas-series-rename-axis).

databricks/koalas/series.py
```diff
@@ -1115,6 +1115,71 @@ def rename(self, index=None, **kwargs):
...
+            Modifies the object directly, instead of creating a new Series.
+
+        Returns
+        -------
+        Series, or None if `inplace` is True.
+
+        See Also
+        --------
+        Series.rename : Alter Series index labels or name.
+        Index.rename : Set new names on index.
```
itholic, line 1139: Maybe we can also have `DataFrame.rename` here ?

databricks/koalas/series.py
```diff
@@ -1115,6 +1115,71 @@ def rename(self, index=None, **kwargs):
...
+                monkey    2
+        Name: num_legs, dtype: int64
+        >>> s.rename_axis(index=str.upper).sort_index()  # doctest: +NORMALIZE_WHITESPACE
+        TYPE    NAME
+        mammal  cat       4
+                dog       4
+                monkey    2
+        Name: num_legs, dtype: int64
+        """
+        return first_series(self.to_frame().rename_axis(mapper=mapper, index=index))
```
itholic, line 1181: Maybe `inplace=inplace` is missing?

databricks/koalas/tests/test_dataframe.py
```diff
@@ -581,6 +581,132 @@ def mul10(x) -> int:
...
+        self.assert_eq(
+            pdf.rename_axis(["index2"], axis=1).sort_index(),
+            kdf.rename_axis(["index2"], axis=1).sort_index(),
+        )
+
+        self.assertRaises(ValueError, lambda: kdf.rename_axis(["index2", "index3"], axis=0))
+        self.assertRaises(ValueError, lambda: kdf.rename_axis(["cols2", "cols3"], axis=1))
+
+        # index/columns parameters and dict_like/functions mappers introduced in pandas 0.24.0
+        if LooseVersion(pd.__version__) >= LooseVersion("0.24.0"):
```
itholic, line 615: Would you make tests for `pandas < 0.24.0` by manually declaring the expected result like we did before ??

databricks/koalas/tests/test_dataframe.py
```diff
@@ -581,6 +581,132 @@ def mul10(x) -> int:
...
+
+        self.assertRaises(
+            ValueError, lambda: kdf.rename_axis(["index3", "index4", "index5"], axis=0)
+        )
+        self.assertRaises(
+            ValueError, lambda: kdf.rename_axis(["columns3", "columns4", "columns5"], axis=1)
+        )
+
+        # index/columns parameters and dict_like/functions mappers introduced in pandas 0.24.0
+        if LooseVersion(pd.__version__) >= LooseVersion("0.24.0"):
```
itholic, line 675: ditto ?

itholic (approved):
Otherwise, Seems fine to me.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -581,6 +581,132 @@ def mul10(x) -> int:
...
+            )
+
+        self.assert_eq(
+            pdf.rename_axis(["index2"]).sort_index(), kdf.rename_axis(["index2"]).sort_index(),
+        )
+
+        self.assert_eq(
+            pdf.rename_axis(["index2"], axis=1).sort_index(),
+            kdf.rename_axis(["index2"], axis=1).sort_index(),
+        )
```
itholic, line 609: Why don't we just add tests to the above??

For example,

```python
        for axis in [0, "index"]:
            self.assert_eq(
                pdf.rename_axis("index2", axis=axis).sort_index(),
                kdf.rename_axis("index2", axis=axis).sort_index(),
            )
            self.assert_eq(
                pdf.rename_axis(["index2"], axis=axis).sort_index(),
                kdf.rename_axis(["index2"], axis=axis).sort_index(),
            )

        for axis in [1, "columns"]:
            self.assert_eq(
                pdf.rename_axis("cols2", axis=axis).sort_index(),
                kdf.rename_axis("cols2", axis=axis).sort_index(),
            )
            self.assert_eq(
                pdf.rename_axis(["cols2"], axis=axis).sort_index(),
                kdf.rename_axis(["cols2"], axis=axis).sort_index(),
            )
```

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,179 @@ def gen_new_column_labels_entry(column_labels_entry):
...
+        spark_frame = self._internal.resolved_copy.spark_frame
+        internal = InternalFrame(
+            spark_frame=spark_frame,
+            index_map=index_map,
+            column_labels=self._internal.column_labels,
+            data_spark_columns=[
+                scol_for(spark_frame, col) for col in self._internal.data_spark_column_names
+            ],
+            column_label_names=column_label_names,
+        )
```
itholic, line 9290: Oh I think we don't need to create a new `InternalFrame` and maybe can just simply copy the `index_map` and `column_label_names` here ??

Because this API updates only the name of index or columns rather than any other features.

```python
            index_map=index_map,
            column_label_names=column_label_names,
        )
```

databricks/koalas/tests/test_series.py
itholic: Cool, @LucasG0 . But maybe I think we have a bit more better way ?

Could you look at https://github.com/databricks/koalas/pull/1843#discussion_r507027779 ?

databricks/koalas/frame.py
LucasG0: Yes, I should have precised that pandas does not support using a dict or a function as `mapper`. They say :  
`However, if mapper is dict-like or a function, it will use the deprecated behavior of modifying the axis labels.`
So, I am not sure this is the correct behavior, that is why I added the possibility to use dict / function as `mapper`, and therefore updated the docs.
 I am interested in your opinion on this !

databricks/koalas/frame.py
LucasG0: Yes I did not put them as I split` Series`/`DataFrame ` docs unlike pandas, but I will add them back. :)

databricks/koalas/series.py
LucasG0: This is for the same reason I explained above.

databricks/koalas/series.py
LucasG0: Yes indeed !

databricks/koalas/frame.py
LucasG0: According to [mypy docs](https://mypy.readthedocs.io/en/stable/getting_started.html#the-typing-module) : `Optional[str] is just a shorthand or alias for Union[str, None]. It exists mostly as a convenience to help function signatures look a little cleaner`
So should I let `Optional["DataFrame"]` ?

databricks/koalas/frame.py
LucasG0: Definitely !

databricks/koalas/frame.py
itholic: Oh, I see. Then shall we fix also `Series.sort_values` from `Union["Series", None]` to `Optional["Series"]` ??

LucasG0:
By the way, I actually do not think anymore that a copy of the underlying data is performed in this implementation. However, I am not sure copying the immutable `spark_frame` is relevant here, so I wonder if we should keep `copy` parameter.

databricks/koalas/frame.py
LucasG0: I guess so ! I fixed it

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,170 @@ def gen_new_column_labels_entry(column_labels_entry):
...
+        index_names = (
+            gen_names(index, self.index.names) if index is not None else self._internal.index_names
+        )
+        index_map = OrderedDict(zip(self._internal.index_spark_column_names, index_names))
+
+        internal = self._internal.copy(index_map=index_map, column_label_names=column_label_names)
+        if inplace:
+            self._update_internal_frame(internal)
+            return None
```
HyukjinKwon, line 9285: nit: I think we don't have to return `None`.

HyukjinKwon (approved):

databricks/koalas/frame.py
```diff
@@ -9122,6 +9122,170 @@ def gen_new_column_labels_entry(column_labels_entry):
...
+                    "either dict-like or function type."
+                )
+
+            if len(newnames) != len(curnames):
+                raise ValueError(
+                    "Length of new names must be {}, got {}".format(len(curnames), len(newnames))
+                )
+
+            return [
+                name if name is None or isinstance(name, tuple) else (name,) for name in newnames
```
ueshin, line 9259: nit: shall we use `is_name_like_tuple`? `[name if is_name_like_tuple(name) else (name,) for name in newnames]`

ueshin (approved):
Otherwise, LGTM.

databricks/koalas/frame.py
LucasG0: Updated :)

LucasG0:
Thanks !

PR #1845, opened 2020-10-13
Remove the upper bound of matplotlib

HyukjinKwon:
This PR removes the upper bound of matplotlib in the dependency.

The test failures because of `manage_xticks`. This was [deprecated at 3.1.0](https://matplotlib.org/3.1.1/api/api_changes.html#boxplot-tick-methods) and [removed at 3.3.0](https://matplotlib.org/api/api_changes.html#arguments).

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -167,7 +167,10 @@ class KoalasBoxPlot(BoxPlot):
         meanprops=None,
         capprops=None,
         whiskerprops=None,
-        manage_xticks=True,
+        manage_ticks=None,
+        # manage_xticks is for compatibility of matplotlib < 3.1.0.
+        # Remove this when minimum version is 3.0.0
+        manage_xticks=None,
         autorange=False,
         zorder=None,
         precision=None,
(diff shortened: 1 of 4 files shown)


databricks/koalas/plot.py
```diff
@@ -265,6 +268,11 @@ def update_dict(dictionary, rc_name, properties):
                             if ci[1] is not None:
                                 stats["cihi"] = ci[1]
 
+        if LooseVersion(matplotlib.__version__) < LooseVersion("3.1.0"):
+            extra_args = {"manage_xticks": bool(manage_ticks or manage_xticks)}
+        else:
+            extra_args = {"manage_ticks": bool(manage_ticks or manage_xticks)}
```
HyukjinKwon, line 274: We could make it more fine-grained like considering precedence e.g.) `False` and `True` -> `False`. But I think it's too much ..

HyukjinKwon:
Ah .. it starts to fail today. Will take a look.

HyukjinKwon:
Let me just merge this in @ueshin and @itholic.

PR #1846, opened 2020-10-13
Calculate fliers correctly in Series.plot.box

### LINKED ISSUE #1827: Series Boxplot
Hey, in the Series Boxplot you show the top 1k absolute values.
However, may be some values which are fliers but don't have a big absolute value.
In my opinion, you should take the absolute value of the **difference**.

https://github.com/databricks/koalas/blob/b0e1c41389cd5ec09858cbf2a2f94ea38672a1bb/databricks/koalas/plot.py#L479-L492

I will illustrate another issue by a screen shot -

image

Thanks a lot :)

HyukjinKwon:
Koalas:

```python
from matplotlib import pyplot as plt
import pandas as pd
pdf = pd.DataFrame({"a": [1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 50],}, index=[0, 1, 3, 5, 6, 8, 9, 9, 9, 10, 10])
ks.from_pandas(-pdf).a.plot.box()
plt.savefig('koalas_series_box.png')
```

pandas:

```python
from matplotlib import pyplot as plt
import pandas as pd
pdf = pd.DataFrame({"a": [1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 50],}, index=[0, 1, 3, 5, 6, 8, 9, 9, 9, 10, 10])
(-pdf).a.plot.box()
plt.savefig('pandas_series_box.png')
```

pandas_series_box

**Before:**

koalas_series_box

**After:**

koalas_series_box

Resolves #1827

### DIFF
--- a/databricks/koalas/plot.py
+++ b/databricks/koalas/plot.py
@@ -326,7 +326,7 @@ class KoalasBoxPlot(BoxPlot):
         whiskers = KoalasBoxPlot._calc_whiskers(spark_column_name, outliers)
 
         if showfliers:
-            fliers = KoalasBoxPlot._get_fliers(spark_column_name, outliers)
+            fliers = KoalasBoxPlot._get_fliers(spark_column_name, outliers, whiskers[0])
         else:
             fliers = []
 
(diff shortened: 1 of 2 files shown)


itholic:
Oh, Nice catch !

ueshin:
I tried with the original example at #1827 and got the following plot:

download

This is different from pandas one:

download-1

This might be a different issue than #1827, though.

HyukjinKwon:
@ueshin, I think the problem there is because Spark picks the actual data as median 25%, 50% and 75% from `approx_percentile`  IIRC. It works out of the box when data is a lot.  FYI, using `percentile`  is very expensive in Apache Spark so I think `approx_percentile` was chosen.

ueshin (approved):
LGTM.

PR #1857, opened 2020-10-16
Add return type of to_koalas and to_pandas

xinrong-databricks:
Add return type of to_koalas and to_pandas

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -27,7 +27,19 @@ import types
 from functools import partial, reduce
 import sys
 from itertools import zip_longest
-from typing import Any, Optional, List, Tuple, Union, Generic, TypeVar, Iterable, Dict, Callable
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Optional,
+    Generic,
+    Iterable,
+    List,
+    Tuple,
+    TYPE_CHECKING,
+    TypeVar,
+    Union,
+)
 
 import numpy as np
 import pandas as pd
@@ -303,6 +315,9 @@ rectangle    16.0  2.348543e+108
 """
 
 T = TypeVar("T")
+
+if TYPE_CHECKING:
+    from pandas.core.frame import DataFrame as Pandas_DataFrame
 
 
 def _create_tuple_for_frame_type(params):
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -4250,7 +4265,7 @@ def to_spark(self, index_col: Optional[Union[str, List[str]]] = None):
 
     to_spark.__doc__ = SparkFrameMethods.__doc__
 
-    def to_pandas(self):
+    def to_pandas(self) -> "Pandas_DataFrame":
```
ueshin, line 4268: I guess `-> pd.DataFrame` works?

ueshin: Could you also add it to `Series.to_pandas()`?

ueshin (commented):

databricks/koalas/frame.py
```diff
@@ -27,7 +27,18 @@
...
+    Callable,
+    Dict,
+    Optional,
+    Generic,
+    Iterable,
+    List,
+    Tuple,
+    TypeVar,
+    Union,
+)
```
ueshin, line 41: Shall we revert this change?

databricks/koalas/frame.py
```diff
@@ -4250,7 +4261,7 @@ def to_spark(self, index_col: Optional[Union[str, List[str]]] = None):
 
     to_spark.__doc__ = SparkFrameMethods.__doc__
 
-    def to_pandas(self):
+    def to_pandas(self) -> "pd.DataFrame":
```
ueshin, line 4264: I don't think we need to surround with double-quotes.

xinrong-databricks: `dev/reformat` seem to change `'pd.DataFrame'` to double quotes after each run.
Shall we disable it?

databricks/koalas/frame.py
xinrong-databricks: Reverted.

databricks/koalas/frame.py
xinrong-databricks: Certainly! Updated.

databricks/koalas/frame.py
ueshin: ah, sorry, I mean `def to_pandas(self) -> pd.DataFrame:` should be fine?

xinrong-databricks: Sorry, I misunderstood. Updated. It works great!

ueshin (approved):

ueshin:
Thanks! merging.

xinrong-databricks:
@ueshin Thank you!

PR #1860, opened 2020-10-21
Upgrade PyArrow from 1.0.1 to 2.0.0 in CI

itholic:
This PR proposes upgrade PyArrow from 1.0.1 to 2.0.0 in CI since PyArrow 2.0.0 is released now.


.github/workflows/master.yml
```diff
@@ -117,7 +117,7 @@ jobs:
           - python-version: 3.8
```
HyukjinKwon, line 117: @itholic can we change ` pyarrow-version: 0.15.1` above to ` pyarrow-version: 1.0.1`? I think we should also test PyArrow 1.x as well.

HyukjinKwon:
Looks good otherwise.

.github/workflows/master.yml
itholic: Sounds good! Thanks :)

HyukjinKwon (approved):

HyukjinKwon:
Actually @itholic just for completeness, can you set `PYARROW_IGNORE_TIMEZONE` automatically? To keep the behaviour same looks like we should set this env for Arrow 2.0.0, see also https://github.com/apache/spark/commit/47a6568265525002021c1e5cfa4330f5b1a91469#.

I did similar stuff before at https://github.com/databricks/koalas/commit/05b93247c6384cbafbb818cd3e0b3682966a5fd8. You could refer this PR.

I was thinking we can just ignore it since we haven't officially supported nested timestamps with localization but .. after rethinking, I think it's just best to match with the Apache Spark upstream. cc @ueshin too.

itholic:
Thanks, @HyukjinKwon . Sorry for the late.
Just make it to set the `PYARROW_IGNORE_TIMEZONE` automatically to 1 in `pyarrow` >=2.0.0.

.github/workflows/master.yml
```diff
@@ -125,6 +125,8 @@ jobs:
       GITHUB_OAUTH_KEY: ${{ secrets.GITHUB_TOKEN }}
       # Github Actions' default miniconda
       CONDA_PREFIX: /usr/share/miniconda
+      # Preserve legacy nested timezone behavior for pyarrow>=2, remove after SPARK-32285
+      PYARROW_IGNORE_TIMEZONE: "1",
```
ueshin, line 129: Do we need this?

ueshin (commented):
@itholic Could you push another commit?
Seems like GitHub Action didn't run tests.

.github/workflows/master.yml
itholic: Removed it, thanks for the catch!

databricks/koalas/__init__.py
```diff
@@ -70,6 +70,26 @@ def assert_pyspark_version():
...
+        "Koalas will set it for you but it does not work if there is a Spark context already "
+        "launched."
+    )
+    os.environ["PYARROW_IGNORE_TIMEZONE"] = "1"
+elif "PYARROW_IGNORE_TIMEZONE" in os.environ:
+    raise RuntimeError(
+        "Please explicitly unset 'PYARROW_IGNORE_TIMEZONE' environment variable in both "
+        "driver and executor sides. It is required to set this environment variable only "
+        "when you use pyarrow>=2.0.0."
+    )
```
ueshin, line 92: Do we need to raise the error?

ueshin (approved):
Otherwise, LGTM. cc @HyukjinKwon

databricks/koalas/__init__.py
itholic: Hm... I don't think it's necessary. Will remove it

HyukjinKwon (approved):

ueshin:
Thanks! merging.

PR #1861, opened 2020-10-21
Fix Python version and Koalas version in installation docs

itholic:
This PR proposes:
- Fix supported Python version from `Python 3.5 and above` to `Python 3.5 to 3.8` since we doesn't support Python 3.9 yet.
- Fix Koalas version in installation guide from `0.19.0` to `1.3.0` since it's too old.

### DIFF
--- a/docs/source/getting_started/install.rst
+++ b/docs/source/getting_started/install.rst
@@ -21,7 +21,7 @@ To install PySpark, you can use:
 Python version support
 ----------------------
 
-Officially Python 3.5 and above.
+Officially Python 3.5 to 3.8.
 
 
 Installing Koalas
@@ -49,7 +49,7 @@ following command::
 
 To install a specific Koalas version::
 
-    conda install -c conda-forge koalas=0.19.0
+    conda install -c conda-forge koalas=1.3.0
 
 
 Installing from PyPI


docs/source/getting_started/install.rst
```diff
@@ -21,7 +21,7 @@ To install PySpark, you can use:
 Python version support
 ----------------------
 
-Officially Python 3.5 and above.
+Officially Python 3.5 to 3.8.
```
HyukjinKwon, line 24: Okay, seems like this is what we did in `setup.py`.

docs/source/getting_started/install.rst
```diff
@@ -49,7 +49,7 @@ following command::
 
 To install a specific Koalas version::
 
-    conda install -c conda-forge koalas=0.19.0
+    conda install -c conda-forge koalas=1.3.0
```
HyukjinKwon, line 52: This is just an example of how to specify the version so it doesn't matter if this is old or not. But okay since you already opened a PR.

HyukjinKwon (approved):

docs/source/getting_started/install.rst
itholic: Yep, I also think It's not important but I think maybe someone who very first to Koalas could just copy this line and paste without any concern. 😂

docs/source/getting_started/install.rst
itholic: Right. maybe someone just use Python 3.9 and got confused.

PR #1862, opened 2020-10-22
Introduce spark.analyzed to force resolve InternalFrame.

ueshin:
Introduces `spark.analyzed` to force resolve `InternalFrame`.

After multiple operations, the underlying Spark plan could grow huge and make the Spark planner take a long time to finish the planning.
This function is for the workaround to avoid it.

### DIFF
--- a/databricks/koalas/spark/accessors.py
+++ b/databricks/koalas/spark/accessors.py
@@ -196,6 +196,59 @@ class SparkSeriesMethods(SparkIndexOpsMethods):
         # Lose index.
         return first_series(DataFrame(sdf)).rename(self._data.name)
 
+    @property
+    def analyzed(self) -> "ks.Series":
+        """
+        Returns a new Series with the analyzed Spark DataFrame.
+
+        After multiple operations, the underlying Spark plan could grow huge
+        and make the Spark planner take a long time to finish the planning.
+
+        This function is for the workaround to avoid it.
+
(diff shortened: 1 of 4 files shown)


databricks/koalas/spark/accessors.py
```diff
@@ -851,6 +903,58 @@ def apply(self, func, index_col: Optional[Union[str, List[str]]] = None) -> "ks.
             )
         return output.to_koalas(index_col)
 
+    @property
+    def analyzed(self) -> "ks.DataFrame":
```
HyukjinKwon, line 907: @ueshin, should we maybe just expose Spark's `checkpoint` or `localCheckpoint` instead?

ueshin: As it will invoke another Spark job, I think we should have them separately.

docs/source/reference/frame.rst
```diff
@@ -276,6 +276,7 @@ in Spark. These can be accessed by ``DataFrame.spark.<function/property>``.
    DataFrame.spark.to_spark_io
    DataFrame.spark.explain
    DataFrame.spark.apply
+   DataFrame.spark.analyzed
```
HyukjinKwon, line 279: Maybe we could keep it internal and don't document.

HyukjinKwon (approved):
LGTM otherwise.

PR #1864, opened 2020-10-22
Introduce df.spark.repartition

xinrong-databricks:


databricks/koalas/spark/accessors.py
```diff
@@ -851,6 +851,27 @@ def apply(self, func, index_col: Optional[Union[str, List[str]]] = None) -> "ks.
             )
         return output.to_koalas(index_col)
 
+    def repartition(
+        self, numPartitions: int, index_col: Optional[Union[str, List[str]]] = None
```
ueshin, line 855: `num_partitions` instead of `numPartitions`?

databricks/koalas/frame.py
```diff
@@ -9286,6 +9286,13 @@ def gen_names(v, curnames):
         else:
             return DataFrame(internal)
 
+    def repartition(
+        self, numPartitions: Union[int, str], index_col: Optional[Union[str, List[str]]] = None
+    ) -> "ks.DataFrame":
+        return self.spark.repartition(numPartitions, index_col)
+
+    repartition.__doc__ = SparkFrameMethods.repartition.__doc__
+
```
ueshin, line 9295: I don't think we need to add it to `DataFrame`.

databricks/koalas/spark/accessors.py
ueshin: I think this should work without `index_col`, returns the same DataFrame except for partitioning.

databricks/koalas/spark/accessors.py
```diff
@@ -851,6 +851,27 @@ def apply(self, func, index_col: Optional[Union[str, List[str]]] = None) -> "ks.
...
+        numPartitions : int
+            The target number of partitions.
+        index_col: str or list of str, optional, default: None
+            Column names to be used in Spark to represent Koalas' index. The index name
+            in Koalas is ignored. By default, the index is always lost.
+
+        Returns
+        -------
+        DataFrame
+        """
```
ueshin, line 872: Could you add simple examples?

databricks/koalas/tests/test_dataframe.py
```diff
@@ -4546,6 +4546,17 @@ def test_explain_hint(self):
         finally:
             sys.stdout = prev
 
+    def test_repartition(self):
+        kdf = ks.DataFrame({"age": [5, 5, 2, 2], "name": ["Bob", "Bob", "Alice", "Alice"]})
+
+        new_kdf = kdf.spark.repartition(7)
+        self.assertEqual(new_kdf.to_spark().rdd.getNumPartitions(), 7)
```
ueshin, line 4553: Shall we also test `self.assert_eq(kdf.sort_index(), new_kdf.sort_index())`?

databricks/koalas/tests/test_dataframe.py
```diff
@@ -4546,6 +4546,17 @@ def test_explain_hint(self):
...
+
+        kdf = kdf.set_index("age")
+        new_kdf = kdf.spark.repartition(5)
+        self.assertEqual(new_kdf.to_spark().rdd.getNumPartitions(), 5)
+        self.assertEqual(kdf.index.name, new_kdf.index.name)
```
ueshin, line 4558: Could you add tests with mutli-index columns?

ueshin, line 4558: Shall we put this test in `test_frame_spark`?

databricks/koalas/spark/accessors.py
xinrong-databricks: Certainly! Added.

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Sounds good!

xinrong-databricks: Certainly, added.

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Added!

databricks/koalas/spark/accessors.py
```diff
@@ -851,6 +851,48 @@ def apply(self, func, index_col: Optional[Union[str, List[str]]] = None) -> "ks.
...
+            The target number of partitions.
+
+        Returns
+        -------
+        DataFrame
+
+        Examples
+        --------
+        >>> import databricks.koalas as ks
+        >>> import pyspark
```
ueshin, line 871: nit: we don't need to import those.

xinrong-databricks: Removed.

databricks/koalas/spark/accessors.py
```diff
@@ -851,6 +851,50 @@ def apply(self, func, index_col: Optional[Union[str, List[str]]] = None) -> "ks.
...
+
+        Returns
+        -------
+        DataFrame
+
+        Examples
+        --------
+        >>> kdf = ks.DataFrame({"age": [5, 5, 2, 2],
+        ...         "name": ["Bob", "Bob", "Alice", "Alice"]}).set_index("age")
+        >>> kdf  # doctest: +NORMALIZE_WHITESPACE
```
ueshin, line 872: Shall we use `.sort_index()` here as well so that we can see the same as the repartitioned result?

ueshin (commented):

HyukjinKwon (approved):
LGTM too. We can also add coalesce I guess :-). Also this should be added into doc.

databricks/koalas/spark/accessors.py
xinrong-databricks: Makes sense! Updated.

xinrong-databricks:
@HyukjinKwon Sounds good! Thank you for your approval!

PR #1866, opened 2020-10-26
Set the lowerbound of numpydoc

HyukjinKwon:
This PR proposes to set the lower bound of numpydoc. The main reason to set the bound to 0.8.0 was that `self` argument appeared in the API references (see https://github.com/databricks/koalas/issues/439). This seems fixed now from 1.1.0+

There are a couple of methods not being found in "See Also" via numpydoc but I think they are trivial issues.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -2169,7 +2169,6 @@ def isna(obj):
 
     See Also
     --------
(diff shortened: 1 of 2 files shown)


HyukjinKwon:
Let me merge this @ueshin. I think I am pretty sure about this change.

PR #1867, opened 2020-10-26
Ignore docs/mlruns directory generated during doc creation

HyukjinKwon:
Currently `docs/mlruns` directory can be generated during documentation generation for some reasons. This PR adds it to the `.gitignore`.


HyukjinKwon:
I checked together with @ueshin. Let me merge this - seems like this is new behaviour from latest MLflow (at least 1.11.0).

PR #1869, opened 2020-10-27
Ignore mlruns directory globally

HyukjinKwon:
Looks like mlflow creates that directory when it imports:

```python
>>> import mlflow
```

```bash
ls -al mlruns
```
```
total 0
drwxr-xr-x   4 hyukjin.kwon  staff  128 Oct 27 11:42 .
drwxr-xr-x@ 30 hyukjin.kwon  staff  960 Oct 27 11:42 ..
drwxr-xr-x   2 hyukjin.kwon  staff   64 Oct 27 11:42 .trash
drwxr-xr-x   3 hyukjin.kwon  staff   96 Oct 27 11:42 0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,4 +59,4 @@ pandoc*
 whatsnew
 
 # Mlflow directory during documentation generation
-docs/mlruns
+mlruns


HyukjinKwon:
@harupy FYI :-). Seems like there's new behaviour in latest mlflow - creates a directory when it initially imports.

ueshin:
Thanks! merging.

PR #1875, opened 2020-10-28
Enable no-strict-optional of mypy

xinrong-databricks:
## Proposal
Enable no-strict-optional of mypy

## Motivation
For methods with inplace parameter, their return types are mostly of Optional type. For example:

    def sort_values(
        self,
        by: Union[Any, List[Any], Tuple, List[Tuple]],
        ascending: Union[bool, List[bool]] = True,
        inplace: bool = False,
        na_position: str = "last",
    ) -> Optional["DataFrame"]:

Even if we call these methods with inplace as False, i.e., we are promised to get a new DataFrame, mypy will still complain about None. For example, in 

    def nsmallest(self, n: int, columns: "Any") -> "DataFrame":
        return self.sort_values(by=columns, ascending=True).head(n=n) 

Mypy checks will fail:
databricks/koalas/frame.py: note: In member "nlargest" of class "DataFrame":
databricks/koalas/frame.py:6408: error: Item "None" of "Optional[DataFrame[Any]]" has no attribute "head"

Our current approach is to add type comment “#type: ignore”. There are several disadvantages:
- Not type check for the whole line (if want to ignore type check for a method only, code refactoring is needed)
- Confuse people
- Not trivial considering the growing number of them

Enabling no-strict-optional of mypy will mitigate this problem.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -6739,7 +6739,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                     )
                 left_key_names = list(map(self._internal.spark_column_name_for, to_list(common)))
                 right_key_names = list(map(right._internal.spark_column_name_for, to_list(common)))
-            if len(left_key_names) != len(right_key_names):  # type: ignore
+            if len(left_key_names) != len(right_key_names):
                 raise ValueError("len(left_keys) must equal len(right_keys)")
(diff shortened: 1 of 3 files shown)


ueshin (approved):
LGTM.

HyukjinKwon (approved):

PR #1876, opened 2020-10-28
DataFrame.reindex supports named columns index

### LINKED ISSUE #1872: DataFrame.reindex does not support named columns index
Hi, it seems currently not possible to reindex a `DataFrame` with a named columns `Index`, the new name is missing.

```python
>>> pdf = pd.DataFrame({"values": [1, 2, 3]})
>>> pdf
   values
0       1
1       2
2       3
>>> kdf = ks.from_pandas(pdf)
>>> columns = pd.Index(["values"], name="name")
>>> kdf.reindex(columns=columns)
   values
0       1
1       2
2       3
```

But in pandas :

```python
>>> pdf.reindex(columns=columns)
name  values
0          1
1          2
2          3
```

I will submit a PR for this.

LucasG0:
This PR would close #1872.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7987,7 +7987,17 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 scols.append(F.lit(fill_value).alias(name_like_string(label)))
             labels.append(label)
 
-        return DataFrame(self._internal.with_new_columns(scols, column_labels=labels))
+        if isinstance(columns, pd.Index):
+            column_label_names = [
+                name if is_name_like_tuple(name) else (name,) for name in columns.names
+            ]
+            internal = self._internal.with_new_columns(
+                scols, column_labels=labels, column_label_names=column_label_names
+            )
+        else:
+            internal = self._internal.with_new_columns(scols, column_labels=labels)
+
+        return DataFrame(internal)
 
     def melt(self, id_vars=None, value_vars=None, var_name=None, value_name="value") -> "DataFrame":
         """
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -2923,9 +2923,18 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
 
     def test_reindex(self):
         index = pd.Index(["A", "B", "C", "D", "E"])
-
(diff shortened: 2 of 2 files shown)


databricks/koalas/tests/test_dataframe.py
```diff
@@ -2923,8 +2923,8 @@ def test_drop_duplicates(self):
 
     def test_reindex(self):
         index = pd.Index(["A", "B", "C", "D", "E"])
-
-        pdf = pd.DataFrame({"numbers": [1.0, 2.0, 3.0, 4.0, None]}, index=index)
+        columns = pd.Index(["numbers"], name="cols")
```
xinrong-databricks, line 2926: Shall we reserve one test case for "when the original column is not named"?

databricks/koalas/frame.py
```diff
@@ -7987,7 +7987,12 @@ def _reindex_columns(self, columns, fill_value):
                 scols.append(F.lit(fill_value).alias(name_like_string(label)))
             labels.append(label)
 
-        return DataFrame(self._internal.with_new_columns(scols, column_labels=labels))
+        df = DataFrame(self._internal.with_new_columns(scols, column_labels=labels))
+
+        if isinstance(columns, pd.Index):
+            return df.rename_axis(columns=columns.names)
```
xinrong-databricks, line 7993: How about `df.rename_axis(columns=columns.names, inplace=True)` and returning `df` instead? I believe this would save us from creating a new DataFrame.

ueshin: We should rather specify `columns.names` when `self._internal.with_new_columns`.

LucasG0: Yes indeed !

databricks/koalas/tests/test_dataframe.py
LucasG0: Yes, I also added a test reindexing with a non named column.

xinrong-databricks: Great!

xinrong-databricks:
LGTM.

xinrong-databricks:
Thank you for raising the issue and the solution!

xinrong-databricks (approved):

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

itholic:
Nice !

PR #1880, opened 2020-11-01
Implement DataFrame/Series reindex_like

LucasG0:
Hi, this PR implements both `DataFrame.reindex_like` and `Series.reindex_like`.
I did not add a test about reindexing MultiIndex columns on single Index columns because it is currently not supported in koalas. Also, as for `DataFrame.reindex`, there is no test for reindexing single Index columns/index on MultiIndex columns/index, as it is not supported in pandas.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7992,9 +7992,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         """
         Return a DataFrame with matching indices as other object.
 
-        Conform the object to the same index on all axes. Optional
-        filling logic, placing NaN in locations having no value
-        in the previous index. A new object is produced unless the
+        Conform the object to the same index on all axes. Places NA/NaN in locations
+        having no value in the previous index. A new object is produced unless the
         new index is equivalent to the current one and copy=False.
 
         Parameters
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7988,6 +7988,82 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         return DataFrame(internal)
 
+    def reindex_like(self: "DataFrame", other: "DataFrame", copy: bool = True) -> "DataFrame":
+        """
+        Return a DataFrame with matching indices as other object.
+
+        Conform the object to the same index on all axes. Optional
+        filling logic, placing NaN in locations having no value
+        in the previous index. A new object is produced unless the
+        new index is equivalent to the current one and copy=False.
+
+        Parameters
+        ----------
+        other : DataFrame
+            Its row and column indices are used to define the new indices
+            of this object.
+        copy : bool, default True
+            Return a new object, even if the passed indexes are the same.
+
+        Returns
+        -------
+        DataFrame
+            DataFrame with changed indices on each axis.
+
+        See Also
+        --------
+        DataFrame.set_index : Set row labels.
+        DataFrame.reset_index : Remove row labels or move them to new columns.
+        DataFrame.reindex : Change to new indices or expand indices.
+
+        Notes
+        -----
+        Same as calling
+        ``.reindex(index=other.index, columns=other.columns,...)``.
+
+        Examples
+        --------
+
+        >>> df1 = ks.DataFrame([[24.3, 75.7, 'high'],
+        ...                     [31, 87.8, 'high'],
+        ...                     [22, 71.6, 'medium'],
+        ...                     [35, 95, 'medium']],
+        ...                    columns=['temp_celsius', 'temp_fahrenheit',
+        ...                             'windspeed'],
+        ...                    index=pd.date_range(start='2014-02-12',
+        ...                                        end='2014-02-15', freq='D'))
+        >>> df1
+                    temp_celsius  temp_fahrenheit windspeed
+        2014-02-12          24.3             75.7      high
+        2014-02-13          31.0             87.8      high
+        2014-02-14          22.0             71.6    medium
+        2014-02-15          35.0             95.0    medium
+
+        >>> df2 = ks.DataFrame([[28, 'low'],
+        ...                     [30, 'low'],
+        ...                     [35.1, 'medium']],
+        ...                    columns=['temp_celsius', 'windspeed'],
+        ...                    index=pd.DatetimeIndex(['2014-02-12', '2014-02-13',
+        ...                                            '2014-02-15']))
+        >>> df2
+                    temp_celsius windspeed
+        2014-02-12          28.0       low
+        2014-02-13          30.0       low
+        2014-02-15          35.1    medium
+
+        >>> df2.reindex_like(df1).sort_index() # doctest: +NORMALIZE_WHITESPACE
+                    temp_celsius  temp_fahrenheit windspeed
+        2014-02-12          28.0              NaN       low
+        2014-02-13          30.0              NaN       low
+        2014-02-14           NaN              NaN       None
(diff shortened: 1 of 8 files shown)


itholic:
Nice work, @LucasG0 ! 
Let me take a look after test passing :)

LucasG0:
Despite Codecov report, coverage seems actually ok.

databricks/koalas/series.py
```diff
@@ -1748,6 +1748,72 @@ def reindex(self, index: Optional[Any] = None, fill_value: Optional[Any] = None,
             self.name
         )
 
+    def reindex_like(self: "Series", other: Union["Series", "DataFrame"]) -> "Series":
+        """
+        Return a Series with matching indices as other object.
+
+        Conform the object to the same index on all axes. Optional
+        filling logic, placing NaN in locations having no value
+        in the previous index.
```
itholic, line 1757: ditto ?

databricks/koalas/series.py
```diff
@@ -1748,6 +1748,72 @@ def reindex(self, index: Optional[Any] = None, fill_value: Optional[Any] = None,
...
+
+        Parameters
+        ----------
+        other : Series or DataFrame
+            Its row and column indices are used to define the new indices
+            of this object.
```
itholic, line 1763: Anyway, maybe pandas has `copy` parameter for `Series.reindex_like`, too ??

Even if it doesn't really do any meaningful work, how about adding it for compatibility with pandas?

databricks/koalas/series.py
```diff
@@ -1748,6 +1748,72 @@ def reindex(self, index: Optional[Any] = None, fill_value: Optional[Any] = None,
...
+        See Also
+        --------
+        DataFrame.set_index : Set row labels.
+        DataFrame.reset_index : Remove row labels or move them to new columns.
+        DataFrame.reindex : Change to new indices or expand indices.
+
+        Notes
+        -----
+        Same as calling
+        ``.reindex(index=other.index, columns=other.columns,...)``.
```
itholic, line 1779: I don't think maybe `Series.reindex` has a `columns` parameter ??

```python
>>> pser.reindex(index=pdf.index, columns=pdf.columns)
...
TypeError: reindex() got an unexpected keyword argument "columns"
```

databricks/koalas/tests/test_dataframe.py
```diff
@@ -3087,6 +3087,65 @@ def test_reindex(self):
...
+            [("A", "B"), ("C", "D"), ("E", "F")], names=["name1", "name2"]
+        )
+        pdf = pd.DataFrame(data=data, index=index, columns=columns)
+        kdf = ks.from_pandas(pdf)
+
+        self.assert_eq(
+            pdf.reindex_like(pdf2).sort_index(), kdf.reindex_like(kdf2).sort_index(),
+        )
+
+        self.assertRaises(TypeError, lambda: kdf.reindex_like(index2))
```
itholic, line 3147: How about add "Reindexing single Index on MultiIndex" for an exception case ??

databricks/koalas/tests/test_series.py
```diff
@@ -339,6 +339,62 @@ def test_reindex(self):
...
+            [("A", "B"), ("C", "D"), ("E", "F")], names=["name3", "name4"]
+        )
+        pdf = pd.DataFrame(data=data, index=index2)
+        kdf = ks.from_pandas(pdf)
+
+        self.assert_eq(
+            pser.reindex_like(pdf).sort_index(), kser.reindex_like(kdf).sort_index(),
+        )
+
+        self.assertRaises(TypeError, lambda: kser.reindex_like(index2))
```
itholic, line 396: ditto ?

itholic (approved):
Otherwise, seems fine to me since it's basically based on specific usage of `reindex`.

HyukjinKwon:
awesome, thanks @LucasG0 for working on this.

databricks/koalas/series.py
LucasG0: No indeed !

databricks/koalas/frame.py
```diff
@@ -7988,6 +7988,82 @@ def _reindex_columns(self, columns, fill_value):
...
         return DataFrame(internal)
 
+    def reindex_like(self: "DataFrame", other: "DataFrame", copy: bool = True) -> "DataFrame":
+        """
+        Return a DataFrame with matching indices as other object.
+
+        Conform the object to the same index on all axes. Optional
+        filling logic, placing NaN in locations having no value
+        in the previous index. A new object is produced unless the
+        new index is equivalent to the current one and copy=False.
```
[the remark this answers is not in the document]
LucasG0: In deed, it is from pandas doc, but I agree it seems a bit unclear as pandas does not support `fill_value` parameter.

databricks/koalas/series.py
LucasG0: Yes we can, so we should also add it to `Series.reindex`.

databricks/koalas/tests/test_dataframe.py
LucasG0: I added it !

databricks/koalas/series.py
LucasG0: Yes pandas has `copy` parameter for `Series.reindex_like`. If we add it, we need to update `Series.reindex` too, which does not support `copy` parameter either.
Should I do the change ?

itholic: Hmm... okay let's keep it as it is for now and discuss separately later.
Thanks for the opinion!

itholic:
Great, @LucasG0 !
Looks pretty good to me.
Waiting for others opinion for a while.

databricks/koalas/series.py
```diff
@@ -1748,6 +1748,71 @@ def reindex(self, index: Optional[Any] = None, fill_value: Optional[Any] = None,
             self.name
         )
 
+    def reindex_like(self: "Series", other: Union["Series", "DataFrame"]) -> "Series":
+        """
+        Return a Series with matching indices as other object.
+
+        Conform the object to the same index on all axes. Places NA/NaN in locations
+         having no value in the previous index.
```
ueshin, line 1756: We should remove an extra space before `having`; otherwise, the doc format breaks:

Screen Shot 2020-11-04 at 10 43 14 AM

databricks/koalas/frame.py
```diff
@@ -7988,6 +7988,81 @@ def _reindex_columns(self, columns, fill_value):
 
         return DataFrame(internal)
 
+    def reindex_like(self: "DataFrame", other: "DataFrame", copy: bool = True) -> "DataFrame":
```
ueshin, line 7991: nit: I don't think we need the type hint for `self`.

databricks/koalas/series.py
```diff
@@ -1748,6 +1748,71 @@ def reindex(self, index: Optional[Any] = None, fill_value: Optional[Any] = None,
             self.name
         )
 
+    def reindex_like(self: "Series", other: Union["Series", "DataFrame"]) -> "Series":
```
ueshin, line 1751: ditto.

ueshin (approved):
Otherwise, LGTM.

databricks/koalas/series.py
LucasG0: Oh yes thanks !

databricks/koalas/series.py
LucasG0: Ok :)

ueshin:
Thanks! merging.

PR #1882, opened 2020-11-02
Add type annotations to Index and MultiIndex

xinrong-databricks:


databricks/koalas/indexes.py
```diff
@@ -2297,12 +2308,12 @@ def from_tuples(tuples, sortorder=None, names=None):
                     (2, 'blue')],
                    names=['number', 'color'])
         """
-        return DataFrame(
+        return DataFrame(  # type: ignore
```
xinrong-databricks, line 2311: Otherwise, Mypy expects "Index" as the return type.

databricks/koalas/indexes.py
```diff
@@ -2331,12 +2342,12 @@ def from_arrays(arrays, sortorder=None, names=None):
...
```
xinrong-databricks, line 2345: Otherwise, Mypy expects "Index" as the return type.

databricks/koalas/indexes.py
```diff
@@ -2373,12 +2384,12 @@ def from_product(iterables, sortorder=None, names=None):
                     (2, 'purple')],
                    names=['number', 'color'])
         """
-        return DataFrame(
+        return DataFrame(  # type: ignore
```
xinrong-databricks, line 2387: Otherwise, Mypy expects "Index" as the return type.

databricks/koalas/indexes.py
```diff
@@ -2726,10 +2737,10 @@ def toPandas(self):
     toPandas.__doc__ = to_pandas.__doc__
 
     def nunique(self, dropna=True):
-        raise NotImplementedError("isna is not defined for MultiIndex")
+        raise NotImplementedError("nunique is not defined for MultiIndex")
 
     # TODO: add 'name' parameter after pd.MultiIndex.name is implemented
-    def copy(self, deep=None):
+    def copy(self, deep=None) -> "MultiIndex":  # type: ignore
```
xinrong-databricks, line 2743: Otherwise, Mypy complains it not compatible with superclass `copy`.

databricks/koalas/indexes.py
```diff
@@ -2759,9 +2770,9 @@ def copy(self, deep=None):
                     ('d', 'h')],
                    )
         """
-        return super().copy(deep=deep)
+        return super().copy(deep=deep)  # type: ignore
```
xinrong-databricks, line 2773: Otherwise, Mypy expects "Index" as the return type.

databricks/koalas/indexes.py
```diff
@@ -2520,10 +2531,10 @@ def swaplevel(self, i=-2, j=-1):
         index_map = list(self._internal.index_map.items())
         index_map[i], index_map[j], = index_map[j], index_map[i]
         result = DataFrame(self._kdf._internal.copy(index_map=OrderedDict(index_map))).index
-        return result
+        return result  # type: ignore
```
xinrong-databricks, line 2534: Otherwise, Mypy expects "Index" as the return type.

databricks/koalas/indexes.py
```diff
@@ -2444,7 +2455,7 @@ def from_frame(df, names=None):
 
         index_map = OrderedDict(zip(sdf.columns, names))
         internal = InternalFrame(spark_frame=sdf, index_map=index_map)
-        return DataFrame(internal).index
+        return DataFrame(internal).index  # type: ignore
```
xinrong-databricks, line 2458: Otherwise, Mypy expects "Index" as the return type.

databricks/koalas/indexes.py
```diff
@@ -1220,14 +1231,14 @@ def droplevel(self, level):
             )
 
         int_level = [n if isinstance(n, int) else names.index(n) for n in level]
-        index_map = list(self._internal.index_map.items())
-        index_map = OrderedDict(index_map[c] for c in range(0, nlevels) if c not in int_level)
+        index_map_list = list(self._internal.index_map.items())
```
itholic, line 1234: nit: really nothing, but how about `index_map_items` ?
Maybe `index_map_list` somewhat sounds like there are several `index_map`s in the list ?

databricks/koalas/indexes.py
itholic: Oh, so maybe it seems that `mypy` requires us to add a kind of abstract methods to the `Index` class ?

Maybe can we ignore that rule like we did for `inplace` parameter ??  😲

itholic (approved):
Seems fine to me except one [question](https://github.com/databricks/koalas/pull/1882#discussion_r516363647)

databricks/koalas/indexes.py
```diff
@@ -1365,13 +1376,13 @@ def sort_values(self, ascending=True):
...
 
-    def sort(self, *args, **kwargs):
+    def sort(self, *args, **kwargs) -> None:
         """
         Use sort_values instead.
         """
         raise TypeError("cannot sort an Index object in-place, use sort_values instead")
 
-    def min(self):
+    def min(self) -> Union[Scalar, Tuple]:
```
HyukjinKwon, line 1385: maybe -> `Union[Scalar, Tuple[Scalar, ...]]`

databricks/koalas/indexes.py
```diff
@@ -2077,7 +2088,7 @@ def holds_integer(self):
         """
         return isinstance(self.spark.data_type, IntegralType)
 
-    def item(self):
+    def item(self) -> Union[Scalar, Tuple]:
```
HyukjinKwon, line 2091: Actually using `Scalar` is strictly not correct because this can return a pandas instance too. For exmaple, `pd.Timestamp` which is not in the `Scalar` from my cursory reading. But I think it's okay - we can fix it later separately.

databricks/koalas/indexes.py
```diff
@@ -3096,7 +3110,7 @@ def insert(self, loc: int, item):
         internal = InternalFrame(spark_frame=sdf, index_map=self._internal.index_map,)
         return DataFrame(internal).index
 
-    def item(self):
+    def item(self) -> Tuple:
```
HyukjinKwon, line 3113: ditto: `Tuple[Scalar, ...]`

HyukjinKwon:
Looks pretty good

databricks/koalas/indexes.py
xinrong-databricks: Good idea! Renamed.

databricks/koalas/indexes.py
xinrong-databricks: Sounds good!

databricks/koalas/indexes.py
xinrong-databricks: Modified.

databricks/koalas/indexes.py
xinrong-databricks: Thanks @itholic, this is a good idea!

Unfortunately, I didn't find a parameter in https://mypy.readthedocs.io/en/stable/config_file.html can ignore this rule.

I do agree this should a common problem for other projects. Let me do more search and see if there is a better solution.

databricks/koalas/indexes.py
xinrong-databricks: This makes sense. How about we define a pandas Scalar separately? @HyukjinKwon

databricks/koalas/indexes.py
itholic: Yep, Sure :)

PR #1883, opened 2020-11-03
TypeError when Index.name / Series.name is not a hashable type

itholic:
When set a `name` for `Index` & `Series` or initiate the `Index` with `name`,
pandas raises `TypeError` if the given `name` is not a `hashable`, but Koalas doesn't.

```python
# set a unhashable `name` for `Index`.
>>> pidx = pd.Index([1, 2, 3])
>>> pidx.name = [1]
...
TypeError: Index.name must be a hashable type

# set a unhashable `name` for `Series`.
>>> pser = pd.Series([1, 2, 3])
>>> pser.name = [1]
...
TypeError: Series.name must be a hashable type

# Initiate an `Index` with unhashable `name`.
>>> pd.Index([1, 2, 3], name=[1])
...
TypeError: Index.name must be a hashable type
```


databricks/koalas/indexes.py
```diff
@@ -565,6 +567,8 @@ def name(self) -> Union[Any, Tuple]:
 
     @name.setter
     def name(self, name: Union[Any, Tuple]) -> None:
+        if not is_hashable(name):
```
HyukjinKwon, line 570: is there the same issue in series too?

itholic: Oh yeah. Just fixed Series, too. Thanks!

databricks/koalas/indexes.py
```diff
@@ -115,6 +115,8 @@ class Index(IndexOpsMixin):
     def __new__(cls, data: Union[DataFrame, list], dtype=None, name=None, names=None):
         assert data is not None
 
+        if not is_hashable(name):
+            raise TypeError("Index.name must be a hashable type")
```
ueshin, line 119: Do we need this?
Currently:

```py
>>> ks.Index([1, 2, 3], name=[1])
...
TypeError: Index.name must be a hashable type
```

databricks/koalas/tests/test_series.py
```diff
@@ -135,6 +135,10 @@ def test_rename(self):
         self.assertEqual(kidx.name, "renamed")
         self.assert_eq(kidx, pidx)
 
+        expected_error_message = "Series.name must be a hashable type"
+        with self.assertRaisesRegex(TypeError, expected_error_message):
+            kser.name = ["0", "1"]
```
ueshin, line 140: Shall we also add `ks.Series([1, 2, 3], name=["0", "1"])`?

databricks/koalas/indexes.py
```diff
@@ -565,6 +567,8 @@ def name(self) -> Union[Any, Tuple]:
 
     @name.setter
     def name(self, name: Union[Any, Tuple]) -> None:
+        if not is_hashable(name):
+            raise TypeError("Index.name must be a hashable type")
```
ueshin, line 571: We should check this in `names` setter?
E.g.,:

```py
>>> pidx = pd.Index([1, 2, 3])
>>> pidx.names = [["0"]]
...
TypeError: Int64Index.name must be a hashable type
```

itholic: Thanks for the comments!! 👍

databricks/koalas/indexes.py
```diff
@@ -579,6 +579,8 @@ def names(self) -> List[Union[Any, Tuple]]:
     def names(self, names: List[Union[Any, Tuple]]) -> None:
         if not is_list_like(names):
             raise ValueError("Names must be a list-like")
+        if not is_hashable(names[0]):
```
ueshin, line 582: Why only `names[0]`?

itholic: Oh, sorry it was my mistake.

Btw, I think we should check this in `__new__` since pandas < 1.0.0 doesn't raise error even when name is not hashable.

```python
# pandas < 1.0.0 doesn't raise Exception for this case
>>> pd.Index([1, 2, 3], name=["0", "1"])
Int64Index([1, 2, 3], dtype='int64', name=['0', '1'])

# whereas pandas >= 1.0.0 raises Exception
>>> pd.Index([1, 2, 3], name=["0", "1"])
...
TypeError: Index.name must be a hashable type
```

So, if underlying pandas is lower than 1.0.0, it raises `AssertionError` rather than `TypeError` in `__new__`.

```python
>>> ks.Index([1, 2, 3], name=["0", "1"])
...
AssertionError: OrderedDict([('__index_level_0__', (['0', '1'],))])
```

databricks/koalas/indexes.py
```diff
@@ -579,6 +582,8 @@ def names(self) -> List[Union[Any, Tuple]]:
     def names(self, names: List[Union[Any, Tuple]]) -> None:
         if not is_list_like(names):
             raise ValueError("Names must be a list-like")
+        if not all(is_hashable(name) for name in names):
+            raise TypeError("Index.name must be a hashable type")
```
ueshin, line 586: oh, wait. Do we have to check this in `rename`?

databricks/koalas/series.py
```diff
@@ -1065,6 +1065,8 @@ def name(self) -> Union[Any, Tuple]:
 
     @name.setter
     def name(self, name: Union[Any, Tuple]):
+        if not is_hashable(name):
+            raise TypeError("Series.name must be a hashable type")
```
ueshin, line 1069: ditto.

databricks/koalas/indexes.py
itholic: Oh, Let me check.

databricks/koalas/indexes.py
```diff
@@ -565,6 +568,8 @@ def name(self) -> Union[Any, Tuple]:
 
     @name.setter
     def name(self, name: Union[Any, Tuple]) -> None:
+        if not is_hashable(name):
+            raise TypeError("Index.name must be a hashable type")
```
ueshin, line 572: Why did this come back?

databricks/koalas/series.py
ueshin, line 1069: I don't think we need this if we check it in `rename`?

databricks/koalas/indexes.py
itholic: Otherwise the below case doesn't raise Exception unlike pandas.

```python
>>> kidx.name = ["renamed"]

>>> pidx = pd.Index([1, 2, 3])
>>> p.name = ["renamed"]
...
TypeError: Int64Index.name must be a hashable type
```

databricks/koalas/series.py
itholic: Oh, rIght. I'll remove this. Thanks!

databricks/koalas/indexes.py
itholic: Just added related tests!

ueshin: In my local with the diff from the latest commit:

```
$ git diff
diff --git a/databricks/koalas/indexes.py b/databricks/koalas/indexes.py
index 7b01d4c..673caa1 100644
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -568,8 +568,6 @@ class Index(IndexOpsMixin):

     @name.setter
     def name(self, name: Union[Any, Tuple]) -> None:
-        if not is_hashable(name):
-            raise TypeError("Index.name must be a hashable type")
         self.names = [name]

     @property
```

```py
>>> kidx.name = ["renamed"]
...
TypeError: Index.name must be a hashable type
```

itholic:
Let me investigate this more detail.
Not pretty sure, but maybe, `Index` and `MultiIndex` should have their own setters separately for handling rename properly.

databricks/koalas/tests/test_indexes.py
```diff
@@ -253,6 +263,8 @@ def test_index_rename(self):
         self.assert_eq(kidx.rename(None), pidx.rename(None))
         self.assert_eq(kdf.index.names, pdf.index.names)
 
+        # TODO?: unlike pandas, this case raises ValueError rather than TypeError
+        # since we set both `name` and `names` via `rename` function for now.
```
itholic, line 267: Anyway, it's catching the error properly, so I think it's okay to leave it as it is for now since I don't think it's efficient to spend a lot of time just changing the type of error.

databricks/koalas/indexes.py
itholic: Roll back to the previous fix!

databricks/koalas/tests/test_indexes.py
ueshin: How about:

```diff
diff --git a/databricks/koalas/indexes.py b/databricks/koalas/indexes.py
index ad411a4..1c56eaf 100644
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -580,9 +580,16 @@ class Index(IndexOpsMixin):
     def names(self, names: List[Union[Any, Tuple]]) -> None:
         if not is_list_like(names):
             raise ValueError("Names must be a list-like")
-        if not all(is_hashable(name) for name in names):
-            raise TypeError("Index.name must be a hashable type")
-        self.rename(names, inplace=True)
+        if self._internal.index_level != len(names):
+            raise ValueError(
+                "Length of new names must be {}, got {}".format(
+                    self._internal.index_level, len(names)
+                )
+            )
+        if self._internal.index_level == 1:
+            self.rename(names[0], inplace=True)
+        else:
+            self.rename(names, inplace=True)

     @property
     def nlevels(self) -> int:
@@ -667,20 +674,12 @@ class Index(IndexOpsMixin):
             return DataFrame(internal).index

     def _verify_for_rename(self, name):
-        if is_name_like_tuple(name):
-            return [name]
-        elif is_name_like_value(name):
-            return [(name,)]
-        elif is_list_like(name):
-            if self._internal.index_level != len(name):
-                raise ValueError(
-                    "Length of new names must be {}, got {}".format(
-                        self._internal.index_level, len(name)
-                    )
-                )
-            return [n if is_name_like_tuple(n) else (n,) for n in name]
-        else:
-            raise TypeError("name must be a hashable type")
+        if is_hashable(name):
+            if is_name_like_tuple(name):
+                return [name]
+            elif is_name_like_value(name):
+                return [(name,)]
+        raise TypeError("Index.name must be a hashable type")

     # TODO: add downcast parameter for fillna function
     def fillna(self, value: Scalar) -> "Index":
@@ -2588,6 +2587,8 @@ class MultiIndex(Index):
                         self._internal.index_level, len(name)
                     )
                 )
+            if any(not is_hashable(n) for n in name):
+                raise TypeError("MultiIndex.name must be a hashable type")
             return [n if is_name_like_tuple(n) else (n,) for n in name]
         else:
             raise TypeError("Must pass list-like as `names`.")
```

itholic: Sounds good. Thanks for the suggestion, +10 !

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

itholic:

PR #1890, opened 2020-11-04
Introduce InternalFrame.index_level.

ueshin:
Currently there are different ways to check the number of the index columns. We should introduce `InternalFrame.index_level` and use it.


HyukjinKwon (approved):

ueshin:
Thanks! I'd merge this now.

PR #1893, opened 2020-11-06
Implement 'weights' and 'axis' in sample at DataFrame and Series

### LINKED ISSUE #1887: Implementing the full functionality of the 'sample' function
The current implementation of the sample function is based on the sample function of pyspark, and the parameter n is not supported, and frac cannot be empty.

From the source code of pandas, https://github.com/pandas-dev/pandas/blob/master/pandas/core/generic.py#L5076
the implementation of the sample function relies on methods such as iloc, take and reindex. These methods are currently supported by koalas. Therefore, the implementation of the sample function can be based on the current logic of pandas.

I will try to implement the sample function of frame and series from this idea.

chi2liu:
Implement sample. Resolves #1887

Support the remaining parameters of the sample function of DataFrame, such as n, axis, weights.

Now there are two unsupported situations:
1.does not support axis=1
2.If the value of the frac parameter > 1, the weights parameter is not supported

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7218,7 +7218,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             Fraction of axis items to return.
         replace : bool, default False
             Sample with or without replacement.
-        weights : str or ndarray-like, optional
+        weights : ndarray-like, optional
+            Currently does not support Series and str.
             Default 'None' results in equal probability weighting.
             If passed a Series, will align with target object on index. Index
             values in weights not found in sampled object will be ignored and
@@ -7277,12 +7278,9 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         dog          4          0                  2
         fish         0          0                  8
         """
-        if axis in ("index", "rows", 0, None):
-            axis = 0
-        elif axis in ("columns", 1):
+        axis = validate_axis(axis)
+        if axis == 1:
             raise NotImplementedError("Function sample currently does not support axis=1.")
-        else:
-            raise ValueError("No axis named %s for object type %s." % (axis, type(axis)))
 
         axis_length = self.shape[axis]
 
@@ -7295,25 +7293,15 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         # Check weights for compliance
         if weights is not None:
 
-            # If a series, align with frame
-            if isinstance(weights, ks.Series):
-                weights = weights.reindex(self.axes[axis])
-
-            # Strings acceptable if a dataframe and axis = 0
-            if isinstance(weights, str):
-                if isinstance(self, ks.DataFrame):
-                    if axis == 0:
-                        try:
-                            weights = self[weights]
-                        except KeyError as err:
-                            raise KeyError("String passed to weights not a valid column") from err
-
-            # Because ks.Series currently does not support the Series.__iter__ method,
+            # If a series or str, ks.Series currently does not support the Series.__iter__ method,
             # It cannot be initialized to the pandas Series, so here is to_pandas.
-            if isinstance(weights, ks.Series):
-                weights = pd.Series(weights.to_pandas(), dtype="float64")
-            else:
-                weights = pd.Series(weights, dtype="float64")
+            # Don't support weights as Series for now since it could occur performance degradation.
+            if isinstance(weights, (ks.Series, str)):
+                raise NotImplementedError(
+                    "The weights parameter does not currently support the Series and str."
+                )
+
+            weights = pd.Series(weights, dtype="float64")
 
             if len(weights) != axis_length:
                 raise ValueError("Weights and axis to be sampled must be of same length")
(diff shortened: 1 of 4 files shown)


HyukjinKwon:
@itholic can you review this please?

itholic:
Sure, let me take a look

databricks/koalas/frame.py
```diff
@@ -7237,46 +7260,109 @@ def sample(
...
+                weights = pd.Series(weights, dtype="float64")
+
+            if len(weights) != axis_length:
+                raise ValueError("Weights and axis to be sampled must be of same length")
+
+            if (weights == np.inf).any() or (weights == -np.inf).any():
+                raise ValueError("weight vector may not include `inf` values")
+
+            if (weights < 0).any():
+                raise ValueError("weight vector many not include negative values")
```
itholic, line 7325: nit: many -> may 
Of course I understand that you've just followed pandas' message 👍 , but It looks obviously typo.

databricks/koalas/frame.py
```diff
@@ -7237,46 +7260,109 @@ def sample(
...
+
+            # If has nan, set to zero.
+            weights = weights.fillna(0)
+
+            # Renormalize if don't sum to 1
+            if weights.sum() != 1:
+                if weights.sum() != 0:
+                    weights = weights / weights.sum()
```
itholic, line 7333: I can see `weights.sum()` here and there.
Can we make a variable for `weights.sum()` and just reuse it to avoid running Spark job too many times ??
just like:
```python
weights_sum = weights.sum()
if weights_sum != 1:
    if weights_sum != 0:
        weights = weights / weights_sum
```

databricks/koalas/frame.py
```diff
@@ -7237,46 +7260,109 @@ def sample(
...
-        >>> df.sample(frac=0.25, random_state=1)  # doctest: +SKIP
-                num_legs  num_wings  num_specimen_seen
-        falcon         2          2                 10
-        fish           0          0                  8
-
-        Extract 25% random elements from the ``Series`` ``df['num_legs']``, with replacement,
-        so the same items could appear more than once.
-
-        >>> df['num_legs'].sample(frac=0.4, replace=True, random_state=1)  # doctest: +SKIP
+        >>> df['num_legs'].sample(n=3, random_state=1).sort_index()
```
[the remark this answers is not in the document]
chi2liu: Yeah, there is another reason that the previous doctest was skipped.

databricks/koalas/frame.py
```diff
@@ -7237,46 +7260,109 @@ def sample(
...
-            raise NotImplementedError(
-                "Function sample currently does not support specifying "
-                "exact number of items to return. Use frac instead."
-            )
+        if axis in ("index", "rows", 0, None):
+            axis = 0
+        elif axis in ("columns", 1):
+            raise NotImplementedError("Function sample currently does not support axis=1.")
+        else:
+            raise ValueError("No axis named %s for object type %s." % (axis, type(axis)))
```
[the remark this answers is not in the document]
chi2liu: Good idea!
I did not see this function before.

databricks/koalas/frame.py
```diff
@@ -7237,46 +7260,109 @@ def sample(
...
+                    if axis == 0:
+                        try:
+                            weights = self[weights]
+                        except KeyError as err:
+                            raise KeyError("String passed to weights not a valid column") from err
+
+            # Because ks.Series currently does not support the Series.__iter__ method,
+            # It cannot be initialized to the pandas Series, so here is to_pandas.
+            if isinstance(weights, ks.Series):
+                weights = pd.Series(weights.to_pandas(), dtype="float64")
```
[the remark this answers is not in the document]
chi2liu: You are right!
If the series is relatively large, it may indeed cause performance problems.
The series weight is not supported here right now.

databricks/koalas/frame.py
```diff
@@ -7237,46 +7260,109 @@ def sample(
...
+            raise ValueError("A negative number of rows requested. Please provide positive value.")
+
+        # Because duplicated row selection is not currently supported.
+        # So if frac > 1, use the pyspark implementation.
+        if frac is not None and frac > 1:
+            sdf = self._internal.resolved_copy.spark_frame.sample(
+                withReplacement=replace, fraction=float(frac), seed=random_state
+            )
+            return DataFrame(self._internal.with_new_sdf(sdf))
+        locs = rs.choice(axis_length, size=n, replace=replace, p=weights)
```
ueshin, line 7364: What's the estimated size of `locs`? Will it also be much huge?
e.g., for the case `A random 50% sample of the ``DataFrame`` with replacement`?

chi2liu: The `rs.choice` method is similar to the [numpy.random.RandomState.choice](https://numpy.org/doc/stable/reference/random/generated/numpy.random.RandomState.choice.html).
The return value of `rs.choice` is the size of the `size` parameter, in this case is `n`.
So the size of `locs` here is `n`.
If parameter `n` is set, `locs` is equal to `n`, otherwise `locs` is equal to `n = int(round(frac * axis_length))`.
For the case `A random 50% sample of the ``DataFrame`` with replacement`, the size of `locs` will be one-half of the size of the `DataFrame`, to be precise, it should be `int(round(0.5 * DataFrame.nrows))`.
If `n` is large, or `frac` is large, `locs` will be large.
Performance depends on the performance of the `take` method, which is actually the performance of the `iloc` method

ueshin: In that case, I wouldn't recommend to use `take` or `iloc` for this.
Thinking of Koalas workload, the length of `DataFrame` could be so huge, then `locs` will be too huge for a single Driver node.
Also, row access by its row number is essentially heavy on Spark, and so `iloc` is heavy in Koalas, especially if the `locs` is huge.

chi2liu: Yeah!
I agree with you, because spark does not have the concept of rowindex, iloc is a heavy operation. 
But here the weights parameter of the sample function is the same as iloc, which must depend on the row index. 
There may be no other way to support the weights parameter right now. 
Just like iloc is a heavy operation, but it must also be implemented based on row index right now. 
Maybe the current sample function supports weights operation and must also be based on row index

xinrong-databricks:
Hi @chi2liu , since Koalas is ported to Spark, would you like to migrate this PR to Spark repo? If not, I will port it next week.
https://issues.apache.org/jira/browse/SPARK-36436 is the ticket. Thanks!

xinrong-databricks:
I am porting this.

PR #1897, opened 2020-11-09
Use nullable inferred schema in function apply

### LINKED ISSUE #1885: udf schema inference should not set column nullable=false
When udf is applied to dataframe and no return schema hint is provided - schema is infered based on data sample and makes wrong assumption about columns nullability. Looks like if all values in some column are not null - resulting schema nullable=false. 
But this is a wrong assumption, cause only a sample of data was examined. 
Sample code leading to runtime error:
```
import numpy as np

data = list()
for i in range(1, 10000):
  data.append((str(i % 100), np.nan if i % 9999 == 0 else float(i),))

sdf = spark.createDataFrame(data, 'a string, b float').repartition(10)
kdf = sdf.to_koalas()

def f(df):
  return df

df = kdf\
  .groupby('a')\
  .apply(f)

df.to_spark().printSchema()
df[df['b'] == -1]
```

printSchema output:
```
root
 |-- a: string (nullable = false)
 |-- b: float (nullable = false)
```

Running this code with koalas version 1.3 throws error:
```
Job aborted due to stage failure: Task 0 in stage 67.0 failed 4 times, most recent failure: Lost task 0.3 in stage 67.0 (TID 435, 10.147.224.152, executor 0): java.lang.IllegalStateException: Value at index is null
```

HyukjinKwon:
This PR proposes to use nullable schema in when to apply a function. We take some to infer the schema. Usually types are matched but null or NaN is sparse often. This behaviour is batch with Spark's JSON schema inference as well.

```python
from pyspark.sql import SparkSession
import numpy as np

spark = SparkSession.builder.getOrCreate()
data = list()
for i in range(1, 10000):
  data.append((str(i % 100), np.nan if i % 9999 == 0 else float(i),))

sdf = spark.createDataFrame(data, 'a string, b float').repartition(10)
kdf = sdf.to_koalas()

def f(df):
  return df

df = kdf\
  .groupby('a')\
  .apply(f)

df.to_spark().printSchema()
df[df['b'] == -1]
```

**Before:**

```
root
 |-- a: string (nullable = false)
 |-- b: float (nullable = false)

java.lang.IllegalStateException: Value at index is null
...
```

**After:**

```
root
 |-- a: string (nullable = true)
 |-- b: float (nullable = true)

Empty DataFrame
Columns: [a, b]
Index: []
```

Resolves #1885

### DIFF
--- a/databricks/koalas/accessors.py
+++ b/databricks/koalas/accessors.py
@@ -26,7 +26,6 @@ import pandas as pd
 import pyspark
 from pyspark.sql import functions as F
 from pyspark.sql.functions import pandas_udf, PandasUDFType
-from pyspark.sql.types import StructType, StructField
 
 from databricks.koalas.internal import (
     InternalFrame,
@@ -34,6 +33,7 @@ from databricks.koalas.internal import (
     SPARK_DEFAULT_SERIES_NAME,
 )
 from databricks.koalas.typedef import infer_return_type, DataFrameType, SeriesType
+from databricks.koalas.spark.utils import as_nullable_spark_type
 from databricks.koalas.utils import (
     is_name_like_value,
     is_name_like_tuple,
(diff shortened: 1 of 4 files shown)


ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

PR #1903, opened 2020-11-11
Support typehints for various of ArrayType.

### LINKED ISSUE #1133: Support for Array type hints in APIs that take Python native functions (e.g., DataFrame.apply)
```python

def tokenizeDF(col1) -> ks.Series[np.array(... ???)]:
    pass
```

itholic:
Until now, only arrays of string type could be used as type hints with np.array.
This PR proposes to support many other child types, too.

This should resolve #1133

### DIFF
--- a/databricks/koalas/typedef/typehints.py
+++ b/databricks/koalas/typedef/typehints.py
@@ -22,6 +22,11 @@ import datetime
 import decimal
 from inspect import getfullargspec, isclass
 
+try:
+    from typing import _GenericAlias  # type: ignore
+except ImportError:
+    from typing import GenericMeta as _GenericAlias  # type: ignore
+
 import numpy as np
 import pandas as pd
 from pandas.api.types import is_datetime64_dtype, is_datetime64tz_dtype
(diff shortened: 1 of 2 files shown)


databricks/koalas/typedef/typehints.py
```diff
@@ -22,6 +22,11 @@
 import decimal
 from inspect import getfullargspec, isclass
 
+try:
+    from typing import _GenericAlias  # type: ignore
```
itholic, line 26: FYI: `GenericMeta` is renamed to `_GenericAlias` in Python3.7.

databricks/koalas/typedef/typehints.py
```diff
@@ -104,8 +109,9 @@ def as_spark_type(tpe) -> types.DataType:
     # TODO: Add "boolean" and "string" types.
     # ArrayType
     if tpe in (np.ndarray,):
-        # TODO: support other child types
         return types.ArrayType(types.StringType())
+    elif isinstance(tpe, _GenericAlias) and (tpe.__origin__ in (list, typing.List)):
```
ueshin, line 113: Shall we use `hasattr(tpe, "__origin__")` to use the same way as in `infer_return_type`?

ueshin: Btw, I don't think `list` can be a generic type?
In that case, should we handle `list` as the same as `np.ndarray`? WDYT @HyukjinKwon ?

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

itholic:

PR #1904, opened 2020-11-11
Force to use the Spark's system default precision and scale when inferred data type contains DecimalType.

ueshin:
The return type inference sometimes fails when the inferred type contains `DecimalType` to infer the precision or the scale smaller.

```py
>>> from decimal import Decimal
>>> kdf = ks.DataFrame({'a': [Decimal('1.1'), Decimal('2.2'), Decimal('10.01')]})
>>> kdf
       a
0   1.10
1   2.20
2  10.01
>>> kdf.spark.print_schema()
root
 |-- a: decimal(4,2) (nullable = false)

>>> with ks.option_context("compute.shortcut_limit", 1):
...   kdf.transform(lambda x: x).spark.print_schema()
...
root
 |-- a: decimal(3,2) (nullable = true)

>>> with ks.option_context("compute.shortcut_limit", 1):
...   kdf.transform(lambda x: x)
...
...
pyarrow.lib.ArrowInvalid: Decimal type with precision 4 does not fit into precision inferred from first array element: 3
...
```

We should use the Spark's system default precision and scale instead to avoid as many such cases as possible:

```py
>>> with ks.option_context("compute.shortcut_limit", 1):
...   kdf.transform(lambda x: x).spark.print_schema()
...
root
 |-- a: decimal(38,18) (nullable = true)

>>> with ks.option_context("compute.shortcut_limit", 1):
...   kdf.transform(lambda x: x)
...
                       a
0   1.100000000000000000
1   2.200000000000000000
2  10.010000000000000000
```

### DIFF
--- a/databricks/koalas/spark/utils.py
+++ b/databricks/koalas/spark/utils.py
@@ -16,7 +16,7 @@
 """
 Helpers and utilities to deal with PySpark instances
 """
-from pyspark.sql.types import StructType, MapType, ArrayType, StructField, DataType
+from pyspark.sql.types import DecimalType, StructType, MapType, ArrayType, StructField, DataType
 
 
 def as_nullable_spark_type(dt: DataType) -> DataType:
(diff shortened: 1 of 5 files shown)


databricks/koalas/spark/utils.py
```diff
@@ -65,3 +65,59 @@ def as_nullable_spark_type(dt: DataType) -> DataType:
...
+    ...         StructType([
+    ...             StructField('a',
+    ...                 MapType(DecimalType(5, 0),
+    ...                 ArrayType(DecimalType(20, 0), False), False), False),
+    ...             StructField('b', StringType(), True)])),
+    ...     StructField("B", DecimalType(30, 15), False)]),
+    ...     precision=30, scale=15)  # doctest: +NORMALIZE_WHITESPACE
+    StructType(List(StructField(A,StructType(List(StructField(a,MapType(DecimalType(30,15),\
+ArrayType(DecimalType(30,15),false),false),false),StructField(b,StringType,true))),true),\
+StructField(B,DecimalType(30,15),false)))
```
itholic, line 95: nit: indent ?

itholic (approved):
Everything looks clear to me except one nit.

databricks/koalas/spark/utils.py
ueshin: It's a single line without spaces.

itholic: Oh, got it

HyukjinKwon (approved):

ueshin:
Thanks! merging.

PR #1905, opened 2020-11-12
Fix read_excel to support squeeze argument.

ueshin:
Fixes `read_excel` to support squeeze argument.
Currently it always return `DataFrame` when there are multiple Excel files.

### DIFF
--- a/databricks/koalas/namespace.py
+++ b/databricks/koalas/namespace.py
@@ -64,6 +64,7 @@ from databricks.koalas.frame import DataFrame, _reduce_spark_multi
 from databricks.koalas.internal import (
     InternalFrame,
     DEFAULT_SERIES_NAME,
+    HIDDEN_COLUMNS,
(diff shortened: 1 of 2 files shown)


HyukjinKwon (approved):

databricks/koalas/namespace.py
```diff
@@ -1093,23 +1093,19 @@ def pd_read_excel(io_or_bin, sn):
 
         def read_excel_on_spark(pdf, sn):
 
+            if isinstance(pdf, pd.Series):
```
itholic, line 1096: nit: I think it'd be better to use `pdf_or_pser` or something if `pdf` can also be a `Series`. but not urgent now :)

itholic (approved):
LGTM, pending tests.

itholic:
Let me merge this, pretty sure it's good to go.
Thanks, @ueshin !

Issue #1826, opened 2020-10-06
ModuleNotFoundError: No module named 'databricks' when using apply_batch or apply_transform


maxpagels:
Hi,

I'm using Spark in client mode and I've gotten Koalas working, but the `apply_batch` method seems to indicate that koalas is missing from the executor nodes. It it really so that koalas must be explicitly installed to worker nodes? Or is it another issue / something simple I'm missing? Spark version: 2.4.3, Koalas version: 1.2.0.

Example:
```
  kdf = ks.DataFrame({'a': range(0, 20000), 'i': range(0, 20000)}).set_index('i')
  # --> works
  kdf.head(10)
  # --> also works

  def test_apply(df):
    return df
  kdf.koalas.apply_batch(test_apply)
  # --> fails, see error below
```

Error:
```
...
  File "/var/lib/mesos/slaves/ad6bc800-ab3b-486e-bfa2-cf24ca7aebae-S1/frameworks/7461c35c-4cf7-47a5-ae69-3ba9362cee61-71216/executors/1/runs/71cf1309-75b9-4ac2-b14e-3abc04506810/spark-2.4.3-bin-datalake-hadoop-2.9.2-1/python/lib/pyspark.zip/pyspark/serializers.py", line 580, in loads
    return pickle.loads(obj, encoding=encoding)
ModuleNotFoundError: No module named 'databricks'

	at org.apache.spark.api.python.BasePythonRunner$ReaderIterator.handlePythonException(PythonRunner.scala:452)
	at org.apache.spark.sql.execution.python.ArrowPythonRunner$$anon$1.read(ArrowPythonRunner.scala:172)
	at org.apache.spark.sql.execution.python.ArrowPythonRunner$$anon$1.read(ArrowPythonRunner.scala:122)
	at org.apache.spark.api.python.BasePythonRunner$ReaderIterator.hasNext(PythonRunner.scala:406)
	at org.apache.spark.InterruptibleIterator.hasNext(InterruptibleIterator.scala:37)
	at scala.collection.Iterator$$anon$12.hasNext(Iterator.scala:440)
	at scala.collection.Iterator$$anon$11.hasNext(Iterator.scala:409)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:255)
	at org.apache.spark.sql.execution.SparkPlan$$anonfun$2.apply(SparkPlan.scala:247)
[...]

PR #1910, opened 2020-11-13
Move the words about Koalas in workers

HyukjinKwon:
In some cases such as `apply`, `transform`, `apply_batch`, or `tranform_batch`, Koalas still requires to be installed in worker nodes, see https://github.com/databricks/koalas/issues/1826.
This PR removes some wordings about the installation in workers for now, and makes it deferred to future work.


ueshin (approved):
LGTM.

PR #1912, opened 2020-11-13
Fix to_csv to avoid duplicated option 'path' for DataFrameWriter.

ueshin:
Spark doesn't allow duplicated option `path` when `DataFrameWriter.save()` since Spark 3.1. (apache/spark#29543)


ueshin:
I'd merge this now as I'm quite sure about this change.

PR #1914, opened 2020-11-16
Moved Series.astype to base.py for sharing implementation with Index

itholic:
For some cases, `Index.astype` is not working properly unlike `Series.astype`.

```python
>>> pser = pd.Index(['a', 'b', None], name="x")
>>> kser = ks.Index(pser)
>>> pser.astype(bool)
Index([True, True, False], dtype='object', name='x')
>>> kser.astype(bool)
Index([None, None, None], dtype='object', name='x')
```

This PR would suggest sharing the implementation of `Series.astype` for `Index.astype`, and adding tests for `Index.astype`.


ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

PR #1916, opened 2020-11-17
Add a constructor to GroupBy for attribute type hints

xinrong-databricks:
Add a constructor to GroupBy for attribute type hints. This proposal can save a lot of "# type: ignore" comments.

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -26,7 +26,7 @@ from collections.abc import Callable
 from distutils.version import LooseVersion
 from functools import partial
 from itertools import product
-from typing import Any, List, Set, Tuple, Union
+from typing import Any, List, Set, Tuple, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -83,6 +83,16 @@ class GroupBy(object, metaclass=ABCMeta):
     :ivar _groupkeys: The list of keys that will be used to perform the grouping
     :type _groupkeys: List[Series]
     """
+
+    def __init__(self):
+        self._kser = cast(Series, None)
+        self._kdf = cast(DataFrame, None)
+        self._groupkeys = cast(List[Series], None)
+        self._as_index = cast(bool, None)
+        self._dropna = cast(bool, None)
+        self._column_labels_to_exlcude = cast(Set[Tuple], None)
+        self._agg_columns_selected = cast(bool, None)
+        self._agg_columns = cast(List, None)
 
     @property
     def _groupkeys_scols(self):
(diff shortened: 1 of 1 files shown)


itholic (approved):

xinrong-databricks:
@itholic Thanks for review!

xinrong-databricks:
Add a constructor to GroupBy for attribute type hints.

databricks/koalas/groupby.py
```diff
@@ -84,6 +84,17 @@ class GroupBy(object, metaclass=ABCMeta):
...
+    def __init__(self):
+        """A constructor for attribute type hints."""
+        self._kser = cast(Series, None)
+        self._kdf = cast(DataFrame, None)
+        self._groupkeys = cast(List[Series], None)
+        self._as_index = cast(bool, None)
+        self._dropna = cast(bool, None)
+        self._column_labels_to_exlcude = cast(Set[Tuple], None)
+        self._agg_columns_selected = cast(bool, None)
+        self._agg_columns = cast(List, None)
```
ueshin, line 96: I wonder whether this is a right usage or not.

xinrong-databricks: Oh isn't this "a constructor with type hints" (as we discussed)? Please let me know if there is a better approach, then I can fix it!

ueshin: I thought the constructor would look like:

```py
def __init__(self, kser: Series, kdf: DataFrame, ...):
    self._kser = kser
    self._kdf = kdf
    ...

...

class DataFrameGroupBy(GroupBy):

    def __init__(self, ...):
        ...
        super().__init__(self, ...)
        ...

...

class SeriesGroupBy(GroupBy):

    def __init__(self, ...):
        ...
        super().__init__(self, ...)
        ...

...
```

but it might be just I haven't seen the usage with `cast` to declare instance variables.

@itholic Have you see the usage?

itholic: No, I've never seen exactly same case, but I thought it was looking way better than the general implementation in terms of readability - of course, opinions on this may vary from person to person -.

Because the purpose of this initializer is simply for type check, and the actual implementation of initializer was done in each subclass.

However, after I re-thinking again, maybe it could be better to just follow the general case as Takuya mentioned since `typing.cast` aren't commonly used, it could confuse someone.

xinrong-databricks: Certainly, I will make a PR to follow the general case as Takuya mentioned. Thank you both for the input!

PR #1918, opened 2020-11-17
Use typing.cast in return statements that call functions with union return types

xinrong-databricks:
Use `typing.cast` in return statements that call functions with union return types. This saves us from `# type: ignore` comments.


ueshin (approved):
LGTM except for the failures.

HyukjinKwon (approved):

itholic (approved):

xinrong-databricks:
Thank you! Tests passed, let me merge it.

PR #1931, opened 2020-11-25
Support DataFrame parameter in Series.dot

xinrong-databricks:


databricks/koalas/tests/test_series.py
```diff
@@ -459,6 +459,31 @@ def test_fillna(self):
         self.assert_eq(kser, pser)
         self.assert_eq(kdf, pdf)
 
+    def test_dot(self):
+        pser = pd.Series([0, 1, 2, 3])
+        kser = ks.from_pandas(pser)
+
+        self.assert_eq(kser.dot(kser), pser.dot(pser))
+
+        # other_pser = pd.Series([-1, 2, -3, 4])
```
xinrong-databricks, line 468: @ueshin It seems we don't support the test case commented out for now. Are we going to support this?

xinrong-databricks: `kser.dot(other_kser)` will hit
```
ValueError: Cannot combine the series or dataframe because it comes from a different 
dataframe. In order to allow this operation, enable 'compute.ops_on_diff_frames' opti
on.
```

databricks/koalas/series.py
```diff
@@ -4795,11 +4795,11 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
-            raise ValueError(
-                "Series.dot() is currently not supported with DataFrame since "
-                "it will cause expansive calculation as many as the number "
-                "of columns of DataFrame"
-            )
+            col_results = []
+            with option_context("compute.ops_on_diff_frames", True):
+                for k, v in other._ksers.items():
+                    col_results.append((self * v).sum())
+            return Series(col_results, index=other.columns)
```
ueshin, line 4802: This hits the same issues as https://github.com/databricks/koalas/pull/1136#pullrequestreview-333211768

databricks/koalas/tests/test_series.py
ueshin: Shall we use `option_context` to enable it?

```py
with ks.option_context("compute.ops_on_diff_frames", True):
    ...
```

ueshin (commented):
Could you add an example using `DataFrame` in the docstring?

databricks/koalas/tests/test_series.py
xinrong-databricks: Got it, we don't enable `compute.ops_on_diff_frames` without user acknowledgment.

databricks/koalas/series.py
xinrong-databricks: Modified!

databricks/koalas/series.py
```diff
@@ -4793,19 +4793,56 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
         """
         if isinstance(other, DataFrame):
-            raise ValueError(
-                "Series.dot() is currently not supported with DataFrame since "
-                "it will cause expansive calculation as many as the number "
-                "of columns of DataFrame"
-            )
-        if self._kdf is not other._kdf:
-            if len(self.index) != len(other.index):
+            if not self.index.equals(other.index):
```
ueshin, line 4813: Shall we check `same_anchor(self, other)` to avoid extra Spark jobs?

databricks/koalas/series.py
```diff
@@ -4793,19 +4793,56 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
+                sdf.select([F.sum(sdf[scol]).alias(scol) for scol in that_data_spark_column_names])
+                .toPandas()
+                .transpose()
+            )
+
+            pdf.index = other.columns
+            return cast(Union[Series, Scalar], ks.from_pandas(first_series(pdf)).rename(None))
+
+        elif isinstance(other, Series):
+            if self._kdf is not other._kdf:
```
ueshin, line 4838: maybe `not same_anchor(self, other)`?

databricks/koalas/series.py
```diff
@@ -4793,19 +4793,56 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
-        if self._kdf is not other._kdf:
-            if len(self.index) != len(other.index):
+            if not self.index.equals(other.index):
                 raise ValueError("matrices are not aligned")
-        if isinstance(other, Series):
+
+            combined = combine_frames(self._kdf, other)
+            sdf = combined._internal.spark_frame
+
+            this_data_spark_column_name = combined["this"]._internal.data_spark_column_names[0]
```
ueshin, line 4819: Why can we use `data_spark_column_names[0]`?

databricks/koalas/series.py
```diff
@@ -4793,19 +4793,56 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
+            that_data_spark_column_names = combined["that"]._internal.data_spark_column_names
+
+            sdf = sdf.select(
+                [
+                    (sdf[this_data_spark_column_name] * sdf[scol]).alias(scol)
```
ueshin, line 4824: `scol_for(sdf, this_data_spark_column_name)`?

databricks/koalas/series.py
```diff
@@ -4793,19 +4793,56 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
+                    for scol in that_data_spark_column_names
```
ueshin, line 4825: We usually use `scol` for Spark column object, but this is not. Could you rename it?

databricks/koalas/series.py
```diff
@@ -4793,19 +4793,56 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
+                ]
+            )
+            pdf = (
+                sdf.select([F.sum(sdf[scol]).alias(scol) for scol in that_data_spark_column_names])
+                .toPandas()
+                .transpose()
+            )
+
+            pdf.index = other.columns
```
ueshin, line 4834: Do we need this?

ueshin (commented):
Shall we add a type annotation for the argument `other`?

databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -886,9 +879,26 @@ def test_dot(self):
...
+        pser = pd.Series([0, 1, 2, 3])
+        kser = ks.from_pandas(pser)
+
+        # DataFrame "other" without Index/MultiIndex as columns
+        pdf = pd.DataFrame([[0, 1], [-2, 3], [4, -5], [6, 7]])
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(kser.dot(kdf), pser.dot(pdf))
+
+        # DataFrame "other" with Index as columns
+        pdf.columns = pd.Index(["x", "y"])
```
ueshin, line 893: Could you also use non-empty `columns.names`?

databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -886,9 +879,26 @@ def test_dot(self):
...
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(kser.dot(kdf), pser.dot(pdf))
+
+        # DataFrame "other" with MultiIndex as columns
+        pdf.columns = pd.MultiIndex.from_tuples([("a", "x"), ("b", "y")])
```
ueshin, line 898: ditto.

ueshin:
Could you add more tests like:

```py
pdf = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
kdf = ks.from_pandas(pdf)

self.assert_eq((pdf['b'] * 10).dot(pdf['a']), ...)
self.assert_eq((pdf['b'] * 10).dot(pdf), ...)
self.assert_eq((pdf['b'] * 10).dot(pdf + 1), ...)
```

databricks/koalas/series.py
xinrong-databricks: Were you saying using `same_anchor` to replace index comparison (L4813)?

I may not fully understand. For example, `kdf` and `kser` don't have the same anchor, but they are valid inputs.
```
>>> kdf
   0  1
0  0  1
1 -2  3
2  4 -5
3  6  7
>>> kser
0    0
1    1
2    2
3    3
dtype: int64
>>> same_anchor(kser, kdf)
False
>>> with ks.option_context("compute.ops_on_diff_frames", True):
...     kser.dot(kdf)
... 
0    24
1    14
dtype: int64

```

ueshin: I mean, if `same_anchor(self, other)`, we don't need to compare `self.index.equals(other.index)` because it's always true, otherwise we should check it.

databricks/koalas/series.py
xinrong-databricks: May I ask how is `same_anchor`different from`if self._kdf is not other._kdf` in this case?

databricks/koalas/series.py
xinrong-databricks: That really makes sense. Thank you!

databricks/koalas/series.py
ueshin: The simple case would be `kser.copy().dot(kser.copy())`.

xinrong-databricks: I see. Thanks!

databricks/koalas/series.py
xinrong-databricks: Since `this` is series, can't we assume it to be data_spark_column_names[0]?

ueshin: You are using `self._kdf` to combine frames. In that case, `data_spark_column_names[0]` is not always the column for the `Series`.

Please try:

```py
ks.DataFrame({'a': [1, 2, 3], 'b': [4, 5, 6]}).b.dot(ks.DataFrame({'c': [7, 8, 9]}))
```

databricks/koalas/series.py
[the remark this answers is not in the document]
xinrong-databricks: Just modified and added a test case!

The example above should look like
```
>>> ks.Series([1, 2, 3], index=[1, 2, 3]).dot(ks.DataFrame({'a': [4, 5, 6], 'b': [7, 8, 9]}, index=[3, 1, 2]))
a    29
b    47
dtype: int64

```

databricks/koalas/series.py
xinrong-databricks: Certainly!

databricks/koalas/series.py
ueshin: Could you use `scol_for` for the other places?

xinrong-databricks: Certainly, modified!

databricks/koalas/tests/test_ops_on_diff_frames.py
xinrong-databricks: Makes sense, added!

databricks/koalas/series.py
xinrong-databricks: Makes sense, changed it to "to_frame" instead!

databricks/koalas/series.py
```diff
@@ -4742,7 +4742,7 @@ def combine_first(self, other) -> "Series":
         internal = self._internal.with_new_sdf(sdf)
         return first_series(ks.DataFrame(internal))
 
-    def dot(self, other) -> Union[Scalar, "Series"]:
+    def dot(self, other: Union["Series", ks.DataFrame]) -> Union[Scalar, "Series"]:
```
ueshin, line 4745: `DataFrame` instead of `ks.DataFrame`?

databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -886,9 +879,51 @@ def test_dot(self):
...
+        )
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(kser.dot(kdf), pser.dot(pdf))
+
+        pdf = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+        kdf = ks.from_pandas(pdf)
+
+        self.assert_eq((kdf["b"] * 10).dot(kdf["a"]), (pdf["b"] * 10).dot(pdf["a"]))
+        self.assert_eq((kdf["b"] * 10).dot(kdf), (pdf["b"] * 10).dot(pdf))
+        self.assert_eq((kdf["b"] * 10).dot(kdf + 1), (pdf["b"] * 10).dot(pdf + 1))
```
ueshin, line 919: These tests can be in `test_series.py`?

xinrong-databricks: My implementation will call resolve_copy for `kdf["b"] * 10` and then combine the resolved with the pass-in, so it might still require compute.ops_on_diff_frames to be True.

databricks/koalas/series.py
xinrong-databricks: Modified!

databricks/koalas/tests/test_ops_on_diff_frames.py
ueshin: I don't think we need to use `combine_frames` if `same_anchor(self, other)`.

ueshin: How about:

```diff
diff --git a/databricks/koalas/series.py b/databricks/koalas/series.py
index c2b406f..6ed1a29 100644
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4794,15 +4794,30 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                     raise ValueError("matrices are not aligned")

                 combined = combine_frames(self.to_frame(), other)
+
+                this_data_spark_column_name = combined["this"]._internal.data_spark_column_names[0]
+                that_data_spark_column_names = combined["that"]._internal.data_spark_column_names
             else:
-                new_this = DataFrame(self._internal.resolved_copy)  # type: DataFrame
-                combined = combine_frames(new_this, other)
+                self_column_label = verify_temp_column_name(other, "__self_column__")
+
+                combined = DataFrame(
+                    self._internal.with_new_columns(
+                        (
+                            [
+                                self._internal.data_spark_columns[0].alias(
+                                    name_like_string(self_column_label)
+                                )
+                            ]
+                            + other._internal.data_spark_columns
+                        ),
+                        column_labels=([self_column_label] + other._internal.column_labels),
+                    ).resolved_copy
+                )
+                this_data_spark_column_name = name_like_string(self_column_label)
+                that_data_spark_column_names = other._internal.data_spark_column_names

             sdf = combined._internal.spark_frame

-            this_data_spark_column_name = combined["this"]._internal.data_spark_column_names[0]
-            that_data_spark_column_names = combined["that"]._internal.data_spark_column_names
-
             sdf = sdf.select(
                 [
                     (scol_for(sdf, this_data_spark_column_name) * scol_for(sdf, scol_name)).alias(
```

databricks/koalas/series.py
```diff
@@ -4772,19 +4772,81 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
-            if len(self.index) != len(other.index):
-                raise ValueError("matrices are not aligned")
-        if isinstance(other, Series):
+
+            pdf.index = other.columns
+            return cast(Union[Series, Scalar], ks.from_pandas(first_series(pdf)).rename(self.name))
+
+        elif isinstance(other, Series):
+            if not same_anchor(self, other):
+                if not self.sort_index().index.equals(other.sort_index().index):
```
xinrong-databricks, line 4843: @ueshin Shall we match it with pandas' behavior?
In line 4734 of this file, `To match with pandas', it requires to read the whole data for, for example, counting.`
I may not fully understand `requires to read the whole data`, does `sort_index` requires to read the whole data as well?

databricks/koalas/tests/test_ops_on_diff_frames.py
xinrong-databricks: That's wonderful! Thanks! Updated in 6d92a84.

databricks/koalas/series.py
ueshin: cc @itholic

ueshin: Yeah, actually `self.sort_index().index.equals(other.sort_index().index)` is much more expensive than `len(self.index) != len(other.index)`. We might want to avoid such expensive operations.

xinrong-databricks: Would Index.sort_values() be acceptable in this case?

ueshin: I guess it's the same as `self.sort_index().index`.

xinrong-databricks: Okay I see.

I restored `series.dot(series)`'s behavior; for `series.dot(dataframe)`, it seems more efforts needed to support non-aligned indexes.

itholic: To add a bit more to Takuya's comment,

`DataFrame.sort_index()` uses `pyspark.sql.dataframe.DataFrame.sort` internally,

https://github.com/databricks/koalas/blob/138c7b834b5f00d4d96c8ffb272bc1b1e025bd9c/databricks/koalas/frame.py#L6142

and `Index.sort_values()` uses `pyspark.sql.dataframe.DataFrame.orderBy` internally.

https://github.com/databricks/koalas/blob/138c7b834b5f00d4d96c8ffb272bc1b1e025bd9c/databricks/koalas/indexes.py#L1421-L1423

And I just found that these two functions are defined by the same in PySpark.

https://github.com/apache/spark/blob/f94cb53a90558285541090d484a6ae9938fe02e8/python/pyspark/sql/dataframe.py#L1407

databricks/koalas/series.py
```diff
@@ -4772,20 +4772,49 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
                 raise ValueError("matrices are not aligned")
-        if isinstance(other, Series):
-            result = (self * other).sum()
 
-        return result
+            other = other.copy()
+            column_labels = other._internal.column_labels
+
+            self_column_label = verify_temp_column_name(other, "__self_column__")
+            other[self_column_label] = self
```
itholic, line 4801: This raises `ValueError` when the `compute.ops_on_diff_frames` option is set to `False`.

I'm not sure we should use `option_context` here like `combine_first`.

https://github.com/databricks/koalas/blob/138c7b834b5f00d4d96c8ffb272bc1b1e025bd9c/databricks/koalas/series.py#L4710-L4711

itholic (commented):
Otherwise, seems good to me.

databricks/koalas/series.py
itholic: Oh, sorry we already check `same_anchor` above.

Please ignore this comment.

HyukjinKwon: I wonder if there are a lot of places like this. If there are a lot of such instances to fix, maybe we should just fix all to fail with an exception and let users switch on the configuration.

databricks/koalas/series.py
```diff
@@ -4772,20 +4772,49 @@ def dot(self, other) -> Union[Scalar, "Series"]:
...
         """
         if isinstance(other, DataFrame):
-            raise ValueError(
-                "Series.dot() is currently not supported with DataFrame since "
-                "it will cause expansive calculation as many as the number "
-                "of columns of DataFrame"
-            )
-        if self._kdf is not other._kdf:
-            if len(self.index) != len(other.index):
+            if not same_anchor(self, other) and not self.index.sort_values().equals(
```
HyukjinKwon, line 4792: Ah .. this will be very expensive .. but I have no better idea. But we'll have to fail fast if `compute.ops_on_diff_frames` is disabled. Currently, it looks it's going to sort each DataFrame first and then throw an exception.

databricks/koalas/series.py
xinrong-databricks: Great! Thank you so much for looking into this.

databricks/koalas/series.py
xinrong-databricks: Makes sense! Modified to fail fast here.

databricks/koalas/series.py
xinrong-databricks: I modified this PR to check option_context in the beginning (as 10127ac).

I did a rough grep and found two such places
```
$ git grep "with option_context(\"compute.ops_on_diff_frames\", True):"
databricks/koalas/series.py:            with option_context("compute.ops_on_diff_frames", True):
databricks/koalas/series.py:        with option_context("compute.ops_on_diff_frames", True):
```

Shall we change them to raise an exception instead?

ueshin: Also
```
databricks/koalas/namespace.py:            with ks.option_context("compute.ops_on_diff_frames", True):
```

ueshin: I'm fine with disabling implicit switch-on (in the separate PR).

databricks/koalas/series.py
ueshin: Do we actually need to raise the exception here?
The `sort_values()` is just a planning, so I guess it's not a big deal.
We should rather fix the `Index.equals` which I will work on as a while Index operations.

databricks/koalas/series.py
xinrong-databricks: Sounds good! I will be working on this.

databricks/koalas/series.py
xinrong-databricks: I see. Removed the exception. Thanks!

ueshin (approved):
LGTM.

ueshin:
Thanks! I'd merge this now.

xinrong-databricks:
Thank you!

PR #1936, opened 2020-11-27
Separate matplotlib specific codes and plot core in main codes

HyukjinKwon:
This PR propose to separate matplotlib specific codes and plot core in Koalas:
- `databricks.koalas.plot.core`: other plot backends, the entry point of plotting, shared codes, etc.
- `databricks.koalas.plot.matplotlib`:  `matplotlib` specific codes

PR #1937, opened 2020-11-27
Separate matplotlib specific codes and plot core in test codes

HyukjinKwon:
Similar with https://github.com/databricks/koalas/pull/1936 but tests only. This PR is not dependent on each other, and can be merged alone.
- `tests/plot/test_series_plot.py`/`tests/plot/test_frame_plot.py`: common plot tests
- `tests/plot/test_series_plot_matplotlib.py`/`tests/plot/test_frame_plot_matplotlib.py`: plots with matplotlib


ueshin (approved):
LGTM.

PR #1945, opened 2020-12-01
Implement DataFrame.dot supporting Series parameter

xinrong-databricks:
Compute the matrix multiplication between the DataFrame and other **series only**.

```
        >>> kdf = ks.DataFrame([[0, 1, -2, -1], [1, 1, 1, 1]])
        >>> kser = ks.Series([1, 1, 2, 1])
        >>> kdf.dot(kser)
        0   -4
        1    5
        dtype: int64

        Note how shuffling of the objects does not change the result.

        >>> kser2 = kser.reindex([1, 0, 2, 3])
        >>> kdf.dot(kser2)
        0   -4
        1    5
        dtype: int64
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4059,6 +4059,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         0   -4
         1    5
         dtype: int64
+
+        >>> reset_option("compute.ops_on_diff_frames")
         """
         if not isinstance(other, ks.Series):
             raise TypeError("Unsupported type {}".format(type(other).__name__))
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4065,7 +4065,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         if not isinstance(other, ks.Series):
             raise TypeError("Unsupported type {}".format(type(other).__name__))
         else:
-            return cast(Series, other.dot(self.transpose())).rename(None)
+            return cast(ks.Series, other.dot(self.transpose())).rename(None)
 
     def to_koalas(self, index_col: Optional[Union[str, List[str]]] = None) -> "DataFrame":
         """
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4065,7 +4065,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         if not isinstance(other, ks.Series):
             raise TypeError("Unsupported type {}".format(type(other).__name__))
         else:
-            return other.dot(self.transpose()).rename(None)
+            return cast(Series, other.dot(self.transpose())).rename(None)
 
     def to_koalas(self, index_col: Optional[Union[str, List[str]]] = None) -> "DataFrame":
         """
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4017,6 +4017,25 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         It can also be called using ``self @ other`` in Python >= 3.5.
 
+        .. note:: This method is based on an expensive operation due to the nature
+            of big data. Internally it needs to generate each row for each value, and
+            then group twice - it is a huge operation. To prevent misusage, this method
+            has the 'compute.max_rows' default limit of input length, and raises a ValueError.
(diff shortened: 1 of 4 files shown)


databricks/koalas/frame.py
```diff
@@ -4008,6 +4008,65 @@ def duplicated(self, subset=None, keep="first") -> "Series":
...
+        >>> kser2 = kser.reindex([1, 0, 2, 3])
+        >>> kdf.dot(kser2)
+        0   -4
+        1    5
+        dtype: int64
+
+        >>> reset_option("compute.ops_on_diff_frames")
+        """
+        if not isinstance(other, ks.Series):
+            raise TypeError("Unsupported type {}".format(type(other).__name__))
```
itholic, line 4066: Seems pandas also support many other types as below, not only `Series`.

```python
>>> pdf
   0  1  2  3
0  0  1 -2 -1
1  1  1  1  1
>>> pser
0    1
1    1
2    2
3    1
dtype: int64
>>>
>>> pdf.dot(pser.to_frame())  # DataFrame
   0
0 -4
1  5
>>> pdf.dot(pser.index)  # Index
0   -6
1    6
dtype: int64
>>> pdf.dot([1, 2, 3, 4])  # list
0    -8
1    10
dtype: int64
```

xinrong-databricks: Yeah, I should've clarified in the PR description. This PR targets to support other as `Series` only. 

Considering the implementation complexity and user demand, we might want to support other data types later.

Let me modify the PR description and leave a comment for this.

Thanks for looking into this!

itholic: Oh, sure I got it :)

databricks/koalas/frame.py
```diff
@@ -4008,6 +4008,66 @@ def duplicated(self, subset=None, keep="first") -> "Series":
...
 
+    # TODO: support other as DataFrame or array-like
+    def dot(self, other: "Series") -> "Series":
+        """
+        Compute the matrix multiplication between the DataFrame and other.
+
+        This method computes the matrix product between the DataFrame and the
+        values of an other Series
+
+        It can also be called using ``self @ other`` in Python >= 3.5.
```
ueshin, line 4019: We need to define `__matmul__` for this as same as `Series`?

https://github.com/databricks/koalas/blob/026262dcd7e455be135b7c1e0971768c3c9d37db/databricks/koalas/series.py#L4820-L4824

databricks/koalas/frame.py
```diff
@@ -4008,6 +4008,66 @@ def duplicated(self, subset=None, keep="first") -> "Series":
...
+        Examples
+        --------
+        >>> from databricks.koalas.config import set_option, reset_option
+        >>> set_option("compute.ops_on_diff_frames", True)
+        >>> kdf = ks.DataFrame([[0, 1, -2, -1], [1, 1, 1, 1]])
+        >>> kser = ks.Series([1, 1, 2, 1])
+        >>> kdf.dot(kser)
+        0   -4
+        1    5
+        dtype: int64
```
ueshin, line 4054: Shall we add an example to use `@` operator?

databricks/koalas/frame.py
```diff
@@ -4008,6 +4008,66 @@ def duplicated(self, subset=None, keep="first") -> "Series":
...
+    # TODO: support other as DataFrame or array-like
+    def dot(self, other: "Series") -> "Series":
+        """
+        Compute the matrix multiplication between the DataFrame and other.
+
+        This method computes the matrix product between the DataFrame and the
+        values of an other Series
+
+        It can also be called using ``self @ other`` in Python >= 3.5.
+
```
ueshin, line 4020: Shall we add the same note as `transpose` as this uses it internally?

databricks/koalas/frame.py
xinrong-databricks: Sounds good!

databricks/koalas/frame.py
xinrong-databricks: Added!

databricks/koalas/frame.py
xinrong-databricks: Makes sense. Added it!

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

PR #1947, opened 2020-12-03
Adjust data column names before attaching default index.

ueshin:
Adjust data column names before attaching default index; otherwise `reset_index()` could fail when the `DataFrame` or `Series` is converted from `Index`.

```py
>>> kdf = ks.Index([10, 20, 30]).to_frame()
>>> kdf
     0
10  10
20  20
30  30
>>> kdf.reset_index()
...
AssertionError: Index columns should not appear in columns of the Spark DataFrame. Avoid index column names [re.compile('__index_level_[0-9]+__')].
```

or

```py
>>> kser = ks.Index([10, 20, 30]).to_series()
>>> kser
10    10
20    20
30    30
dtype: int64
>>> kser.reset_index(drop=True)
...
AssertionError: Index columns should not appear in columns of the Spark DataFrame. Avoid index column names [re.compile('__index_level_[0-9]+__')].
```

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3515,7 +3515,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                 raise ValueError("cannot insert {}, already exists".format(name_like_string(label)))
 
         if self._internal.column_labels_level > 1:
-            column_depth = self._internal.column_labels_level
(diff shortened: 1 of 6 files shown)


HyukjinKwon (approved):
Thanks Takuya. I think this looks fine to me.

ueshin:
Thanks! merging.

PR #1953, opened 2020-12-04
Remove implicit switch-ons of "compute.ops_on_diff_frames"

xinrong-databricks:
## Proposal
- Remove implicit switch-ons of "compute.ops_on_diff_frames".
- Fix doctests
- Fix unit tests


xinrong-databricks:
Shall we modify the docstrings of each touched function to reflect this change?

HyukjinKwon:
I tend to agree with this change ... but to confirm WDYT @ueshin and @itholic?

itholic:
I agree, too!

> Shall we modify the docstrings of each touched function to reflect this change?

For this, IMO, I think we don't necessarily need to modify the docstrings since it already shows a proper error message - `Cannot combine the series or dataframe because it comes from a different dataframe. In order to allow this operation, enable 'compute.ops_on_diff_frames' option.` -.

ueshin (approved):
LGTM.

databricks/koalas/series.py
```diff
@@ -5590,11 +5593,12 @@ def compare(
...
+
+        >>> reset_option("compute.ops_on_diff_frames")
         """
-        with option_context("compute.ops_on_diff_frames", True):
-            if not self.index.equals(other.index):
-                raise ValueError("Can only compare identically-labeled Series objects")
-            combined = combine_frames(self.to_frame(), other.to_frame())
+        if not self.index.equals(other.index):
+            raise ValueError("Can only compare identically-labeled Series objects")
+        combined = combine_frames(self.to_frame(), other.to_frame())
```
ueshin, line 5601: Btw, we should consider other is from the same anchor:

```py
>>> s1 = ks.Series(["a", "b", "c", "d", "e"])
>>> s1.compare(s1)
...
AssertionError: We don't need to combine. `this` and `that` are same.
```

xinrong-databricks: Good idea! Let me file a new PR for this.

PR #1955, opened 2020-12-08
Enable index operations.

ueshin:
Currently Koalas can't handle index operations very well.

```py
>>> kidx = ks.Index([1, 2, 3, 4, 5])
>>> kidx + kidx
Int64Index([2, 4, 6, 8, 10], dtype='int64')
>>> kidx + kidx + kidx
...
AssertionError: args should be single DataFrame or single/multiple Series
```

or

```py
>>> ks.Index([1, 2, 3, 4, 5]) + ks.Index([6, 7, 8, 9, 10])
...
AssertionError: args should be single DataFrame or single/multiple Series
```

This PR enables those operations:

```py
>>> kidx + kidx + kidx
Int64Index([3, 6, 9, 12, 15], dtype='int64')

>>> ks.Index([1, 2, 3, 4, 5]) + ks.Index([6, 7, 8, 9, 10])
Int64Index([7, 9, 13, 11, 15], dtype='int64')
```

### DIFF
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -1153,6 +1153,8 @@ class Index(IndexOpsMixin):
                         scol_for(sdf, col) for col in self._internal.index_spark_column_names
                     ],
                     index_names=self._internal.index_names,
+                    column_labels=[],
+                    data_spark_columns=[],
                 )
             )
         )
@@ -3161,6 +3163,8 @@ class MultiIndex(Index):
                         scol_for(sdf, col) for col in internal.index_spark_column_names
                     ],
                     index_names=internal.index_names,
+                    column_labels=[],
+                    data_spark_columns=[],
                 )
             )
         )
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -170,7 +170,7 @@ class Index(IndexOpsMixin):
         if get_option("compute.ops_on_diff_frames"):
             # This could cause as many counts, reset_index calls, joins for combining
             # as the number of `Index`s in `args`. So far it's fine since we can assume the ops
-            # only work between two `Index`s at the most. We might need to fix it in the future.
+            # only work between at most two `Index`s. We might need to fix it in the future.
             self_len = len(self)
             if any(len(col) != self_len for col in args if isinstance(col, IndexOpsMixin)):
                 raise ValueError("operands could not be broadcast together with shapes")
--- a/databricks/koalas/indexes.py
+++ b/databricks/koalas/indexes.py
@@ -47,7 +47,7 @@ from pyspark.sql.types import TimestampType, IntegralType, DataType
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
 from databricks.koalas.config import get_option, option_context
 from databricks.koalas.exceptions import PandasNotImplementedError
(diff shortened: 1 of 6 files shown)


databricks/koalas/indexes.py
```diff
@@ -154,14 +156,50 @@ def _with_new_scol(self, scol: spark.Column) -> "Index":
...
+        return self._internal.spark_frame is not other._internal.spark_frame
+
+    def _align_and_column_op(self, f, *args) -> "Index":
+        if get_option("compute.ops_on_diff_frames"):
+            # This could cause as many counts, reset_index calls, joins for combining
+            # as the number of `Index`s in `args`. So far it's fine since we can assume the ops
+            # only work between at most two `Index`s. We might need to fix it in the future.
+            self_len = len(self)
+            if any(len(col) != self_len for col in args if isinstance(col, IndexOpsMixin)):
+                raise ValueError("operands could not be broadcast together with shapes")
```
xinrong-databricks, line 176: How does length comparison relate to broadcast in the error message?

databricks/koalas/utils.py
```diff
@@ -233,15 +240,17 @@ def fill_label(label):
...
         )
     else:
-        raise ValueError(
-            "Cannot combine the series or dataframe because it comes from a different dataframe. "
-            "In order to allow this operation, enable 'compute.ops_on_diff_frames' option."
-        )
+        raise ValueError(ERROR_MESSAGE_CANNOT_COMBINE)
 
 
 def align_diff_frames(
```
xinrong-databricks, line 246: Great, thank you!

databricks/koalas/utils.py
```diff
@@ -376,12 +385,11 @@ def align_diff_frames(
     return kdf
 
 
-def align_diff_series(func, this_series, *args, how="full"):
-    from databricks.koalas.base import IndexOpsMixin
+def align_diff_series(func, this_series: "Series", *args, how: str = "full") -> "Series":
```
xinrong-databricks, line 388: I'm just wondering whether we should keep `align_diff_series` here or move to `series.py` since it is only for Series.

databricks/koalas/indexes.py
ueshin: The term "broadcast" here is not the same as Spark's "broadcast".
Maybe the term should've been "broad-cast", which means cast to broader types, or in this case cast to larger size.

databricks/koalas/utils.py
ueshin: Yeah, maybe we can move it, but I'd leave it to the future PRs.

xinrong-databricks: Sounds good!

databricks/koalas/indexes.py
xinrong-databricks: I see, maybe that's what we are talking about
```
>>> z= np.arange(12).reshape(3,4)
>>> m= np.arange(9).reshape(3,3)
>>> z * m
  File "<input>", line 1, in <module>
    z * m
ValueError: operands could not be broadcast together with shapes (3,4) (3,3)
```

The current error message also looks good.

xinrong-databricks (approved):

ueshin:
Thanks! I'd merge this now. Please feel free to leave comments if any. @HyukjinKwon @itholic

PR #1956, opened 2020-12-08
Compare Series from the same anchor

xinrong-databricks:
Enable Series.compare to compare Series from the same anchor 

Preivous behavior:
```
>>> kser = ks.Series([1, 2])
>>> kser.compare(kser)
...
AssertionError: We don't need to combine. `this` and `that` are same.

>>> kser.compare(kser + 1)
...
AssertionError: We don't need to combine. `this` and `that` are same.
```

Current behavior:
```
>>> kser = ks.Series([1, 2])
>>> kser.compare(kser)
Empty DataFrame
Columns: [self, other]
Index: []

>>> kser.compare(kser + 1)
   self  other
0     1      2
1     2      3
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -5627,7 +5627,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                 self._internal.with_new_columns(
                     [self._internal.data_spark_columns[0]]
                     + [other._internal.data_spark_columns[0]],
-                    column_labels=[self_column_label] + [other_column_label],
+                    column_labels=[self_column_label, other_column_label],
                 )
             )  # type: DataFrame
         else:
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -5629,7 +5629,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                     + [other._internal.data_spark_columns[0]],
                     column_labels=[self_column_label] + [other_column_label],
                 )
-            )
+            )  # type: DataFrame
         else:
             if not self.index.equals(other.index):
                 raise ValueError("Can only compare identically-labeled Series objects")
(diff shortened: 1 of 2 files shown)


databricks/koalas/series.py
```diff
@@ -5620,9 +5620,21 @@ def compare(
...
-            raise ValueError("Can only compare identically-labeled Series objects")
-        combined = combine_frames(self.to_frame(), other.to_frame())
+        if same_anchor(self, other):
+            self_column_label = verify_temp_column_name(other.to_frame(), "__self_column__")
+            other_column_label = verify_temp_column_name(self.to_frame(), "__other_column__")
+            combined = DataFrame(
+                self._internal.with_new_columns(
+                    [self._internal.data_spark_columns[0]]
+                    + [other._internal.data_spark_columns[0]],
+                    column_labels=[self_column_label] + [other_column_label],
```
ueshin, line 5630: nit: we can use one list for each: `[self..., other...]` instead of `[self...] + [other...]`.

ueshin (approved):
Otherwise, LGTM.

databricks/koalas/series.py
xinrong-databricks: 👍  Modified!

databricks/koalas/series.py
```diff
@@ -5620,9 +5620,21 @@ def compare(
...
-        if not self.index.equals(other.index):
-            raise ValueError("Can only compare identically-labeled Series objects")
-        combined = combine_frames(self.to_frame(), other.to_frame())
+        if same_anchor(self, other):
+            self_column_label = verify_temp_column_name(other.to_frame(), "__self_column__")
+            other_column_label = verify_temp_column_name(self.to_frame(), "__other_column__")
+            combined = DataFrame(
+                self._internal.with_new_columns(
+                    [self._internal.data_spark_columns[0]]
+                    + [other._internal.data_spark_columns[0]],
```
ueshin, line 5629: nit: this line, too?

ueshin:
Thanks! merging.

Issue #1929, opened 2020-11-24
Implement Koalas Missing APIs


xinrong-databricks:
Help wanted! A few popular pandas APIs are missing in Koalas. We are going to implement them!

Please use this thread to comment on which function you will be working so we don't duplicate work. Please mention this issue in your PR so that the list below can be updated.

- DataFrame
  - [ ] [cov](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.cov.html)
  - [ ] [first](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.first.html)
  - [ ] [itertuples](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.itertuples.html)
  - [ ] [resample](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.resample.html)
  - [ ] [combine_first](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.combine_first.html)
  - [ ] [between_time](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.between_time.html)
  - [ ] [boxplot](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.boxplot.html)
  - [ ] [insert](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.insert.html)
  - [ ] [swapaxes](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.swapaxes.html)

- Series
  - [ ] [ewm](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.ewm.html)
  - [ ] [resample](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.resample.html)
  - [ ] [factorize](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.factorize.html)
  - [ ] [last](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.last.html)
  - [ ] [swapaxes](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.swapaxes.html)

- Index
[...]

PR #1957, opened 2020-12-09
Implemented GroupBy.median()

itholic:
This PR proposes `GroupBy.median()`.

Note: the result can be slightly different from pandas since we use an approximated median based upon approximate percentile computation because computing median across a large dataset is extremely expensive.

```python
>>> kdf = ks.DataFrame({'a': [1., 1., 1., 1., 2., 2., 2., 3., 3., 3.],
...                     'b': [2., 3., 1., 4., 6., 9., 8., 10., 7., 5.],
...                     'c': [3., 5., 2., 5., 1., 2., 6., 4., 3., 6.]},
...                    columns=['a', 'b', 'c'],
...                    index=[7, 2, 4, 1, 3, 4, 9, 10, 5, 6])
>>> kdf
      a     b    c
7   1.0   2.0  3.0
2   1.0   3.0  5.0
4   1.0   1.0  2.0
1   1.0   4.0  5.0
3   2.0   6.0  1.0
4   2.0   9.0  2.0
9   2.0   8.0  6.0
10  3.0  10.0  4.0
5   3.0   7.0  3.0
6   3.0   5.0  6.0

>>> kdf.groupby('a').median().sort_index()  # doctest: +NORMALIZE_WHITESPACE
       b    c
a
1.0  2.0  3.0
2.0  8.0  2.0
3.0  7.0  4.0

>>> kdf.groupby('a')['b'].median().sort_index()
a
1.0    2.0
2.0    8.0
3.0    7.0
Name: b, dtype: float64
```

ref #1929

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -71,6 +71,7 @@ from databricks.koalas.utils import (
 from databricks.koalas.spark.utils import as_nullable_spark_type, force_decimal_precision_scale
 from databricks.koalas.window import RollingGroupby, ExpandingGroupby
 from databricks.koalas.exceptions import DataError
+from databricks.koalas.spark import functions as SF
 
 # to keep it the same as pandas
 NamedAgg = namedtuple("NamedAgg", ["column", "aggfunc"])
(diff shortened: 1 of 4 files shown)


itholic:
FYI: we're doing same thing in median for DataFrame and Series.

https://github.com/databricks/koalas/blob/9e8d99b208b3dfbb28407798372f776a66108932/databricks/koalas/generic.py#L1898-L1904

databricks/koalas/tests/test_groupby.py
```diff
@@ -2608,3 +2608,16 @@ def test_get_group(self):
         self.assertRaises(
             ValueError, lambda: kdf.groupby([("B", "class"), ("A", "name")]).get_group("mammal")
         )
+
+    def test_median(self):
```
xinrong-databricks, line 2612: I'm wondering if a positive test case is needed here? Then we might compare the result DataFrame with a pre-created Koalas DataFrame.

itholic: Sounds good. I'll add some positive test cases!

xinrong-databricks: Thank you!

xinrong-databricks:
Otherwise, LGTM!

databricks/koalas/groupby.py
```diff
@@ -2277,6 +2278,73 @@ def get_group(self, name) -> Union[DataFrame, Series]:
 
         return DataFrame(internal)
 
+    def median(self, numeric_only=True, accuracy=10000):
```
ueshin, line 2281: Shall we add type annotations?

ueshin (approved):
Could you fix the conflicts as well?
Otherwise, LGTM.

databricks/koalas/groupby.py
itholic: Oh, sure. Thanks!

databricks/koalas/groupby.py
```diff
@@ -2343,6 +2344,73 @@ def get_group(self, name) -> Union[DataFrame, Series]:
 
         return DataFrame(internal)
 
+    def median(self, numeric_only=True, accuracy=10000) -> Union[DataFrame, Series]:
```
ueshin, line 2347: Are `numeric_only` and `accuracy` unknown types?

ueshin: We should add more type hints for function arguments in the separate PR.

ueshin:
Thanks! I'd merge this now.
Let's add more type hints for function arguments later.

databricks/koalas/groupby.py
itholic: Sure, I'll add in the separated PR.

Thanks for the review!

itholic: Oh, anyway `numeric_only` only supports for `True` since this is only for pandas compatibility.

Should we add type hints as `bool` though ?

ueshin: Yep, it must be `bool`, and `accuracy` must be `int`?
There are other similar places in the file or the other files, we can do it in a batch.

itholic: Sounds good. I will take a look and add type hints throughout the whole file next week !

PR #1962, opened 2020-12-10
Fix DataFrame.replace with NaN/None values #1907

### LINKED ISSUE #1907: Function "replace" Case
I got an unexpected result when using funciton "replace" to replace "np.nan"

    >>> import numpy as np
    >>> df = ks.DataFrame([[1, np.inf], [2, -np.inf], [3, np.nan]], columns=['c1', 'c2'])
    >>> print(df)
         c1    c2
     0    1    inf
     1    2    -inf
     2    3    NaN
    >>> print(df.replace([np.inf, -np.inf, np.nan], 0))
         c1    c2
     0    1    0
     1    2    0
     2    3    NaN
    >>> print(df.mask(df['c2'].isin([np.inf, -np.inf, np.nan]), 0))
         c1    c2
     0    1    0
     1    2    0
     2    3    0

LucasG0:
Hi, this PR would close #1907.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -4264,7 +4264,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         dtype: float64
         """
         if to_replace is None:
-            return self
+            return self.fillna(method="ffill")
         if not isinstance(to_replace, (str, list, dict, int, float)):
             raise ValueError("'to_replace' should be one of str, list, dict, int, float")
         if regex:
@@ -4283,14 +4283,23 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                 current = self.spark.column
             else:
                 for to_replace_, value in to_replace.items():
+                    cond = (
+                        (F.isnan(self.spark.column) | self.spark.column.isNull())
+                        if pd.isna(to_replace_)
+                        else (self.spark.column == F.lit(to_replace_))
+                    )
                     if is_start:
-                        current = F.when(self.spark.column == F.lit(to_replace_), value)
+                        current = F.when(cond, value)
                         is_start = False
                     else:
-                        current = current.when(self.spark.column == F.lit(to_replace_), value)
+                        current = current.when(cond, value)
                 current = current.otherwise(self.spark.column)
         else:
-            current = F.when(self.spark.column.isin(to_replace), value).otherwise(self.spark.column)
+            cond = self.spark.column.isin(to_replace)
+            # to_replace may be a scalar
+            if np.array(pd.isna(to_replace)).any():
+                cond = cond | F.isnan(self.spark.column) | self.spark.column.isNull()
+            current = F.when(cond, value).otherwise(self.spark.column)
 
         return self._with_new_scol(current)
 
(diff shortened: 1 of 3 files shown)


databricks/koalas/series.py
```diff
@@ -4283,14 +4283,23 @@ def replace(self, to_replace=None, value=None, regex=False) -> "Series":
                 current = self.spark.column
             else:
                 for to_replace_, value in to_replace.items():
+                    cond = (
+                        (F.isnan(self.spark.column) | self.spark.column.isNull())
```
itholic, line 4287: Not pretty sure, but I think maybe we don't need `F.isnan(self.spark.column)` because both `None` and `np.nan` cast to `null` in the internal Spark DataFrame ??

```python
>>> kser_with_nan
0    1.0
1    2.0
2    NaN
3    4.0
dtype: float64

>>> kser_with_none
0       a
1       b
2    None
3       d
dtype: object

>>> kser_with_nan._internal.spark_frame.select(kser_with_nan.spark.column).show()
+----+
|   0|
+----+
| 1.0|
| 2.0|
|null|
| 4.0|
+----+

>>> kser_with_none._internal.spark_frame.select(kser_with_none.spark.column).show()
+----+
|   0|
+----+
|   a|
|   b|
|null|
|   d|
+----+
```

databricks/koalas/series.py
```diff
@@ -4283,14 +4283,23 @@ def replace(self, to_replace=None, value=None, regex=False) -> "Series":
...
                     else:
-                        current = current.when(self.spark.column == F.lit(to_replace_), value)
+                        current = current.when(cond, value)
                 current = current.otherwise(self.spark.column)
         else:
-            current = F.when(self.spark.column.isin(to_replace), value).otherwise(self.spark.column)
+            cond = self.spark.column.isin(to_replace)
+            # to_replace may be a scalar
+            if np.array(pd.isna(to_replace)).any():
+                cond = cond | F.isnan(self.spark.column) | self.spark.column.isNull()
```
itholic, line 4301: ditto ?

itholic (commented):
Otherwise, looks pretty fine to me.

databricks/koalas/series.py
LucasG0: I think we can keep using `F.isnan`, some use cases may produce NaN in the internal Spark DataFrame : 

```python
>>> kdf = ks.DataFrame({"A": [-1, 2, 3, 4, np.nan]})
>>> kdf._internal.spark_frame.withColumn("sqrt(A)", sqrt("A")).show()
+------------------+
|           sqrt(A)|
+------------------+
|               NaN|
|1.4142135623730951|
|1.7320508075688772|
|               2.0|
|              null|
+------------------+
```

itholic: Great! let's keep it 👍

itholic:
Seems good to me. Let's wait a while for reviews from others :)

itholic (approved):

HyukjinKwon (approved):

HyukjinKwon:
Thanks @LucasG0. Merged.

itholic:
Thanks, @LucasG0 !! :D

LucasG0:
Thanks. :)

PR #1712, opened 2020-08-17
Support non-named Series.

ueshin:
Adding support non-named Series.
Currently Koalas Series must have its name, but this PR allows it.

```py
>>> ks.Series([1, 2, 3])
0    1
1    2
2    3
dtype: int64

>>> ks.Series([1, 2, 3], name='x')
0    1
1    2
2    3
Name: x, dtype: int64
```

PR #1964, opened 2020-12-11
Removed comment for Series.rename test

itholic:
Since #1712, Koalas is able to support non-names Series.

So now we can remove the comment from related test.


ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

PR #1968, opened 2020-12-12
between_time api for koalas dataframe

shril:
Implement Koalas Missing APIs #1929

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -2977,20 +2977,23 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
             If the datatype of index is not datetime
         Examples
         --------
-        >>> i = pd.date_range('2018-04-09', periods=4, freq='1D20min')
-        >>> ts = ks.DataFrame({'A': [1, 2, 3, 4]}, index=i)
+        >>> idx = pd.date_range('2018-04-09', periods=4, freq='1D20min')
+        >>> ts = ks.DataFrame({'A': [1, 2, 3, 4]}, index=idx)
         >>> ts
                              A
         2018-04-09 00:00:00  1
         2018-04-10 00:20:00  2
         2018-04-11 00:40:00  3
         2018-04-12 01:00:00  4
+
         >>> ts.between_time('0:15', '0:45')
                              A
         2018-04-10 00:20:00  2
         2018-04-11 00:40:00  3
+
         You get the times that are *not* between two times by setting
         ``start_time`` later than ``end_time``:
+
         >>> ts.between_time('0:45', '0:15')
                              A
         2018-04-09 00:00:00  1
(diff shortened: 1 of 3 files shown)


shril:
@ueshin, @HyukjinKwon can you take a look at this PR?

databricks/koalas/frame.py
```diff
@@ -2950,6 +2950,87 @@ class  locomotion
...
+            Time passed either as object (datetime.time) or as string in
+            appropriate format ("%H:%M", "%H%M", "%I:%M%p", "%I%M%p",
+            "%H:%M:%S", "%H%M%S", "%I:%M:%S%p","%I%M%S%p").
+        include_start : bool, default True
+        include_end : bool, default True
+        Returns
+        -------
+        values_between_time : array of integers
+        """
+        return self.index.to_pandas().indexer_between_time(
```
HyukjinKwon, line 3030: @shril can we avoid calling `to_pandas()`? It will bring all data from other nodes to the single client node which can easily OOM.

shril: @HyukjinKwon, my idea here was to convert just the index column to pandas since there is API support for Datetime index hence the `indexer_between_time()` function.

If you want me to, I can also start with Datetime index support in koalas, and then work on this PR.

itholic: Thanks for the contribution , @shril . :D

Yeah, I'd say we need Datetime index support in Koalas, not use pandas'.

shril: @itholic , starting with a new PR for `DatetimeIndex`. 
@HyukjinKwon can we leave this PR open till I implement the other PR?
Thanks. :)

vmdhhh: I am fairly new to this so I am sorry if I sound too naive, but doesn't `is_all_dates` function from `koalas` provide support for `Datetimeindex`? @itholic

itholic: @shril,
Sure, we can just leave this PR as it is and let's revisit after finishing `DatetimeIndex` ! :D

@vmdhhh, 
Thanks for your interest to Koalas !!
`DatetimeIndex` in Koalas now actually not `DatetimeIndex`, but just `Index`.
It's only shown as `DatetimeIndex` only when calling repr as below.
```python
>>> idx = ks.Index([datetime(2019, 1, 1, 0, 0, 0), datetime(2019, 2, 3, 0, 0, 0)])
>>> idx  # It's repr is `DatetimeIndex` since we internally convert this to pandas and use pandas' repr.
DatetimeIndex(['2019-01-01', '2019-02-03'], dtype='datetime64[ns]', freq=None)
>>> type(idx)  # Actually it's instance of `Index`, not `DatetimeIndex`
<class 'databricks.koalas.indexes.Index'>
>>> type(idx.to_pandas())
<class 'pandas.core.indexes.datetimes.DatetimeIndex'>
```

`is_all_dates` just check if all data included in the Index is Spark `TimestampType` type.

https://github.com/databricks/koalas/blob/b81afcc7a461eaad6489e51200862ecdadd04b3c/databricks/koalas/indexes.py#L1973

vmdhhh: Thank you @itholic for clearing this.

xinrong-databricks:
Currently `between_time` is implemented https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.DataFrame.between_time.html. May I close the pull request?

PR #1970, opened 2020-12-14
Simplify plot backend support

HyukjinKwon:
This PR proposes to simplify plot implementation. Current Koalas implementation attempts to map the argument between other plotting backends (e.g., matplotlib vs plotly). Keeping this map is a huge maintenance cost and it's unrealistic to track their change and keep updating this map.

Pandas itself does not keep this map either:

```python
>>> import pandas as pd
>>> pd.DataFrame([1,2,3]).plot.line(logx=1)
<AxesSubplot:>
>>> pd.options.plotting.backend = "plotly"
>>> pd.DataFrame([1,2,3]).plot.line(logx=1)
  File "/.../opt/miniconda3/envs/python3.8/lib/python3.8/site-packages/pandas/plotting/_core.py", line 1017, in line
    return self(kind="line", x=x, y=y, **kwargs)
  File "/.../opt/miniconda3/envs/python3.8/lib/python3.8/site-packages/pandas/plotting/_core.py", line 879, in __call__
    return plot_backend.plot(self._parent, x=x, y=y, kind=kind, **kwargs)
  File "/.../miniconda3/envs/python3.8/lib/python3.8/site-packages/plotly/__init__.py", line 102, in plot
    return line(data_frame, **kwargs)
TypeError: line() got an unexpected keyword argument 'logx'
```

### DIFF
--- a/databricks/koalas/plot/core.py
+++ b/databricks/koalas/plot/core.py
@@ -189,10 +189,10 @@ class KoalasPlotAccessor(PandasObject):
                 "area": SampledPlot().get_sampled,
                 "line": SampledPlot().get_sampled,
             }
-            if data_preprocessor_map[kind]:
+            if not data_preprocessor_map[kind]:
                 raise NotImplementedError(
                     "'%s' plot is not supported with '%s' plot "
-                    "backend yet." % plot_backend.__name__
+                    "backend yet." % (kind, plot_backend.__name__)
                 )
             plot_data = data_preprocessor_map[kind](plot_data)
             return plot_backend.plot(plot_data, kind=kind, **kwargs)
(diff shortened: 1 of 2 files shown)


databricks/koalas/plot/core.py
```diff
@@ -176,128 +176,26 @@ def _get_plot_backend(backend=None):
...
-                "xlim": "range_x",
-                "ylim": "range_y",
-                "yerr": "error_y",
-                "xerr": "error_x",
-            }
-        }
-
-        if isinstance(data, Series):
-            positional_args = {
-                ("ax", None),
```
HyukjinKwon, line 210: I think this mapping was copied from pandas, which is actually not needed in our case since we're dispatching it in each method e.g.) `KoalasPlotAccessor.area`.

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

xinrong-databricks:
Awesome!

PR #1972, opened 2020-12-16
Implement Series.factorize()

xinrong-databricks:
ref #1929
```
        >>> kser = ks.Series(['b', None, 'a', 'c', 'b'])
        >>> codes, uniques = kser.factorize()
        >>> codes
        0    1
        1   -1
        2    0
        3    2
        4    1
        dtype: int64
        >>> uniques

        >>> codes, uniques = kser.factorize(na_sentinel=None)
        >>> codes
        0    1
        1    3
        2    0
        3    2
        4    1
        dtype: int64
        >>> uniques
        Index(['a', 'b', 'c', None], dtype='object')

        >>> codes, uniques = kser.factorize(na_sentinel=-2)
        >>> codes
        0    1
        1   -2
        2    0
        3    2
        4    1
        dtype: int64
        >>> uniques
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -2018,17 +2018,20 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
             chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
         )
 
-        map_scol = F.create_map(kvs)
-
         scol = self.spark.column
+
         if isinstance(self.spark.data_type, (FloatType, DoubleType)):
             cond = scol.isNull() | F.isnan(scol)
         else:
             cond = scol.isNull()
-
         null_scol = F.when(cond, F.lit(na_sentinel_code))
-        mapped_scol = map_scol.getItem(scol)
-        new_scol = null_scol.otherwise(mapped_scol)
+
+        map_scol = None if len(kvs) <= 0 else F.create_map(kvs)
+
+        if map_scol is not None:
+            new_scol = null_scol.otherwise(map_scol.getItem(scol))
+        else:
+            new_scol = null_scol
 
         internal = self._internal.with_new_columns(
             [new_scol.alias(self._internal.data_spark_column_names[0])]
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -2003,28 +2003,21 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
             )
 
         uniques_list = first_series(uniq_pdf).tolist()
-        uniques_list = sorted(uniques_list, key=lambda x: (x is None, x))
+        uniques_list = sorted(uniques_list, key=lambda x: (pd.isna(x), x))
 
-        # Constructs `unique_to_code` mapping unique to code
+        # Constructs `unique_to_code` mapping non-na unique to code
         unique_to_code = {}
         if na_sentinel is not None:
-            unique_to_code = {np.nan: na_sentinel}
             na_sentinel_code = na_sentinel
         code = 0
         for unique in uniques_list:
             if unique not in unique_to_code:
-                if unique is None and na_sentinel is not None:
-                    continue
-
-                if unique is None:
-                    na_sentinel_code = code
+                if pd.isna(unique):
+                    if na_sentinel is None:
+                        na_sentinel_code = code
                 else:
                     unique_to_code[unique] = code
                 code += 1
-
-        sdf = self._internal.resolved_copy.spark_frame
-        scol_name = self._internal.data_spark_column_names[0]
-        new_scol_name = cast(str, verify_temp_column_name(sdf, "__new_column__"))
 
         kvs = list(
             chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
@@ -2038,7 +2031,8 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         codes = first_series(DataFrame(internal))
 
         if na_sentinel is not None:
-            uniques = ks.Index([x for x in uniques_list if x is not None])
+            # Drops the NaN from the uniques of the values
+            uniques = ks.Index([x for x in uniques_list if not pd.isna(x)])
         else:
             uniques = ks.Index(uniques_list)
 
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1996,12 +1996,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                     )
                 )
         else:
-            raise ValueError(
-                "Please set 'compute.max_rows' by using 'databricks.koalas.config.set_option' "
-                "to restrict the total number of unique values of the current Series."
-                "Note that, before changing the 'compute.max_rows', "
-                "this operation is considerably expensive."
-            )
+            uniq_pdf = uniq_sdf.toPandas()
 
         uniques_list = first_series(uniq_pdf).tolist()
         uniques_list = sorted(uniques_list, key=lambda x: (pd.isna(x), x))
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1997,8 +1997,9 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                 )
         else:
             uniq_pdf = uniq_sdf.toPandas()
-
-        uniques_list = first_series(uniq_pdf).tolist()
+        # pandas takes both NaN and null in Spark to np.nan, so de-duplication is required
+        uniq_series = first_series(uniq_pdf).drop_duplicates()
+        uniques_list = uniq_series.tolist()
         uniques_list = sorted(uniques_list, key=lambda x: (pd.isna(x), x))
 
         # Constructs `unique_to_code` mapping non-na unique to code
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1914,7 +1914,9 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
             )
         )._kser_for(self._column_label)
 
-    def factorize(self, sort: bool = True, na_sentinel: Optional[int] = -1):
+    def factorize(
+        self, sort: bool = True, na_sentinel: Optional[int] = -1
+    ) -> Tuple[pd.Series, "ks.Index"]:
         """
         Encode the object as an enumerated type or categorical variable.
         This method is useful for obtaining a numeric representation of an
@@ -1930,11 +1932,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         Returns
         -------
         codes : Series
-            An Series that's an indexer into `uniques`.
+            A Series that's an indexer into `uniques`.
             ``uniques.take(codes)`` will have the same values as `values`.
         uniques : Index
-            The unique valid values. When `values` is Koalas object, an
-            `Index` is returned.
+            The unique valid values.
 
             .. note ::
 
@@ -2031,13 +2032,9 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
 
         if na_sentinel is not None:
             # Drops the NaN from the uniques of the values
-            non_na_list = [x for x in uniques_list if not pd.isna(x)]
-            if len(non_na_list) == 0:
-                uniques = pd.Index(non_na_list)
-            else:
-                uniques = ks.Index(non_na_list)
-        else:
-            uniques = ks.Index(uniques_list)
+            uniques_list = [x for x in uniques_list if not pd.isna(x)]
+
+        uniques = pd.Index(uniques_list)
 
         return codes, uniques
 
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -2018,20 +2018,18 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
             chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
         )
 
-        scol = self.spark.column
-
-        if isinstance(self.spark.data_type, (FloatType, DoubleType)):
-            cond = scol.isNull() | F.isnan(scol)
+        if len(kvs) == 0:  # uniques are all missing values
+            new_scol = F.lit(na_sentinel_code)
         else:
-            cond = scol.isNull()
-        null_scol = F.when(cond, F.lit(na_sentinel_code))
+            scol = self.spark.column
+            if isinstance(self.spark.data_type, (FloatType, DoubleType)):
+                cond = scol.isNull() | F.isnan(scol)
+            else:
+                cond = scol.isNull()
+            map_scol = F.create_map(kvs)
 
-        map_scol = None if len(kvs) <= 0 else F.create_map(kvs)
-
-        if map_scol is not None:
+            null_scol = F.when(cond, F.lit(na_sentinel_code))
             new_scol = null_scol.otherwise(map_scol.getItem(scol))
-        else:
-            new_scol = null_scol
 
         internal = self._internal.with_new_columns(
             [new_scol.alias(self._internal.data_spark_column_names[0])]
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -2011,13 +2011,12 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
             na_sentinel_code = na_sentinel
         code = 0
         for unique in uniques_list:
-            if unique not in unique_to_code:
-                if pd.isna(unique):
-                    if na_sentinel is None:
-                        na_sentinel_code = code
-                else:
-                    unique_to_code[unique] = code
-                code += 1
+            if pd.isna(unique):
+                if na_sentinel is None:
+                    na_sentinel_code = code
+            else:
+                unique_to_code[unique] = code
+            code += 1
 
         kvs = list(
             chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -1951,7 +1951,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         2    0
         3    2
         4    1
-        dtype: int64
+        dtype: int32
         >>> uniques
         Index(['a', 'b', 'c'], dtype='object')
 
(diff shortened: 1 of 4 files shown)


databricks/koalas/series.py
```diff
@@ -1907,6 +1907,93 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+            uniq_pdf = uniq_sdf.limit(max_compute_count + 1).toPandas()
+            if len(uniq_pdf) > max_compute_count:
+                raise ValueError(
+                    "Current Series has more then {0} unique values. "
+                    "Please set 'compute.max_rows' by using 'databricks.koalas.config.set_option' "
+                    "to more than {0} rows. Note that, before changing the "
+                    "'compute.max_rows', this operation is considerably expensive.".format(
+                        max_compute_count
+                    )
+                )
```
ueshin, line 1950: `else` for `if max_compute_count is not None` is missing?

databricks/koalas/tests/test_series.py
```diff
@@ -2251,6 +2251,45 @@ def test_first_valid_index(self):
...
+
+        pcodes, puniques = pser.factorize(sort=True, na_sentinel=-2)
+        kcodes, kuniques = kser.factorize(na_sentinel=-2)
+        self.assert_eq(pcodes.tolist(), kcodes.to_list())
+        self.assert_eq(puniques, kuniques)
+
+        pcodes, puniques = pser.factorize(sort=True, na_sentinel=2)
+        kcodes, kuniques = kser.factorize(na_sentinel=2)
+        self.assert_eq(pcodes.tolist(), kcodes.to_list())
+        self.assert_eq(puniques, kuniques)
```
ueshin, line 2284: Could you try with `(kser + 1).factorize()`?

databricks/koalas/series.py
xinrong-databricks: Yeah, added!

databricks/koalas/tests/test_series.py
xinrong-databricks: Sounds good! Added.

databricks/koalas/series.py
```diff
@@ -1907,6 +1908,136 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+        kvs = list(
+            chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
+        )
+        sdf = sdf.withColumn(new_scol_name, F.create_map(kvs))
+
+        null_scol = F.when(scol_for(sdf, scol_name).isNull(), F.lit(na_sentinel_code))
+        mapped_scol = scol_for(sdf, new_scol_name).getItem(scol_for(sdf, scol_name))
+        sdf = sdf.withColumn(scol_name, null_scol.otherwise(mapped_scol))
+
+        codes = first_series(DataFrame(self._internal.with_new_sdf(sdf))).astype(np.int64)
```
ueshin, line 2032: I guess we can make it not create a new Spark DataFrame for `codes`.

xinrong-databricks: Thank you! Updated.

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,136 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+                "this operation is considerably expensive."
+            )
+
+        uniques_list = first_series(uniq_pdf).tolist()
+        uniques_list = sorted(uniques_list, key=lambda x: (x is None, x))
+
+        # Constructs `unique_to_code` mapping unique to code
+        unique_to_code = {}
+        if na_sentinel is not None:
+            unique_to_code = {np.nan: na_sentinel}
```
ueshin, line 2011: Do we need `np.nan` here?

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,136 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+            na_sentinel_code = na_sentinel
+        code = 0
+        for unique in uniques_list:
+            if unique not in unique_to_code:
```
ueshin, line 2015: Why do we need this `if` case?

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,136 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+
+                if unique is None:
+                    na_sentinel_code = code
+                else:
+                    unique_to_code[unique] = code
+                code += 1
+
+        sdf = self._internal.resolved_copy.spark_frame
+        scol_name = self._internal.data_spark_column_names[0]
+        new_scol_name = cast(str, verify_temp_column_name(sdf, "__new_column__"))
```
ueshin, line 2027: Seems like we don't use these?

databricks/koalas/tests/test_series.py
```diff
@@ -2283,6 +2283,74 @@ def test_first_valid_index(self):
...
+            [0, 1, 0, 2] if is_lower_pandas_version else pcodes.tolist(), kcodes.to_list()
+        )
+        self.assert_eq(puniques, kuniques)
+
+        if not is_lower_pandas_version:
+            pcodes, puniques = pser.factorize(sort=True, na_sentinel=None)
+            kcodes, kuniques = kser.factorize(na_sentinel=None)
+            self.assert_eq(pcodes.tolist(), kcodes.to_list())
+            # puniques is Index(['a', 'b', nan], dtype='object')
+            self.assert_eq(ks.Index(["a", "b", None]), kuniques)
```
ueshin, line 2352: I guess we need more cases using `None`, e.g., `ks.Series([1, None, 3, 2, 1])`.

databricks/koalas/series.py
ueshin: I guess we should handle `np.nan` case in `null_scol = F.when(self.spark.column.isNull(), F.lit(na_sentinel_code))` below.

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,136 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+            Value to mark "not found". If None, will not drop the NaN
+            from the uniques of the values.
+
+        Returns
+        -------
+        codes : Series
+            An Series that's an indexer into `uniques`.
+            ``uniques.take(codes)`` will have the same values as `values`.
+        uniques : Index
+            The unique valid values. When `values` is koalas object, an
```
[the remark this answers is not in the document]
xinrong-databricks: Thanks!

databricks/koalas/series.py
xinrong-databricks: Removed, thanks!

databricks/koalas/series.py
xinrong-databricks: Makes sense, I excluded na values from `unique_to_code`.

databricks/koalas/series.py
xinrong-databricks: This statement was for my last approach where `unique_to_code` handles na values.  Let me remove this `if` case.

databricks/koalas/tests/test_series.py
xinrong-databricks: Sounds good! `ks.Series([1, None, 3, 2, 1])` is added. Let me add more.

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,133 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
             )
         )._kser_for(self._column_label)
 
+    def factorize(self, sort: bool = True, na_sentinel: Optional[int] = -1):
```
ueshin, line 1917: Shall we add the return type annotation?

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,133 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+            else:
+                unique_to_code[unique] = code
+            code += 1
+
+        kvs = list(
+            chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
+        )
+
+        map_scol = F.create_map(kvs)
+        null_scol = F.when(self.spark.column.isNull(), F.lit(na_sentinel_code))
```
ueshin, line 2026: We should handle `NaN` when the data type is `FloatType` or `DoubleType`?

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,133 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+
+        if na_sentinel is not None:
+            # Drops the NaN from the uniques of the values
+            non_na_list = [x for x in uniques_list if not pd.isna(x)]
+            if len(non_na_list) == 0:
+                uniques = pd.Index(non_na_list)
+            else:
+                uniques = ks.Index(non_na_list)
+        else:
+            uniques = ks.Index(uniques_list)
```
ueshin, line 2040: I think we can always return `pd.Index` as `uniques` ..? cc @HyukjinKwon

ueshin (commented):
Otherwise, LGTM.

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,133 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+            code += 1
+
+        kvs = list(
+            chain(*([(F.lit(unique), F.lit(code)) for unique, code in unique_to_code.items()]))
+        )
+
+        map_scol = F.create_map(kvs)
+        null_scol = F.when(self.spark.column.isNull(), F.lit(na_sentinel_code))
+        mapped_scol = map_scol.getItem(self.spark.column)
+        internal = self._internal.with_new_columns([null_scol.otherwise(mapped_scol)])
```
ueshin, line 2028: It's safer to use `alias` for the new column.

databricks/koalas/series.py
xinrong-databricks: Sounds good!

databricks/koalas/series.py
xinrong-databricks: Modified it to `pd.Index` for now.

databricks/koalas/series.py
xinrong-databricks: May I ask what value in Koalas dataframe would be represented by `NaN` in the Spark dataframe?

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,136 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+                        max_compute_count
+                    )
+                )
+        else:
+            raise ValueError(
+                "Please set 'compute.max_rows' by using 'databricks.koalas.config.set_option' "
+                "to restrict the total number of unique values of the current Series."
+                "Note that, before changing the 'compute.max_rows', "
+                "this operation is considerably expensive."
+            )
```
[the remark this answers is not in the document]
xinrong-databricks: What does it mean by `collect all the data`?

databricks/koalas/series.py
xinrong-databricks: Is d33d1d3 what you meant?

databricks/koalas/series.py
ueshin: Spark has `F.isnan` to check whether the value is `NaN` or not. e.g., https://github.com/databricks/koalas/blob/d33d1d3ab6cba12dbcbd79f00eae8b93516a2399/databricks/koalas/series.py#L1876

databricks/koalas/series.py
ueshin: ah, yes, but we can reuse `self._internal.data_spark_column_names[0]`?

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,134 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+        na_sentinel : int or None, default -1
+            Value to mark "not found". If None, will not drop the NaN
+            from the uniques of the values.
+
+        Returns
+        -------
+        codes : Series
+            A Series that's an indexer into `uniques`.
+            ``uniques.take(codes)`` will have the same values as `values`.
+        uniques : Index
```
ueshin, line 1937: `pd.Index`?

databricks/koalas/series.py
```diff
@@ -1914,7 +1914,9 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
             )
         )._kser_for(self._column_label)
 
-    def factorize(self, sort: bool = True, na_sentinel: Optional[int] = -1):
+    def factorize(
+        self, sort: bool = True, na_sentinel: Optional[int] = -1
+    ) -> Tuple[pd.Series, "ks.Index"]:
```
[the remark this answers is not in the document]
xinrong-databricks: Sorry, corrected it!

databricks/koalas/series.py
xinrong-databricks: Modified!

databricks/koalas/series.py
xinrong-databricks: Makes sense.
Why do we say using `alias` is safer?

databricks/koalas/series.py
xinrong-databricks: In which case, would `NaN` be in the spark dataframe?
```
>>> kser
0    1.0
1    NaN
2    3.0
3    2.0
4    1.0
dtype: float64

>>> kser._internal.spark_frame.show()
+-----------------+----+-----------------+
|__index_level_0__|   0|__natural_order__|
+-----------------+----+-----------------+
|                0| 1.0|      17179869184|
|                1|null|      34359738368|
|                2| 3.0|      60129542144|
|                3| 2.0|      77309411328|
|                4| 1.0|      94489280512|
+-----------------+----+-----------------+
```

databricks/koalas/series.py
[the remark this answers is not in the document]
xinrong-databricks: Got it! Modified to `toPandas()` without limits for now.

databricks/koalas/series.py
ueshin: e.g.,:

```py
>>> kser = ks.Series(range(5))
>>> kser.loc[3] = np.nan
>>> kser._internal.resolved_copy.spark_frame.show()
+-----------------+---+-----------------+
|__index_level_0__|  0|__natural_order__|
+-----------------+---+-----------------+
|                0|0.0|      17179869184|
|                1|1.0|      34359738368|
|                2|2.0|      60129542144|
|                3|NaN|      77309411328|
|                4|4.0|      94489280512|
+-----------------+---+-----------------+
```

databricks/koalas/series.py
ueshin: What happens if we don't use alias internally?
The Spark column name would be something like `CASE WHEN (...) THEN ... ELSE ... END`.

```py
>>> spark.range(2).select(F.when(F.col('id').isNull(), F.lit(-1)).otherwise(F.create_map(F.lit(1), F.lit(0)).getItem(F.col('id')))).show()
+-----------------------------------------------------+
|CASE WHEN (id IS NULL) THEN -1 ELSE map(1, 0)[id] END|
+-----------------------------------------------------+
|                                                 null|
|                                                    0|
+-----------------------------------------------------+
```

databricks/koalas/series.py
ueshin: Also please don't forget that Koalas objects can be converted from existing Spark DataFrames which can include `NaN`s.

xinrong-databricks: For the `kser.loc[3] = np.nan` example, is `NaN` in the spark_frame as a result of `resolved_copy`?

databricks/koalas/series.py
xinrong-databricks: That makes sense to me. Thank you!

databricks/koalas/series.py
```diff
@@ -1913,6 +1914,139 @@ def _fillna(self, value=None, method=None, axis=None, limit=None, part_cols=()):
...
+        else:
+            cond = scol.isNull()
+        null_scol = F.when(cond, F.lit(na_sentinel_code))
+
+        map_scol = None if len(kvs) <= 0 else F.create_map(kvs)
+
+        if map_scol is not None:
+            new_scol = null_scol.otherwise(map_scol.getItem(scol))
+        else:
+            new_scol = null_scol
```
ueshin, line 2034: Seems like we can branch by the condition `len(kvs) <= 0`?

```py
if len(kvs) == 0:
    new_scol = F.lit(na_sentinel_code)
else:
    scol = self.spark.column
    if isinstance(self.spark.data_type, (FloatType, DoubleType)):
        ...
    null_scol = ...
    map_scol = ...
    new_scol = null_scol.otherwise(map_scol.getItem(scol))
```

databricks/koalas/series.py
ueshin: `kser.loc[3]` doesn't update the internal Spark DataFrame but only update the `data_spark_columns`. We need to use `resolved_copy` to see the actual change.

databricks/koalas/series.py
xinrong-databricks: Good idea, that makes code clearer!

ueshin (approved):
LGTM, pending tests.

ueshin:
@xinrong-databricks Could you try the following as well?

```py
>>> kser = ks.Series([1, 2, np.nan, 4, 5])
>>> kser.loc[3] = np.nan
>>> kser.factorize(na_sentinel=None)
(0    0
1    1
2    4
3    4
4    2
dtype: int32, Float64Index([1.0, 2.0, 5.0, nan, nan], dtype='float64'))

>>> kser.to_pandas().factorize(na_sentinel=None)
(array([0, 1, 3, 3, 2]), Float64Index([1.0, 2.0, 5.0, nan], dtype='float64'))
```

databricks/koalas/series.py
xinrong-databricks: Got it, thanks!

xinrong-databricks:
@ueshin Good catch! Let me look into this.

databricks/koalas/tests/test_series.py
```diff
@@ -2364,13 +2364,16 @@ def test_factorize(self):
 
         pcodes, puniques = pser.factorize(sort=True, na_sentinel=-2)
         kcodes, kuniques = kser.factorize(na_sentinel=-2)
-
-        self.assert_eq(pcodes.tolist(), kcodes.to_list())
+        self.assert_eq(
+            [0, 1, 0, -2, -2] if is_lower_pandas_version else pcodes.tolist(), kcodes.to_list()
```
ueshin, line 2369: This behavior difference only happens with pandas<0.24, right?
We should check with the actual pandas if the behavior is the same.

xinrong-databricks: Sounds good! Modified.

ueshin:
Thanks! merging.

xinrong-databricks:
Thank you for reviewing and merging the PR! @ueshin  :)

PR #1980, opened 2020-12-22
Fix Frame.abs to support bool type and disallow non-numeric types.

ueshin:
Fixes `Frame.abs` to support bool type and disallow non-numeric types.

### DIFF
--- a/databricks/koalas/generic.py
+++ b/databricks/koalas/generic.py
@@ -1757,7 +1757,18 @@ class Frame(object, metaclass=ABCMeta):
         2  6  30   30
         3  7  40   50
         """
-        return self._apply_series_op(lambda kser: kser._with_new_scol(F.abs(kser.spark.column)))
+
+        def abs(kser):
+            if isinstance(kser.spark.data_type, BooleanType):
+                return kser
+            elif isinstance(kser.spark.data_type, NumericType):
+                return kser.spark.transform(F.abs)
+            else:
+                raise TypeError(
+                    "bad operand type for abs(): {}".format(kser.spark.data_type.simpleString())
+                )
+
+        return self._apply_series_op(abs)
 
     # TODO: by argument only support the grouping name and as_index only for now. Documentation
     # should be updated when it's supported.
--- a/databricks/koalas/tests/test_stats.py
+++ b/databricks/koalas/tests/test_stats.py
@@ -98,12 +98,22 @@ class StatsTest(ReusedSQLTestCase, SQLTestUtils):
                 "B": [1.0, -2, 3, -4, 5],
                 "C": [-6.0, -7, -8, -9, 10],
                 "D": ["a", "b", "c", "d", "e"],
+                "E": [True, False, False, True, True],
             }
         )
         kdf = ks.from_pandas(pdf)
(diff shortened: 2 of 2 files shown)


databricks/koalas/tests/test_stats.py
```diff
@@ -98,12 +98,22 @@ def test_abs(self):
...
                 "D": ["a", "b", "c", "d", "e"],
+                "E": [True, False, False, True, True],
             }
         )
         kdf = ks.from_pandas(pdf)
         self.assert_eq(kdf.A.abs(), pdf.A.abs())
         self.assert_eq(kdf.B.abs(), pdf.B.abs())
+        self.assert_eq(kdf.E.abs(), pdf.E.abs())
+        # pandas' bug?
+        # self.assert_eq(kdf[["B", "C", "E"]].abs(), pdf[["B", "C", "E"]].abs())
```
ueshin, line 109: The pandas result is:

```py
>>> pdf[["B", "C", "E"]].abs()
   B   C  E
0  1   6  1
1  2   7  0
2  3   8  0
3  4   9  1
4  5  10  1
```

whereas calculating separately:

```py
>>> pdf[["B", "C"]].abs()
     B     C
0  1.0   6.0
1  2.0   7.0
2  3.0   8.0
3  4.0   9.0
4  5.0  10.0
>>> pdf[["E"]].abs()
       E
0   True
1  False
2  False
3   True
4   True
```

I believe this is a pandas bug.

The current Koalas behavior:

```py
>>> ks.from_pandas(pdf)[["B", "C", "E"]].abs()
     B     C      E
0  1.0   6.0   True
1  2.0   7.0  False
2  3.0   8.0  False
3  4.0   9.0   True
4  5.0  10.0   True
```

databricks/koalas/generic.py
```diff
@@ -1757,7 +1757,18 @@ def abs(self) -> Union["DataFrame", "Series"]:
...
-        return self._apply_series_op(lambda kser: kser._with_new_scol(F.abs(kser.spark.column)))
+
+        def abs(kser):
+            if isinstance(kser.spark.data_type, BooleanType):
+                return kser
+            elif isinstance(kser.spark.data_type, NumericType):
+                return kser.spark.transform(F.abs)
+            else:
+                raise TypeError(
+                    "bad operand type for abs(): {}".format(kser.spark.data_type.simpleString())
```
HyukjinKwon, line 1768: Hm, should we show spark's dtype instead of Spark data type?

HyukjinKwon (approved):
LGTM otherwise

databricks/koalas/generic.py
ueshin: In that case, we should address it separately. The same error message will appear in the other places.

databricks/koalas/tests/test_stats.py
itholic: Seems pandas bug to me, too. Let me report this to pandas dev.

itholic (approved):
LGTM. Let's address https://github.com/databricks/koalas/pull/1980#discussion_r547607957 in separately.

ueshin:
Thanks! merging.

PR #1985, opened 2020-12-23
Use Python type name instead of Spark's in error messages.

ueshin:
Addressing https://github.com/databricks/koalas/pull/1980#discussion_r547607957 to add pandas dtypes.

### DIFF
--- a/databricks/koalas/typedef/typehints.py
+++ b/databricks/koalas/typedef/typehints.py
@@ -103,7 +103,7 @@ def as_spark_type(tpe) -> types.DataType:
     """
     # TODO: Add "boolean" and "string" types.
     # ArrayType
-    if tpe in (np.ndarray,):
+    if tpe in (list, np.ndarray,):
         return types.ArrayType(types.StringType())
     elif hasattr(tpe, "__origin__") and issubclass(tpe.__origin__, list):
         return types.ArrayType(as_spark_type(tpe.__args__[0]))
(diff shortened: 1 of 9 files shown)


databricks/koalas/typedef/typehints.py
```diff
@@ -103,7 +103,7 @@ def as_spark_type(tpe) -> types.DataType:
     """
     # TODO: Add "boolean" and "string" types.
     # ArrayType
-    if tpe in (np.ndarray,):
+    if tpe in (list, np.ndarray,):
```
ueshin, line 106: Is this reasonable?

HyukjinKwon: Yeah, I think it's fine.

databricks/koalas/typedef/typehints.py
```diff
@@ -142,7 +142,46 @@ def as_spark_type(tpe) -> types.DataType:
         raise TypeError("Type %s was not understood." % tpe)
 
 
-def spark_type_to_pandas_dtype(spark_type):
+def spark_type_to_python_type(spark_type: types.DataType) -> type:
+    """ Return the given Spark DataType to Python type. """
+    # ArrayType
+    if isinstance(spark_type, types.ArrayType):
+        return list
```
HyukjinKwon, line 149: Maybe `numpy.ndarray` instead of `list`.. ?

databricks/koalas/typedef/typehints.py
```diff
@@ -142,7 +142,46 @@ def as_spark_type(tpe) -> types.DataType:
...
+    elif isinstance(spark_type, types.StringType):
+        return str
+    # TimestampType
+    elif isinstance(spark_type, types.TimestampType):
+        return datetime.datetime
+    else:
+        raise TypeError("Type %s was not understood." % spark_type.simpleString())
+
+
+def spark_type_to_pandas_dtype(spark_type: types.DataType) -> np.dtype:
```
HyukjinKwon, line 184: @ueshin, btw what about we just use `spark_type_to_pandas_dtype` in the error message?

ueshin: We can use it, but in that case, it will show only `object` for most cases.

HyukjinKwon: Hm .. what about we show both pandas dtype and spark type? for example ...

```
Could not convert pandas type (spark type) to numeric?
```

ueshin: Sure.

HyukjinKwon (approved):

PR #1987, opened 2020-12-28
Upgrade pandas to 1.2 in CI

itholic:
Since pandas 1.2 was released, upgraded pandas version in CI to follow pandas' behavior.

https://pandas.pydata.org/pandas-docs/version/1.2.0/whatsnew/v1.2.0.html
https://pandas.pydata.org/pandas-docs/version/1.2.1/whatsnew/v1.2.1.html


databricks/koalas/tests/test_indexing.py
```diff
@@ -1196,10 +1196,12 @@ def test_series_iloc_setitem(self):
...
-        pser.iloc[[1]] = -pdf.b
-        kser.iloc[[1]] = -kdf.b
-        self.assert_eq(kser, pser)
-        self.assert_eq(kdf, pdf)
+        # TODO: matching the behavior with pandas 1.2 and uncomment below test.
+        # with self.assertRaisesRegex(
+        #     ValueError,
+        #     "cannot set using a list-like indexer with a different length than the value",
+        # ):
+        #     kser.iloc[[1]] = -kdf.b
```
itholic, line 1204: Let me resolve this in the separated PR because it seems quite complicated to fix, so I think we better to concentrate this one separately.

itholic:
Ready to get review. Let me handle the `iLocIndexer.__setter__` part in the separated PR as I commented at https://github.com/databricks/koalas/pull/1987#discussion_r553752763

databricks/koalas/frame.py
```diff
@@ -2049,10 +2050,16 @@ def to_latex(
...
+        >>> print(df.to_latex(index=False)) # doctest: +NORMALIZE_WHITESPACE
+        \begin{tabular}{lll}
+        \toprule
+              name &    mask &    weapon \\
+        \midrule
+           Raphael &     red &       sai \\
+         Donatello &  purple &  bo staff \\
+        \bottomrule
+        \end{tabular}
+        <BLANKLINE>
```
itholic, line 2062: Just print statement was changed for `to_latex()` in pandas 1.2

databricks/koalas/indexes.py
```diff
@@ -2220,20 +2220,20 @@ def intersection(self, other) -> "Index":
         if isinstance(other, DataFrame):
             raise ValueError("Index data must be 1-dimensional")
         elif isinstance(other, MultiIndex):
-            # Always returns an empty MultiIndex if `other` is MultiIndex.
-            return other.to_frame().head(0).index
+            # Always returns a no-named empty Index if `other` is MultiIndex.
+            return self._kdf.head(0).index.rename(None)
```
itholic, line 2224: The behavior of `Index.intersection` was changed in pandas 1.2 when the `other` is `MultiIndex`.

It returned empty `MultiIndex` before, but it returns empty `Index` now.

databricks/koalas/indexes.py
```diff
@@ -2220,20 +2220,20 @@ def intersection(self, other) -> "Index":
...
         elif isinstance(other, Index):
             spark_frame_other = other.to_frame().to_spark()
             keep_name = self.name == other.name
         elif isinstance(other, Series):
             spark_frame_other = other.to_frame().to_spark()
-            keep_name = self.name == other.name
+            keep_name = True
```
itholic, line 2230: Now `Index.intersection` always keeps the name of `Index` when `other` is `Series` from pandas 1.2.

databricks/koalas/indexes.py
```diff
@@ -2220,20 +2220,20 @@ def intersection(self, other) -> "Index":
...
         elif is_list_like(other):
             other = Index(other)
             if isinstance(other, MultiIndex):
                 return other.to_frame().head(0).index
             spark_frame_other = other.to_frame().to_spark()
-            keep_name = False
+            keep_name = True
```
itholic, line 2236: Now `Index.intersection` always keeps the name of `Index` when `other` is `list_like` from pandas 1.2.

databricks/koalas/frame.py
```diff
@@ -673,6 +673,7 @@ def _reduce_for_stat_function(self, sfun, name, axis=None, numeric_only=True, **
         else:
             # Here we execute with the first 1000 to get the return type.
             # If the records were less than 1000, it uses pandas API directly for a shortcut.
+            numeric_only = None if name in ["sum", "product", "prod"] else numeric_only
```
itholic, line 676: The behavior when `numeric_only=None` is changed for some cases.

Please refer to https://pandas.pydata.org/pandas-docs/dev/whatsnew/v1.2.0.html#consistency-of-dataframe-reductions

HyukjinKwon: Can you elabourate a bit more? which functions were changed and how?

databricks/koalas/tests/test_indexing.py
```diff
@@ -1196,11 +1196,18 @@ def test_series_iloc_setitem(self):
         pser = pdf.a
         kser = kdf.a
 
-        pser.iloc[[1]] = -pdf.b
-        kser.iloc[[1]] = -kdf.b
+        pser.iloc[[0, 1, 2]] = -pdf.b
+        kser.iloc[[0, 1, 2]] = -kdf.b
```
HyukjinKwon, line 1200: what are these diff?

databricks/koalas/tests/test_stats.py
```diff
@@ -80,7 +80,41 @@ def test_stat_functions_with_no_numeric_columns(self):
             }
         )
         kdf = ks.from_pandas(pdf)
-        self._test_stat_functions(pdf, kdf)
```
HyukjinKwon, line 83: Can you add a flag that that excludes `std` in the test instead of copying this method?

databricks/koalas/tests/test_indexing.py
itholic: The length of parameter to `iloc` should be the same with the length of the assigning values from pandas 1.1.2.

In short, `len([0, 1, 2])` and `len(-pdf.b)` should be the same.

databricks/koalas/tests/test_stats.py
itholic: But `std` is still valid for the `pandas < 1.2.0`.

Don't we need to care about that case ??

databricks/koalas/frame.py
ueshin: Btw, let's avoid a logic depending on a specific function in a common function. We should pick a proper value in each function before passing it to the common function.

itholic: From pandas 1.1.2, the `dtype` of `sum()`, `product()` is `<class 'numpy.object_'>` which cannot inferred in Koalas when `numeric_only=True`.

```python
>>> pdf.sum(axis=1, numeric_only=True).dtype.type
<class 'numpy.object_'>  # object type cannot be inferred in Koalas, raise Exception in `as_spark_type`
>>> pdf.sum(axis=1, numeric_only=None).dtype.type
<class 'numpy.float64'>  # Should use `None` rather than `True` for `sum`, `prod`
```

itholic: @ueshin 

Sure, let me address it

databricks/koalas/tests/test_stats.py
ueshin: Seems like this modification is not following the purpose of this test which is to check the behavior for non-numeric columns.

Maybe we should use string type data instead of date time here, and add a TODO to support `std` for the date time type. https://github.com/pandas-dev/pandas/pull/37436

databricks/koalas/frame.py
ueshin: > ```py
> >>> pdf.sum(axis=1, numeric_only=True).dtype.type
> <class 'numpy.object_'>  # object type cannot be inferred in Koalas, raise Exception in `as_spark_type`
> ```

What's the `pdf` like?

I tried in my local:

```py
>>> pd.__version__
'1.2.0'

>>> pdf
   a  b  c
0  1  4  a
1  2  5  b
2  3  6  c
>>> pdf.sum(axis=1, numeric_only=True)
0    5
1    7
2    9
dtype: int64
>>> pdf.sum(axis=1, numeric_only=True).dtype.type
<class 'numpy.int64'>

>>> pdf
     a    b  c
0  1.0  4.0  a
1  2.0  5.0  b
2  3.0  6.0  c
>>> pdf.sum(axis=1, numeric_only=True)
0    5.0
1    7.0
2    9.0
dtype: float64
>>> pdf.sum(axis=1, numeric_only=True).dtype.type
<class 'numpy.float64'>
```

databricks/koalas/tests/test_indexes.py
```diff
@@ -1583,46 +1583,76 @@ def test_intersection(self):
...
-        )
-        self.assert_eq(
-            (pidx + 1).intersection(pmidx),
-            (kidx + 1).intersection(kmidx).sort_values(),
-            almost=True,
-        )
+        if LooseVersion(pd.__version__) < LooseVersion("1.2.0"):
+            self.assert_eq(
+                kidx.intersection(kmidx).sort_values(),
+                kidx._kdf.head(0).index.rename(None),
```
ueshin, line 1589: The difference must be only its name, right?
Then we should use `pidx` and `pmidx` for the expected values? e.g.,

```py
pidx.intersection(pmidx).rename(None)
```

databricks/koalas/tests/test_indexes.py
```diff
@@ -1583,46 +1583,76 @@ def test_intersection(self):
...
+                almost=True,
+            )
+            self.assert_eq(
+                (kidx + 1).intersection(kmidx).sort_values(),
+                kidx._kdf.head(0).index.rename(None),
```
ueshin, line 1594: ditto, and the following tests.

databricks/koalas/frame.py
itholic: Sorry for the late reply.

`pdf` looks like as below.

```python
>> pd.__version__
1.2.0

>>> pdf
   A    B     C      D
0  1  1.0  -6.0   True
1 -2 -2.0  -7.0  False
2  3  3.0  -8.0   True
3 -4 -4.0  -9.0  False
4  5  5.0  10.0  False

>>> pdf.sum(axis=1, numeric_only=True)
0    -3.0
1   -11.0
2    -1.0
3   -17.0
4    20.0
dtype: object

>>> pdf.sum(axis=1, numeric_only=True).dtype.type
<class 'numpy.object_'>
```

databricks/koalas/tests/test_indexes.py
itholic: Sure, thanks! Looks more clearer.

itholic: @ueshin Oh, no this is because the expected result is `Int64Index`, but the actual result of pandas < 1.2.0 is `MultiIndex`.

databricks/koalas/tests/test_indexes.py
itholic: They have all same reason with above

databricks/koalas/tests/test_stats.py
itholic: Addressed it. Thanks!

databricks/koalas/frame.py
ueshin: ah, seems like if there is a bool type column, the return type will be `object`.
I'd suspect it's a bug..

databricks/koalas/generic.py
```diff
@@ -1207,6 +1207,8 @@ def sum(
         >>> df['b'].sum(min_count=3)
         nan
         """
+        axis = validate_axis(axis)
+        numeric_only = None if axis == 1 else numeric_only
```
ueshin, line 1211: `numeric_only = None if axis == 1 and numeric_only is True else numeric_only`?

databricks/koalas/generic.py
```diff
@@ -1285,6 +1287,8 @@ def product(
         >>> ks.Series([]).prod(min_count=1)
         nan
         """
+        axis = validate_axis(axis)
+        numeric_only = None if axis == 1 else numeric_only
```
ueshin, line 1291: ditto.

databricks/koalas/generic.py
```diff
@@ -1467,12 +1471,15 @@ def min(
         >>> df['a'].min()
         1.0
         """
+        axis = validate_axis(axis)
+        numeric_only = None if axis == 1 else numeric_only
```
ueshin, line 1475: ditto.

databricks/koalas/generic.py
```diff
@@ -1424,7 +1428,7 @@ def kurtosis(spark_column, spark_type):
     kurt = kurtosis
 
     def min(
-        self, axis: Union[int, str] = None, numeric_only: bool = None
+        self, axis: Union[int, str] = None, numeric_only: bool = True
```
ueshin, line 1431: This should remain as `None`, right?

databricks/koalas/generic.py
```diff
@@ -1467,12 +1471,15 @@ def min(
...
+        axis = validate_axis(axis)
+        numeric_only = None if axis == 1 else numeric_only
+
         return self._reduce_for_stat_function(
             F.min, name="min", axis=axis, numeric_only=numeric_only
         )
 
     def max(
-        self, axis: Union[int, str] = None, numeric_only: bool = None
+        self, axis: Union[int, str] = None, numeric_only: bool = True
```
ueshin, line 1482: ditto.

ueshin (commented):
Could you also remove the upper bound for pandas?

databricks/koalas/generic.py
```diff
@@ -1236,7 +1238,7 @@ def product(
         ----------
         axis : {index (0), columns (1)}
             Axis for the function to be applied on.
-        numeric_only : bool, default True
+        numeric_only : bool, default None
```
ueshin, line 1241: We should also use `None` for the default value?

```py
    def product(
        self, axis: Union[int, str] = None, numeric_only: bool = None, min_count: int = 0
    ) -> Union[Scalar, "Series"]:
```

ueshin: We should modify for `sum` as well?

databricks/koalas/tests/test_indexes.py
ueshin: Got it.

databricks/koalas/generic.py
itholic: Right. Thanks for the correction

itholic:

databricks/koalas/tests/test_stats.py
```diff
@@ -73,13 +73,24 @@ def test_stat_functions_multiindex_column(self):
...
     def test_stat_functions_with_no_numeric_columns(self):
+        # TODO: pandas support datetime64 or datetime64tz dtypes for `std` from pandas 1.2
+        # to return Timedelta Series which is Koalas currently cannot support.
+        # pdf = pd.DataFrame(
+        #     {
+        #         "A": pd.date_range("2020-01-01", periods=3),
+        #         "B": pd.date_range("2021-01-01", periods=3),
+        #     }
+        # )
+        # kdf = ks.from_pandas(pdf)
```
ueshin, line 84: Shall we file an issue for that instead of comment out here?

itholic: Sounds good. Let me file up to the issue. Thanks :)

databricks/koalas/generic.py
itholic: Maybe I think we can use `True` for pandas compatibility?? I just matched it to pandas'. 

for `sum`, `prod`, `mix`, `max`, they uses `True` as a default value for `numeric_only`.

In addition, we can apply same logic for `numeric_only` for all methods - `numeric_only = None if axis == 1 and numeric_only is True else numeric_only` - if we use same default value.

WDYT?

ueshin: pandas uses `None` as a default value, right?
e.g., https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.sum.html

.github/workflows/master.yml
```diff
@@ -107,7 +107,7 @@ jobs:
             pyarrow-version: 1.0.1
           - python-version: 3.8
             spark-version: 3.0.1
-            pandas-version: 1.1.5
+            pandas-version: 1.2.0
```
ueshin, line 110: Shall we maybe use `1.2.1` btw?

ueshin: @itholic up to you, we can do it separately.

itholic: Sure, I'd upgrade this here.

databricks/koalas/generic.py
itholic: Oops, sorry for annoying. It was my misread. I'll address it  😅

itholic: Addressed it. Thanks, @ueshin

ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

gsheni:
@itholic @ueshin Should the upper limit of pandas 1.2.0 have been removed here?
- https://github.com/databricks/koalas/blob/master/setup.py#L59
It was mentioned in the above comment...

itholic:
I thought we should remove upper bound after resolving https://github.com/databricks/koalas/pull/1987/files#r553752763. Let me take a look at this.
Thanks for the notice, @gsheni :)

PR #1989, opened 2020-12-28
Support to set attributes for DataFrame.

ueshin:
Support to set attributes for DataFrame.

```py
>>> kdf = ks.DataFrame({'A': [1, 2, 3, None]})
>>> kdf.A = kdf.A.fillna(kdf.A.median())
>>> kdf
     A
0  1.0
1  2.0
2  3.0
3  2.0
```

PR #1991, opened 2020-12-30
Use object.__setattr__ in Series.

ueshin:
This is a follow-up of #1989.
There are some more places where setting attributes with the overwritten `DataFrame.__setattr__`.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -86,6 +86,7 @@ from databricks.koalas.utils import (
     default_session,
     is_name_like_tuple,
     is_name_like_value,
+    is_testing,
     name_like_string,
     same_anchor,
     scol_for,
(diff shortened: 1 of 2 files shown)


itholic (approved):
LGTM

ueshin:
Thanks! merging.

databricks/koalas/frame.py
```diff
@@ -11097,9 +11098,11 @@ def __setattr__(self, key: str, value) -> None:
         if (key,) in self._internal.column_labels:
             self[key] = value
         else:
-            warnings.warn(
-                "Koalas doesn't allow columns to be created via a new attribute name", UserWarning
-            )
+            msg = "Koalas doesn't allow columns to be created via a new attribute name"
```
HyukjinKwon, line 11101: Quick question. What about we just throw an exception as is for what we don't allow?

PR #1996, opened 2021-01-08
Support operations between Series and Index.

ueshin:
Supports operations between `Series` and `Index`.

```py
>>> kser = ks.Series([1, 2, 3, 4, 5, 6, 7])
>>> kidx = ks.Index([0, 1, 2, 3, 4, 5, 6])

>>> (kser + 1 + 10 * kidx).sort_index()
0     2
1    13
2    24
3    35
4    46
5    57
6    68
dtype: int64
>>> (kidx + 1 + 10 * kser).sort_index()
0    11
1    22
2    33
3    44
4    55
5    66
6    77
dtype: int64
```

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -41,20 +41,126 @@ from pyspark.sql.types import (
 
 from databricks import koalas as ks  # For running doctests and reference resolution in PyCharm.
 from databricks.koalas import numpy_compat
+from databricks.koalas.config import get_option, option_context
 from databricks.koalas.internal import (
     InternalFrame,
+    DEFAULT_SERIES_NAME,
     NATURAL_ORDER_COLUMN_NAME,
     SPARK_DEFAULT_INDEX_NAME,
 )
 from databricks.koalas.spark import functions as SF
 from databricks.koalas.spark.accessors import SparkIndexOpsMethods
 from databricks.koalas.typedef import as_spark_type, spark_type_to_pandas_dtype
-from databricks.koalas.utils import scol_for, validate_axis
+from databricks.koalas.utils import (
+    combine_frames,
+    same_anchor,
+    scol_for,
+    validate_axis,
+    ERROR_MESSAGE_CANNOT_COMBINE,
+)
 from databricks.koalas.frame import DataFrame
 
 if TYPE_CHECKING:
     from databricks.koalas.indexes import Index
     from databricks.koalas.series import Series
+
+
+def need_alignment_for_column_op(self: "IndexOpsMixin", other: "IndexOpsMixin") -> bool:
+    from databricks.koalas.series import Series
+
+    if isinstance(self, Series) and isinstance(other, Series):
+        return not same_anchor(self, other)
+    else:
+        return self._internal.spark_frame is not other._internal.spark_frame
+
+
+def align_diff_series(func, this_series: "Series", *args) -> "Series":
+    from databricks.koalas.series import Series
+
+    cols = [arg for arg in args if isinstance(arg, Series)]
+    combined = combine_frames(this_series.to_frame(), *cols, how="full")
+
+    return column_op(func)(
+        combined["this"]._kser_for(combined["this"]._internal.column_labels[0]),
+        *[combined["that"]._kser_for(label) for label in combined["that"]._internal.column_labels]
+    )
+
+
+def align_diff_index(func, this_index: "Index", *args) -> "Index":
+    from databricks.koalas.indexes import Index
+    from databricks.koalas.series import Series
+
+    with option_context("compute.default_index_type", "distributed-sequence"):
+        # Directly using Series from both self and other seems causing
+        # some exceptions when 'compute.ops_on_diff_frames' is enabled.
+        # Working around for now via using frame.
+        return (
+            cast(
+                Series,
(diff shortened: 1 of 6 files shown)


databricks/koalas/base.py
```diff
@@ -41,22 +41,128 @@
...
+    ERROR_MESSAGE_CANNOT_COMBINE,
+)
 from databricks.koalas.frame import DataFrame
 
 if TYPE_CHECKING:
     from databricks.koalas.indexes import Index
     from databricks.koalas.series import Series
 
 
+def need_alignment_for_column_op(self: "IndexOpsMixin", other: "IndexOpsMixin") -> bool:
```
HyukjinKwon, line 68: nit: maybe `need_alignment_for_column_op` -> `should_alignment_for_column_op`

databricks/koalas/base.py
```diff
@@ -41,22 +41,128 @@
...
+                Series,
+                column_op(func)(
+                    this_index.to_series().reset_index(drop=True),
+                    *[
+                        arg.to_series().reset_index(drop=True) if isinstance(arg, Index) else arg
+                        for arg in args
+                    ]
+                ),
+            )
+            .sort_index()
```
HyukjinKwon, line 108: Hm, why should we sort by index here?

databricks/koalas/base.py
```diff
@@ -41,22 +41,128 @@
...
+                first_series(combined["this"]),
+                *[
+                    combined["that"]._kser_for(label)
+                    for label in combined["that"]._internal.column_labels
+                ]
+            )
+        else:
+            this = cast(Index, this_index_ops).to_frame().reset_index(drop=True)
+
+            that_series = next(col for col in cols if isinstance(col, Series))
```
HyukjinKwon, line 147: no big deal but can we add an assert on `cols` that it always has `Series`?

databricks/koalas/base.py
```diff
@@ -41,22 +41,128 @@
...
+
+
+def align_diff_index_ops(func, this_index_ops: "IndexOpsMixin", *args) -> "Series":
+    from databricks.koalas.indexes import Index
+    from databricks.koalas.series import Series, first_series
+
+    cols = [arg for arg in args if isinstance(arg, IndexOpsMixin)]
+
+    with option_context("compute.default_index_type", "distributed-sequence"):
+        if isinstance(this_index_ops, Series):
```
HyukjinKwon, line 122: @ueshin, can we briefly explain what `align_diff_series`, `align_diff_index` and `align_diff_index_ops` do in terms of input/output? From reading the names, looks like we can just use `align_diff_index_ops` instead of `align_diff_series` and `align_diff_index`.

HyukjinKwon:
Looks fine otherwise

databricks/koalas/base.py
```diff
@@ -94,27 +200,43 @@ def column_op(f):
...
+                if isinstance(self, Index) and all(isinstance(col, Index) for col in cols):
+                    index_ops = align_diff_index(f, self, *args)
+                else:
+                    index_ops = align_diff_index_ops(f, self, *args)
         else:
-            index_ops = self._align_and_column_op(f, *args)
+            raise ValueError(ERROR_MESSAGE_CANNOT_COMBINE)
 
         if not all(self.name == col.name for col in cols):
             index_ops = index_ops.rename(None)
```
xinrong-databricks, line 242: I am wondering whether `index_ops`'s name should always be `None`.

xinrong-databricks: And I may have some concern about what `index_ops` means. It seems to be the result we return finally.

itholic: @xinrong-databricks When the name is same, seems they keep it in pandas.

```python
>>> pd.Series([1, 2, 3], name="Koalas") + pd.Index([2, 3, 4], name="Koalas")
0    3
1    5
2    7
Name: Koalas, dtype: int64
```

databricks/koalas/base.py
ueshin: Otherwise, as the result will lost the order, we can't know which row was which. Then the following operations will be non-deterministic and not make sense.

```py
>>> spark = ks.utils.default_session({'spark.master': 'local-cluster[3,2,4096]'})
>>> ks.Index([1,2,3]*100) + ks.Index([4,5,6]*100) + ks.Index([7,8,9]*100)
Int64Index([16, 16, 18, 18, 12, 16, 18, 12, 16, 17,
            ...
            14, 12, 14, 18, 14, 15, 15, 12, 14, 17],
           dtype='int64', length=300)
```

This should be:

```py
>>> ks.Index([1,2,3]*100) + ks.Index([4,5,6]*100) + ks.Index([7,8,9]*100)
Int64Index([12, 15, 18, 12, 15, 18, 12, 15, 18, 12,
            ...
            18, 12, 15, 18, 12, 15, 18, 12, 15, 18],
           dtype='int64', length=300)
```

HyukjinKwon: 👌

HyukjinKwon (approved):

ueshin:
Thanks! merging.

databricks/koalas/base.py
xinrong-databricks: Got it, thanks!

PR #1997, opened 2021-01-08
Optimize histogram calculation as a single pass

HyukjinKwon:
This PR optimizes histogram plot in Koalas by unioning the transformed results and making it single pass.

Previously, when the `DataFrame.plot.hist` is called, each column had to trigger each job.
Now, we can do it in single pass even for `DataFrame`s.

I also tested that it still result same:

Screen Shot 2021-01-08 at 1 15 47 PM

Screen Shot 2021-01-08 at 1 16 19 PM

### DIFF
--- a/databricks/koalas/plot/matplotlib.py
+++ b/databricks/koalas/plot/matplotlib.py
@@ -24,7 +24,6 @@ from pandas.core.dtypes.inference import is_integer, is_list_like
 from pandas.io.formats.printing import pprint_thing
 
 from databricks.koalas.plot import TopNPlot, SampledPlot
-import pyspark
 from pyspark.ml.feature import Bucketizer
 from pyspark.mllib.stat import KernelDensity
 from pyspark.sql import functions as F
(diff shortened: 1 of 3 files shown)


databricks/koalas/plot/matplotlib.py
```diff
@@ -536,7 +529,25 @@ def _get_bins(sdf, bins):
         return np.linspace(boundaries[0], boundaries[1], bins + 1)
 
     @staticmethod
-    def _compute_hist(sdf, bins):
+    def _compute_hist(kdf, bins):
+        if LooseVersion(pyspark.__version__) < LooseVersion("3.0.0"):
+            sdf = kdf._internal.spark_frame
+            for i, label in enumerate(kdf._internal.column_labels):
```
ueshin, line 535: nit: `i` is not used?

databricks/koalas/plot/matplotlib.py
```diff
@@ -536,7 +529,25 @@ def _get_bins(sdf, bins):
...
+                # 'y' is a Spark DataFrame that selects one column.
+                yield KoalasHistPlot._compute_hist_single_column(
+                    sdf.select(kdf._internal.spark_column_for(label)), bins
+                )
+        else:
+            sdf = kdf._internal.spark_frame
+            scols = []
+            for i, label in enumerate(kdf._internal.column_labels):
```
ueshin, line 543: ditto.

databricks/koalas/plot/matplotlib.py
```diff
@@ -566,6 +577,142 @@ def _compute_hist(sdf, bins):
...
+            if output_df is None:
+                output_df = bucket_df.select(
+                    F.lit(group_id).alias("__group_id"), F.col(bucket_name).alias("__bucket")
+                )
+            else:
+                output_df = output_df.union(
+                    bucket_df.select(
+                        F.lit(group_id).alias("__group_id"), F.col(bucket_name).alias("__bucket")
+                    )
+                )
```
ueshin, line 644: If we try to flatten the output like this, we can just union single-column transformations for each column? Then we can apply it to older Spark?

HyukjinKwon: Oh hm, yeah I think we can. Let me try to combine both.

databricks/koalas/plot/matplotlib.py
```diff
@@ -536,35 +529,144 @@ def _get_bins(sdf, bins):
...
-        bucket_name = "__{}_bucket".format(colname)
-        # creates a Bucketizer to get corresponding bin of each value
-        bucketizer = Bucketizer(
-            splits=bins, inputCol=colname, outputCol=bucket_name, handleInvalid="skip"
-        )
-        # after bucketing values, groups and counts them
+        colnames = sdf.columns
+        bucket_names = ["__{}_bucket".format(colname) for colname in colnames]
+
+        if LooseVersion(pyspark.__version__) >= LooseVersion("3.0.0"):
```
ueshin, line 576: I'm wondering whether we need this case? I guess we can just use the `else` case.

ueshin (approved):
LGTM.

databricks/koalas/plot/matplotlib.py
HyukjinKwon: Hm .. true. I don't think there's a notable benefit by doing this.

itholic:
Cool. Thanks for working on this!

xinrong-databricks:
Great!

PR #1999, opened 2021-01-08
Implement (DataFrame|Series).plot.hist in plotly

HyukjinKwon:
This PR implements `(DataFrame|Series).plot.hist` in plotly:

This can be tested via: https://mybinder.org/v2/gh/HyukjinKwon/koalas/plotly-histogram?filepath=docs%2Fsource%2Fgetting_started%2F10min.ipynb

Example:

```python
# Koalas
ks.options.plotting.backend = "plotly"
kdf = ks.DataFrame({
    'a c': [1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 50],
    'b': [2, 3, 4, 5, 7, 9, 10, 15, 10, 20, 20]
})
(kdf + 100).plot.hist()

# pandas
import pandas as pd
pd.options.plotting.backend = "plotly"
pdf = pd.DataFrame({
    'a c': [1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 50],
    'b': [2, 3, 4, 5, 7, 9, 10, 15, 10, 20, 20]
})
(pdf + 100).plot.hist()
```

Screen Shot 2021-01-12 at 10 12 47 PM

Screen Shot 2021-01-12 at 10 12 52 PM

 NOTE that the output is a bit different because:
 - We use Spark for histogram calculation and it's a bit different from pandas'
 - Histogram plot in plotly cannot be directly used in our case but should work around by leveraging bar charts because we don't use plotly to calculate histogram, see https://plotly.com/python/histograms/.

### DIFF
--- a/databricks/koalas/plot/core.py
+++ b/databricks/koalas/plot/core.py
@@ -18,6 +18,8 @@ import importlib
 
 import pandas as pd
 import numpy as np
+
+from databricks.koalas.utils import name_like_string
 from pyspark.ml.feature import Bucketizer
 from pyspark.sql import functions as F
 from pandas.core.base import PandasObject
@@ -143,8 +145,11 @@ class HistogramPlotBase:
 
         sdf = kdf._internal.spark_frame
         scols = []
+        input_column_names = []
         for label in kdf._internal.column_labels:
-            scols.append(kdf._internal.spark_column_for(label))
+            input_column_name = name_like_string(label)
+            input_column_names.append(input_column_name)
+            scols.append(kdf._internal.spark_column_for(label).alias(input_column_name))
         sdf = sdf.select(*scols)
 
         # 1. Make the bucket output flat to:
@@ -243,7 +248,7 @@ class HistogramPlotBase:
         #     |0                |
         #     +-----------------+
         output_series = []
-        for i, bucket_name in enumerate(bucket_names):
+        for i, (input_column_name, bucket_name) in enumerate(zip(input_column_names, bucket_names)):
             current_bucket_result = result[result["__group_id"] == i]
             # generates a pandas DF with one row for each bin
             # we need this as some of the bins may be empty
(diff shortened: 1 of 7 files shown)


databricks/koalas/plot/core.py
```diff
@@ -338,31 +344,41 @@ def _get_plot_backend(backend=None):
 
             KoalasPlotAccessor._backends["matplotlib"] = module
 
+        if backend == "plotly":
```
HyukjinKwon, line 347: Here I avoided to monkey patch plotly. I will need to refactor codes here in another PR .. here needs some refactoring.

databricks/koalas/plot/matplotlib.py
```diff
@@ -16,7 +16,7 @@
 
 from distutils.version import LooseVersion
 
-import matplotlib
+import matplotlib as mat
```
HyukjinKwon, line 19: Here is due to the conflict when you import, `from databricks.koalas.plot import matplotlib`
As a module `databricks.koalas.plot.matplotlib` vs `matplotlib` library being imported at `databricks.koalas.plot`. I will fix it back in the next PR that refactors.

databricks/koalas/plot/core.py
```diff
@@ -338,31 +343,42 @@ def _get_plot_backend(backend=None):
...
+            try:
+                # test if plotly can be imported
+                import plotly  # noqa: F401
+                from databricks.koalas.plot import plotly as module
+            except ImportError:
+                raise ImportError(
+                    "matplotlib is required for plotting when the "
+                    "default backend 'matplotlib' is selected."
```
ueshin, line 354: The error message sounds weird? We should replace "matplotlib" with "plotly" here?
Also "default" is not needed?

databricks/koalas/plot/plotly.py
```diff
@@ -15,17 +15,21 @@
...
-    return plot
+    # Koalas specific plots
+    if kind == "pie":
+        return plot_pie(data, **kwargs)
+    if kind == "hist":
+        # Note that here data is a Koalas DataFrame or Series unlike other type of plots.
+        return plot_histogram(data, **kwargs)
+
+    # Other plots.
+    return plotly.plot(data, kind, **kwargs)
```
ueshin, line 32: Nice! 👍

ueshin (commented):
Seems good.

databricks/koalas/plot/core.py
HyukjinKwon: oops

PR #2005, opened 2021-01-13
Refactor to use one similar logic to call plot backends

HyukjinKwon:
This PR proposes:
- Remove `koalas_plotting_backends`. We don't currently have such mechanism like https://github.com/pandas-dev/pandas/commit/e9a60bbc4f70820630f12666c273d14bcf83f1a7
- Load and use plotting backend in the same way:
  - If a plot module has `plot`, we use it after converting it to pandas instance.
  - if a plot module has `plot_koalas` (Koalas' `matplotlib` and `plotly` modules for example), we just pass Koalas instances to plot.

Now `databricks.koalas.plot.plotly` and `databricks.koalas.plot.matplotlib` modules work like external plotting backends.

### DIFF
--- a/databricks/koalas/plot/core.py
+++ b/databricks/koalas/plot/core.py
@@ -324,7 +324,7 @@ class KoalasPlotAccessor(PandasObject):
             # Because matplotlib is an optional dependency and first-party backend,
             # we need to attempt an import here to raise an ImportError if needed.
             try:
-                # test if plotly can be imported
+                # test if matplotlib can be imported
                 import matplotlib  # noqa: F401
                 from databricks.koalas.plot import matplotlib as module
             except ImportError:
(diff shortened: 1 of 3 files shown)


databricks/koalas/plot/core.py
```diff
@@ -325,15 +315,17 @@ def _find_backend(backend):
...
+        # Shortcut
+        if backend in KoalasPlotAccessor._backends:
+            return KoalasPlotAccessor._backends[backend]
 
-        # TODO: leverage koalas_plotting_backends, and remove the codes below.
         if backend == "matplotlib":
             # Because matplotlib is an optional dependency and first-party backend,
             # we need to attempt an import here to raise an ImportError if needed.
             try:
+                # test if plotly can be imported
```
xinrong-databricks, line 327: nit: matplotlib?

xinrong-databricks (approved):

ueshin (approved):
LGTM.

databricks/koalas/plot/core.py
HyukjinKwon, line 327:
```suggestion
                # test if matplotlib can be imported
```

HyukjinKwon:
Thanks guys!

itholic:
Nice!

PR #2011, opened 2021-01-16
Fix as_spark_type to not support "bigint".

ueshin:
Fix `as_spark_type` to not support "bigint".

The string "bigint" is not recognizable by `np.dtype` and it causes an unexpected error:

```py
>>> import numpy as np
>>> from databricks.koalas.typedef import as_spark_type
>>> as_spark_type(np.dtype("datetime64[ns]"))
...
TypeError: data type "bigint" not understood
```

Also, it doesn't work in pandas:

```py
>>> pd.Series([1, 2, 3], dtype="bigint")
...
TypeError: data type "bigint" not understood
```

### DIFF
--- a/databricks/koalas/typedef/typehints.py
+++ b/databricks/koalas/typedef/typehints.py
@@ -21,7 +21,6 @@ import typing
 import datetime
 import decimal
 from inspect import getfullargspec, isclass
-import warnings
 
 import numpy as np
 import pandas as pd
@@ -102,8 +101,6 @@ def as_spark_type(tpe) -> types.DataType:
     - dictionaries of field_name -> type
     - Python3's typing system
     """
-    from databricks.koalas.utils import is_testing
-
     # TODO: Add "boolean" and "string" types.
     # ArrayType
     if tpe in (np.ndarray,):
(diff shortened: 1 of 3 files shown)


databricks/koalas/typedef/typehints.py
```diff
@@ -128,8 +131,18 @@ def as_spark_type(tpe) -> types.DataType:
         return types.FloatType()
     elif tpe in (np.int32, "int32", "i"):
         return types.IntegerType()
-    elif tpe in (int, np.int, np.int64, "int", "int64", "long", "bigint"):
+    elif tpe in (int, np.int, np.int64, "int", "int64", "long"):
         return types.LongType()
+    elif isinstance(tpe, str) and tpe in ("bigint",):
```
HyukjinKwon, line 136: I think we can even remove this and add this in the release note as a breaking change since `bigint` isn't really a numpy type, and we did not document that `astype` can take DDL formatted type Spark types. But deprecating is fine too. I will leave it to you.

HyukjinKwon (approved):
LGTM

itholic (approved):
LGTM

databricks/koalas/typedef/typehints.py
ueshin: Then shall we remove this for now.

ueshin:
Thanks! merging.

PR #2012, opened 2021-01-20
Reuse as_spark_type in infer_pd_series_spark_type.

ueshin:
Now that `as_spark_type` is good enough for Koalas, we should reuse it in `infer_pd_series_spark_type` to avoid inconsistency.


HyukjinKwon (approved):

itholic (approved):
LGTM if test passes

ueshin:
Thanks! merging.

PR #2013, opened 2021-01-20
Use NullType for empty or null dataset.

ueshin:
Experimental.

### DIFF
--- a/databricks/koalas/tests/test_series.py
+++ b/databricks/koalas/tests/test_series.py
@@ -84,26 +84,38 @@ class SeriesTest(ReusedSQLTestCase, SQLTestUtils):
         self.assertEqual(s.__repr__(), s.rename("a").__repr__())
 
     def test_empty_series(self):
-        a = pd.Series([], dtype="i1")
-        b = pd.Series([], dtype="str")
+        pser_a = pd.Series([], dtype="i1")
+        pser_b = pd.Series([], dtype="str")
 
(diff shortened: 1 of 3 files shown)


itholic (approved):

databricks/koalas/tests/test_series.py
```diff
@@ -84,28 +84,38 @@ def test_repr_cache_invalidation(self):
...
-        b = pd.Series([], dtype="str")
+        pser_a = pd.Series([], dtype="i1")
+        pser_b = pd.Series([], dtype="str")
 
-        self.assert_eq(ks.from_pandas(a), a)
-        self.assertRaises(ValueError, lambda: ks.from_pandas(b))
+        self.assert_eq(ks.from_pandas(pser_a), pser_a)
+
+        kser_b = ks.from_pandas(pser_b)
+        if LooseVersion(pyspark.__version__) >= LooseVersion("2.4"):
```
HyukjinKwon, line 93: @ueshin, is it supported lower versions? I thought we fixed it only for 3.2 at https://github.com/apache/spark/pull/31285 (arrow code path though).

HyukjinKwon: Ah, but does it work for empty data?

ueshin: It still works with a fallback mode.

HyukjinKwon (approved):

PR #2020, opened 2021-01-25
Add python 3.9 support

gsheni:
- Add support for python 3.9 by fixing matplotlib, and sphinx requirements
- Add CI testing for python 3.9


gsheni:
- Waiting on conda-forge to release pyarrow 3.0.0 which supports python 3.9

HyukjinKwon:
Nice, thanks @gsheni.

itholic:
Great work, @gsheni

gsheni:
@HyukjinKwon I am have trouble getting the dev dependencies to work with Python <3.9 
- If I remove the pinning of (sphinx, nbconvert, nbformat), python 3.5, 3.6, 3.7, 3.8 tests fail
- If I don't pin those reqs, python 3.9 test fail
What do you suggest doing?

HyukjinKwon:
Can you push the changes for Python 3.9 build only in this PR for now? Will take a look how to make it compatible then.

gsheni:
@HyukjinKwon I have pushed only Python 3.9 changes to this build.

ueshin:
As conda for py3.9 seems not stable yet, how about adding a metrix for py3.9 build in `pip_build`?

gsheni:
@ueshin I added a matrix for python 3.9 build but the CI continues to fail:
- https://github.com/databricks/koalas/pull/2020/checks?check_run_id=1787967253

ueshin:
@gsheni Yeah, but at least now we can install dependencies and it's failing the Sphinx build test in `lint-python`.
Could you take a look at the failure? Thanks!

ueshin:
or skip to install Sphinx for py3.9 build, then at least we can see if the unit tests pass or not.

gsheni:
@ueshin The following tests failed 
```
FAILED databricks/koalas/tests/test_dataframe_conversion.py::DataFrameConversionTest::test_to_excel
FAILED databricks/koalas/tests/test_dataframe_spark_io.py::DataFrameSparkIOTest::test_read_excel
```

.github/workflows/master.yml
```diff
@@ -67,6 +72,7 @@ jobs:
         # It forces the sphinx version to 2.x.
         if [[ "$PYTHON_VERSION" < "3.6" ]]; then pip install "sphinx<3.0.0"; fi
         pip list
+        sudo apt install -y pandoc
```
ueshin, line 75: If the Sphinx build passes by installing `pandoc`, how about explicitly including it in `requirement-dev.txt` instead of `apt install` here?

requirements-dev.txt
```diff
@@ -36,4 +36,4 @@ openpyxl
 # by leveraging openpyxl, see also
 # https://stackoverflow.com/questions/65254535/xlrd-biffh-xlrderror-excel-xlsx-file-not-supported
 # We can remove this upperbound when our minimum pandas version is 0.25+.
-xlrd<2.0.0
+xlrd<2.0.0
```
ueshin, line 39: Seems like there is no difference here? I guess a line break was removed? Could you revert this?

ueshin (commented):
Awesome!

I guess we can skip the tests if Python > 3.8 for now with an issue filed as TODO. cc @HyukjinKwon

requirements-dev.txt
gsheni: Fixed

ueshin:
@gsheni As seems not working with including `pandoc` in `requirements-dev.txt` or the docker way, shall we go back to install it with `pip` as you did before for now?
Then could you make the two tests skip when py3.9?

.github/workflows/master.yml
gsheni: I believe pandoc has to be installed via `apt install`. The official instructions do not mention a way to install it via pip install
- https://pandoc.org/installing.html

PR #2022, opened 2021-01-27
Supports list-like Python objects for Series comparison.

### LINKED ISSUE #2018: Can't use __eq__ with list
In pandas doing
```python
pandas_series == [1, 2, 3]
```
does element wise comparison if the length match.
I koalas, this doesn't seem to be supported. Is that something you would consider adding?

Thanks!

itholic:
Currently Series doesn't support the comparison to list-like Python objects such as `list`, `tuple`, `dict`, `set`.

```python
>>> kser
0    1
1    2
2    3
dtype: int64

>>> kser == [3, 2, 1]
...
    raise Py4JJavaError(
py4j.protocol.Py4JJavaError: An error occurred while calling o77.equalTo.
...
```

This PR proposes supporting them as well for Series comparison.

```python
>>> kser
0    1
1    2
2    3
dtype: int64

>>> kser == [3, 2, 1]
0    False
1     True
2    False
dtype: bool
```

This should resolve #2018

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -658,6 +658,11 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         d    False
         Name: b, dtype: bool
         """
+        return self == other
+
+    equals = eq
+
+    def __eq__(self, other):
         if isinstance(other, (list, tuple)):
             if len(self) == len(other):
                 other = ks.Series(other)
@@ -666,9 +671,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         # pandas always returns False for all items with dict and set.
         elif isinstance(other, (dict, set)):
             return self != self
-        return self == other
-
-    equals = eq
+        return IndexOpsMixin.__eq__(self, other)
 
     def gt(self, other) -> "Series":
         """
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -658,6 +658,14 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         d    False
         Name: b, dtype: bool
         """
+        if isinstance(other, (list, tuple)):
+            if len(self) == len(other):
+                other = ks.Series(other)
+            else:
+                raise ValueError("Lengths must be equal")
+        # pandas always returns False for all items with dict and set.
+        elif isinstance(other, (dict, set)):
+            return self != self
         return self == other
 
     equals = eq
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -658,6 +658,13 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
         d    False
         Name: b, dtype: bool
         """
+        # pandas won't keep the name with `eq` when other is list of tuple,
+        # whereas `__eq__` always keeps the name.
+        if isinstance(other, (list, tuple)):
+            if len(self) == len(other):
+                other = ks.Series(other)
+            else:
+                raise ValueError("Lengths must be equal")
         return self == other
 
     equals = eq
@@ -665,7 +672,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
     def __eq__(self, other):
         if isinstance(other, (list, tuple)):
             if len(self) == len(other):
-                other = ks.Series(other)
+                other = ks.Series(other, name=self.name)
             else:
                 raise ValueError("Lengths must be equal")
         # pandas always returns False for all items with dict and set.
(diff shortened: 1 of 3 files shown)


ueshin:
Found a bug:

```py
>>> pser = pd.Series([1,2,3], index=[10,20,30])
>>> pser == [3, 2, 1]
10    False
20     True
30    False
dtype: bool
```

whereas:

```py
>>> kser = ks.Series([1,2,3], index=[10,20,30])
>>> kser == [3, 2, 1]
0     False
1     False
10    False
2     False
30    False
20    False
dtype: bool
```

ueshin:
Btw, do we also want to support binary operations with list-like Python objects? cc @HyukjinKwon 

```py
>>> pser + [3, 2, 1]
10    4
20    4
30    4
dtype: int64
>>> pser - [3, 2, 1]
10   -2
20    0
30    2
dtype: int64
>>> [3, 2, 1] + pser
10    4
20    4
30    4
dtype: int64
```

itholic:
FYI: Seems like pandas has some inconsistent behavior as below.

```python
>>> a = pd.Series([1, 1, 1, np.nan], index=['a', 'b', 'c', 'd'])
>>> b = pd.Series([1, np.nan, 1, np.nan], index=['a', 'b', 'd', 'e'])

>>> a.eq(b)
a     True
b    False
c    False
d    False
e    False
dtype: bool

>>> a == b
...
ValueError: Can only compare identically-labeled Series objects
```

I posted question to pandas repo, and will share if they response.

itholic:
> Btw, we might also want to support binary operations with list-like Python objects? cc @HyukjinKwon
> 
> ```python
> >>> pser + [3, 2, 1]
> 10    4
> 20    4
> 30    4
> dtype: int64
> >>> pser - [3, 2, 1]
> 10   -2
> 20    0
> 30    2
> dtype: int64
> >>> [3, 2, 1] + pser
> 10    4
> 20    4
> 30    4
> dtype: int64
> ```

Let me do this in the separated PR since there may be inconsistent cases like `eq`.

ueshin:
The `eq` case sounds different from the topic here which is binary operations between `Series` and `list`.

databricks/koalas/indexes/base.py
```diff
@@ -353,8 +353,10 @@ def equals(self, other) -> bool:
                     # some exceptions when 'compute.ops_on_diff_frames' is enabled.
                     # Working around for now via using frame.
                     return (
-                        self.to_series("self").reset_index(drop=True)
-                        == other.to_series("other").reset_index(drop=True)
+                        IndexOpsMixin.__eq__(
+                            self.to_series("self").reset_index(drop=True),
+                            other.to_series("other").reset_index(drop=True),
+                        )
```
itholic, line 359: Series comparison with `==` operator uses `Index.equals` internally.

So it falls eternal recursive if we use `==` for Series comparison here.

databricks/koalas/series.py
```diff
@@ -657,10 +657,27 @@ def eq(self, other) -> bool:
...
+            other = ks.Index(other, name=self.name)
+        # pandas always returns False for all items with dict and set.
+        elif isinstance(other, (dict, set)):
+            return self != self
+        # pandas doesn't support `==` for different Index, but support for `eq` function.
+        elif isinstance(other, ks.Series):
+            if not self.index.equals(other.index):
+                raise ValueError("Can only compare identically-labeled Series objects")
+        return IndexOpsMixin.__eq__(self, other)
+
```
itholic, line 680: we cannot use same code for both `eq` and `__eq__` because the behavior is different in pandas.

databricks/koalas/series.py
```diff
@@ -657,10 +657,27 @@ def eq(self, other) -> bool:
...
+        elif isinstance(other, (dict, set)):
+            return self != self
+        return IndexOpsMixin.__eq__(self, other)
 
     equals = eq
 
+    def __eq__(self, other):
+        if isinstance(other, (list, tuple)):
+            other = ks.Index(other, name=self.name)
+        # pandas always returns False for all items with dict and set.
```
HyukjinKwon, line 672: I wonder why pandas behaves like this ..

databricks/koalas/series.py
```diff
@@ -657,10 +657,27 @@ def eq(self, other) -> bool:
...
+        elif isinstance(other, (dict, set)):
+            return self != self
+        return IndexOpsMixin.__eq__(self, other)
 
     equals = eq
 
+    def __eq__(self, other):
```
HyukjinKwon, line 669: Does `Index` support this case too? it might be best to move to `base.py`.

databricks/koalas/series.py
```diff
@@ -635,7 +635,7 @@ def rfloordiv(self, other) -> "Series":
     koalas = CachedAccessor("koalas", KoalasSeriesMethods)
 
     # Comparison Operators
-    def eq(self, other) -> bool:
+    def eq(self, other) -> "Series":
```
xinrong-databricks, line 638: Good catch! Thanks!

xinrong-databricks: I'm wondering if we want to add a docstring example here for list-like Python objects.

itholic: Added, thanks :)

xinrong-databricks:
https://issues.apache.org/jira/browse/SPARK-36438

PR #2026, opened 2021-02-01
Support NumPy 1.20.0

itholic:
The several `mypy` tests are failing on Python >= 3.7 in CI since [NumPy has been upgraded to 1.20.0.](https://github.com/numpy/numpy/releases)

Since NumPy dropped Python 3.6 from NumPy 1.20.0, only the tests run on Python >= 3.7 are affected.

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4773,8 +4773,8 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
     @staticmethod
     def from_records(
-        data: Union[np.array, List[tuple], dict, pd.DataFrame],
-        index: Union[str, list, np.array] = None,
+        data: Union[np.ndarray, List[tuple], dict, pd.DataFrame],
+        index: Union[str, list, np.ndarray] = None,
         exclude: list = None,
         columns: list = None,
         coerce_float: bool = False,
--- a/databricks/koalas/indexing.py
+++ b/databricks/koalas/indexing.py
@@ -1449,14 +1449,14 @@ class iLocIndexer(LocIndexerLike):
     ) -> Tuple[Optional[spark.Column], Optional[int], Optional[int]]:
         sdf = self._internal.spark_frame
 
-        if any(isinstance(key, (int, np.int, np.int64, np.int32)) and key < 0 for key in rows_sel):
+        if any(isinstance(key, (int, np.int64, np.int32)) and key < 0 for key in rows_sel):
             offset = sdf.count()
         else:
             offset = 0
 
         new_rows_sel = []
         for key in list(rows_sel):
-            if not isinstance(key, (int, np.int, np.int64, np.int32)):
+            if not isinstance(key, (int, np.int64, np.int32)):
                 raise TypeError(
                     "cannot do positional indexing with these indexers [{}] of {}".format(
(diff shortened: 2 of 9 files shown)


itholic:
Here is the summary of investigation:

```python
>>> pdf = pd.DataFrame(
...     {
...         "a": [1, 2, 3, 4, 5, 6] * 100,
...         "b": [1.0, 1.0, 2.0, 3.0, 5.0, 8.0] * 100,
...         "c": [1, 4, 9, 16, 25, 36] * 100,
...     },
...     columns=["a", "b", "c"],
...     index=np.random.rand(600),
... )
>>> kdf = ks.DataFrame(pdf)
>>> kdf.koalas.transform_batch(lambda pdf: pdf.c + 1)
org.apache.spark.api.python.PythonException: Traceback (most recent call last):
...
pyarrow.lib.ArrowTypeError: Did not pass numpy.dtype object
...
RuntimeError: ('Exception thrown when converting pandas.Series (int64) to Arrow Array (int64). It can be caused by overflows or other unsafe conversions warned by Arrow. Arrow safe type check can be disabled by using SQL config `spark.sql.execution.pandas.convertToArrowArraySafely`.', ArrowTypeError('Did not pass numpy.dtype object'))
```

Seems like there is an incompatible issue between PyArrow and latest NumPy (1.20.0).

See https://github.com/Azure/MachineLearningNotebooks/issues/1314.

Until now, people seem to choose the solution not using the latest version of NumPy.

So, I'd say maybe it would be better to keep the NumPy upperbound and pending this PR until the issue is resolved ?

WDYT, @ueshin @HyukjinKwon @xinrong-databricks ??

HyukjinKwon:
Can you file an issue in Apache Arrow side and confirm that this is an issue?

Cedric-Magnan:
@itholic, I've run the following code within a databricks cluster using the test_mypy branch, python 3.7.6, PyArrow 4.0.0, numpy 1.20.3, pandas 1.2.3 and pyspark 3.0.1

I can't seem to reproduce your issue :

```python
import pandas as pd
import numpy as np

pdf = pd.DataFrame(
    {
        "a": [1, 2, 3, 4, 5, 6] * 100,
        "b": [1.0, 1.0, 2.0, 3.0, 5.0, 8.0] * 100,
        "c": [1, 4, 9, 16, 25, 36] * 100,
    },
    columns=["a", "b", "c"],
    index=np.random.rand(600),
)
kdf = ks.DataFrame(pdf)
kdf.koalas.transform_batch(lambda pdf: pdf.c + 1)
```

Output :
```sh
Out[1]: 0.519803     2
0.381068     5
0.928598    10
0.324069    17
0.600385    26
0.559973    37
0.196704     2
0.272351     5
0.659746    10
0.337685    17
...
...
```

I've also checked what is wrong within the CI :
For the failed build with python3.7 with pandas 0.25.3, it looks like pandas version prior to 1.0.4 are having issues with numpy >= 1.20.0.
Link to the issue
https://github.com/pandas-dev/pandas/issues/39520

Hope this can help !

xinrong-databricks:
numpy 1.21.1 has been supported in pyspark-pandas module of Build and test. I will close the PR now. @itholic let me know if there are any concerns. :)

PR #2029, opened 2021-02-01
Change matplotlib as an optional dependency

HyukjinKwon:
This PR proposes to change matplotlib as an optional dependency.

```python
>>> from databricks import koalas as ks
>>> ks.range(100).plot.bar()
  ...
ImportError: matplotlib is required for plotting when the default backend 'matplotlib' is selected.
```

Resolves #issues


ueshin (approved):
Shall we move it to `Optional dependencies in Koalas.` in `requirements-dve.txt` as well?
Otherwise, LGTM.

xinrong-databricks:
Great, thanks!

PR #2037, opened 2021-02-02
Use iloc to make sure it retrieves the first element.

ueshin:
Use `iloc` in `infer_pd_series_spark_type` to make sure it retrieves the first element.

Before:

```py
>>> infer_pd_series_spark_type(pd.Series(['a','b','c'], index=[10,20,30]))
...
KeyError: 0
```

After:

```py
>>> infer_pd_series_spark_type(pd.Series(['a','b','c'], index=[10,20,30]))
StringType
```

Note that currently the function in only called from `InternalFrame.from_pandas` which always resets index before calling it.


HyukjinKwon (approved):

itholic:
Nice

PR #2041, opened 2021-02-05
Show a warning message when Python 3.5.

ueshin:
Shows a warning message when Python 3.5.

### DIFF
--- a/databricks/koalas/__init__.py
+++ b/databricks/koalas/__init__.py
@@ -21,23 +21,24 @@ from databricks.koalas.version import __version__  # noqa: F401
 
 
 def assert_python_version():
-    import logging
+    import warnings
 
-    _major = 3
-    _minor = 5
-    _deprecated_version = (_major, _minor)
-    _min_supported_version = (_major, _minor + 1)
+    major = 3
+    minor = 5
+    deprecated_version = (major, minor)
+    min_supported_version = (major, minor + 1)
 
-    if sys.version_info[:2] == _deprecated_version:
-        logging.warning(
+    if sys.version_info[:2] <= deprecated_version:
+        warnings.warn(
             "Koalas support for Python {dep_ver} is deprecated and will be dropped in "
             "the future release. At that point, existing Python {dep_ver} workflows "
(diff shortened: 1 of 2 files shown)


databricks/koalas/__init__.py
```diff
@@ -20,10 +20,30 @@
...
+    if sys.version_info[:2] == _deprecated_version:
+        logging.warning(
+            "Koalas support for Python {dep_ver} is deprecated and will be dropped in "
+            "the future release. At that point, existing Python {dep_ver} workflows "
+            "that use Koalas will continue to work without modification, but Python {dep_ver} "
+            "users will no longer get access to the latest Koalas features and bugfixes. "
+            "We recommend that you upgrade to Python {min_ver} or newer.".format(
+                dep_ver=".".join(map(str, _deprecated_version)),
+                min_ver=".".join(map(str, _min_supported_version)),
+            )
```
HyukjinKwon, line 40: @ueshin shall we add `FutureWarning` warning type?

ueshin: Currently it's using `logging.warning` as same as the following warnings. Should we use `warnings` instead here?

databricks/koalas/__init__.py
```diff
@@ -20,10 +20,30 @@
 from databricks.koalas.version import __version__  # noqa: F401
 
 
+def assert_python_version():
+    import logging
+
+    _major = 3
```
HyukjinKwon, line 26: I think it's fine to don't use leading understore because they are not exposed to the outer scope anyway.

databricks/koalas/__init__.py
HyukjinKwon: If we're worried it becomes too noisy, we could use `DeprecatedWarning` type too which is not printed out by default.

HyukjinKwon: Yeah, if they are deprecation, we should use such types instead of UserWarning

HyukjinKwon (approved):
Looks good (except the doc update to be done)

xinrong-databricks (approved):

PR #2047, opened 2021-02-10
Fix bug on pow and rpow

itholic:
Fixed incompatible behavior for `pow` and `rpow`.

In pandas:

```python
>>> pd.Series([1, 2, 3]) ** np.nan
0    1.0
1    NaN
2    NaN
dtype: float64

>>> 1 ** pd.Series([np.nan, 2, 3])
0    1.0
1    1.0
2    1.0
dtype: float64
```

In Koalas:

```python
>>> ks.Series([1, 2, 3]) ** np.nan
0    NaN  # doesn't match
1    NaN
2    NaN
dtype: float64

>>> 1 ** ks.Series([np.nan, 2, 3])
0    NaN  # doesn't match
1    1.0
2    1.0
dtype: float64
```

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -581,13 +581,13 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
 
     def __pow__(self, other) -> Union["Series", "Index"]:
         def pow_func(left, right):
-            return F.when(F.lit(right is np.nan) & (left == 1), left).otherwise(pow(left, right))
+            return F.when(left == 1, left).otherwise(Column.__pow__(left, right))
 
         return column_op(pow_func)(self, other)
 
     def __rpow__(self, other) -> Union["Series", "Index"]:
         def rpow_func(left, right):
-            return F.when(left.isNull() & F.lit(right == 1), right).otherwise(pow(right, left))
+            return F.when(F.lit(right == 1), right).otherwise(Column.__rpow__(left, right))
 
         return column_op(rpow_func)(self, other)
 
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -579,8 +579,18 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
 
         return column_op(rmod)(self, other)
 
-    __pow__ = column_op(Column.__pow__)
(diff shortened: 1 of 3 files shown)


databricks/koalas/base.py
```diff
@@ -579,8 +579,18 @@ def rmod(left, right):
 
         return column_op(rmod)(self, other)
 
-    __pow__ = column_op(Column.__pow__)
-    __rpow__ = column_op(Column.__rpow__)
+    def __pow__(self, other) -> Union["Series", "Index"]:
+        def pow_func(left, right):
+            return F.when(F.lit(right is np.nan) & (left == 1), left).otherwise(pow(left, right))
```
ueshin, line 584: I guess just checking `(left == 1)` is enough?

Otherwise:

```py
>>> ks.Series([1, 2, 3]) ** ks.Series([np.nan, 2, 3])
0     NaN
1     4.0
2    27.0
dtype: float64
>>> pd.Series([1, 2, 3]) ** pd.Series([np.nan, 2, 3])
0     1.0
1     4.0
2    27.0
dtype: float64
```

ueshin, line 584: We should use `Column.__pow__(left, right)` or `left.__pow__(right)` instead of builtin `pow`?

databricks/koalas/base.py
```diff
@@ -579,8 +579,18 @@ def rmod(left, right):
...
+
+        return column_op(pow_func)(self, other)
+
+    def __rpow__(self, other) -> Union["Series", "Index"]:
+        def rpow_func(left, right):
+            return F.when(left.isNull() & F.lit(right == 1), right).otherwise(pow(right, left))
```
ueshin, line 590: I guess just checking `(right == 1)` is enough?

ueshin, line 590: We should use `Column.__rpow__(left, right)` or `left.__rpow__(right)` instead of builtin `pow`?

itholic:
Thanks for the review, @ueshin :)

ueshin (approved):
LGTM, pending tests.

ueshin:
Thanks! merging.

PR #2054, opened 2021-02-15
Enabling binary operations with list-like Python objects.

itholic:
So far, Koalas doesn't support list-like Python objects for Series binary operations.

```python
>>> kser
0    1
1    2
2    3
3    4
4    5
5    6
Name: x, dtype: int64

>>> kser + [10, 20, 30, 40, 50, 60]
...
```

This PR enables this.

```python
>>> kser
0    1
1    2
2    3
3    4
4    5
5    6
Name: x, dtype: int64
>>> kser + [10, 20, 30, 40, 50, 60]
0    11
1    22
2    33
3    44
4    55
5    66
Name: x, dtype: int64
>>> kser - [10, 20, 30, 40, 50, 60]
0    -9
1   -18
2   -27
3   -36
4   -45
5   -54
Name: x, dtype: int64
>>> kser * [10, 20, 30, 40, 50, 60]
0     10
1     40
2     90
3    160
4    250
5    360
Name: x, dtype: int64
>>> kser / [10, 20, 30, 40, 50, 60]
0    0.1
1    0.1
2    0.1
3    0.1
4    0.1
5    0.1
Name: x, dtype: float64
```

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -452,6 +452,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
     def add(self, other) -> "Series":
         return self + other
 
+    def __add__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__add__(self, other)
+
     add.__doc__ = _flex_doc_SERIES.format(
         desc="Addition",
         op_name="+",
@@ -462,6 +466,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
 
     def radd(self, other) -> "Series":
         return other + self
+
+    def __radd__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__radd__(self, other)
 
     radd.__doc__ = _flex_doc_SERIES.format(
         desc="Reverse Addition",
@@ -498,6 +506,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
     def truediv(self, other) -> "Series":
         return self / other
 
+    def __truediv__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__truediv__(self, other)
+
     truediv.__doc__ = _flex_doc_SERIES.format(
         desc="Floating division",
         op_name="/",
@@ -509,6 +521,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
     def rtruediv(self, other) -> "Series":
         return other / self
 
+    def __rtruediv__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__rtruediv__(self, other)
+
     rtruediv.__doc__ = _flex_doc_SERIES.format(
         desc="Reverse Floating division",
         op_name="/",
@@ -519,6 +535,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
 
     def mul(self, other) -> "Series":
         return self * other
+
+    def __mul__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__mul__(self, other)
 
     mul.__doc__ = _flex_doc_SERIES.format(
         desc="Multiplication",
@@ -533,6 +553,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
     def rmul(self, other) -> "Series":
         return other * self
 
+    def __rmul__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__rmul__(self, other)
+
     rmul.__doc__ = _flex_doc_SERIES.format(
         desc="Reverse Multiplication",
         op_name="*",
@@ -543,6 +567,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
 
     def sub(self, other) -> "Series":
         return self - other
+
+    def __sub__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__sub__(self, other)
 
     sub.__doc__ = _flex_doc_SERIES.format(
         desc="Subtraction",
@@ -557,6 +585,10 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
     def rsub(self, other) -> "Series":
         return other - self
 
+    def __rsub__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
+        return IndexOpsMixin.__rsub__(self, other)
+
     rsub.__doc__ = _flex_doc_SERIES.format(
         desc="Reverse Subtraction",
         op_name="-",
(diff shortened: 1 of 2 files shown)


databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -1528,6 +1528,595 @@ def test_pow_and_rpow(self):
...
+        if LooseVersion(pd.__version__) >= LooseVersion("1.2"):
+            self.assert_eq(pser.sub(other), kser.sub(other).sort_index())
+            self.assert_eq(pser - other, (kser - other).sort_index())
+            # self.assert_eq(pser.rsub(other), kser.rsub(other).sort_index())
+            # self.assert_eq(other - pser, (other - kser).sort_index())
+        else:
+            self.assert_eq(pser.sub(other).rename("x"), kser.sub(other).sort_index())
+            self.assert_eq((pser - other).rename("x"), (kser - other).sort_index())
+            # self.assert_eq(pser.rsub(other).rename("x"), kser.rsub(other).sort_index())
+            # self.assert_eq((other - pser).rename("x"), (other - kser).sort_index())
```
itholic, line 1659: `rsub` doesn't work properly for now.

I'm investigating

databricks/koalas/base.py
```diff
@@ -495,7 +495,7 @@ def __rsub__(self, other) -> Union["Series", "Index"]:
                 return -column_op(F.datediff)(self, F.lit(other)).astype("long")
             else:
                 raise TypeError("date subtraction can only be applied to date series.")
-        return column_op(Column.__rsub__)(self, other)
+        return column_op(lambda left, right: right - left)(self, other)
```
itholic, line 498: FYI: `Column.__rsub__` doesn't support `pyspark.sql.column.Column` for second parameter.

```python
>>> kdf = ks.DataFrame({"A": [1, 2, 3, 4], "B": [10, 20, 30, 40]})
>>> sdf = kdf.to_spark()
>>> col1 = sdf.A
>>> col2 = sdf.B
>>> Column.__rsub__(col1, col2)
...
TypeError: Column is not iterable
```

databricks/koalas/tests/test_ops_on_diff_frames.py
```diff
@@ -1528,6 +1528,595 @@ def test_pow_and_rpow(self):
...
+            kser.rmod(other)
+        with self.assertRaisesRegex(
+            ValueError, "operands could not be broadcast together with shapes"
+        ):
+            other % kser
+
+        # other = tuple with the different length
+        other = (np.nan, 1, 3, 4, np.nan)
+        with self.assertRaisesRegex(
+            ValueError, "operands could not be broadcast together with shapes"
```
HyukjinKwon, line 1936: The error message looks weird. Is it matched with pandas'?

databricks/koalas/base.py
HyukjinKwon: It does support:

```python
>>> Column.__rsub__(df.id, 1)
Column<'(1 - id)'>
```

It doesn't work in your case above because the instance is Spark column. In practice, that wouldn't happen because it will only be called when the first operand doesn't know how to handle Spark column e.g.) `1 - df.id`.

HyukjinKwon: Does it cause any exception?

databricks/koalas/series.py
```diff
@@ -452,6 +452,10 @@ def spark_type(self):
     def add(self, other) -> "Series":
         return self + other
 
+    def __add__(self, other):
+        other = ks.Index(other, name=self.name) if isinstance(other, (list, tuple)) else other
```
HyukjinKwon, line 456: - Why do we create Index to compare instead of Series?
- Looks like Index support this too. Can't we move this into `base.py`?

databricks/koalas/tests/test_ops_on_diff_frames.py
itholic: The original error message from pandas looks like :

```python
ValueError: operands could not be broadcast together with shapes (4,) (8,)
```

@ueshin , maybe we don't include the `(4,) (8,)` part since it requires compute length of both objects which can be expensive ??

databricks/koalas/series.py
itholic: > * Why do we create Index to compare instead of Series?

If we use Series here, we cannot match the Index as below.

```python
>>> kser = ks.Series([1, 2, 3, 4], index=[10, 20, 30, 40])
>>> kser + ks.Series([3, 4, 5, 6])
0    NaN
1    NaN
10   NaN
3    NaN
2    NaN
30   NaN
20   NaN
40   NaN
dtype: float64
```

so we use Index.

```python
>>> kser + ks.Index([3, 4, 5, 6])
10     4
20     6
30     8
40    10
dtype: int64
```

> * Looks like Index support this too. Can't we move this into `base.py`?

Yes, think we can. Let me move this into `base.py`

Thanks!

databricks/koalas/base.py
itholic: If we use `column_op(Column.__rsub__)(self, other)` as it is, it raises `TypeError: Column is not iterable` for the case below.

```python
>>> kser = ks.Series([1, 2, 3, 4])
>>> [10, 20, 30, 40] - kser
...
TypeError: Column is not iterable
```

databricks/koalas/utils.py
```diff
@@ -813,3 +813,31 @@ def compare_disallow_null(left, right, comp):
 
 def compare_allow_null(left, right, comp):
     return left.isNull() | right.isNull() | comp(left, right)
+
+
+def check_same_length(left: "IndexOpsMixin", right: Union[list, tuple]):
```
xinrong-databricks, line 818: Nice utility! The function name might be misleading while concluding its return type. Would it be possible to annotate the return type?

databricks/koalas/base.py
ueshin: This case must be handled in lines 490-492. We can move back to `Column.__rsub__`.

databricks/koalas/base.py
```diff
@@ -560,6 +593,11 @@ def floordiv(left, right):
         return numpy_column_op(floordiv)(self, other)
 
     def __rfloordiv__(self, other) -> Union["Series", "Index"]:
+        if isinstance(other, (list, tuple)):
+            pindex_ops, other = check_same_length(self, other)
+            return ks.from_pandas(other // pindex_ops)  # type: ignore
+        if isinstance(other, (list, tuple)):
+            other = ks.Index(other, name=self.name)  # type: ignore
```
ueshin, line 600: not needed?

ueshin (commented):
Also could you try to reduce the amount of test codes by using loop or parameterizing?

databricks/koalas/utils.py
```diff
@@ -813,3 +813,31 @@ def compare_disallow_null(left, right, comp):
...
+    with ks.option_context("compute.ordered_head", True):
+        len_right = len(right)
+        if isinstance(left, ks.Series):
+            pindex_ops = left.head(len_right + 1)._to_internal_pandas()
+        elif isinstance(left, ks.Index):
+            pindex_ops = left._kdf.head(len_right + 1).index._to_internal_pandas()
+            # pandas < 1.2.0 doesn't fully support binary operations with list or tuple for Index.
+            # So, we convert list or tuple to the Index for this case.
+            if LooseVersion(pd.__version__) < LooseVersion("1.2.0"):
+                right = pd.Index(right, name=pindex_ops.name)
```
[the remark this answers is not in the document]
itholic: Ohh,,, seems like It doesn't work for only `rmod` operations.

```python
>>> [4, 5, 6] % pd.Index([1,2,3])
```

Let me address for only this case.

Thanks!

xinrong-databricks:
https://issues.apache.org/jira/browse/SPARK-36437

PR #2061, opened 2021-02-20
Fix to delegate dtypes.

ueshin:
Fixes to delegate dtypes properly.
The parameter `preserve_dtypes` for the workaround was removed.
Also maked as `TODO` where the dtypes are unknown.

### DIFF
--- a/databricks/koalas/series.py
+++ b/databricks/koalas/series.py
@@ -372,7 +372,7 @@ class Series(Frame, IndexOpsMixin, Generic[T]):
                 )
             internal = InternalFrame.from_pandas(pd.DataFrame(s))
             if s.name is None:
-                internal = internal.copy(column_labels=[None], preserve_dtypes=True)
+                internal = internal.copy(column_labels=[None])
             anchor = DataFrame(internal)
 
             self._anchor = anchor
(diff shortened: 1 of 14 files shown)


databricks/koalas/series.py
```diff
@@ -2715,7 +2713,9 @@ def add_prefix(self, prefix) -> "Series":
             ]
             + internal.data_spark_columns
         )
-        return first_series(DataFrame(internal.with_new_sdf(sdf)))
+        return first_series(
+            DataFrame(internal.with_new_sdf(sdf, index_dtypes=([None] * internal.index_level)))
```
xinrong-databricks, line 2717: Why don't we pass `object`s as `index_dtypes` directly?

xinrong-databricks (approved):

xinrong-databricks:
I'm wondering if we wanted to call out `spark_type_to_pandas_dtype` is where the type mapping defined in `internal.py`.
I'm afraid contributors with little context might not be clear about this.

xinrong-databricks:
Looks great! Left some questions. Thank you!

databricks/koalas/series.py
ueshin: Why `object`?

If we pass `None`, the dtype will be inferred during the initialization of `InternalFrame`.

ueshin:
Thanks! let me merge this now. Please feel free to leave comments later if any.

PR #2065, opened 2021-02-25
Upgrade PySpark to 3.0.2 in CI.

ueshin:
As PySpark 3.0.2 has been released, we should upgrade it in CI.
- http://spark.apache.org/news/spark-3-0-2-released.html


ueshin:
As tests passed, I'd just merge this.

PR #2070, opened 2021-03-01
Support subscription to denote expected types for Index

xinrong-databricks:
Support subscription to denote expected types for Index

From
```
>>> def func() -> Index[np.int64]:
...     pass
... 
...
TypeError: 'ABCMeta' object is not subscriptable
```

To
```
>>> def func() -> Index[np.int64]:
...     pass
...
>>>
```

### DIFF
--- a/databricks/koalas/indexes/base.py
+++ b/databricks/koalas/indexes/base.py
@@ -15,7 +15,8 @@
 #
 
 from functools import partial
-from typing import Any, List, Optional, Tuple, Union
+from typing import Any, Generic, List, Optional, Tuple, TypeVar, Union
+import sys
 import warnings
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


databricks/koalas/indexes/base.py
```diff
@@ -116,6 +119,11 @@ def __new__(cls, data=None, dtype=None, copy=False, name=None, tupleize_cols=Tru
             )
         )
 
+    if sys.version_info >= (3, 7):
+        # In order to support the type hints such as Index[...]. See DataFrame.__class_getitem__.
+        def __class_getitem__(cls, tpe):
+            return IndexType[tpe]
```
ueshin, line 125: Shall we put this at the bottom of the class definition?

PR #2077, opened 2021-03-03
Return actual values via toPandas.

### LINKED ISSUE #2068: datetime support for min and max functions
It looks to me I'm doing something wrong, but I can't find the reason.

image

It seems that when I select the row using min / max functions, an additional hour is added to the date. The same example using pandas works fine. Could you please help me?

ueshin:
As `sdf.toPandas()` has a special handling for the timestamp type, we should always use `sdf.toPandas()` when returning the actual values.


databricks/koalas/frame.py
```diff
@@ -11588,9 +11590,11 @@ def _reduce_spark_multi(sdf, aggs):
     """
     assert isinstance(sdf, spark.DataFrame)
     sdf0 = sdf.agg(*aggs)
-    l = sdf0.head(2)
+    with sql_conf({SPARK_CONF_ARROW_ENABLED: False}):
+        # Disable Arrow to keep row ordering.
```
HyukjinKwon, line 11594: Hm .. I thought we guaranteed to keep the row ordering even when Arrow optimization is enabled e.g., https://github.com/apache/spark/blob/9b875ceada60732899053fbd90728b4944d1c03d/python/pyspark/sql/pandas/conversion.py#L288-L293.

ueshin: Yeah, but `sdf.limit(n).toPandas()` with Arrow enabled doesn't.

HyukjinKwon: BTW, why does the order matter here? It throws an exception if the results are more than one.

databricks/koalas/generic.py
```diff
@@ -2336,18 +2338,22 @@ def first_valid_index(self) -> Optional[Union[Scalar, Tuple[Scalar, ...]]]:
...
-            self._internal.spark_frame.filter(cond)
-            .select(self._internal.index_spark_columns)
-            .first()
-        )
+        with sql_conf({SPARK_CONF_ARROW_ENABLED: False}):
+            # Disable Arrow to keep row ordering.
+            first_valid_row = (
+                self._internal.spark_frame.filter(cond)
+                .select(self._internal.index_spark_columns)
+                .limit(1)
```
HyukjinKwon, line 2346: Can we mimic what we do in Koalas' DataFrame.head?

ueshin: Do you mean we should add `.orderBy(NATURAL_ORDER_COLUMN_NAME)` instead of disabling Arrow execution?

databricks/koalas/frame.py
ueshin: ah, that's right. good catch. removed.

databricks/koalas/generic.py
HyukjinKwon: I was just wondering ...

HyukjinKwon (approved):
I am good with as is. Looks fine.

databricks/koalas/generic.py
HyukjinKwon: Hm, I still don't get why the order gets changed though.

ueshin: The underlying plans are different between with and without Arrow as well as the execution method:

e.g.,

```py
sdf = spark.range(10)
sdf.limit(5).toPandas()
```

----

- Without Arrow

`collect()` -> `plan.executeCollect().map(fromRow)`

The plan is:

```
== Physical Plan ==
CollectLimit 5
+- *(1) Range (0, 10, step=1, splits=12)
```

There is no shuffle and `CollectLimitExec.executeCollect()` runs several Spark jobs **from the first partition** until the specified number of rows are collected. Thus the row order will be preserved.

----

- With Arrow:

`collectAsArrowToPython()` -> `toArrowBatchRdd(plan)` -> `plan.execute().mapPartitionsInternal ...`

The plan is:

```
== Physical Plan ==
*(2) Project [id#0L AS col_0#12L]
+- *(2) GlobalLimit 5
   +- Exchange SinglePartition, true, [id=#63]
      +- *(1) LocalLimit 5
         +- *(1) Range (0, 10, step=1, splits=12)
```

There is a shuffle to limit the number of rows which won't preserve the order of rows and `plan.execute()` will just return the generated RDD as-is.

HyukjinKwon: gotya. we gotta fix it in Apache Spark I guess .. anyway thanks. LGTM

ueshin:
Thanks! Let me merge this now.

Issue #2072, opened 2021-03-01
groupby and compute.max_rows ?


lfdversluis:
When using `groupby` and then `apply` will the pandas dataframe supplied to the function to be applied have at most `compute.max_rows` rows? I cannot reliably tell from [the documentation](https://koalas.readthedocs.io/en/latest/user_guide/options.html) if this also applies to groupby. 

I am applying a function per group, where a group may very well exceed the 100k rows. Will the function then get chunks (and thus lead to incorrect results for me) or will in such a case a Pandas dataframe with all the rows be passed to my function?

ueshin:
Hi @lfdversluis,

No, `groupby-apply` doesn't check `compute.max_rows` and the whole data in each group will be passed to the function.

Could you share the reproducible example that seems incorrect?

Thanks.

lfdversluis:
Thanks for your answer. It was just a question as I am struggling with Memory issues, and I was checking the documentation on how to manage this. Knowing the groupby does provide the full group the my function is good to know. May I suggest that this is emphasized in the documentation, just in case anyone else has this question :)

PR #2084, opened 2021-03-04
Add a note for GroupBy and compute.max_rows

lfdversluis:
This clarifies a question raised in #2072 regarding the impact of max_rows on the size of dataframes/series received by function when using, e.g., `groupby` and `apply`, where the groups can exceed `compute.max_rows` rows.

### DIFF
--- a/databricks/koalas/config.py
+++ b/databricks/koalas/config.py
@@ -134,9 +134,9 @@ _options = [
     Option(
         key="compute.max_rows",
         doc=(
-            "'compute.max_rows' sets the limit of the current DataFrame. Set `None` to unlimit "
-            "the input length. When the limit is set, it is executed by the shortcut by "
-            "collecting the data into driver side, and then using pandas API. If the limit is "
+            "'compute.max_rows' sets the limit of the current Koalas DataFrame. Set `None` to "
+            "unlimit the input length. When the limit is set, it is executed by the shortcut by "
+            "collecting the data into the driver, and then using the pandas API. If the limit is "
             "unset, the operation is executed by PySpark. Default is 1000."
         ),
         default=1000,
--- a/dev/gendoc.py
+++ b/dev/gendoc.py
@@ -94,7 +94,7 @@ def gen_release_notes(path):
     shutil.rmtree(whatsnew_dir, ignore_errors=True)
     os.mkdir(whatsnew_dir)
 
-    with open("%s/index.rst" % whatsnew_dir, "a") as index_file:
+    with open("%s/index.rst" % whatsnew_dir, "a", encoding="utf-8") as index_file:
         title = "Release Notes"
 
         index_file.write("=" * len(title))
@@ -126,7 +126,8 @@ def gen_release_notes(path):
             index_file.write("\n")
             index_file.write("\n")
 
-            with open("%s/%s.rst" % (whatsnew_dir, tag_name), "a") as release_file:
+            with open("%s/%s.rst" % (whatsnew_dir, tag_name), "a", encoding="utf-8") \
+                    as release_file:
                 release_file.write("=" * len(name))
                 release_file.write("\n")
                 release_file.write(name)
--- a/docs/source/user_guide/options.rst
+++ b/docs/source/user_guide/options.rst
@@ -236,10 +236,10 @@ display.max_rows                1000           This sets the maximum number of r
                                                at the repr() in a dataframe. Set `None` to unlimit
                                                the input length. Default is 1000.
 compute.max_rows                1000           'compute.max_rows' sets the limit of the current
-                                               DataFrame. Set `None` to unlimit the input length.
-                                               When the limit is set, it is executed by the shortcut
-                                               by collecting the data into driver side, and then
-                                               using pandas API. If the limit is unset, the
+                                               Koalas DataFrame. Set `None` to unlimit the input
+                                               length. When the limit is set, it is executed by the
+                                               shortcut by collecting the data into the driver, and
+                                               then using the pandas API. If the limit is unset, the
                                                operation is executed by PySpark. Default is 1000.
 compute.shortcut_limit          1000           'compute.shortcut_limit' sets the limit for a
                                                shortcut. It computes specified number of rows and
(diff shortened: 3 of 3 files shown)


itholic (approved):
Thanks for the improve the documents :)
Seems fine to me.

docs/source/user_guide/options.rst
```diff
@@ -240,7 +240,10 @@ compute.max_rows                1000           'compute.max_rows' sets the limit
                                                When the limit is set, it is executed by the shortcut
```
HyukjinKwon, line 240: I think it's easier to just document that `'compute.max_rows' sets the limit of the current DataFrame.` -> `compute.max_rows' sets the limit of the current Koalas DataFrame.`.

It is documented that `groupby.apply` passes pandas DataFrame so explicitly mentioning it is a Koalas DataFrame will clarify all similar cases.

HyukjinKwon (changes requested):
Can you also update https://github.com/databricks/koalas/blob/e6f90e880ca44d8e73f8480dc4d2e03ae910aeee/databricks/koalas/config.py#L137-L140? This table is supposed to be generated via:

https://github.com/databricks/koalas/blob/e6f90e880ca44d8e73f8480dc4d2e03ae910aeee/databricks/koalas/config.py#L116-L117

docs/source/user_guide/options.rst
lfdversluis: If I understand correctly, even though `compute.max_rows` rows will be loaded into memory, Koalas will still ensure that if a group exceeds `compute.max_rows` rows, it will pass a Pandas DataFrame with all rows the the function(s). I think emphasizing this is an important, too. Any suggestions?

lfdversluis:
I'll fix the lint errors but on Windows I am getting.
```
  File "c:\users\l\appdata\local\programs\python\python37\lib\site-packages\sphinx\config.py", line 327, in eval_config_file
    execfile_(filename, namespace)
  File "c:\users\l\appdata\local\programs\python\python37\lib\site-packages\sphinx\util\pycompat.py", line 88, in execfile_
    exec(code, _globals)
  File "C:\Users\L\Documents\koalas\docs\source\conf.py", line 59, in <module>
    gendoc()
  File "C:\Users\L\Documents\koalas\docs\source\conf.py", line 56, in gendoc
    gendoc.gen_release_notes(source_dir)
  File "C:\Users\L\Documents\koalas\docs\source/../../dev/gendoc.py", line 137, in gen_release_notes
    release_file.write(release_doc)
  File "c:\users\l\appdata\local\programs\python\python37\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\u200b' in position 1486: character maps to <undefined>
``` 

I will check this out quickly.

docs/source/user_guide/options.rst
```diff
@@ -270,6 +273,5 @@ plotting.sample_ratio           None           'plotting.sample_ratio' sets the
                                                'plotting.max_rows' option.
 plotting.backend                'plotly'       Backend to use for plotting. Default is plotly.
                                                Supports any package that has a top-level `.plot`
-                                               method. Some options are: [matplotlib, plotly,
```
lfdversluis, line 273: I double checked this because I didn't change a line, but it looks like a previous change in config.py wasn't synced. I figured I could add it in this PR as well.

lfdversluis:
For safety, the hardcoded paths should perhaps be refactored using `os.path.join`, but that is for another PR :)

lfdversluis:
Ah I see Koalas still supports Python 3.5, f-strings were introduced in 3.6. I thought 3.5 was dropped along with [Spark 3.1.1.](https://spark.apache.org/releases/spark-release-3-1-1.html), my bad.

lfdversluis:
@HyukjinKwon please have another look.

dev/gendoc.py
```diff
@@ -19,21 +19,20 @@
...
+import re
 import shutil
 import sys
 from distutils.version import LooseVersion
 from urllib.error import HTTPError
 from urllib.request import urlopen, Request
-import json
-import os
-import re
-import platform
```
HyukjinKwon, line 30: I would remove unrelated changes here.

dev/gendoc.py
```diff
@@ -19,21 +19,20 @@
...
 
+import databricks.koalas as ks
 import pypandoc
 from pypandoc import pandoc_download
 
-import databricks.koalas as ks
```
HyukjinKwon, line 35: and PEP8 specifies the import order. I believe the previous import order was correct. Let's revert this change back.

docs/source/user_guide/options.rst
HyukjinKwon: Yeah, sounds good!

databricks/koalas/config.py
```diff
@@ -134,10 +134,12 @@ def validate(self, v: Any) -> None:
...
         doc=(
-            "'compute.max_rows' sets the limit of the current DataFrame. Set `None` to unlimit "
-            "the input length. When the limit is set, it is executed by the shortcut by "
-            "collecting the data into driver side, and then using pandas API. If the limit is "
-            "unset, the operation is executed by PySpark. Default is 1000."
+            "'compute.max_rows' sets the limit of the current Koalas DataFrame. Set `None` to "
+            "unlimit the input length. When the limit is set, it is executed by the shortcut by "
+            "collecting the data into the driver, and then using the pandas API. If the limit is "
+            "unset, the operation is executed by PySpark. Note that this option does not impact "
+            "the sizes of the Pandas DataFrames/Series received by functions of GroupBy. "
```
HyukjinKwon, line 141: I suggested to remove this because other friends like pandas DataFrame/Series via `DataFrame.apply_batch` also don't get affected by this configuration. Elaborating these cases all looks slightly an overkill, and I think "Koalas DataFrame on driver side" clarifies it enough.

lfdversluis:
@HyukjinKwon I understand now what you mean by mentioning explicitly Koalas Dataframe. From this you can derive that other DataFrames are not impacted. I made the changes and rebased.

docs/source/user_guide/options.rst
```diff
@@ -270,6 +270,7 @@ plotting.sample_ratio           None           'plotting.sample_ratio' sets the
...
-                                               pandas_bokeh, pandas_altair].
+                                               method. Known options are: [matplotlib, plotly].
 =============================== ============== =====================================================
+
+
```
HyukjinKwon, line 276:
```suggestion
```

HyukjinKwon (approved):
Looks good. Thank you @lfdversluis for clarifying this.

PR #2085, opened 2021-03-05
Add note to MultiIndex.is_monotonic*

itholic:
`MultiIndex.is_monotonic_increasing` and `MultiIndex.is_monotonic_decreasing` won't work in Koalas < 1.7.0 with PySpark 3.1.1 without disabling the `spark.sql.optimizer.nestedSchemaPruning.enabled`.

We should mention this to the docs.

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -767,6 +767,9 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
             which is potentially expensive. In case of multi-index, all data are
             transferred to single node which can easily cause out-of-memory error currently.
 
+        .. note:: Disable the `spark.sql.optimizer.nestedSchemaPruning.enabled`
+            for multi-index if you're using Koalas < 1.7.0 with PySpark 3.1.1.
+
         Returns
         -------
         is_monotonic : bool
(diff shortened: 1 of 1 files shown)


databricks/koalas/base.py
```diff
@@ -767,6 +767,9 @@ def is_monotonic(self) -> bool:
             which is potentially expensive. In case of multi-index, all data are
             transferred to single node which can easily cause out-of-memory error currently.
 
+        .. note:: Disable the `spark.sql.optimizer.nestedSchemaPruning.enabled`
```
HyukjinKwon, line 770: I think you should mention that this is a Spark configuration.

databricks/koalas/base.py
```diff
@@ -767,6 +767,9 @@ def is_monotonic(self) -> bool:
...
+            for multi-index if you're using Koalas < 1.7.0 with PySpark 3.1.1.
```
HyukjinKwon, line 771: Ideally we should mention the JIRA issue link (`SPARK-XXXXX`). Shall we make a reproducer, report to Apache Spark and add the issue link here?

itholic: Sure, let me create & add a link. Thanks!

itholic:
Let me handle https://github.com/databricks/koalas/pull/2085#discussion_r588997987 after release.

PR #2094, opened 2021-03-08
Bump up version to 1.7.0

itholic:


HyukjinKwon (approved):
Can we remove

"The chart below is the expected release dates of minor releases." and the chart at https://github.com/databricks/koalas/blob/master/docs/source/development/contributing.rst#release-cadence?

PR #2097, opened 2021-03-10
Move Series.factorize to IndexOpsMixin.

ueshin:
We can reuse `Series.factorize` in `Index`.

```py
>>> kidx = ks.Index(['b', None, 'a', 'c', 'b'])
>>> codes, uniques = kidx.factorize()
>>> codes
Int64Index([1, -1, 0, 2, 1], dtype='int64')
>>> uniques
```


databricks/koalas/tests/indexes/test_base.py
```diff
@@ -2168,3 +2168,16 @@ def test_index_ops(self):
             self.assert_eq(kidx1 * 10 + kidx2, pidx1 * 10 + pidx2)
         else:
             self.assert_eq(kidx1 * 10 + kidx2, (pidx1 * 10 + pidx2).rename(None))
+
+    def test_factorize(self):
```
xinrong-databricks, line 2172: I have a general question not related to this PR.

Some APIs in this file cannot be applied to subclasses of Index, e.g. DatetimeIndex, while others can.

For these APIs that can apply to DatetimeIndex, shall we add a test case for DatetimeIndex in this file?

ueshin: Basically, if the implementation is different in `DatatimeIndex`, we should add the tests in `test_datetime.py`; otherwise we can put tests in either this file or `test_datetime.py`.
What APIs we can't apply to `DatatimeIndex`?

xinrong-databricks (approved):

databricks/koalas/tests/indexes/test_base.py
xinrong-databricks: Got it!

Sorry I meant we have lots of index + 1' in test cases which is not suitable for DatetimeIndex.

xinrong-databricks:
Looks great, thank you!

databricks/koalas/tests/indexes/test_base.py
ueshin: Then we should explicitly disable the operations. Would you like to work on it?

ueshin:
Thanks! I'd merge this now.

databricks/koalas/tests/indexes/test_base.py
xinrong-databricks: Certainly!

PR #2100, opened 2021-03-11
Check index_dtype and data_dtypes more strictly.

ueshin:
Checks explicitly `index_dtype` and `data_dtypes` are the instance of `np.dtype` or `ExtensionDtype`.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -540,7 +540,7 @@ class InternalFrame(object):
         ]
 
         assert all(
-            isinstance(dtype, Dtype.__args__)
+            isinstance(dtype, Dtype.__args__)  # type: ignore
             and (dtype == np.dtype("object") or as_spark_type(dtype, raise_error=False) is not None)
             for dtype in index_dtypes
         ), index_dtypes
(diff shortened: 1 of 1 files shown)


databricks/koalas/internal.py
```diff
@@ -593,13 +595,15 @@ def __init__(
...
             spark_type_to_pandas_dtype(spark_frame.select(scol).schema[0].dataType)
-            if dtype is None
+            if dtype is None or dtype == np.dtype("object")
             else dtype
             for dtype, scol in zip(data_dtypes, data_spark_columns)
         ]
 
         assert all(
-            as_spark_type(dtype, raise_error=False) is not None for dtype in data_dtypes
+            isinstance(dtype, Dtype.__args__)  # type: ignore
```
xinrong-databricks, line 604: Can I take it as pandas dtype consists of np.dtype and ExtensionDtype?

xinrong-databricks (approved):

xinrong-databricks:
Looks great, thank you!

databricks/koalas/internal.py
ueshin: I believe so, so far. https://github.com/pandas-dev/pandas/blob/7af47c9/pandas/_typing.py#L126

xinrong-databricks: Got it, thanks!

ueshin:
Thanks! merging.

PR #2101, opened 2021-03-11
Standardize exceptions of arithmetic operations on Datetime-like data

xinrong-databricks:
Standardize exceptions of arithmetic operations on Datetime-like data


ueshin (approved):
LGTM.

ueshin:
Thanks! merging.

PR #2103, opened 2021-03-15
Fix (DataFrame|Series).isin to pass numpy array

### LINKED ISSUE #2098: Series.isin() throws exception when fed numpy array
Running Spark 3.1.1 with Pyspark 3.1.1. and Koalas 1.7.0
```
df = pd.DataFrame({'a': [1, 2, 3], 'b': [2, 3, 4]})
numpy_arr = np.array([3, 4, 5])
df['a'].isin(numpy_arr)
```
works

```
kdf = ks.from_pandas(df)
kdf[kdf['a'].isin(numpy_arr)]
```

throws `AttributeError: 'numpy.int64' object has no attribute '_get_object_id'`

using `numpy_arr.tolist()` fixes it. Is this deviation from the Pandas API intended?

itholic:
`(Series|DataFrame).isin` don't work properly when passing numpy array as a parameter.

```python
>>> df = pd.DataFrame({'a': [1, 2, 3], 'b': [2, 3, 4]})
>>> numpy_arr = np.array([3, 4, 5])
>>> df['a'].isin(numpy_arr)
0    False
1    False
2     True
Name: a, dtype: bool

>>> kdf = ks.from_pandas(df)
>>> kdf[kdf['a'].isin(numpy_arr)]
...
AttributeError: 'numpy.int64' object has no attribute '_get_object_id'
```

This should resolve #2098

### DIFF
--- a/databricks/koalas/base.py
+++ b/databricks/koalas/base.py
@@ -1161,7 +1161,7 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
 
         Parameters
         ----------
-        values : list or set
+        values : set or list-like
             The sequence of values to test.
 
         Returns
@@ -1202,7 +1202,8 @@ class IndexOpsMixin(object, metaclass=ABCMeta):
                 " to isin(), you passed a [{values_type}]".format(values_type=type(values).__name__)
             )
 
-        return self._with_new_scol(self.spark.column.isin(list(values)))
+        values = values.tolist() if isinstance(values, np.ndarray) else list(values)
+        return self._with_new_scol(self.spark.column.isin(values))
 
     def isnull(self) -> Union["Series", "Index"]:
         """
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -7090,9 +7090,11 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         if isinstance(values, dict):
             for i, col in enumerate(self.columns):
                 if col in values:
+                    item = values[col]
+                    item = item.tolist() if isinstance(item, np.ndarray) else list(item)
                     data_spark_columns.append(
                         self._internal.spark_column_for(self._internal.column_labels[i])
-                        .isin(values[col])
+                        .isin(item)
                         .alias(self._internal.data_spark_column_names[i])
                     )
                 else:
@@ -7100,9 +7102,10 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
                         F.lit(False).alias(self._internal.data_spark_column_names[i])
                     )
         elif is_list_like(values):
+            values = values.tolist() if isinstance(values, np.ndarray) else list(values)
             data_spark_columns += [
                 self._internal.spark_column_for(label)
-                .isin(list(values))
+                .isin(values)
                 .alias(self._internal.spark_column_name_for(label))
                 for label in self._internal.column_labels
             ]
--- a/databricks/koalas/tests/test_dataframe.py
+++ b/databricks/koalas/tests/test_dataframe.py
@@ -1848,9 +1848,15 @@ class DataFrameTest(ReusedSQLTestCase, SQLTestUtils):
         kdf = ks.from_pandas(pdf)
 
         self.assert_eq(kdf.isin([4, "six"]), pdf.isin([4, "six"]))
+        # Seems like pandas has a bug when passing `np.array` as parameter
+        self.assert_eq(kdf.isin(np.array([4, "six"])), pdf.isin([4, "six"]))
         self.assert_eq(
             kdf.isin({"a": [2, 8], "c": ["three", "one"]}),
             pdf.isin({"a": [2, 8], "c": ["three", "one"]}),
+        )
+        self.assert_eq(
+            kdf.isin({"a": np.array([2, 8]), "c": ["three", "one"]}),
+            pdf.isin({"a": np.array([2, 8]), "c": ["three", "one"]}),
         )
 
         msg = "'DataFrame' object has no attribute {'e'}"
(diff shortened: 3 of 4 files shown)


databricks/koalas/tests/test_dataframe.py
```diff
@@ -1848,10 +1848,16 @@ def test_isin(self):
         kdf = ks.from_pandas(pdf)
 
         self.assert_eq(kdf.isin([4, "six"]), pdf.isin([4, "six"]))
+        # Seems like pandas has a bug when passing `np.array` as parameter
```
lfdversluis, line 1851: `pandas` should be `koalas` :)

itholic: Oh, seems like pandas also has bug when passing `np.array` to pandas `DataFrame.isna`.

For example, let's say we have pandas DataFrame as below.

```python
>>> pdf
          a  b      c
0.475873  4  1    one
0.659894  2  2  three
0.581484  3  9    six
0.226334  4  4  seven
0.349045  8  2    one
0.464269  6  4      5
```

Then use `isin` to check if each column of DataFrame has `4` or `"six"`.

```python
>>> pdf.isin(np.array([4, "six"]))
              a      b      c
0.475873  False  False  False
0.659894  False  False  False
0.581484  False  False   True
0.226334  False  False  False
0.349045  False  False  False
0.464269  False  False  False
```

I think this should be

```python
>>> pdf.isin([4, "six"])
              a      b      c
0.475873   True  False  False
0.659894  False  False  False
0.581484  False  False   True
0.226334   True   True  False
0.349045  False  False  False
0.464269  False   True  False
```

itholic: Ohh... thanks, good catch!

This is because Koalas internally uses PySpark's `isin` directly,

but seems like PySpark's `isin` doesn't distinguish the type between `4` and `"4"`...

```python
>>> sdf
DataFrame[__index_level_0__: double, a: bigint, b: bigint, c: string, __natural_order__: bigint]

>>> sdf.select(sdf.a).show()
+---+
|  a|
+---+
|  4|
|  2|
|  3|
|  4|
|  8|
|  6|
+---+

>>> sdf.select(sdf.a.isin(4)).show()
+----------+
|(a IN (4))|
+----------+
|      true|
|     false|
|     false|
|      true|
|     false|
|     false|
+----------+

>>> sdf.select(sdf.a.isin('4')).show()  # I think this should return `false` for all values since the type of column `a` is `bigint`
+----------+
|(a IN (4))|
+----------+
|      true|
|     false|
|     false|
|      true|
|     false|
|     false|
+----------+
```

databricks/koalas/base.py
```diff
@@ -1202,7 +1202,8 @@ def isin(self, values) -> Union["Series", "Index"]:
                 " to isin(), you passed a [{values_type}]".format(values_type=type(values).__name__)
             )
 
-        return self._with_new_scol(self.spark.column.isin(list(values)))
+        values = values.tolist() if isinstance(values, np.ndarray) else list(values)
```
xinrong-databricks, line 1205:
```
>>> np.array([3, 4, 5]).tolist() == list(np.array([3, 4, 5]))
True
```
I'm wondering why we check `isinstance(values, np.ndarray)`?

itholic: This is because the type of items is different.

```python
>>> for item in np.array([3, 4, 5]).tolist():
...     type(item)  # int
...
<class 'int'>
<class 'int'>
<class 'int'>

>>> for item in list(np.array([3, 4, 5])):
...     type(item)  # numpy.int64
...
<class 'numpy.int64'>
<class 'numpy.int64'>
<class 'numpy.int64'>
```

And, seems like PySpark's `isin` doesn't understand `numpy.int64`.

```python
>>> sdf.a.isin(np.int64)  # `sdf` is PySpark DataFrame
...
AttributeError: type object 'numpy.int64' has no attribute '_get_object_id'
```

xinrong-databricks: Thanks for the explanation! That's clear.

xinrong-databricks (approved):

databricks/koalas/frame.py
```diff
@@ -7090,19 +7090,22 @@ def isin(self, values) -> "DataFrame":
         if isinstance(values, dict):
             for i, col in enumerate(self.columns):
                 if col in values:
+                    item = values[col]
+                    item = item.tolist() if isinstance(item, np.ndarray) else list(item)
```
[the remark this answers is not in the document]
itholic: Thanks for the comment :D

IIRC, `list()` itself calls the `__iter__` internally, so I believe `list(a)` and `list(iter(a))` exactly same.

And I think we still need `tolist()` since https://github.com/databricks/koalas/pull/2103#discussion_r594781305 ?

```python
>>> a = np.array([1,2,3])
>>> list(iter(a))
[1, 2, 3]
>>> type(list(iter(a))[0])
<class 'numpy.int64'>  # this should be `int`
>>> type(a.tolist()[0])
<class 'int'>
```

Please let me know if there is anything I missing!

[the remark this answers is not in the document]
lfdversluis: You are right, I think your method is correct.

databricks/koalas/tests/test_dataframe.py
ueshin: It must be caused by type-coercion rule. We can leave it as-is. cc @HyukjinKwon

itholic: Thanks, @ueshin :)

itholic:
I'm pretty sure for this fix. Please feel free to leave comment if any!

PR #2108, opened 2021-03-18
Improve documentations of transform/apply functions

xinrong-databricks:
Improve documentations of transform/apply functions.

Some transform/apply functions will add a default index when the given UDF has return type annotated. This PR raises users' attention on that.

### DIFF
--- a/docs/source/user_guide/transform_apply.rst
+++ b/docs/source/user_guide/transform_apply.rst
@@ -90,7 +90,7 @@ Koalas combines the pandas DataFrames as a Koalas DataFrame.
 
 Note that :func:`DataFrame.koalas.transform_batch` has the length restriction - the length of input and output should be
 the same whereas :func:`DataFrame.koalas.apply_batch` does not.  However, it is important to know that
-the output belongs to the same DataFrame when :func:`DataFrame.koalas.transform_batch` can a Series, and
+the output belongs to the same DataFrame when :func:`DataFrame.koalas.transform_batch` returns a Series, and
 you can avoid a shuffle by the operations between different DataFrames. In case of :func:`DataFrame.koalas.apply_batch`, its output is always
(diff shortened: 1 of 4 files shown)


itholic (approved):

docs/source/user_guide/transform_apply.rst
```diff
@@ -100,6 +100,9 @@ treated that it belongs to a new different DataFrame. See also
   :align: center
   :width: 650
 
+In case of :func:`DataFrame.koalas.apply_batch`, when the given function has the return type annotated, the original index of the DataFrame will be lost and then a default index will be attached to the result DataFrame. Please be careful about configuring default index. See also
```
ueshin, line 103: I guess `DataFrame.apply` as well? Could you confirm?

ueshin: I guess we should rather put this in the each function's doc?

- https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.DataFrame.apply.html
- https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.DataFrame.koalas.apply_batch.html

ueshin: What about `DataFrame.koalas.transform_batch` with the UDF returning `DataFrame`?

xinrong-databricks: Updated!

I added GroupBy.transform as well.

ueshin: ah, then GroupBy.apply too?

xinrong-databricks: Ah good catch! I was greping 'retain_index=False` so missed this one. Updated!

ueshin (approved):
LGTM.

xinrong-databricks:
Thanks, merging!

PR #2109, opened 2021-03-18
Implement DatetimeIndex.indexer_at_time()

xinrong-databricks:
```py
>>> kidx = ks.date_range("2000-01-01", periods=3, freq="T")
>>> kidx
DatetimeIndex(['2000-01-01 00:00:00', '2000-01-01 00:01:00',
               '2000-01-01 00:02:00'],
              dtype='datetime64[ns]', freq=None)

>>> kidx.indexer_at_time("00:00")
Int64Index([0], dtype='int64')

>>> kidx.indexer_at_time("00:01")
Int64Index([1], dtype='int64')
```

### DIFF
--- a/databricks/koalas/indexes/datetimes.py
+++ b/databricks/koalas/indexes/datetimes.py
@@ -690,6 +690,49 @@ class DatetimeIndex(Index):
             kdf = kdf.koalas.apply_batch(pandas_between_time)
         return ks.Index(first_series(kdf).rename(self.name))
 
+    def indexer_at_time(self, time: Union[datetime.time, str], asof: bool = False) -> Index:
+        """
+        Return index locations of values at particular time of day
+        (e.g. 9:30AM).
+
+        Parameters
+        ----------
+        time : datetime.time or str
+            Time passed in either as object (datetime.time) or as string in
+            appropriate format ("%H:%M", "%H%M", "%I:%M%p", "%I%M%p",
+            "%H:%M:%S", "%H%M%S", "%I:%M:%S%p", "%I%M%S%p").
+
+        Returns
+        -------
+        values_at_time : Index of integers
+
+        Examples
+        --------
+        >>> kidx = ks.date_range("2000-01-01", periods=3, freq="T")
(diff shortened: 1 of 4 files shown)


databricks/koalas/indexes/datetimes.py
```diff
@@ -634,6 +636,49 @@ def strftime(self, date_format: str) -> Index:
         """
         return Index(self.to_series().dt.strftime(date_format))
 
+    def indexer_at_time(self, time: Union[datetime.time, str], asof: bool = False) -> Index:
```
xinrong-databricks, line 639: `asof` is not implemented in pandas yet, so there is not docstring for it.
https://github.com/pandas-dev/pandas/blob/v1.2.3/pandas/core/indexes/datetimes.py#L828-L865

ueshin: Shall we check it before we call `apply_batch`; otherwise the error will be messy.

databricks/koalas/indexes/datetimes.py
```diff
@@ -634,6 +636,49 @@ def strftime(self, date_format: str) -> Index:
...
+            "%H:%M:%S", "%H%M%S", "%I:%M:%S%p", "%I%M%S%p").
+
+        Returns
+        -------
+        values_at_time : Index of integers
+
+        Examples
+        --------
+        >>> kidx = ks.date_range("2000-01-01", periods=3, freq="T")
+        >>> kidx # doctest: +NORMALIZE_WHITESPACE
```
ueshin, line 658: nit: 2 spaces between `kidx` and `# doctest`.

ueshin (approved):
Otherwise, LGTM.
Could you also fix the conflicts?

databricks/koalas/indexes/datetimes.py
xinrong-databricks: Good idea!

databricks/koalas/indexes/datetimes.py
xinrong-databricks: Updated.

ueshin:
Thanks! merging.

PR #2111, opened 2021-03-19
Implemented dateframe.between_time

LSturtew:
ref #1929

Implement `DataFrame.between_time`

```py
>>> i = pd.date_range('2018-04-09', periods=4, freq='1D20min')
>>> ts = pd.DataFrame({'A': [1, 2, 3, 4]}, index=i)
>>> kts = ks.from_pandas(ts)
>>> kts
                     A
2018-04-09 00:00:00  1
2018-04-10 00:20:00  2
2018-04-11 00:40:00  3
2018-04-12 01:00:00  4

>>> kts.between_time('0:15', '0:45')
                     A
2018-04-10 00:20:00  2
2018-04-11 00:40:00  3

You get the times that are *not* between two times by setting
``start_time`` later than ``end_time``:

>>> kts.between_time('0:45', '0:15')
                     A
2018-04-09 00:00:00  1
2018-04-12 01:00:00  4
```


xinrong-databricks:
Thanks for your PR!

Would you add `DataFrame.between_time` to https://github.com/databricks/koalas/blob/master/docs/source/reference/frame.rst?

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,86 @@ class  locomotion
             ).resolved_copy
             return DataFrame(internal)
 
+    def between_time(
+        self, start_time, end_time, include_start=True, include_end=True, axis: Union[int, str] = 0,
```
xinrong-databricks, line 2981: Would you complete type annotations of parameters?

```
        start_time: Union[datetime.time, str],
        end_time: Union[datetime.time, str],
        include_start: bool = True,
        include_end: bool = True,
```

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,86 @@ class  locomotion
...
+        at_time : Select values at a particular time of the day.
+        first : Select initial periods of time series based on a date offset.
+        last : Select final periods of time series based on a date offset.
+        DatetimeIndex.indexer_between_time : Get just the index locations for
+            values between particular times of the day.
+
+        Examples
+        --------
+        >>> i = pd.date_range('2018-04-09', periods=4, freq='1D20min')
+        >>> ts = pd.DataFrame({'A': [1, 2, 3, 4]}, index=i)
```
xinrong-databricks, line 3023:
```
        >>> ts = pd.DataFrame({'A': [1, 2, 3, 4]}, index=i)
        >>> kts = ks.from_pandas(ts)
```
can be
```
        >>> kts = ks.DataFrame({'A': [1, 2, 3, 4]}, index=i)
```

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,86 @@ class  locomotion
...
+        else:
+            raise NotImplementedError("between_time currently only works for axis=0")
+
+        if not isinstance(index, DatetimeIndex):
+            raise TypeError("Index must be DatetimeIndex")
+
+        indexer = index.indexer_between_time(
+            start_time, end_time, include_start=include_start, include_end=include_end
+        ).to_numpy()
+        return self.copy().take(indexer)
```
xinrong-databricks, line 3058: Would `return self.iloc[indexer]` be better?

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,86 @@ class  locomotion
...
+            index = self.index
+        else:
+            raise NotImplementedError("between_time currently only works for axis=0")
+
+        if not isinstance(index, DatetimeIndex):
+            raise TypeError("Index must be DatetimeIndex")
+
+        indexer = index.indexer_between_time(
+            start_time, end_time, include_start=include_start, include_end=include_end
+        ).to_numpy()
```
xinrong-databricks, line 3057: I am not sure `to_numpy()` is a good idea since all the data will be loaded into the driver's memory.
Let me think about this.

LSturtew: The reason I used `to_numpy()` is because without it I got this not implemented error.

`databricks.koalas.exceptions.PandasNotImplementedError: The method `pd.Index.__iter__()` is not implemented. If you want to collect your data as an NumPy array, use 'to_numpy()' instead.`

xinrong-databricks: How about
```
       ...
        def pandas_between_time(pdf):
            return pdf.between_time(start_time, end_time, include_start, include_end)

        return self.koalas.apply_batch(pandas_between_time)
```
?

xinrong-databricks: We might also want to add a test
```
    def test_between_time_no_shortcut(self):
        with ks.option_context("compute.shortcut_limit", 0):
            i = pd.date_range("2018-04-09", periods=4, freq="1D20min")
            ts = pd.DataFrame({"A": [1, 2, 3, 4]}, index=i)
            kts = ks.DataFrame({"A": [1, 2, 3, 4]}, index=i)
            self.assert_eq(
                ts.between_time("0:15", "0:45"), kts.between_time("0:15", "0:45"), almost=True
            )
```

databricks/koalas/frame.py
```diff
@@ -43,7 +43,7 @@
     cast,
     TYPE_CHECKING,
 )
-
+import datetime
```
ueshin, line 46: nit: Could you add an empty line between `import datetime` (built-in library block) and `import numpy as np` (third-party library block)?

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,90 @@ class  locomotion
...
             return DataFrame(internal)
 
+    def between_time(
+        self,
+        start_time: Union[datetime.time, str],
+        end_time: Union[datetime.time, str],
+        include_start: bool = True,
+        include_end: bool = True,
+        axis: Union[int, str] = 0,
+    ) -> "Union[Series[Any], DataFrame[Any]]":
```
ueshin, line 2987: nit: `Union["Series", "DataFrame"]` should be fine.

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,90 @@ class  locomotion
...
+        at_time : Select values at a particular time of the day.
+        first : Select initial periods of time series based on a date offset.
+        last : Select final periods of time series based on a date offset.
+        DatetimeIndex.indexer_between_time : Get just the index locations for
+            values between particular times of the day.
+
+        Examples
+        --------
+        >>> i = pd.date_range('2018-04-09', periods=4, freq='1D20min')
+        >>> kts = ks.DataFrame({'A': [1, 2, 3, 4]}, index=i)
```
ueshin, line 3028: We usually use `idx` for Index, `kdf` or `df` for DataFrame. Shall we follow it?

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,90 @@ class  locomotion
...
+        ``start_time`` later than ``end_time``:
+
+        >>> kts.between_time('0:45', '0:15')
+                             A
+        2018-04-09 00:00:00  1
+        2018-04-12 01:00:00  4
+        """
+        from databricks.koalas.indexes import DatetimeIndex
+
+        if axis == 0:
```
ueshin, line 3051: Could you use `validate_axis` which validates the given value is valid and returns only int even in case the value is string before check the value? Otherwise, `axis='column'` passes here.

databricks/koalas/frame.py
```diff
@@ -2977,6 +2977,90 @@ class  locomotion
...
+            index = self.index
+        else:
+            raise NotImplementedError("between_time currently only works for axis=0")
+
+        if not isinstance(index, DatetimeIndex):
+            raise TypeError("Index must be DatetimeIndex")
```
ueshin, line 3057: Maybe:

```py
if axis != 0:
   raise NotImplementedError ...

if not isinstance(self.index, DatetimeIndex):
   raise TypeError ...
```

is clearer?

databricks/koalas/tests/test_dataframe.py
```diff
@@ -5422,3 +5422,31 @@ def test_align(self):
             pdf_l, pdf_r = pdf1.align(pdf2, join=join, axis=1)
             self.assert_eq(kdf_l.sort_index(), pdf_l.sort_index())
             self.assert_eq(kdf_r.sort_index(), pdf_r.sort_index())
+
+    def test_between_time(self):
+        i = pd.date_range("2018-04-09", periods=4, freq="1D20min")
+        ts = pd.DataFrame({"A": [1, 2, 3, 4]}, index=i)
+        kts = ks.DataFrame({"A": [1, 2, 3, 4]}, index=i)
```
ueshin, line 5429: We usually use `pdf` for pandas DataFrame and `kdf` for Koalas DataFrame. Could you follow it?

ueshin: btw, you can use `kdf = ks.from_pandas(pdf)`.

databricks/koalas/tests/test_dataframe.py
```diff
@@ -5422,3 +5422,31 @@ def test_align(self):
...
+    def test_between_time_no_shortcut(self):
+        with ks.option_context("compute.shortcut_limit", 0):
+            i = pd.date_range("2018-04-09", periods=4, freq="1D20min")
+            ts = pd.DataFrame({"A": [1, 2, 3, 4]}, index=i)
+            kts = ks.DataFrame({"A": [1, 2, 3, 4]}, index=i)
+            self.assert_eq(
+                ts.between_time("0:15", "0:45"),
+                kts.between_time("0:15", "0:45").sort_index(),
+                almost=True,
+            )
```
ueshin, line 5452: Shall we just call `self.test_between_time` here to reuse the tests?

ueshin (commented):
@LSturtew Thanks for working on this!

I left comments, most of them are nits. Otherwise, LGTM so far.

databricks/koalas/tests/test_dataframe.py
xinrong-databricks: Good idea!

databricks/koalas/tests/test_dataframe.py
```diff
@@ -5422,3 +5422,26 @@ def test_align(self):
...
             self.assert_eq(kdf_r.sort_index(), pdf_r.sort_index())
+
+    def test_between_time(self):
+        idx = pd.date_range("2018-04-09", periods=4, freq="1D20min")
+        pdf = pd.DataFrame({"A": [1, 2, 3, 4]}, index=idx)
+        kdf = ks.from_pandas(pdf)
+        self.assert_eq(
+            pdf.between_time("0:15", "0:45"),
+            kdf.between_time("0:15", "0:45").sort_index(),
+            almost=True,
```
ueshin, line 5433: Do we need `almost=True`?

ueshin: ah, seems needed for now.

ueshin (approved):
LGTM.

ueshin:
@LSturtew Thanks! Let me merge this now.
@xinrong-databricks Please feel free to leave additional comments or submit a follow-up PR if any.

xinrong-databricks:
Thanks @LSturtew @ueshin !

PR #2112, opened 2021-03-20
Disable check_freq in testing util.

ueshin:
We'd disable `check_freq` for checking the `freq` in `DatetimeIndex` because so far we don't provide `freq` in `DatetimeIndex`.


ueshin:
Let merge this now.

xinrong-databricks:
Thanks, the change is super helpful!

PR #2113, opened 2021-03-22
Save pandas shortcut of DataFrame.between_time()

xinrong-databricks:
The UDF in DataFrame.between_time() was without return type hint, so the pandas shortcut was required to infer the return type. The PR is proposed to add type hint to the UDF in DataFrame.between_time().

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3066,7 +3066,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         kdf.index.name = verify_temp_column_name(kdf, "__index_name__")
         return_types = [kdf.index.dtype] + list(kdf.dtypes)
 
-        def pandas_between_time(pdf,) -> ks.DataFrame[return_types]:  # type: ignore
+        def pandas_between_time(pdf) -> ks.DataFrame[return_types]:  # type: ignore
             return pdf.between_time(start_time, end_time, include_start, include_end).reset_index()
 
         # apply_batch will remove the index of the Koalas DataFrame and attach a default index,
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3073,12 +3073,7 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         # which will never be used. So use "distributed" index as a dummy to avoid overhead.
         with option_context("compute.default_index_type", "distributed"):
             # Get the new column name from pandas reset_index()
-            if self.index.name is None:
-                index_name = "index"
-            elif self.index.name == "index":
-                index_name = "level_0"
-            else:
-                index_name = self.index.name
+            index_name = self.reset_index().columns[0]
             kdf = self.koalas.apply_batch(pandas_between_time).set_index(index_name)
 
         kdf.index.rename(self.index.name, inplace=True)
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -3069,7 +3069,10 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
         ]:
             return pdf.between_time(start_time, end_time, include_start, include_end).reset_index()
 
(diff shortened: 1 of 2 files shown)


databricks/koalas/frame.py
```diff
@@ -3062,10 +3062,14 @@ def between_time(
...
 
-        def pandas_between_time(pdf):
-            return pdf.between_time(start_time, end_time, include_start, include_end)
+        def pandas_between_time(
+            pdf,
+        ) -> ks.DataFrame[zip(self.reset_index().columns, self.reset_index().dtypes)]:
+            return pdf.between_time(start_time, end_time, include_start, include_end).reset_index()
 
-        return self.koalas.apply_batch(pandas_between_time)
+        kdf = self.koalas.apply_batch(pandas_between_time).set_index("index")
```
ueshin, line 3070: Don't we need to consider the default index?

ueshin: Can we always assume the index column name is `"index"`?

xinrong-databricks: Updated, thanks!

databricks/koalas/frame.py
```diff
@@ -3062,10 +3062,23 @@ def between_time(
...
+            zip(self.reset_index().columns, self.reset_index().dtypes)
+        ]:
+            return pdf.between_time(start_time, end_time, include_start, include_end).reset_index()
+
+        # apply_batch will remove the index of the Koalas DataFrame and attach a default index,
+        # which will never be used. So use "distributed" index as a dummy to avoid overhead.
+        with option_context("compute.default_index_type", "distributed"):
+            # In pandas reset_index(), the new column is assigned the name 'index'
+            # if the original index has no name
+            index_name = "index" if self.index.name is None else self.index.name
```
ueshin, line 3077: I'm afraid this is not good enough. 

```py
>>> kdf = ks.DataFrame({('X', 'A'): [1, 2, 3, 4]}, index=idx)
>>> kdf.between_time('0:15', '0:45')
...
TypeError: not all arguments converted during string formatting
```
or
```py
>>> kdf = ks.DataFrame({'index': [1, 2, 3, 4]}, index=idx)
>>> kdf.between_time('0:15', '0:45')
              level_0
2 2018-04-10 00:20:00
3 2018-04-11 00:40:00
```

ueshin: @xinrong-databricks Could you add the tests with the cases I posted above and the following?

```py
>>> idx = pd.date_range('2018-04-09', periods=4, freq='1D20min', name='index')
>>> kdf = ks.DataFrame({'index': [1, 2, 3, 4]}, index=idx)
>>> kdf.between_time('0:15', '0:45')
...
ValueError: cannot insert index, already exists
```

xinrong-databricks: Sounds good!

ueshin (approved):
LGTM, pending tests.

databricks/koalas/frame.py
```diff
@@ -3062,10 +3062,25 @@ def between_time(
         if not isinstance(self.index, DatetimeIndex):
             raise TypeError("Index must be DatetimeIndex")
 
-        def pandas_between_time(pdf):
-            return pdf.between_time(start_time, end_time, include_start, include_end)
+        kdf = self.copy()
+        kdf.index.name = verify_temp_column_name(kdf, "__index_name__")
+        return_types = [kdf.index.dtype] + list(kdf.dtypes)
+
+        def pandas_between_time(pdf,) -> ks.DataFrame[return_types]:  # type: ignore
```
ueshin, line 3069: nit: remove an extra `,` after `pdf`.

ueshin (commented):
Shall we remove the test `test_between_time_no_shortcut`? It won't go into the shortcut path anymore.

databricks/koalas/frame.py
xinrong-databricks: Thanks!

ueshin:
Thanks! merging.

PR #2122, opened 2021-03-26
Use raise_error=False instead of try-catch.

ueshin:
Now that `as_spark_type` takes an argument `raise_error`, we don't need to use try-catch.


xinrong-databricks (approved):

xinrong-databricks:
Looks good! Thanks!

ueshin:
Thanks! merging.

PR #2123, opened 2021-03-26
Make SeriesType, DataFrameType, and ScalarType hold the dtypes as well.

ueshin:
After the return type inference, `SeriesType`, `DataFrameType`, and `ScalarType` hold the dtypes as well as Spark DataTypes.

### DIFF
--- a/databricks/koalas/typedef/typehints.py
+++ b/databricks/koalas/typedef/typehints.py
@@ -327,6 +327,11 @@ def infer_pd_series_spark_type(pser: pd.Series, dtype: Dtype) -> types.DataType:
 
 def infer_return_type(f) -> Union[SeriesType, DataFrameType, ScalarType, UnknownType]:
     """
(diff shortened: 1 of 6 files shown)


databricks/koalas/typedef/typehints.py
```diff
@@ -324,86 +329,134 @@ def infer_return_type(f) -> Union[SeriesType, DataFrameType, ScalarType, Unknown
     """
```
xinrong-databricks, line 329: Would it be helpful to add a docstring for `infer_return_type` in order to explain return values? For example:

```
Return a type class, which is furthermore understood by Spark.
This type class indicates both dtypes(a pandas only dtype object or a numpy dtype object) and its corresponding Spark DataType.
```

xinrong-databricks:
Looks great! Thank you!

xinrong-databricks (approved):

ueshin:
Thanks! merging.

PR #2124, opened 2021-03-29
Fix groupby-apply and transform to support additional dtypes.

ueshin:
Fix groupby-apply and transform to support additional dtypes.

After this, additional dtypes can be specified in the return type annotation of the UDFs for groupby-apply and transform.

```py
>>> kdf = ks.DataFrame(
...     {
...         "a": pd.Categorical([1, 2, 3, 1, 2, 3]),
...         "b": pd.Categorical(
...             ["b", "a", "c", "c", "b", "a"], categories=["c", "b", "d", "a"]
...         ),
...     },
... )
>>> def identity(df) -> ks.DataFrame[zip(kdf.columns, kdf.dtypes)]:
...     return df
...
>>> applied = kdf.groupby("a").apply(identity)
>>> applied
   a  b
0  2  a
1  2  b
2  3  c
3  3  a
4  1  b
5  1  c
>>> applied.dtypes
a    category
b    category
dtype: object
```

FYI: without the fix:

```py
>>> applied
   a  b
0  1  3
1  1  1
2  2  0
3  2  3
4  0  1
5  0  0
>>> applied.dtypes
a    int64
b    int64
dtype: object
```

### DIFF
--- a/databricks/koalas/groupby.py
+++ b/databricks/koalas/groupby.py
@@ -1182,6 +1182,7 @@ class GroupBy(object, metaclass=ABCMeta):
 
             if isinstance(return_type, DataFrameType):
                 return_schema = cast(DataFrameType, return_type).spark_type
+                data_dtypes = cast(DataFrameType, return_type).dtypes
             else:
                 should_return_series = True
                 return_schema = cast(Union[SeriesType, ScalarType], return_type).spark_type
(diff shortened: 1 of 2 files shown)


databricks/koalas/groupby.py
```diff
@@ -2162,7 +2169,9 @@ def pandas_transform(pdf):
                 retain_index=False,
             )
             # Otherwise, it loses index.
```
xinrong-databricks, line 2171: I may not understand `# Otherwise, ` in this context. I was thinking `transform` would always lose the index.

databricks/koalas/tests/test_categorical.py
```diff
@@ -23,15 +23,27 @@
 
 
 class CategoricalTest(ReusedSQLTestCase, TestUtils):
-    def test_categorical_frame(self):
-        pdf = pd.DataFrame(
+    @property
```
xinrong-databricks, line 26: Nice refactoring!

xinrong-databricks:
Looks great! Thank you!

xinrong-databricks (approved):

PR #2142, opened 2021-04-08
implemented dataframe.cov

LSturtew:
ref #1929

Implement DataFrame.cov

>>> kdf = ks.DataFrame([(1, 2), (0, 3), (2, 0), (1, 1)],
...                   columns=['dogs', 'cats'])
>>> kdf.cov()
                  dogs      cats
        dogs  0.666667 -1.000000
        cats -1.000000  1.666667

### DIFF
--- a/databricks/koalas/frame.py
+++ b/databricks/koalas/frame.py
@@ -4325,6 +4325,111 @@ defaultdict(<class 'list'>, {'col..., 'col...})]
 
         return self._apply_series_op(op)
 
+    def cov(self, min_periods: Optional[int] = None, ddof: Optional[int] = 1) -> "DataFrame":
+        """
+        Compute pairwise covariance of columns, excluding NA/null values.
+        Compute the pairwise covariance among the series of a DataFrame.
+        The returned data frame is the `covariance matrix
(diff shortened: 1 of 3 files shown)


databricks/koalas/frame.py
```diff
@@ -4325,6 +4325,113 @@ def op(kser):
...
+        c -0.148156  0.164205  0.895202
+        """
+        num_cols = [
+            label
+            for label in self._internal.column_labels
+            if isinstance(self._internal.spark_type_for(label), (NumericType))
+        ]
+        kdf = self[num_cols]
+        names = [name for t in num_cols for name in t]
+        mat = kdf.to_pandas().to_numpy(dtype=float, copy=False)
```
ueshin, line 4420: I'm afraid using `to_pandas()` without any restriction is not a good idea. It will cause OOM if the data side doesn't fit in a driver's memory.

xinrong-databricks:
Hi @LSturtew, since Koalas has been ported to Spark as pandas API on Spark, would you like to migrate this PR to [the Spark repository](https://github.com/apache/spark)? I may also do that for you.

PR #2144, opened 2021-04-08
Set SPARK_LOCAL_IP env var.

ueshin:
Try to fix unstable build by explicitly setting `SPARK_LOCAL_IP` env var.

https://github.com/databricks/koalas/runs/2297499540

```
java.net.BindException: Cannot assign requested address: Service 'sparkDriver' failed after 16 retries (on a random free port)! Consider explicitly setting the appropriate binding address for the service 'sparkDriver' (for example spark.driver.bindAddress for SparkDriver) to the correct binding address.
```

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -32,6 +32,7 @@ jobs:
       PYARROW_VERSION: ${{ matrix.pyarrow-version }}
       DEFAULT_INDEX_TYPE: ${{ matrix.default-index-type }}
       KOALAS_TESTING: 1
+      SPARK_LOCAL_IP: 127.0.0.1
       # DISPLAY=0.0 does not work in Github Actions with Python 3.5. Here we work around with xvfb-run
       PYTHON_EXECUTABLE: xvfb-run python
       # Github token is required to auto-generate the release notes from Github release notes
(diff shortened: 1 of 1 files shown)


ueshin:
The error didn't happen during the five builds.
I'd merge this now. cc @HyukjinKwon

HyukjinKwon:
Interesting. Thanks @ueshin for taking a look for this!

HyukjinKwon:
This is a good idea.  is it because `localhost` fails to get 127.0.01 loopback properly for whatever reason (malformed `/etc/hosts`?). One way is to manually fix `/etc/hosts` but I think this is good enough.

ueshin:
If we set the env var, actually Spark skips some logic in `org.apache.spark.util.Utils.findLocalInetAddress` to find a local net address, which sometimes fails for some reason, and force to use the value.

HyukjinKwon:
Nice, yeah, probably we're talking about the same thing I suspect. Thanks for fixing it!!

harupy:
I applied the same fix in the mlflow repo, but it didn't work 😅

HyukjinKwon:
😢

ueshin:
Using `local-cluster` might be a different story..

harupy:
@HyukjinKwon @ueshin 

I found a mistake in my PR 😅 My github actions config looke like below.

```yml
# github actions config

      - name: Install dependencies
        env:
          SPARK_LOCAL_IP: 127.0.0.1  # this should be specified in the `run tests` step.
        run: |
          ...
      - name: Run tests
        run: |
          ...
```

PR #2174, opened 2021-06-17
Fix backward compatibility for Python version 3.5.*

xinrong-databricks:
In Python 3.5.2, errors as below are raised for `Dtype.__args__`. 

```py
databricks/koalas/internal.py
    543             isinstance(dtype, Dtype.__args__)  # type: ignore
    544             and (dtype == np.dtype("object") or as_spark_type(dtype, raise_error=False) is not None)
--> 545             for dtype in index_dtypes
    546         ), index_dtypes
    547 

AttributeError: type object 'Union' has no attribute '__args__'
```

We fix backward compatibility for Python version 3.5.* by introducing `DtypeDataTypes` with `__union_params__` for 3.5.0~3.5.2.

### DIFF
--- a/databricks/koalas/internal.py
+++ b/databricks/koalas/internal.py
@@ -47,8 +47,8 @@ if TYPE_CHECKING:
 from databricks.koalas.config import get_option
 from databricks.koalas.typedef import (
     Dtype,
+    DtypeDataTypes,
     as_spark_type,
-    dtype_data_types,
     extension_dtypes,
     infer_pd_series_spark_type,
     spark_type_to_pandas_dtype,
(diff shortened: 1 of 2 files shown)


databricks/koalas/internal.py
```diff
@@ -77,6 +78,8 @@
 DEFAULT_SERIES_NAME = 0
 SPARK_DEFAULT_SERIES_NAME = str(DEFAULT_SERIES_NAME)
 
+dtype_data_types = (np.dtype, ExtensionDtype)
```
ueshin, line 81: Shall we use the same way as https://github.com/databricks/koalas/blob/master/databricks/koalas/typedef/typehints.py#L497-L502 with `__union_params__` instead of `__tuple_params__`?

ueshin: And shall we maybe put this right after the definition of `Dtype`?

ueshin (commented):
Good catch!

databricks/koalas/internal.py
xinrong-databricks: Sounds good! `__union_params__` is used, and it is put right after the definition of `Dtype`.

ueshin (approved):
LGTM, pending tests.

PR #2196, opened 2021-08-27
[docs] Fix plot was showing twice the same data.

eavilaes:
On API docs, `databricks.koalas.DataFrame.plot.bar`, an example plot was showing on its first row two bars per element, both of them showing the same data. Changed it to show different data, so it is visually clearer.

### DIFF
--- a/databricks/koalas/plot/core.py
+++ b/databricks/koalas/plot/core.py
@@ -638,7 +638,7 @@ class KoalasPlotAccessor(PandasObject):
             ...                    'lifespan': lifespan}, index=index)
             >>> fig = (make_subplots(rows=2, cols=1)
             ...        .add_trace(df.plot.bar(y='speed').data[0], row=1, col=1)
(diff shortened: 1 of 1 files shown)


xinrong-databricks:
Thanks for the fix! Since we have migrated Koalas to Apache Spark, would you like to port the fix to Spark instead?

itholic:
I don't see any wired part in the docs https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.DataFrame.plot.bar.html ??

Can you add before/after figure in the PR description when you find some time ??

eavilaes:
> Thanks for the fix! Since we have migrated Koalas to Apache Spark, would you like to port the fix to Spark instead?

Sure! Are these docs also on the Spark repository?

> I don't see any wired part in the docs https://koalas.readthedocs.io/en/latest/reference/api/databricks.koalas.DataFrame.plot.bar.html ??
> 
> Can you add before/after figure in the PR description when you find some time ??

This is before the PR (notice same data being plotted on the first row):
image

And after the small modification:
image

